PR #10, opened 2016-03-26
Ngbinh update dependencies

melrief:


melrief:
@ngbinh I usually ask to people that contribute to the code to be part of the costributors list for the project. You are the first contributor of pureconfig except me, so the file doesn't exist yet (see [the one in the purecsv library](https://github.com/melrief/PureCSV/blob/master/CONTRIBUTORS) as an example), but I'll ask anyway: would you like to be mentioned in the CONTRIBUTORS file?

ngbinh:
@melrief that sounds great! Hope that I will be able to contribute more. :)

PR #22, opened 2016-08-14
Add `StringConvert[URL]`

leifwickland:


src/main/scala/pureconfig/StringConvert.scala
```diff
@@ -33,3 +34,10 @@ object StringConvert {
   implicit val readShort = fromUnsafe[Short](_.toShort, _.toString)
   implicit val readString = fromUnsafe[String](identity, identity)
 }
+
+/**
+ * Implicit [[StringConvert]] instances defined such that they can be overriden by library consumer via a locally defined implementation.
+ */
+trait LowPriorityStringConvertImplicits {
+  implicit val readURL = StringConvert.fromUnsafe[URL](new URL(_), _.toString)
```
melrief, line 42: I wonder if we should move some of the instances for primitives types here. For some types, like `String`, there is only one instance that make sense but for others, like `Boolean`, you could have different instances, e.g. a project could also accept `0` for `false` and `1` for `true`. What do you think?

leifwickland, line 42: @melrief Yeah, I definitely think that's the right thing to do. However, I think if it's done, it should be done as with an accompanying version bump that indicates a possible backward compatibility break.

I'd be happy to send another PR for that if you like.

melrief:
Sure, that would be nice. I'll wait for that PR for the next release.

PR #24, opened 2016-08-31
Add resolveFiles to make more convenient loading default configs.

leifwickland:


src/main/scala/pureconfig/package.scala
```diff
@@ -111,4 +111,25 @@ package object pureconfig {
...
+  /**
+   * Loads `files` in order, allowing values in later files to backstop missing values from prior, and converts them into a `Config`.
+   *
+   * This is a convenience method which enables having default configuration which backstops local configuration.
+   *
+   * Note: If an element of `files` references a file which doesn't exist or can't be read, it will silently be ignored.
+   *
+   * @param files Files ordered in decreasing priority containing part or all of a `Config`. Must not be empty.
+   */
+  def resolveFiles[Config: ConfigConvert](files: Traversable[java.io.File]): Try[Config] = {
```
melrief: I wonder if the name of the method should be changed to match `loadConfig`. I get that `resolveFiles` is called like that because of how it works, but wouldn't something like `loadConfigFromFiles` be more consistent with the naming of `pureconfig`?

leifwickland: @melrief Absolutely. I'll update the PR. 
Sorry for the inconsistency.

leifwickland:
@melrief I updated the PR with the rename you requested.

melrief:
Thanks, LGTM and I'll merge as soon as this completes and then we can release the next version of pureconfig.

leifwickland:
Awesome. I'd really appreciate having an updated release to depend on.

melrief:
Released version 0.2.0 and added you to the [authors page](https://github.com/melrief/pureconfig/blob/master/AUTHORS). Thanks for all the help :-).

leifwickland:
Thanks, @melrief. Much appreciated.

PR #25, opened 2016-09-02
Fix the way config keys are filtered regarding their namespaces

jcazevedo:
This PR fixes the way config keys are filtered regarding the provided namespaces. Using only `... startsWith namespace` would also consider all namespaces which have the provided argument as a prefix. This fix changes the way keys are filtered by expecting the namespace separator to be present after the `String` supplied as namespace.


melrief:
Right, nice catch. Thanks for the patch.

Do you mind if I add you to the CONTRIBUTORS file?

jcazevedo:
> Do you mind if I add you to the CONTRIBUTORS file?

Not at all. Thanks!

PR #27, opened 2016-09-05
Add a loadConfig method for typesafe Config objects

jcazevedo:
This PR adds a `loadConfig` method for Typesafe `Config` objects. I'm not sure if this is something you'd want in the library, but it is very helpful in my use case. I previously had an implicit conversion from `Config` to `RawConfig` in my codebase, and this change avoids exposing `RawConfig` outside of PureConfig.

### DIFF
--- a/src/main/scala/pureconfig/package.scala
+++ b/src/main/scala/pureconfig/package.scala
@@ -7,7 +7,7 @@
 import java.io.{ OutputStream, PrintStream }
 import java.nio.file.{ Files, Path }
 
-import com.typesafe.config.ConfigFactory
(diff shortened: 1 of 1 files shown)


melrief:
Well, `pureconfig` is using Typesafe config right now and, while I have plans to extend it to other libraries, the current state works only with Typesafe library and could use those utility methods. When `pureconfig` will support different configuration libraries, we will split the `pureconfig` project into the core and the various backends projects and we will move the methods that you have implemented to `pureconfig-typesafeconfig` (name is temporary, I don't know how to call it).

So thanks for the contribution!

melrief:
I'll release `pureconfig` `0.2.2` probably tonight with your changes, @jcazevedo. Stay tuned.

PR #33, opened 2016-09-12
Remove intrusive scalac options in console

jcazevedo:
This PR removes the `-Xfatal-warnings` and the `-Ywarn-unused-import` scalac options when using the sbt console. The two of them combined would render the console unusable, as every isolated import would raise an error, and not import anything. The latter, without the `-Xfatal-warnings` option would, in my opinion, contribute to an unnecessary level of warnings on the console.


melrief:
I'm not sure where my comment went but before merging I wrote that I agree with your comment. I'm using my phone and that's making things more difficult

PR #36, opened 2016-09-17
add jcazevedo to the authors

melrief:
@jcazevedo this PR  adds you as author. Let me know if the info are correct and if you want to add a public email, else I can merge and we are done.


jcazevedo:
The info is correct. You can add my public email: joao.c.azevedo@gmail.com. Thanks!

PR #38, opened 2016-09-17
Add tests to establish behavior of "+=" with loadConfigFromFiles()

leifwickland:
I'll admit adding these tests to Pureconfig's suite has somewhat tenuous justification. I went for it anyway since I think the result is a bit of a quirk of the way `loadConfigFromFiles` works.

Essentially the tests encode the somewhat surprising (to me) behavior that Typesafe Config's `withFallback` produces when files use `+=` to append to a list.


src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -446,4 +446,38 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
...
+    )
+    val expected = List(
+      98, // These three surprise me; they appear first despite coming from the lower precedence file, rather than at the end of the list.
+      99,
+      100, // 98 and 100 are included via HOCON `include`. Their order within the file is preserved.
+      8, // The next three come in order from the greater precedence file, as I'd expect.
+      9,
+      12,
+      50 // This one is slurped in via a HOCON `include`.
+    ).map(Foo(_))
```
melrief: I'm actually not sure this is correct. Maybe the [code that we are using for this](https://github.com/melrief/pureconfig/blob/fb415c71952c22c1ab071fb08662970212ac761f/src/main/scala/pureconfig/package.scala#L123) is somehow broken?

melrief: Ok I think I see a potential issue: in the code we are using `reduce` instead of `reduceLeft` which means that  ["The order in which operations are performed on elements is unspecified and may be nondeterministic."](http://www.scala-lang.org/api/current/index.html#scala.collection.immutable.List@reduce[A1%3E:A]%28op:%28A1,A1%29=%3EA1%29:A1). We should fix this ASAP.

leifwickland:
@melrief It looks like the build failed due to a transient download error. 
Is there a way to force the build to happen again?

melrief:
Restarted

leifwickland:
@melrief Thanks!
That demonstrated that I currently have the unit test marked as ignored. I'll fix that up at some point and pester you again. Sorry about that.

melrief:
No worries!

Issue #34, opened 2016-09-14
Allow configuring the word delimiters to be used in implicitly derived ConfigConverts


jcazevedo:
It would be nice if `pureconfig` would allow customizing the word delimiters for the `ConfigConverts` it derives implicitly. 

This would allow, for example, to have a class with fields defined in camel case:

``` scala
case class Conf(standardTimeout: Duration, numberOfNodes: Int)
```

and the corresponding config using hyphen-separated strings:

```
{ 
  standard-timeout = 5s
  number-of-nodes = 10
}
```

I have a working implementation of this in [this branch](https://github.com/jcazevedo/pureconfig/tree/word-delimiters), but it depends on #28. 

I approached it by requiring an implicit in scope for deriving products that defines how words are delimited. There's a default implicit available that doesn't do anything to the provided fields, but users can provide a different (with higher priority) one to customize the behavior.

For example, if we have the following class:

``` scala
case class Conf(standardTimeout: Duration, numberOfNodes: Int)
val conf = Conf(5.seconds, 10)
```

The default behaviour would be:

``` scala
println(conf.toConfig)
// SimpleConfigObject({"numberOfNodes":10,"standardTimeout":"5s"})
```

But declaring a different way to delimit words would yield:

``` scala
implicit val delim = new WordDelimiterConverter[Conf](CamelCaseWordDelimiter, HyphenWordDelimiter)
println(conf.toConfig)
// SimpleConfigObject({"number-of-nodes":10,"standard-timeout":"5s"})
```

melrief:
The more the library is used the more people ask me if it is configurable which makes me think this issue is an important one.

I personally don't like implicits options and I think that  passing the options to use to `loadConfig` could be easier to read compared to the implicit options. For instance

``` scala
val configLoaderOptions = ConfigLoaderOptions(CamelCaseWordDelimiter, HyphenWordDelimiter)
pureconfig.loadConfig[Conf](configLoaderOptions)
```

jcazevedo:
> I personally don't like implicits options and I think that passing the options to use to loadConfig could be easier to read compared to the implicit options.

OK. The `loadConfig` method should still support being called without options, using a default value (in this case, not doing anything special regarding word delimiters), right? 

I'll prepare a PR during the weekend and we can discuss further details there.

jcazevedo:
I think we can close this now that #39 is merged in.

PR #39, opened 2016-09-18
Allow configuring config field mappings to be used in implicitly derived converters

jcazevedo:
This PR implements the idea suggested in #34. 

@melrief: I understand you prefer explicitly providing the options in the `loadConfig` method, but I think that providing the converters for word delimiters through implicits has some advantages, as you can have different word delimiters for different implicitly derived types and, since implicit converters are only picked up by the implicit derivations of `pureconfig`, the behaviour for custom `ConfigConvert` implementations is more predictable.


src/main/scala/pureconfig/WordDelimiter.scala
```diff
@@ -0,0 +1,42 @@
+package pureconfig
+
+trait WordDelimiter {
+  def toTokens(s: String): Seq[String]
```
ruippeixotog: Instead of restricting config<->field mappings to word-based rules, wouldn't it be more flexible if you required only an object able to convert from field names to config keys and vice-versa? `WordDelimiter` could then be a subclass that implemented the conversion via word tokenization.

The use case I have in mind is when users have config schemas they do not control and they want to do fully custom mappings of specific keys. Of course common sense suggests that such flexibility should not be abused, but it would be good to have that choice if the need arises.

EDIT: of course you can do that with the current interface by returning a single token, but it makes the API a bit strange :)

jcazevedo: > The use case I have in mind is when users have config schemas they do not control and they want to do fully custom mappings of specific keys.

That's currently possible via a custom `ConfigConvert` for those types, but I'm not against adding that flexibility on the config<->field mappings. 

@melrief: what do you think?

melrief: I think @ruippeixotog means that you want to base the behaviour of the `ConfigConvert` on the key instead of the type. `ConfigConvert` works for types, not specific keys. @ruippeixotog the canonical way to achieve this is to use a custom new type that wraps another type for those specific keys and then define a `ConfigConvert` for it which differs from the standard one. For instance, let's say that you have a `case class Conf(n: Int, pos: Int)` and you want to have different codecs for `n` and `pos` because `pos` must be positive. The way to do this is to define `case class Pos(int: Int)`, then change `Conf` to `case class Conf(n: Int, pos: Pos)` and then make a codec for `Pos` which checks the value.
This is what other libraries do, for instance `ScalaCheck`.
Now, the question is more if we want to give to the user the ability to define somehow specific codecs for certain keys. I'm against this because I think using a new type is the way to go but I may change idea if there is a compelling use case.

ruippeixotog: Ok, it would be more verbose as you'd have to extract all fields of the target class individually rather than provide per-field "exceptions". But now that you suggested that, I think I would prefer to have a custom `ConfigConvert` rather than add a custom field conversion implicit :)

melrief: Sure :). I'm still not sure how the library should handle exceptional behaviour for some fields. This PR is a first attempt to have a configuration but there are many things to decide.

melrief:
@jcazevedo because this is a big change, I would like to first merge the fix at #43, release one version of `pureconfig` and then work on the review of this one. As it stand, I mostly like this PR but I would like to try it first.

jcazevedo:
Sure. No problem. I'll try to review #43 tonight. I also have a small change I'd like to make to the converters' interface in this PR.

jcazevedo:
I changed the `WordDelimiterConverter` to a more generic `ConfigFieldMapping` type class, so that we don't restrict mappings to word based rules. It's still possible to build a `ConfigFieldMapping` from two `WordDelimiters` using the `WordDelimiterConfigFieldMapping` class.

src/main/scala/pureconfig/ConfigFieldMapping.scala
```diff
@@ -0,0 +1,24 @@
+package pureconfig
+
+trait ConfigFieldMapping[T] {
+  def toConfigField(fieldName: String): String
+}
+
+class NoopConfigFieldMapping[T] extends ConfigFieldMapping[T] {
```
melrief: I prefer `IdentityConfigFieldMapping`, it's common to use `identity` in Scala for the identity function.

src/main/scala/pureconfig/ConfigFieldMapping.scala
```diff
@@ -0,0 +1,24 @@
+package pureconfig
+
+trait ConfigFieldMapping[T] {
```
melrief: I may have to do some tests before we merge this because this gives a lot of flexibility that maybe is not required. I understand that one could use different styles for different `T`s within the same file, hypothetically, but in reality how many times this will happen? If we decide for a more simple system where keys of all types are treated in the same way, then the interface would be much easier and accessible. We would just need a function `String => String` with default `identity` and no `Wrapper`.

jcazevedo: I have a use case for that. I want to load my full configuration into a domain object. Within the full configuration, there's an object with Spark properties. My application keys are hyphen separated in the configuration file, but the Spark keys retain their original name, which uses [camel case](http://spark.apache.org/docs/latest/configuration.html#application-properties). I want to retain the original names of the Spark properties because I want to build a `Config` from that domain object later on, whose values are to be dealt directly by Spark.

We can simplify the interface by just requiring a function `String => String` and I can still implement my use case with a custom `ConfigConvert` for my Spark properties class, so that's not a big issue. In that case, do you think it still makes sense for `pureconfig` to provide those helpers for different word delimiters, or should we leave that completely for the user to deal with?

src/main/scala/pureconfig/ConfigFieldMapping.scala
jcazevedo: I agree. I've changed it in 1f9e6e4.

src/main/scala/pureconfig/ConfigFieldMapping.scala
jcazevedo: I was thinking of an alternative to allow customizing the field mappings without having to provide implicit options while also keeping the flexibility to have different mapping rules for different types, and I think there's a better solution for it than providing a `String => String` function in the `loadConfig` method.

I think we can provide utilities to create `ConfigConvert` instances using a `ConfigFieldMapping` (which will probably become untyped then). That way, we can provide a system-wide conversion with a generic `ConfigConvert` and still customize more specific types.

@melrief: What do you think? If you like that idea, I can adapt this pull request.

melrief: @jcazevedo sorry to keep you wait for this PR. The `String => String` was my first idea for this feature and I think I like it because it's very simple. I'm not sure what do you mean with the  `ConfigFieldMapping` and how you can configure for specific types. What kind of Map do you have in mind?

jcazevedo: The current implementation in this PR assumes that you have an implicit `ConfigFieldMapping` in scope that tells you how fields of classes should map to fields in the config. There's always one available that maps the fields using the identity function. This allows you to have different mappings for different types. For example:

``` scala
import pureconfig._
import pureconfig.syntax._
import com.typesafe.config._

case class ConfWithCamelCaseInner(thisIsAnInt: Int, thisIsAnotherInt: Int)
case class ConfWithCamelCase(camelCaseInt: Int, camelCaseString: String, camelCaseConf: ConfWithCamelCaseInner)

implicit val conv = new WordDelimiterConfigFieldMapping[ConfWithCamelCase](CamelCaseWordDelimiter, HyphenWordDelimiter)

  camel-case-int = 1
  camel-case-string = "bar"
  camel-case-conf {
    thisIsAnInt = 3
    thisIsAnotherInt = 10
  }
}""")

conf.to[ConfWithCamelCase]
// returns Success(ConfWithCamelCase(1,bar,ConfWithCamelCaseInner(3,10)))
```

I like to have this flexibility and I have a use case for it in a project. However, I understand that using implicit options is error-prone and a bit magical for the end user. What I was suggesting in my last comment was to have a method to create a `ConfigConvert` from a `ConfigFieldMapping`, so that we no longer have to provide it implicitly (having an implicit `ConfigConvert` in scope is already a reasonable compromise of this library). This would still allow having different mappings for different types, but would simplify the interface:

``` scala
import pureconfig._
import pureconfig.syntax._
import com.typesafe.config._

case class ConfWithCamelCaseInner(thisIsAnInt: Int, thisIsAnotherInt: Int)
case class ConfWithCamelCase(camelCaseInt: Int, camelCaseString: String, camelCaseConf: ConfWithCamelCaseInner)

implicit val conv = ConfigConvert.fromConfigFieldMapping[ConfWithCamelCase](new WordDelimiterConfigFieldMapping(CamelCaseWordDelimiter, HyphenWordDelimiter))

  camel-case-int = 1
  camel-case-string = "bar"
  camel-case-conf {
    thisIsAnInt = 3
    thisIsAnotherInt = 10
  }
}""")

conf.to[ConfWithCamelCase]
// returns Success(ConfWithCamelCase(1,bar,ConfWithCamelCaseInner(3,10)))
```

leifwickland: For comparison, what would this (or the former approach) look like to acheivehyphen-delimiter style throughout the config?

leifwickland (changes requested):

src/main/scala/pureconfig/ConfigFieldMapping.scala
jcazevedo: To make sure that all types use hyphen-delimited fields in the config, if we rely on implicit `ConfigFieldMapping`, one would have to have an implicit in scope for all types:

``` scala
implicit def conv[T] = new WordDelimiterConfigFieldMapping[T](CamelCaseWordDelimiter, HyphenWordDelimiter)
```

In the approach I'm suggesting, one has to have a `ConfigConvert` in scope that overrides the default mapping rules:

``` scala
implicit def conv[T] = ConfigConvert.fromConfigFieldMapping[T](new WordDelimiterConfigFieldMapping(CamelCaseWordDelimiter, HyphenWordDelimiter))
```

The main difference is that the `ConfigConvert` implicit is a bit less "magical", as users of the library are already expecting it to exist when using the `loadConfig` methods, whereas the `ConfigFieldMapping` implicit is a bit more hidden in the API.

melrief: @jcazevedo I like the new proposal. I would like to see the code and review it but, if @leifwickland agreed, I would merge it as soon as the code is ready. Do you want to do it now? I can wait to release pureconfig for a couple of days or I can release it now and then release the configuration system in the next version.

jcazevedo: I think I can make the necessary changes for the new proposal today. I think it's OK to release a new version now and have the configuration system included in the next version, though.

jcazevedo: Well, easier said than done. I think I won't be able to provide an implementation for my proposal until tomorrow, sorry.

jcazevedo: I was trying to implement the proposal I suggested, but unfortunately I don't think we can provide an usage as straightforward as I put it. By doing:

``` scala
implicit def conv[T] = ConfigConvert.fromConfigFieldMapping[T](new WordDelimiterConfigFieldMapping(CamelCaseWordDelimiter, HyphenWordDelimiter))
```

We're effectively providing a `ConfigConvert` for all `T`, which is not possible nor something I think we would want to expose. The workaround I see for it is to have a specific type for `ConfigConverts` we derive implicitly and be able to customize those. However, that would expose an internal type via the API, which is something I'm also not comfortable with.

That being said, I think the current implementation offers the best compromise: it exposes a type that is meant to be used by `ConfigConverts` we derive implicitly and doesn't pollute user-defined `ConfigConverts`, as those don't need a `ConfigFieldMapping`.

@melrief and @leifwickland: What do you guys think?

melrief: I think the current implementation is fine. Again, the main reason why this is not yet merged is not the implementation chosen but because I wanted to review properly the code. Let's stay with the current implementation and I'll have a look again. I may suggest some renaming and some small refactoring but I think it's a good compromise. @jcazevedo do you mind if I take the weekend to have a look, again? Last time I checked this PR it was two weeks ago.

melrief: Also, thanks for all the work @jcazevedo

jcazevedo: > @jcazevedo do you mind if I take the weekend to have a look, again? Last time I checked this PR it was two weeks ago.

Sure. No problem.

melrief:
I had a look again and it's good to merge to me. @jcazevedo thanks for the tests you wrote, very nice and descriptive. @leifwickland what do you think, can we merge and release the 3.3 with the configuration system?

src/main/scala/pureconfig/ConfigFieldMapping.scala
leifwickland: @jcazevedo and @melrief,
Here are a few suggestions that are horrifyingly late in the game. Apologies in advance for not presenting these sooner.
- `object ConfigFieldMapping` could have an `def apply(f: String => String): ConfigFieldMapping`
  - Scalaz often does this or something similar and I find it convenient.
  - If it were `apply` then `ConfigFieldMapping` could extend `String => String`. 
- `object ConfigFieldMapping` could have an: 

```
def apply(typeFieldDelimiter: WordDelimiter, configFieldDelimiter: WordDelimiter): ConfigFieldMapping = {
  if (typeFieldDelimiter == configFieldDelimiter) {
    apply(identity)
  } else {
    apply(typeFieldDelimiter.toTokens _ andThen configFieldDelimiter.fromTokens _)
  }
}
```
- That'd obviate the need for explicitly making a user choose `IdentityConfigFieldMapping` or `WordDelimiterConfigFieldMapping`. We could simplify pureconfig's API by removing them entirely.
- `NamingConvention` might be a better name for `WordDelimiter`. Please pardon the pedantry: a "delimiter" is that which separates. Calling "CamelCase" a delimiter feels a bit awkward. 
- Assuming that you went with the above suggestion, then you could end up with:
  - `CamelCase`: https://en.wikipedia.org/wiki/Camel_case
  - `SnakeCase`: https://en.wikipedia.org/wiki/Snake_case
  - `KebobCase`: http://wiki.c2.com/?KebabCase 
  - `PascalCase`: https://en.wikipedia.org/wiki/PascalCase (This is probably excessive, but I include it just to illustrate the point that your design allows for greater flexibility than just swapping out the delimiter.)
- Please make it clearer what the default naming convention is via something like`:
  - `val default: ConfigFieldMapping = apply(CamelCase, CamelCase)` in `object ConfigFieldMapping`
  - define `implicit def configFieldMapping ... = ConfigFieldMapping.default`
- I'd like a comment on `WordDelimiterConfigFieldMapping` or the `apply` described above explaining the roles of the arguments. E.g.:

```
@param typeFieldDelimiter Naming convention used by the fields of the object which receive config values.
@param configFieldDelimiter Naming convention used in the source configuration (e.g. the config file)
```

src/main/scala/pureconfig/WordDelimiter.scala
```diff
@@ -0,0 +1,37 @@
...
+  def fromTokens(l: Seq[String]): String
+}
+
+object CamelCaseWordDelimiter extends WordDelimiter {
+  def toTokens(s: String): Seq[String] = {
+    s.replaceAll(
```
leifwickland: `replaceAll` recompiles the string into a Pattern each time it's called.
See http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/lang/String.java#2209

I'm pretty sure that means it's recompiling a regex for every field.

I suggest compiling that regex once and storing it in a `private val`.

Then this call site would look like:

```
wordBreakPattern.matcher(this).replaceAll(" ")
```

src/main/scala/pureconfig/ConfigFieldMapping.scala
jcazevedo: Thanks for your review. I agree with all your suggestions. I believe I've addressed them all. I also added a description of how to use the `ConfigFieldMapping` to the README.

src/main/scala/pureconfig/WordDelimiter.scala
jcazevedo: That was indeed costly. I've fixed it in d0643b0.

src/test/scala/pureconfig/NamingConventionTest.scala
```diff
@@ -0,0 +1,110 @@
+package pureconfig
+
+import org.scalatest.{ FlatSpec, Matchers }
+
+class NamingConventionTest extends FlatSpec with Matchers {
+  "A CamelCase" should "properly tokenize words" in {
```
leifwickland: Super nitpick: "A CamelCase" should probably drop the "A" to read a little better.
Ditto all the cases below.

README.md
```diff
@@ -116,6 +117,55 @@ An almost comprehensive example is:
 res0: util.Try[MyClass] = Success(MyClass(1,AdtB(1),List(1.0, 0.2),Map(key -> value),None))
 ```
 
+## Defining naming conventions
+
+In case the naming convention you use in your source configuration files differs
```
leifwickland: Suggestion: I'd drop "source" from this sentence.

README.md
```diff
@@ -116,6 +117,55 @@ An almost comprehensive example is:
...
+from the one used in the objects you're loading the configs into, PureConfig
+allows you to define proper mappings. That configuration should be done by an
+implicit `ConfigFieldMapping` that should be in scope when loading or writing
+configurations. The `ConfigFieldMapping` trait has a single `apply` method that
+maps field names in Scala objects to field names in the source configuration
```
leifwickland, line 127: Suggestion: I'd get rid of "source" here too. I realize this is my fault for how I suggested the comment be written on the `apply`. Sorry. I just couldn't think of a way to avoid saying "config file" in the comment block, where it seemed slightly over specific. Perhaps it'd be less confusing to just say "config file" directly in the comment on `apply`.

README.md
```diff
@@ -116,6 +117,55 @@ An almost comprehensive example is:
...
+file. For instance, here's a contrived example where the configuration file has
+all keys in upper case and we're loading it into a type whose fields are all in
+lower case:
```
leifwickland, line 130: Nice example.

README.md
```diff
@@ -116,6 +117,55 @@ An almost comprehensive example is:
...
+loadConfig[SampleConf](conf)
+// returns Success(SampleConf(2, "two"))
+```
+
+PureConfig provides a way to create a `ConfigFieldMapping` by defining the
+naming conventions of the fields in the Scala object and in the source
+configuration file. Some of the most used naming conventions are supported
+directly in the library:
+
+* [`CamelCase`](https://en.wikipedia.org/wiki/Camel_case)
```
leifwickland: Suggestion: let's add an example on each of the naming conventions.

```
* [`CamelCase`](https://en.wikipedia.org/wiki/Camel_case): `camelCase`, `useMorePureconfig`
* [`SnakeCase`](https://en.wikipedia.org/wiki/Snake_case): `snake_case`, `use_more_pureconfig`
* [`KebabCase`](http://wiki.c2.com/?KebabCase): `kebab-case`, `use-more-pureconfig`
* [`PascalCase`](https://en.wikipedia.org/wiki/PascalCase): `PascalCase`, `UseMorePureconfig`
```

src/main/scala/pureconfig/NamingConvention.scala
```diff
@@ -0,0 +1,46 @@
...
+  def fromTokens(l: Seq[String]): String = {
+    l match {
+      case Seq() => ""
+      case h +: Seq() => h.toLowerCase
+      case h +: t => h.toLowerCase + t.map(_.capitalize).mkString
+    }
+  }
+}
+
+object PascalCase extends CapitalizedWordsNamingConvention {
```
leifwickland, line 40: Suggestion:

```
/**
 * PascalCase identifiers look like e.g.`PascalCase` and `UseMorePureconfig`
 * @see https://en.wikipedia.org/wiki/PascalCase
 */
```

src/main/scala/pureconfig/NamingConvention.scala
```diff
@@ -0,0 +1,46 @@
...
+
+class StringDelimitedNamingConvention(d: String) extends NamingConvention {
+  def toTokens(s: String): Seq[String] =
+    s.split(d).map(_.toLowerCase)
+
+  def fromTokens(l: Seq[String]): String =
+    l.map(_.toLowerCase).mkString(d)
+}
+
+object KebabCase extends StringDelimitedNamingConvention("-")
```
leifwickland, line 56: Suggestion:

```
/**
 * KebabCase identifiers look like `kebab-case` and `use-more-pureconfig`
 * @see http://wiki.c2.com/?KebabCase
 */
```

src/main/scala/pureconfig/NamingConvention.scala
```diff
@@ -0,0 +1,46 @@
...
+
+object SnakeCase extends StringDelimitedNamingConvention("_")
```
leifwickland, line 62: Suggestion:

```
/**
 * SnakeCase identifiers look like `snake_case` and `use_more_pureconfig`
 * @see https://en.wikipedia.org/wiki/Snake_case
 */
```

src/main/scala/pureconfig/NamingConvention.scala
```diff
@@ -0,0 +1,46 @@
...
+
+object CapitalizedWordsNamingConvention {
+  private val wordBreakPattern = String.format(
+    "%s|%s|%s",
+    "(?<=[A-Z])(?=[A-Z][a-z])",
+    "(?<=[^A-Z])(?=[A-Z])",
+    "(?<=[A-Za-z])(?=[^A-Za-z])").r
+}
+
+object CamelCase extends CapitalizedWordsNamingConvention {
```
leifwickland, line 26: Suggestion:

```
/**
 * CamelCase identifiers look like `camelCase` and `useMorePureconfig`
 * @see https://en.wikipedia.org/wiki/Camel_case
 */
```

src/main/scala/pureconfig/ConfigFieldMapping.scala
```diff
@@ -1,24 +1,48 @@
...
-    typeFieldDelimiter: WordDelimiter,
-    configFieldDelimiter: WordDelimiter) extends ConfigFieldMapping[T] {
-  def toConfigField(fieldName: String): String =
-    (typeFieldDelimiter.toTokens _ andThen configFieldDelimiter.fromTokens _)(fieldName)
+trait ConfigFieldMapping[T] extends (String => String) {
+  def apply(fieldName: String): String
 }
 
 object ConfigFieldMapping extends LowPriorityConfigFieldMappingImplicits {
   def apply[T](implicit mapping: ConfigFieldMapping[T]): ConfigFieldMapping[T] = mapping
```
leifwickland: Is this needed? If so is it worth documenting why it exists? (Not a rhetorical question; I really don't know.)

src/main/scala/pureconfig/ConfigFieldMapping.scala
```diff
@@ -1,24 +1,48 @@
...
+
+  /**
+   * Creates a ConfigFieldMapping from the provided function, mapping names in
+   * the object that will receive config values to names in the source
```
leifwickland: Perhaps just "config file".
Ditto other places in the area.

leifwickland (approved):
Thanks again for doing this, @jcazevedo!

src/test/scala/pureconfig/NamingConventionTest.scala
jcazevedo: I did a blind replace and this became a little awkward. Fixed it in 2df58fb.

README.md
jcazevedo: I agree it sounds better to just say "config file" or "configuration file". I've fixed it in ea38720.

README.md
jcazevedo: Good idea. Thanks!

src/main/scala/pureconfig/NamingConvention.scala
jcazevedo: Thanks for the suggestion. It makes sense to add these comments, yes.

jcazevedo:
I think I've addressed all your suggestions, @leifwickland. I've also updated the PR title, as this is no longer about "delimiters".

src/main/scala/pureconfig/ConfigFieldMapping.scala
jcazevedo: It is not necessary, and I've removed it in 590e8bb. This is just a different implementation of `implicitly`. It allows you to get the implicit `ConfigFieldMapping` in scope by just writing, for example (in this case for `Int`), `ConfigFieldMapping[Int]`. This is more useful for `ConfigConvert` than it is for `ConfigFieldMapping`, so I've removed it.

melrief:
@jcazevedo nice job!

melrief:
@leifwickland all the comments have been addressed and it's mergeble for me. What do you think, should we merge and release the new pureconfig?

leifwickland:
Yes!

melrief:
@jcazevedo one last question: there are many commits in this PR but the final changes are not too many. I was going to squash and merge it but maybe you want to keep track of the single commits. What do you prefer?

jcazevedo:
The implementation changed a bit since the PR was opened, so I think it's ok to squash and merge.

PR #41, opened 2016-09-19
Add support for loading Doubles and Floats from a percentage

jcazevedo:
This PR adds support for `Doubles` to be loaded from strings representing a percentage (i.e. that end with a `%`).

### DIFF
--- a/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/src/main/scala/pureconfig/ConfigConvert.scala
@@ -90,7 +90,7 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
             tail <- tConfigConvert.value.from(config)
           } yield field[K](v) :: tail
         case other =>
-          Failure(new Exception(s"Couldn't derive hlist from $other."))
+          Failure(new IllegalArgumentException(s"Couldn't derive hlist from $other."))
       }
     }
 
@@ -114,7 +114,7 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
   }
 
   case class NoValidCoproductChoiceFound(config: ConfigValue)
-    extends RuntimeException(s"No valid coproduct type choice found for configuration $config")
+    extends RuntimeException(s"No valid coproduct type choice found for configuration $config.")
 
   implicit def cNilConfigConvert: ConfigConvert[CNil] = new ConfigConvert[CNil] {
     override def from(config: ConfigValue): Try[CNil] =
@@ -176,7 +176,7 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
 
           tryBuilder.map(_.result())
         case other =>
-          Failure(new Exception(s"Couldn't derive traversable from $other."))
+          Failure(new IllegalArgumentException(s"Couldn't derive traversable from $other."))
       }
     }
 
@@ -201,7 +201,7 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
               } yield acc + (key -> value)
           }
         case other =>
-          Failure(new Exception(s"Couldn't derive map from $other."))
+          Failure(new IllegalArgumentException(s"Couldn't derive map from $other."))
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -246,7 +246,9 @@ trait LowPriorityConfigConvertImplicits {
 
   implicit val readString = fromString[String](identity)
   implicit val readBoolean = fromString[Boolean](_.toBoolean)
-  implicit val readDouble = fromString[Double](_.toDouble)
+  implicit val readDouble = fromString[Double]({
+    v => if (v.last == '%') v.dropRight(1).toDouble / 100.0 else v.toDouble
```
leifwickland: Do you think we should trim the `v` just to ensure that accidental whitespace doesn't mess up the value?

leifwickland: Suggestion: use pattern matching to separate the cases:

```
case "" => throw new Exception("Cannot read a Double from an empty string")
case v if v.last == '%' => v.dropRight(1).toDouble / 100.0
case v => v.toDouble
```

leifwickland: This `v.last` will fail with
`java.util.NoSuchElementException: next on empty iterator`
if `v` is an empty string. Do you think we should make an effort to provide a better error message?

jcazevedo: Yeah, I think we should provide better error messages. I improved them (also for the other "basic" types in 3e43534.

jcazevedo:
I added a similar converter for `Floats` in 3e43534.

src/main/scala/pureconfig/ConfigConvert.scala
melrief: Nice improvement!

melrief: Maybe we could group the code for non empty strings together and provide it as utility to build instances. Something like:

``` scala
def fromNonEmptyString[T](name: String, fromF: String => T): ConfigConvert[T] = new ConfigConvert[T] {
    override def from(config: ConfigValue): Try[T] = fromFConvert { _ match {
      case "" => throw new Exception(s"Cannot read a $name from an empty string")
      case x => fromF(x)
    }(config)
    override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(t)
}
```

It would remove some code and group together the login for empty String errors. What do you think?

leifwickland: I like the idea of having a `fromNonEmptyString`. I'd suggest defining it more like:

```
def fromNonEmptyString[T](typeName: String)(fromF: String => T): ConfigConvert[T] = new ConfigConvert[T] {
    override def from(config: ConfigValue): Try[T] = fromFConvert {
      case "" => throw new IllegalArgumentException(s"Cannot read a $typeName from an empty string.")
      case x => fromF(x)
    }(config)
    override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(t)
}
```

I'd rather see a separate argument list for the function. That'd facilitate writing 

```
fromNonEmptyString("Double") { 
  case v if ... => v.dropRight(1).toDouble
  case v => v.toDouble
}
```

That `_ match {` bit is generally considered bad style, as far as I know. You can just skip straight to the `case`s.

leifwickland: Now that I already suggested using `Exception` in response to an empty string, it seems like we should be using `IllegalArgumentException` instead.

jcazevedo: I added the `fromNonEmptyString` method in eb3ffb1b2487c8dbd233e2221f6bfcde0cacdd86. However, instead of requiring an explicit typename, I've added a context bound for `ClassTag`, in order to extract the typename.

src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -56,6 +57,15 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
     override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(t)
   }
 
+  def fromNonEmptyString[T: ClassTag](fromF: String => T): ConfigConvert[T] = new ConfigConvert[T] {
+    lazy val typeName = implicitly[ClassTag[T]]
```
leifwickland: I really like the idea of using the `ClassTag`. Kudos.
My preference would be to use a `val` or a `def` instead of a `lazy val` given its overhead. The conventional wisdom I subscribe to is that unless the value is very expensive to compute or that you need `lazy val` for correctness, don't use `lazy val`.

src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -104,6 +104,18 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
     config.success.value shouldBe FlatConfig(false, -234.234d, -34.34f, -56, 88L, "QWERTY", None)
   }
 
+  it should "fail when trying to convert to basic types from an empty string" in {
+    import pureconfig.syntax._
+
+    val conf = ConfigFactory.parseString("""{ v: "" }""")
```
leifwickland, line 110: Nice addition of test cases.

leifwickland (commented):
👍

src/main/scala/pureconfig/ConfigConvert.scala
jcazevedo: > Now that I already suggested using `Exception` in response to an empty string, it seems like we should be using `IllegalArgumentException` instead.

I agree. I've changed some of them to `IllegalArgumentException`.

src/main/scala/pureconfig/ConfigConvert.scala
jcazevedo: In that case, it's better to make the implicit parameter "explicit". See 42ac2de for the change.

melrief:
I really like the new code, thanks @jcazevedo!

PR #43, opened 2016-09-20
fix NPE when key not present and restore key not found error

melrief:
@jcazevedo @leifwickland if you have some time please have a look at this fix to the key not found and wrong type issues. I'm still not sure about the implementation and feedback is welcomed. I'll leave this PR open for some days at least. To see the behaviour this PR implements before checking the implementation, you can check the [new test](https://github.com/melrief/pureconfig/compare/fix_npe_on_key_not_found_and_better_error?expand=1#diff-e8e47c5f8227db17829b61b757b0ee0eR472).

### DIFF
--- a/src/test/scala/pureconfig/PureconfSuite.scala
+++ b/src/test/scala/pureconfig/PureconfSuite.scala
@@ -5,21 +5,21 @@ package pureconfig
 
 import java.io.PrintWriter
 import java.net.URL
-import java.nio.file.{ Path, Files }
+import java.nio.file.{ Files, Path }
 import java.util.concurrent.TimeUnit
 
 import com.typesafe.config.{ Config => TypesafeConfig, _ }
 import org.joda.time.DateTime
 import org.joda.time.format.ISODateTimeFormat
-import shapeless.{ Coproduct, :+:, CNil }
+import shapeless.{ :+:, CNil, Coproduct }
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable._
-import scala.util.Success
-import scala.util.{ Failure, Try, Success }
-
+import scala.util.{ Failure, Success, Try }
+import com.typesafe.config.ConfigFactory
 import org.scalatest._
 import pureconfig.ConfigConvert.{ fromString, stringConvert }
+import pureconfig.error.{ KeyNotFoundException, WrongTypeForKeyException }
 
 import scala.concurrent.duration.{ Duration, FiniteDuration }
 
@@ -99,7 +99,7 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
       "i" -> 56,
       "l" -> -88,
       "s" -> "qwerTy").asJava).toConfig)
-
+    
     config.success.value shouldBe FlatConfig(false, -234.234d, -34.34f, -56, 88L, "QWERTY", None)
   }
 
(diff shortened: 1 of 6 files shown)


jcazevedo:
I like the idea of the `improveFailure` method, but I'm worried that it produces confusing messages in some situations. One of them is when you have an user-defined converter (that doesn't use the `improveFailure` method) somewhere in the path of objects you're deserializing:

``` scala
case class Qux(v: Int)
case class Baz(qux: Qux)
case class Bar(baz: Baz)
case class Foo(bar: Bar)

implicit val confconv = new ConfigConvert[Bar] {
  implicit val bazConvert = implicitly[ConfigConvert[Baz]]
  def from(v: ConfigValue) = v match {
    case c: ConfigObject => Try(c.get("baz")).flatMap(bazConvert.from(_)).map(Bar(_))
  }
  def to(bar: Bar) = ???
}

val conf = ConfigFactory.parseMap(Map("bar.baz.qux.u" -> 4).asJava)
val foo = loadConfig[Foo](conf)
// returns Failure(pureconfig.error.KeyNotFoundException: Could not find the key bar.qux.v)
```

The exception that is created misses the `baz` key (it should be `bar.baz.qux.v`).

Another case is when you have objects inside a config list:

``` scala
case class Bar(v: Int)
case class Foo(bar: List[Bar])

  bar = [{
    u: 4
  }, {
    v: 2
  }]
}""")

val foo = loadConfig[Foo](conf)
// returns Failure(pureconfig.error.KeyNotFoundException: Could not find the key bar.v)
```

In this case, it's not the `bar.v` key that is missing, but only the `v` when trying to build an object of type `Bar`.

I don't know how to provide the full path to keys in the errors when interacting with user-defined converters without changing (and possibly complicating) the `ConfigConvert` interface. However, I think it is doable to provide error messages in the form: `"Could not find the key 'v' when trying to convert to type Bar"`. Even though they will miss the full path to the desired key, I think that this kind of messages will be more consistent throughout the whole possible set of HOCON documents.

melrief:
@jcazevedo I see your point. What if we make the `ConfigConvert` constructor private and then provide a method `ConfigConvert.instance` such that we can control the context of the `ConfigConvert`? Basically we could decorate the codec such that the parsing will be left to the user or the generic derivation but all the converters will carry info about they key on which they are operating on. The user won't have to deal with the key and the key in the error will always be the full path.
I honestly don't like the idea of not having the errors in logs and even your proposal of having just the key name is not enough for me. For the sake of usability, I think it's important to inform the user of what's going on with the maximum amount of details possible. This is an easy thing to control, after all.
What do you think, should I do a prototype and see if it works?

melrief:
@jcazevedo I had another read of your comment after my last attempt to fix this but the patch that I'm doing will require even more time.

About your two examples:
- the first one is a problem because custom converters can override the way we control the context. We know that and we can fix this in future by deny the access to the main constructor of `ConfigConvert` so that the `from` method will always have the correct context
- the second one is correct because the first object doesn't have `v` field so the error is perfectly correct imho. Fields are actually referred as `<class>.<field>` so it actually works as expected, no?

I personally think the system proposed in this PR is good enough and I really need to get rid of the NPE and strange errors. I propose to release this PR as it stand and eventually move to a more sophisticated system once we have an idea of how it should look like. I'll personally work on that. What do you think?

jcazevedo:
> - the second one is correct because the first object doesn't have `v` field so the error is perfectly correct imho. Fields are actually referred as `<class>.<field>` so it actually works as expected, no?

I don't agree. Using typesafe's `Config` object, you can't access that field with `conf.getValue("bar.v")`, which is something that you can using the complete keys provided with the other errors. I think we should try to provide paths that are usable by typesafe config directly, or give a bit more context in the error when a path can't be built.

> I personally think the system proposed in this PR is good enough and I really need to get rid of the NPE and strange errors. I propose to release this PR as it stand and eventually move to a more sophisticated system once we have an idea of how it should look like.

Sure, this already is a major improvement over what we have right now.

src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -101,6 +100,7 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
       "l" -> -88,
       "s" -> "qwerTy").asJava).toConfig)
 
+    println(config)
```
jcazevedo: I think you forgot this debug line here.

src/main/scala/pureconfig/config/package.scala
```diff
@@ -0,0 +1,14 @@
+package pureconfig
+
+import com.typesafe.config.{ ConfigObject, ConfigValue }
+import pureconfig.error.KeyNotFoundException
+
+import scala.util.{ Failure, Success, Try }
+
+package object config {
+  private[pureconfig] def lookupKey(key: String, configObject: ConfigObject): Try[ConfigValue] =
```
jcazevedo: This seems to be unused code.

src/test/scala/pureconfig/PureconfSuite.scala
melrief: removed

src/main/scala/pureconfig/config/package.scala
melrief: removed the file

melrief:
@jcazevedo I removed the unused code and rebase in the new master.

jcazevedo:
This looks good to merge as is. We can improve on the other points later.

melrief:
Thanks @jcazevedo for having a look at this, I'm going to publish 0.3.1 then

PR #44, opened 2016-09-26
Added capability to fall back to default configuration is keys are no…

aecc:
This PR adds capability to have fallback to defaults when loading a typesafe configuration


src/main/scala/pureconfig/package.scala
```diff
@@ -76,6 +76,20 @@ package object pureconfig {
   }
 
   /**
+    * Load a configuration of type `Config` from the given `Config`, falling back to the default configuration */
```
melrief: Can I ask you to format the doc as for the other methods and to add a description for the parameters (copy-pasted from the other methods).

melrief:
I like this idea because one thing that we miss in `pureconfig` are methods with a fallback. @jcazevedo @leifwickland what do you think? I know that `pureconfig` exposed API is growing but people are asking for many different ways to load the configuration including falling back to the `application.conf` and `reference.conf`.
One thing to do in future is to clean a bit the API to make it more consistent. I can work on this.

jcazevedo:
I personally prefer calling the `loadConfig` methods with the fallback explicit on the parameter (i.e. `loadConfig[T](conf.withFallback(ConfigFactory.load()))`) because it's more explicit what we're doing with typesafe config, enables configuring the parameters on the `ConfigFactory.load` method and also gives more control to what fallback to use (at our company we use a custom Config loader, for example). Having said that, I'm not against adding these methods for convenience.

melrief:
I think I prefer to have some utility methods because I would like to give the user more flexibility about what they can do. I just wonder if we can put together a minimum but complete set of functions for this API to cover all the common use cases.

src/main/scala/pureconfig/package.scala
```diff
@@ -76,6 +76,20 @@ package object pureconfig {
...
+  def loadConfigWithFallBack[Config](conf: TypesafeConfig)(implicit conv: ConfigConvert[Config]): Try[Config] = {
```
melrief: I also would like to use the same naming of typesafe config. Can we change `FallBack` to `Fallback`?

aecc:
ok, changes were made as you asked @melrief

melrief:
@aecc merged, thanks for the contribution. As contributor, I would like to add you to the [CONTRIBUTORS](https://github.com/melrief/pureconfig/blob/master/CONTRIBUTORS) file if you agree.

aecc:
@melrief it would be my honor :)

melrief:
[Done](https://github.com/melrief/pureconfig/blob/master/CONTRIBUTORS), welcome!

src/main/scala/pureconfig/package.scala
```diff
@@ -76,6 +76,33 @@ package object pureconfig {
   }
 
   /**
+    * Load a configuration of type `Config` from the given `Config`, falling back to the default configuration
```
leifwickland, line 79: Should that second `Config` be `conf`?
Ditto on the method below.

PR #45, opened 2016-09-28
pureconfig-argonaut

mgzuber:
- Modification of build.sbt to allow for a multi-project build
- A new module/project called pureconfig-argonaut allowing for ConfigConvert creation from argonaut Json codcs (`EncodeJson[T]` and `DecodeJson[T]`)


pureconfig-argonaut/src/main/scala/pureconfig/contrib/implicits/package.scala
```diff
@@ -0,0 +1,22 @@
...
+import com.typesafe.config.{ConfigFactory, ConfigRenderOptions, ConfigValue}
+import pureconfig.ConfigConvert
+
+import scala.reflect.ClassTag
+import scala.util.Try
+
+object `package` {
+  implicit def configConvertFromArgonaut[T: EncodeJson : DecodeJson](implicit ct: ClassTag[T]): ConfigConvert[T] = new ConfigConvert[T] {
+    def from(config: ConfigValue): Try[T] = Try(
+      config.render(ConfigRenderOptions.concise())
```
melrief, line 15: Maybe you want to `setJson(true)` in the `ConfigRenderOptions`? I don't think argonaut can parse HOCON and the default should be HOCON for [`render`](https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigValue.html#render--).

pureconfig-argonaut/src/test/scala/pureconfig/PureconfigArgonautTest.scala
```diff
@@ -0,0 +1,16 @@
...
+
+class PureconfigArgonautTest extends FreeSpec {
+  def readConfig: Config = loadConfig[Config].get
+  "Config should be read successfully" in readConfig
+  "Config has good values" in {
+    val config = readConfig
+    assert(config.wrapper.string.s == "Hello pureconfig-argonaut")
+    assert(config.wrapper.int.i == 123)
+    assert(!config.wrapper.bool.b)
+  }
```
melrief, line 15: I would add an identity test that creates a `Config`, convert it to `ConfigValues` with `to`, convert the result back to `Config` with `from` and checks that the values are not changed.

PR #47, opened 2016-09-29
add info about supported types, extending and override in README.md

melrief:
@leifwickland @jcazevedo I decided to extend the README.md with some more info about the library and in particular with info about what is supported, how to add new instances of `ConfigConvert` and how to override an instance of `ConfigConvert`. I made this PR to know what do you think about the changes and if you have input.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,22 @@ A boilerplate-free Scala library for loading configuration files
 
 ![](http://i.imgur.com/S5QUS8c.gif)
 
+## Table of Contents
+- [Why](#why)
+- [Not Yet Another Configuration Library](#not-yet-another-configuration-library)
+- [Add PureConfig to your project](#add-pureconfig-to-your-project)
+- [Use PureConfig](#use-pureconfig)
+- [Supported types](#supported-types)
+- [Extends the library to support new types](#extend-the-library-to-support-new-types)
+- [Override behaviour for types](#override-behaviour-for-types)
+- [Example](#example)
+- [Whence the config files](#whence-the-config-files)
+- [Contribute](#contribute)
+- [License](#license)
+- [Special thanks](#special-thanks)
+
+
+<a name="why" />
 ## Why
 Loading configurations has always been a tedious and error-prone procedure. A common way to do it
 consists in writing code to deserialize each fields of the configuration. The more fields there are,
@@ -21,6 +37,7 @@ The goal of this library is to create at compile-time the boilerplate necessary 
 certain type. In other words, you define **what** to load and PureConfig provides **how** to load it.
 
 
+<a name="not-yet-another-configuration-library" />
 ## Not yet another configuration library
 PureConfig is not a configuration library in the sense that it doesn't search for files or parse them.
 It can be seen as a better front-end for the existing libraries.
@@ -28,6 +45,7 @@ It uses [typesafe config][typesafe-config] library for loading raw configuration
 uses the raw configurations to do its magic.
 
 
+<a name="add-pureconfig-to-your-project" />
 ## Add PureConfig to your project
 
 In the sbt configuration file:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,22 @@ A boilerplate-free Scala library for loading configuration files
...
+- [Extends the library to support new types](#extend-the-library-to-support-new-types)
+- [Override behaviour for types](#override-behaviour-for-types)
+- [Example](#example)
+- [Whence the config files](#whence-the-config-files)
+- [Contribute](#contribute)
+- [License](#license)
+- [Special thanks](#special-thanks)
+
+
+<a name="why" />
```
jcazevedo: You don't need to add these anchors, as GitHub adds them automatically (see https://github.com/melrief/pureconfig/blob/master/README.md#why, for example).

README.md
```diff
@@ -61,6 +79,81 @@ For a full example of `build.sbt` you can have a look at this [build.sbt](https:
...
+
+Currently supported types for fields are:
+- `String`, `Boolean`, `Double` (standard
+and percentage format ending with `%`), `Float` (also supporting percentage),
+`Int`, `Long`, `Short`, `URL`, `Duration`
+- all collections implementing the `TraversableOnce` trait where the elements type
+is in this list
+- `Option` for optional values, i.e. value that can or cannot be in the configuration
+- `Map` with `String` keys and any value type that is in this list
+- typesafe `ConfigValue`, `ConfigObject` and `ConfigList`.
```
jcazevedo: Don't you also want to add case classes and sealed families of case classes to this list?

README.md
```diff
@@ -61,6 +79,81 @@ For a full example of `build.sbt` you can have a look at this [build.sbt](https:
...
+
+
+<a name="extend-the-library-to-support-new-types" />
+## Extend the library to support new types
+
+Not all types are supported automatically by `pureconfig`. For instance, classes
+with mutable fields are not supported out-of-the-box:
```
jcazevedo: I don't think this is accurate. `pureconfig` supports case classes with mutable fields. I think we can refer to classes that are not case classes in this example.

jcazevedo (commented):
This is a great improvement to the project's README file. I left some minor comments inline, but overall this looks great.

README.md
```diff
@@ -61,6 +79,81 @@ For a full example of `build.sbt` you can have a look at this [build.sbt](https:
...
+```
+
+
+<a name="supported-types" />
+## Supported Types
+
+Currently supported types for fields are:
+- `String`, `Boolean`, `Double` (standard
+and percentage format ending with `%`), `Float` (also supporting percentage),
+`Int`, `Long`, `Short`, `URL`, `Duration`
```
leifwickland: Nitpick: `FiniteDuration` is also supported independently of `Duration`.
(Thanks to @jcazevedo for doing that. I needed it the other day!)

README.md
leifwickland: It also supports ADTs (i.e. a sealed family of case classes as @jcazevedo mentioned):
Something like the following works and is awesome:

```
sealed trait MyAdt
case class AdtA(a: String) extends MyAdt
case class AdtB(b: Int) extends MyAdt

case class HasMyAdt(a: MyAdt)

loadConfig[HasMyAdt]()
```

leifwickland (approved):
Thanks for doing this!

README.md
melrief: Added them in 9e86fd4 plus a small example showing all the features

README.md
melrief: Agreed, corrected with 9e86fd4

README.md
melrief: Added with 9e86fd4

melrief:
It should be better now, thanks for the comments.

README.md
melrief: @leifwickland I've stolen your example for adt and used it as base for the example :sweat_smile:

README.md
```diff
@@ -61,6 +76,91 @@ For a full example of `build.sbt` you can have a look at this [build.sbt](https:
...
+
+Currently supported types for fields are:
+- `String`, `Boolean`, `Double` (standard
+and percentage format ending with `%`), `Float` (also supporting percentage),
+`Int`, `Long`, `Short`, `URL`, `Duration`, `FiniteDuration`
+- all collections implementing the `TraversableOnce` trait where the elements type
+is in this list
+- `Option` for optional values, i.e. value that can or cannot be in the configuration
+- `Map` with `String` keys and any value type that is in this list
+- typesafe `ConfigValue`, `ConfigObject` and `ConfigList`.
```
jcazevedo: I'm probably nitpicking, but the period at the end of this line seems a bit off, specially since there are two more bullet points after.

melrief: Well the README is the first thing people read when they open our github page, it must be well written and formatted so nitpicking is right :smile:
Fixed with 68be5b7

jcazevedo:
LGTM

Thanks for doing this!

PR #49, opened 2016-10-04
Improve Travis configuration

jcazevedo:
This PR improves the `.travis.yml` configuration by compiling against Scala 2.11.8 (which is the one defined in `build.sbt`) and also against Scala 2.10.5, which is also defined as a cross Scala version in `build.sbt`.


melrief:
Thanks also for this one, nice to have Scala 2.10 in Travis tests!

PR #51, opened 2016-10-10
Add support for default arguments

jcazevedo:
This PR adds support for default arguments in types derived using shapeless' `LabelledGeneric`. This allows one to load a config missing some values into a type that has those values as default arguments in its constructor:

``` scala
scala> case class Foo(a: Int, b: String = "default")
defined class Foo

scala> loadConfig[Foo](ConfigFactory.parseString("{ a = 2 }"))
res0: scala.util.Try[Foo] = Success(Foo(2,default))
```

### DIFF
--- a/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/src/main/scala/pureconfig/ConfigConvert.scala
@@ -75,11 +75,6 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
     override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(toF(t))
   }
 
-  implicit def hNilConfigConvert = new ConfigConvert[HNil] {
-    override def from(config: ConfigValue): Try[HNil] = Success(HNil)
-    override def to(t: HNil): ConfigValue = ConfigFactory.parseMap(Map().asJava).root()
-  }
-
   private[pureconfig] def improveFailure[Z](result: Try[Z], keyStr: String): Try[Z] =
     result recoverWith {
       case CannotConvertNullException => Failure(KeyNotFoundException(keyStr))
@@ -87,6 +82,11 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
       case WrongTypeException(typ) => Failure(WrongTypeForKeyException(typ, keyStr))
       case WrongTypeForKeyException(typ, suffix) => Failure(WrongTypeForKeyException(typ, keyStr + "." + suffix))
     }
+
+  implicit def hNilConfigConvert = new ConfigConvert[HNil] {
+    override def from(config: ConfigValue): Try[HNil] = Success(HNil)
+    override def to(t: HNil): ConfigValue = ConfigFactory.parseMap(Map().asJava).root()
+  }
 
   implicit def hConsConfigConvert[K <: Symbol, V, T <: HList](
     implicit key: Witness.Aux[K],
--- a/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/src/main/scala/pureconfig/ConfigConvert.scala
@@ -83,23 +83,36 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
       case WrongTypeForKeyException(typ, suffix) => Failure(WrongTypeForKeyException(typ, keyStr + "." + suffix))
     }
 
-  implicit def hNilConfigConvert = new ConfigConvert[HNil] {
-    override def from(config: ConfigValue): Try[HNil] = Success(HNil)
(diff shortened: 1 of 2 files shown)


src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -229,12 +242,13 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
   }
 
   // used for products
-  implicit def deriveInstanceWithLabelledGeneric[F, Repr <: HList](
+  implicit def deriveInstanceWithLabelledGeneric[F, Repr <: HList, DefaultRepr <: HList](
     implicit gen: LabelledGeneric.Aux[F, Repr],
-    cc: Lazy[ConfigConvert[Repr]]): ConfigConvert[F] = new ConfigConvert[F] {
+    default: Default.AsOptions.Aux[F, DefaultRepr],
```
melrief, line 247: This is brilliant @jcazevedo! I learned something new today, thanks.

melrief:
This PR is great, if @leifwickland is fine with it I would like to merge it asap. Thanks @jcazevedo for the great work, I didn't even know it was possible to do something like this!

src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -88,18 +83,36 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
       case WrongTypeForKeyException(typ, suffix) => Failure(WrongTypeForKeyException(typ, keyStr + "." + suffix))
     }
 
-  implicit def hConsConfigConvert[K <: Symbol, V, T <: HList](
+  private[pureconfig] trait DefaultValueConfigConvert[Repr <: HList, DefaultRepr <: HList] extends ConfigConvert[Repr] {
+    def fromWithDefault(config: ConfigValue, default: DefaultRepr): Try[Repr]
+    override def from(config: ConfigValue): Try[Repr] =
+      Failure(
+        new UnsupportedOperationException("Cannot call 'from' from a DefaultValueConfigConvert."))
```
leifwickland: Suggestion: Reword as "...call 'from' on a..." to be slightly less awkward.

src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -561,4 +561,24 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
...
+
+    val conf3 = ConfigFactory.parseMap(Map("c" -> 50).asJava)
+    loadConfig[Conf](conf3).failure.exception shouldEqual KeyNotFoundException("a")
+
+    val conf4 = ConfigFactory.parseMap(Map("a" -> 2, "d.e" -> 5).asJava)
+    loadConfig[Conf](conf4).failure.exception shouldEqual KeyNotFoundException("d.g")
+
+    val conf5 = ConfigFactory.parseMap(Map("a" -> 2, "d.e" -> 5, "d.g" -> 6).asJava)
+    loadConfig[Conf](conf5).success.value shouldBe Conf(2, "default", 42, InnerConf(5, 6))
+  }
```
leifwickland, line 586: I suggest we add an additional test ensuring that the wrong kind of value for a field with a default argument causes conversion to fail with a `WrongTypeForKeyException`.  E.g.: 

```
val conf6 = ConfigFactory.parseMap(Map("a" -> 2, "d" -> "notAnInnerConf").asJava)
loadConfig[Conf](conf6).failure.exception shouldEqual WrongTypeForKeyException("OBJECT", "d")
// I'm not sure what that first argument to the excepiton constructor is supposed to be.
```

leifwickland (approved):
@jcazevedo This is really clever. Excellent work.
I offered a few suggestions for some slight improvements.

src/main/scala/pureconfig/ConfigConvert.scala
jcazevedo: Yeah. It's definitely less awkward. I've fixed it in 0da5155.

src/test/scala/pureconfig/PureconfSuite.scala
jcazevedo: Added that extra test in 008ac89.

leifwickland:
Thanks for the additions, @jcazevedo! 👍 without reservations.

jcazevedo:
The credit for this should go to @alexarchambault, who [added the `Default`](https://github.com/milessabin/shapeless/pull/462) type class to Shapeless.

melrief:
Thanks @jcazevedo, I'm going to release the next version of pureconfig today.

melrief:
@jcazevedo @leifwickland I will add the support for default values in the README in the next days

PR #52, opened 2016-10-17
Improve error messages when using Duration and FiniteDuration converters

jcazevedo:
This PR improves the error messages when using the `Duration` and `FiniteDuration` converters, following the work done in #43.

PR #53, opened 2016-10-25
Fix NPE when converting a `Duration` from nonexistent key.

leifwickland:
Also, improve error messages when converting a `FiniteDuration`.

Without the fix, the assertion on 548 of `PureconfSuite.scala` fails.


jcazevedo:
This is similar to #52. #52 adds one extra type check when attempting to derive `Duration`, as it is never possible to derive it from a `ConfigObject` or `ConfigList`.

melrief:
@leifwickland I reviewed #52 and merged it before reaching this PR. I didn't have the opportunity to see the differences but for what I can read now, #52 already contains this fix. @leifwickland what do you think?

leifwickland:
@melrief Welcome back from holiday!
Sorry I didn't notice that @jcazevedo had already submitted #52. The good news is that our solutions were almost identical. The better news is that his was better than mine. I think we can just close this PR.

PR #55, opened 2016-11-02
Scalacheck should be a test-only dep. Segregate test deps. Fixes #54

### LINKED ISSUE #54: compile dependent on scalacheck?
https://github.com/melrief/pureconfig/blob/master/build.sbt#L32-L33 makes pureconfig `compile` depends on scalacheck. Shouldn't it be only in `test`?

leifwickland:


leifwickland:
@melrief If you don't like the segregation of test dependencies, I'd be happy to revise this to a more traditional encoding.

ngbinh:
lgtm

PR #58, opened 2016-11-04
Support scala 2.12

melrief:
I had to remove the tests using Stack because Stack is deprecated in scala 2.12

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,13 +48,13 @@ uses the raw configurations to do its magic.
 
 In the sbt configuration file:
 
-use scala `2.10` or `2.11`:
+use scala `2.10`, `2.11` or `2.12`:
 
 ```scala
 scalaVersion := "2.11.8" // or "2.10.5"
 ```
 
-Add the library. For scala `2.11`
+Add the library. For scala `2.11` and `2.12`
 
 ```scala
 libraryDependencies ++= Seq(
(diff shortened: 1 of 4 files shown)


jcazevedo:
We should also update the Travis config and the README accordingly.

jcazevedo:
We should also update the Travis config and the README accordingly.

jcazevedo:
LGTM

src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -373,12 +373,6 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
     saveAndLoadIsIdentity(ConfWithQueueOfFoo(Queue(Foo(1), Foo(2))))
   }
 
-  case class ConfWithStackOfFoo(stack: collection.mutable.Stack[Foo])
```
leifwickland, line 376: To preserve a test on a mutable data structure, we could switch to the mutable queue.

```
  case class ConfWithMutableQueueOfFoo(stack: collection.mutable.Queue[Foo])

  it should s"be able to save and load configurations containing mutable.Queue" in {
    saveAndLoadIsIdentity(ConfWithMutableQueueOfFoo(collection.mutable.Queue(Foo(1))))
  }
```

README.md
```diff
@@ -48,13 +48,13 @@ uses the raw configurations to do its magic.
 
 In the sbt configuration file:
 
-use scala `2.10` or `2.11`:
+use scala `2.10`, `2.11` or `2.12`:
 
 ```scala
 scalaVersion := "2.11.8" // or "2.10.5"
```
leifwickland, line 54: Nitpick: 2.12 could appear on this line too.

Alternatively, I think you could probably just drop this section, as it seems:
- A given that a `scalaVersion` need to be specified.
- overspecific insofar as pureconfig doesn't have a dependency on a specific patch version of scala.

leifwickland (commented):
LGTM 👍 
Thanks for doing this!

PR #60, opened 2016-11-15
Prefer Scala's Duration string conversion to Typesafe Config's.

### LINKED ISSUE #59: Prefer Scala's Duration string conversion to Typesafe Config's implementation.
I tried using pureconfig in a Spark 1.x app and discovered that I couldn't parse a `FiniteDuration`. Spark 1.x includes Typesafe Config 1.2.1 , which doesn't include the `Config.getDuration` method. (Pureconfig uses 1.3.1.) At runtime I get an error about that method not existing. Of course the Spark version of typesafe supersedes whatever I ship.

The specific error message is:
```
Exception in thread "main" java.lang.NoSuchMethodError: com.typesafe.config.Config.getDuration(Ljava/lang/String;)Ljava/time/Duration;
        at pureconfig.DurationConvert$$anonfun$from$2.apply(DurationConvert.scala:20)
...
```

Well, in the process of working around that (which kudos to everybody here that I can easily enough work around that) I realized that Scala's `Duration` includes a `def apply(String): Duration` that is (almost) a superset of `SimpleConfig.parseDuration`.

See
https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/impl/SimpleConfig.java#L622
vs
https://github.com/scala/scala/blob/v2.11.8/src/library/scala/concurrent/duration/Duration.scala#L78

The "almost" is that for abbreviating microseconds Scala used a proper _mu_ while typesafe config used a _u_, i.e. "µs" vs "us".

When I originally implemented pureconfig's converter for `Duration` I didn't realize that Scala had its own implementation of something similar. I was shooting for compatibility with Typesafe config.

Pureconfig's `Duration` converter is one of its most complex. It also includes the "config to string" converter I wrote, which Scala's duration includes.

**Proposal**:
- Use Scala's `Duration.apply` instead of `Config.getDuration`
- Maintain compatibility with typesafe config by translating "us" to "µs".
- Discard most of pureconfig's duration converter. It could be replaced by something like ConfigConvert.fromNonEmptyString(Duration.apply(_)). (With the addition of the unit converter mentioned above.)

@melrief @jcazevedo, if you guys are marginally supportive, I'll send a PR for you to consider further.

An interesting side effect of this change is that it'd help #5, support for Java 7. Pureconfig could offer converters for `Duration` and `FiniteDuration` without typesafe config 1.3.x.

leifwickland:
Fixes #59

- The primary intention is to reduce compexity in Pureconfig by making the
  duration `ConfigConvert` instances more similar to the others.
- A secondary benefit is that Pureconfig in a Spark 1.x app should be able
  to parse Duration and FiniteDuration because it no longer depends on a
  newer version of typesafe config than spark ships with.
- Also add a few tests around error messages.
- Also handle infinite durations.

### DIFF
--- a/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/src/main/scala/pureconfig/ConfigConvert.scala
@@ -288,9 +288,13 @@ trait LowPriorityConfigConvertImplicits {
   }
 
   implicit val finiteDurationConfigConvert: ConfigConvert[FiniteDuration] = {
-    val fromString: String => FiniteDuration = (s: String) => DurationConvert.fromString(s, implicitly[ClassTag[FiniteDuration]]).get match {
-      case v if v.isFinite() => Duration(v.length, v.unit)
-      case _ => throw new IllegalArgumentException(s"Couldn't parse '$s' into a finite duration because it's infinite.")
+    val fromString: String => FiniteDuration = { (s: String) =>
+      DurationConvert.fromString(s, implicitly[ClassTag[FiniteDuration]])
+        .flatMap {
+          case d: FiniteDuration => Success(d)
+          case _ => Failure(new IllegalArgumentException(s"Couldn't parse '$s' into a FiniteDuration because it's infinite."))
+        }
+        .get
     }
     nonEmptyStringConvert(fromString, DurationConvert.fromDuration)
   }
--- a/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/src/main/scala/pureconfig/ConfigConvert.scala
@@ -53,13 +53,11 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
       } else {
         // Because we can't trust `fromF` or Typesafe Config not to throw, we wrap the
         // evaluation in one more `Try` to prevent an unintentional exception from escaping.
-        Try {
-          val configValueAsString = config.valueType match {
-            case ConfigValueType.STRING => config.unwrapped.toString
-            case _ => config.render(ConfigRenderOptions.concise)
-          }
-          fromF(configValueAsString)
-        }.flatten
+        // `Try.flatMap(f)` captures any non-fatal exceptions thrown by `f`.
+        Try(config.valueType match {
+          case ConfigValueType.STRING => config.unwrapped.toString
+          case _ => config.render(ConfigRenderOptions.concise)
+        }).flatMap(fromF)
       }
     }
 
--- a/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/src/main/scala/pureconfig/ConfigConvert.scala
@@ -46,39 +46,44 @@ trait ConfigConvert[T] {
 object ConfigConvert extends LowPriorityConfigConvertImplicits {
   def apply[T](implicit conv: ConfigConvert[T]): ConfigConvert[T] = conv
 
-  private def fromFConvert[T](fromF: String => T): ConfigValue => Try[T] =
+  private def fromFConvert[T](fromF: String => Try[T]): ConfigValue => Try[T] =
     config => {
       if (config == null) {
         Failure(CannotConvertNullException)
       } else {
-        Try(fromF(config.valueType() match {
-          case ConfigValueType.STRING => config.unwrapped().toString
-          case _ => config.render(ConfigRenderOptions.concise())
-        }))
+        // Because we can't trust `fromF` or Typesafe Config not to throw, we wrap the
+        // evaluation in one more `Try` to prevent an unintentional exception from escaping.
+        Try {
+          val configValueAsString = config.valueType match {
+            case ConfigValueType.STRING => config.unwrapped.toString
+            case _ => config.render(ConfigRenderOptions.concise)
+          }
+          fromF(configValueAsString)
+        }.flatten
       }
     }
 
-  def fromString[T](fromF: String => T): ConfigConvert[T] = new ConfigConvert[T] {
+  def fromString[T](fromF: String => Try[T]): ConfigConvert[T] = new ConfigConvert[T] {
     override def from(config: ConfigValue): Try[T] = fromFConvert(fromF)(config)
     override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(t)
   }
 
-  def fromNonEmptyString[T](fromF: String => T)(implicit ct: ClassTag[T]): ConfigConvert[T] = {
-    fromString(ensureNonEmpty andThen fromF)
+  def fromNonEmptyString[T](fromF: String => Try[T])(implicit ct: ClassTag[T]): ConfigConvert[T] = {
+    fromString(ensureNonEmpty(ct)(_).flatMap(fromF))
   }
 
-  private def ensureNonEmpty[T](implicit ct: ClassTag[T]): String => String = {
-    case "" => throw new IllegalArgumentException(s"Cannot read a $ct from an empty string.")
-    case x => x
+  private def ensureNonEmpty[T](implicit ct: ClassTag[T]): String => Try[String] = {
+    case "" => Failure(new IllegalArgumentException(s"Cannot read a $ct from an empty string."))
+    case x => Success(x)
   }
 
-  def stringConvert[T](fromF: String => T, toF: T => String): ConfigConvert[T] = new ConfigConvert[T] {
+  def stringConvert[T](fromF: String => Try[T], toF: T => Try[String]): ConfigConvert[T] = new ConfigConvert[T] {
     override def from(config: ConfigValue): Try[T] = fromFConvert(fromF)(config)
-    override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(toF(t))
+    override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(toF(t).get)
   }
 
-  def nonEmptyStringConvert[T](fromF: String => T, toF: T => String)(implicit ct: ClassTag[T]): ConfigConvert[T] = {
-    stringConvert(ensureNonEmpty andThen fromF, toF)
+  def nonEmptyStringConvert[T](fromF: String => Try[T], toF: T => Try[String])(implicit ct: ClassTag[T]): ConfigConvert[T] = {
+    stringConvert(ensureNonEmpty(ct)(_).flatMap(fromF), toF)
   }
 
   private[pureconfig] trait WrappedDefaultValueConfigConvert[Wrapped, SubRepr <: HList, DefaultRepr <: HList] extends ConfigConvert[SubRepr] {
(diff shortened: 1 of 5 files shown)


src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -62,19 +63,24 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
...
   }
 
-  def fromNonEmptyString[T](fromF: String => T)(implicit ct: ClassTag[T]): ConfigConvert[T] = new ConfigConvert[T] {
-    override def from(config: ConfigValue): Try[T] = fromFConvert {
-      case "" => throw new IllegalArgumentException(s"Cannot read a $ct from an empty string.")
-      case x => fromF(x)
-    }(config)
-    override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(t)
+  def fromNonEmptyString[T](fromF: String => T)(implicit ct: ClassTag[T]): ConfigConvert[T] = {
+    fromString(ensureNonEmpty andThen fromF)
```
melrief: I like this

src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -611,6 +612,7 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
     loadConfig[ConfWithConfigList](emptyConf).failure.exception shouldEqual KeyNotFoundException("conf")
     loadConfig[ConfWithDuration](emptyConf).failure.exception shouldEqual KeyNotFoundException("i")
     loadConfig[SparkNetwork](emptyConf).failure.exception shouldEqual KeyNotFoundException("timeout")
+    loadConfig[ConfWithDuration](emptyConf).failure.exception shouldEqual KeyNotFoundException("i")
```
jcazevedo: This line seems repeated from [L613](https://github.com/melrief/pureconfig/pull/60/files#diff-e8e47c5f8227db17829b61b757b0ee0eR613).

src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -277,32 +283,16 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
...
-  implicit val finiteDurationConfigConvert: ConfigConvert[FiniteDuration] = new ConfigConvert[FiniteDuration] {
-    override def from(config: ConfigValue): Try[FiniteDuration] = durationConfigConvert.from(config) match {
-      case Success(v) =>
-        if (v.isFinite())
-          Success(Duration(v.length, v.unit))
-        else
-          Failure(new IllegalArgumentException(s"Couldn't parse a finite duration from a duration that is not finite."))
-      case Failure(f) => Failure(f)
+  implicit val finiteDurationConfigConvert: ConfigConvert[FiniteDuration] = {
+    val fromString: String => FiniteDuration = (s: String) => DurationConvert.fromString(s, implicitly[ClassTag[FiniteDuration]]).get match {
```
melrief: I don't like the `.get` for `Try` here, this can throw an exception and it makes no sense in the context of `ConfigConvert`. Can you use `DurationConvert.fromString().flatMap(/* Success(v) if finite else Failure(IllegalArgumentException(v))*/)`.

leifwickland: The signature of `nonEmptyStringConvert` means I have to throw on failure, right? (I hope I'm not misunderstanding you.)

I could rewrite it as:

```
  implicit val finiteDurationConfigConvert: ConfigConvert[FiniteDuration] = {
    val fromString: String => FiniteDuration = { (s: String) =>
      DurationConvert.fromString(s, implicitly[ClassTag[FiniteDuration]])
        .flatMap { d =>
          if (d.isFinite()) Success(Duration(d.length, d.unit))
          else Failure(new IllegalArgumentException(s"Couldn't parse '$s' into a finite duration because it's infinite."))
        }
        .get
    }
    nonEmptyStringConvert(fromString, DurationConvert.fromDuration)
  }
```

Is that what you were thinking?

src/main/scala/pureconfig/ConfigConvert.scala
leifwickland: Thanks! I did too. :-)

src/test/scala/pureconfig/PureconfSuite.scala
leifwickland: You're correct. Thanks for catching that.

src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -288,9 +288,13 @@ trait LowPriorityConfigConvertImplicits {
...
-    val fromString: String => FiniteDuration = (s: String) => DurationConvert.fromString(s, implicitly[ClassTag[FiniteDuration]]).get match {
-      case v if v.isFinite() => Duration(v.length, v.unit)
-      case _ => throw new IllegalArgumentException(s"Couldn't parse '$s' into a finite duration because it's infinite.")
+    val fromString: String => FiniteDuration = { (s: String) =>
+      DurationConvert.fromString(s, implicitly[ClassTag[FiniteDuration]])
+        .flatMap {
+          case d: FiniteDuration => Success(d)
+          case _ => Failure(new IllegalArgumentException(s"Couldn't parse '$s' into a FiniteDuration because it's infinite."))
+        }
+        .get
```
jcazevedo: This is a bit weird (i.e. unwrapping an `Exception` just to wrap it again later in a `Try`). I think it is better to change the signature of `fromFConvert`, `fromString`, `fromNonEmptyString`, `stringConvert` and `nonEmptyStringConvert` so that the `fromF` parameter becomes `String => Try[T]` instead of `String => T`.

leifwickland: I agree that it's weird. 
I don't relish the idea of making that breaking change to the API. Technically the new methods added in this PR could be defined with `String => Try[T]`, but it'd be odd to introduce inconsistency.
I could see an argument for adding `*Safe` versions of `fromString`, etc (e.g. `def fromStringSafe`) which took a `String => Try[T]`. On the other hand, that means a sizeable expansion of the pureconfig API.
I'm pretty sure if you had both, you'd end up implementing the `*Safe` methods in terms of the existing API because you have to protect against someone who thinks they're catching all the exceptions failing to catch all of them.
In short, I don't really care and can give an economist a run for his money on telling three sides of the story. 
@melrief, Do you have an opinion?

melrief: I agree that we should change the signature to `String => Try[T]`. We could define the methods `String => T` using the safe versions easily. I'm usually ok with breaking the API if it makes sense. Great projects like Circe can change the API over time and so far the result has been a better API with better error handling.

leifwickland: @melrief & @jcazevedo 
Do you guys thing we should have both safe and unsafe versions of the API? If so, do you want to:
1) Keep the existing methods as is and introduce the new safe versions with the `Safe` suffix
2) Rename the existing to have an `Unsafe` suffix and introduce `Safe` method
3) Rename the existing to `Unsafe` and introduce safe versions with the names of the existing methods?

I think my preference among those options is 3, 1, 2.

melrief: I'm fine with 3 but we could avoid having `Unsafe` by just wrapping the `String => T` methods inside a `Try`. I see no point in having the unsafe version of those function to be honest, I would rather prefer exceptions to be trapped in a `Failure`.

leifwickland: @melrief I'm fine skipping the `Unsafe` if you don't see a reason to have them. 

In general the `Unsafe` are more convenient to call for most usages. For example

```
  implicit val readURL = stringConvert[URL](new URL(_), _.toString)
```
becomes
```
  implicit val readURL = stringConvert[URL](x => Try(new URL(x)), x => Try(x.toString))
```

I think there ultimately has be to a `Try` at the bottom of the call stack where all of these methods are called to avoid an unintentional exception from escaping, even if the lambda's type signature would suggest that shouldn't happen.

jcazevedo: Since we end up wrapping everything in a `Try` when the final `ConfigConvert` is built using those helper methods, I'm fine with dropping the unsafe API.

src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -45,34 +46,44 @@ trait ConfigConvert[T] {
...
+    fromString(ensureNonEmpty(ct)(_).flatMap(fromF))
+  }
+
+  private def ensureNonEmpty[T](implicit ct: ClassTag[T]): String => Try[String] = {
+    case "" => Failure(new IllegalArgumentException(s"Cannot read a $ct from an empty string."))
+    case x => Success(x)
   }
 
-  def stringConvert[T](fromF: String => T, toF: T => String): ConfigConvert[T] = new ConfigConvert[T] {
+  def stringConvert[T](fromF: String => Try[T], toF: T => Try[String]): ConfigConvert[T] = new ConfigConvert[T] {
```
jcazevedo: Following `ConfigConvert`'s interface, whose `to` method doesn't return a `Try`, and since we end up calling `get` on this, I think the `toF` could remain typed as `T => String`.

melrief: +1

src/main/scala/pureconfig/ConfigConvert.scala
leifwickland: Updated to remove `Try` on the `T => String` side.

src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -45,34 +46,44 @@ trait ConfigConvert[T] {
...
-        }))
+        // Because we can't trust `fromF` or Typesafe Config not to throw, we wrap the
+        // evaluation in one more `Try` to prevent an unintentional exception from escaping.
+        Try {
+          val configValueAsString = config.valueType match {
+            case ConfigValueType.STRING => config.unwrapped.toString
+            case _ => config.render(ConfigRenderOptions.concise)
+          }
+          fromF(configValueAsString)
+        }.flatten
```
melrief: I would keep the same structure as before and just `.flatMap(fromF)`:

```scala
Try(config.valueType match {
         case ConfigValueType.STRING => config.unwrapped.toString
         case _ => config.render(ConfigRenderOptions.concise)
}).flatMap(fromF)
```

melrief:
Except for a comment, this looks good to me.

src/main/scala/pureconfig/ConfigConvert.scala
leifwickland: Done.

leifwickland:
@melrief Would you be willing to double check my update of `README.md`? There's not a friendly compiler to tell me I'm doing it correctly, without which I'm highly fallible.

melrief:
@leifwickland sure, I'll check it tonight

melrief:
@leifwickland README.md looks ok

leifwickland:
@melrief, thanks for looking.

Are you happy enough with it to merge?

When we go to release after this is merged, we need to remember to bump the
version in a way which indicates a breaking change.

On Wed, Nov 30, 2016 at 10:59 AM, Mario Pastorelli <notifications@github.com

melrief:
Yes, nice job! @jcazevedo what do you think?

jcazevedo:
I think it's good to merge, yes!

melrief:
Thanks, I will publish the next version tomorrow. What version should we use? 0.4?

leifwickland:
@melrief Well semver.org says that you'd bump to 1.x for a breaking API change, but since we're playing in 0.x series, 0.4 seems fine to me.

melrief:
Released `0.4.0`

PR #62, opened 2016-11-20
switch to multi-project

melrief:
@jcazevedo @leifwickland first version of the multi-project structure for pureconfig. It seems to work but I'm not sure about the OSGI plugin nor I've tried to publish core.

I left many options outside the common `settings` because I'm not sure which one would make sense for all the projects. For instance, Java 8 is required by pureconfig because of Lightbend config but maybe other projects don't need it.

For continuity I think we should let `pureconfig-core` keep the name `pureconfig` for the published artifact, while all the modules will have names in the form `pureconfig-<module_name>`.

One thing that it's missing because I'm not yet sure about it is to define dependencies versions at the root level such that all the subproject would be forced to use the same version for the dependencies. In this way you don't have conflicts. I may do this in a nother PR but first I need to do some testing.

Let me know what do you think and what would you change. We don't have to merge this right now and we can discuss a bit on how to do it if you have concerns or you have in mind another way to structure the project.


melrief:
Tests are failing, I think because loading some files failed for some reason. I'm investigating the issue.

core/build.sbt
```diff
@@ -0,0 +1,69 @@
...
+      <developer>
+        <id>melrief</id>
+        <name>Mario Pastorelli</name>
+        <url>https://github.com/melrief</url>
+      </developer>
+    </developers>)
+
+osgiSettings
+
+OsgiKeys.exportPackage := Seq("pureconfig", "pureconfig.syntax", "pureconfig.error")
```
jcazevedo: This is outside the scope of this PR, but I think we can modify this to:

```scala
OsgiKeys.exportPackage := Seq("pureconfig.*")
```

and avoid having to declare every package (which is quite error prone, as we've seen in the past).

jcazevedo:
> For continuity I think we should let `pureconfig-core` keep the name `pureconfig` for the published artifact, while all the modules will have names in the form `pureconfig-<module_name>`.

I agree.

> One thing that it's missing because I'm not yet sure about it is to define dependencies versions at the root level such that all the subproject would be forced to use the same version for the dependencies. In this way you don't have conflicts.

The sbt documentation [suggests](http://www.scala-sbt.org/0.13/docs/Organizing-Build.html#Tracking+dependencies+in+one+place) defining them under a `*.scala` file in the `project` directory so that it becomes part of the build definition. I think I'm OK with that solution.

core/build.sbt
melrief: Done

melrief:
@jcazevedo I've rebased on new master, made the change to osgi that you proposed (good idea btw) and put the dependencies in `project/Dependencies.scala`. When you have sometime can you have a look at this again? Thanks!

@leifwickland do you have sometime to approve the changes? This PR changes completely the project and I would feel more confortable with two more eyes checking it.

leifwickland:
@melrief I'll give it a look by Monday.

build.sbt
```diff
@@ -90,12 +14,22 @@ val newerCompilerLintSwitches = Seq(
...
-
-scalacOptions in (Compile, console) ~= (_ filterNot (Set("-Xfatal-warnings", "-Ywarn-unused-import").contains))
-
-scalacOptions in (Test, console) := (scalacOptions in (Compile, console)).value
-
-scalacOptions ++= PartialFunction.condOpt(CrossVersion.partialVersion(scalaVersion.value)){
+lazy val settings = Seq(
+  scalaVersion := "2.12.0",
+  crossScalaVersions := Seq("2.10.5", "2.11.8", "2.12.0"),
+  scalacOptions ++= Seq("-feature"),
```
leifwickland: Nitpick: I suggest sticking this in `allVersionCompilerLintSwitches` both for tidiness and because its description sounds "lint-y":
`Emit warning and location for usages of features that should be imported explicitly.`

build.sbt
```diff
@@ -90,12 +14,22 @@ val newerCompilerLintSwitches = Seq(
...
+  scalacOptions ++= allVersionCompilerLintSwitches,
+  scalacOptions in (Compile, console) ~= (_ filterNot (Set("-Xfatal-warnings", "-Ywarn-unused-import").contains)),
+  scalacOptions in (Test, console) := (scalacOptions in (Compile, console)).value,
+  scalacOptions ++= PartialFunction.condOpt(CrossVersion.partialVersion(scalaVersion.value)){
     case Some((2, scalaMajor)) if scalaMajor >= 11 => newerCompilerLintSwitches
-}.toList.flatten
+  }.toList.flatten,
+  initialize := {
+    val required = "1.8"
```
leifwickland, line 30: Do you also want to tell `scalac` to target 1.8, i.e. `scalacOptions += "-target:jvm-1.8"`?

core/build.sbt
```diff
@@ -0,0 +1,68 @@
+import Dependencies._
+
+scalariformSettings
+
+name := "pureconfig"
+
+organization := "com.github.melrief"
+
+version := "0.4.0"
```
leifwickland: Should the version be externalized from any one project so that it's less tedious to keep all sub-projects uniform?
Perhaps it ought to be in a `version.sbt` at the root.

build.sbt
```diff
@@ -90,12 +14,22 @@ val newerCompilerLintSwitches = Seq(
...
+    val current  = sys.props("java.specification.version")
+    assert(current == required, s"Unsupported JDK: java.specification.version $current != $required")
+  }
+)
+
+lazy val core = (project in file("core")).
```
leifwickland: Suggestion: Move this to the top of the file to set the precedent that the project are listed up there.

build.sbt
```diff
@@ -1,79 +1,3 @@
...
-
-osgiSettings
-
-OsgiKeys.exportPackage := Seq("pureconfig", "pureconfig.syntax", "pureconfig.error")
-
-OsgiKeys.privatePackage := Seq()
-
-OsgiKeys.importPackage := Seq(s"""scala.*;version="[${scalaBinaryVersion.value}.0,${scalaBinaryVersion.value}.50)"""", "*")
-
 val allVersionCompilerLintSwitches = Seq(
```
leifwickland: Suggestion: For consistency, make this a `lazy val`

core/build.sbt
```diff
@@ -0,0 +1,68 @@
...
+  scalaMacrosParadise,
+  typesafeConfig,
+  scalaTest,
+  joda,
+  jodaConvert,
+  scalaCheck,
+  scalaCheckShapeless
+)
+
+initialize := {
```
leifwickland: I believe this is redundant. I think `settings(settings)` in the root build applies the `initialize` in `settings`.

leifwickland (approved):
Thanks for doing this, @melrief! LGTM 👍

build.sbt
melrief: Done

melrief: No wait, not yet. For some reason it doesn't compile with `jvm-1.8 is not valid for -target`. Let me check how to do it properly.

melrief: Scala 2.10.5 doesn't have `jvm-1.8` as valid target, apparently. I'm investigating to see if I can do it anyway else I'll just skip this change.

melrief: The target `jvm-1.8` [has been added from Scala `2.11.5`](https://issues.scala-lang.org/browse/SI-8966), which means that we cannot use it because we still compile for Scala `2.10.5`

melrief: Reverted the commit with this line

melrief:
There is a strange issue when Travis tries to compile the branch:

```scala
module not found: org.scala-lang.modules#scala-pickling_2.10;0.10.0
```

Not sure why, I've changed nothing related to scala-lang modules.

melrief:
I'm not sure how the error got fixed but it is fixed.

@leifwickland thanks for the review, very useful

@jcazevedo what do you think, shall we merge this?

jcazevedo (approved):
LGTM

melrief:
Thank you guys

Issue #65, opened 2016-12-13
Add support for java.time data structures


melrief:
Add `ConfigConvert` instances for `LocalDate`, `DateTime`...

Most of them needs to be somehow configurable because the string format of a date is not fixed. We could let the user import a template of the instance and the configure it:

```scala
implicit val dateTimeInstance = makeDateTimeInstance(format)
```

leifwickland:
I think this is a great idea. 

Do you think it'd be worth adding support for Joda structures in a separate library?

melrief:
I think for structures in the standard libraries of Java and Scala we should provide support in core so that you need only one libraries to start using it. What do you think?

On a side note, we could support joda-time and that should go in a separate library and be published as `pureconfig-joda`.

leifwickland:
@melrief Yes, I think it makes sense for core Java/Scala date/time types to be in the pureconfig core, in the same way that we support `Duration`/`FiniteDuration` today.

PR #66, opened 2016-12-13
add configurable instances for date, time and datetime

melrief:
@leifwickland @jcazevedo an initial attempt to implement #65. I have implemented configurable instances for java 8's `LocalDate`, `LocalTime` and `LocalDateTime`. What is missing but I would like to do in another PR because I don't have time right now:

1. `TimeZone` and `Period` are not yet covered but they are easy to add because not configurable. I'll do that in another PR
2. `OffsetTime` and `OffsetDateTime` because I never used them. As above, another PR is the right place to put them, I think

I would like to add everything and add a section to the README.md before we release the next version.

### DIFF
--- a/core/src/main/scala/pureconfig/configurable/package.scala
+++ b/core/src/main/scala/pureconfig/configurable/package.scala
@@ -14,7 +14,7 @@ import java.time.format.DateTimeFormatter
  * }}}
  *
  * @example we cannot provide a [[ConfigConvert]] for [[java.time.LocalDate]] because traditionally there are many different
- * [[java.time.format.DateTimeFormatter]] to parse a [[java.time.LocalDate]] from a [[java.lang.String]]. This package
+ * [[java.time.format.DateTimeFormatter]]s to parse a [[java.time.LocalDate]] from a [[java.lang.String]]. This package
  * provides a method that takes an input [[java.time.format.DateTimeFormatter]] and returns a [[ConfigConvert]] for
  * [[java.time.LocalDate]] which will use that [[java.time.format.DateTimeFormatter]] to parse a [[java.time.LocalDate]].
  */
--- a/core/src/main/scala/pureconfig/configurable/package.scala
+++ b/core/src/main/scala/pureconfig/configurable/package.scala
@@ -4,17 +4,31 @@ import scala.util.Try
 import java.time.{ LocalDate, LocalDateTime, LocalTime }
 import java.time.format.DateTimeFormatter
 
+/**
+ * Provides methods that create [[ConfigConvert]] instances from a set of parameters used to configure the instances.
+ *
+ * The result of calling one of the methods can be assigned to an `implicit val` so that `pureconfig` will be able to
+ * use it:
+ * {{{
+ *   implicit val localDateConfigConvert = makeLocalDateConfigConvert(DateTimeFormatter.ISO_TIME)
+ * }}}
+ *
+ * @example we cannot provide a [[ConfigConvert]] for [[java.time.LocalDate]] because traditionally there are many different
+ * [[java.time.format.DateTimeFormatter]] to parse a [[java.time.LocalDate]] from a [[java.lang.String]]. This package
+ * provides a method that takes in input a [[java.time.format.DateTimeFormatter]] and returns a [[ConfigConvert]] for
+ * [[java.time.LocalDate]] which will use that [[java.time.format.DateTimeFormatter]] to parse a [[java.time.LocalDate]].
+ */
 package object configurable {
 
-  def makeLocalDateInstance(formatter: DateTimeFormatter): ConfigConvert[LocalDate] =
+  def makeLocalDateConfigConvert(formatter: DateTimeFormatter): ConfigConvert[LocalDate] =
     ConfigConvert.nonEmptyStringConvert[LocalDate](
       s => Try(LocalDate.parse(s, formatter)), _.format(formatter))
 
-  def makeLocalTimeInstance(formatter: DateTimeFormatter): ConfigConvert[LocalTime] =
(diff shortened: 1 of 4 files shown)


jcazevedo:
This looks good to me! 

We should probably try to normalize the method names of some of the helpers we provide, though. Right now we have `fromString`, `fromNonEmptyString`, `stringConvert` and `nonEmptyStringConvert` as helpers to build `ConfigConvert` from `String`; some of the implicit `ConfigConvert` we provide are named `read<type>` while others are named `<type>ConfigConvert` and this `make<type>Instance` seems to introduce another naming style.

melrief:
@jcazevedo you are right and I would prefer to use the `ConfigConvert` but I don't have a strong opinion about it. What do you prefer?

jcazevedo:
I don't have a strong preference towards a specific alternative. I think I also like the `<type>ConfigConvert` names, so that we would have `stringConfigConvert`, `intConfigConvert`, `localDateConfigConvert(...)`, etc.

leifwickland:
I'll second the idea of using `localDateConfigConvert`, etc for the naming convention.

core/src/main/scala/pureconfig/configurable/package.scala
```diff
@@ -0,0 +1,20 @@
+package pureconfig
+
+import scala.util.Try
+import java.time.{ LocalDate, LocalDateTime, LocalTime }
+import java.time.format.DateTimeFormatter
+
+package object configurable {
```
leifwickland, line 21: I aspire to add a `pureconfig-joda` JAR once this gets merged. 
@melrief Would you expect that the similar methods in that JAR would go in a `pureconfig.joda.configurable` package or would you have them load them into this namespace?
I ask more from a philosophical standpoint of how we want to structure the library's API. I don't really have a good answer to that question.

leifwickland, line 21: Do you want to add a `/** ... */` comment declaring the purpose of this package? At first glance, it'd seem to be limited to date/time types, but I suspect you intend it more broadly than that.

leifwickland (approved):
👍 
Thanks for doing this, @melrief!
Are you planning to add a README section about this?

core/src/main/scala/pureconfig/configurable/package.scala
melrief: I think the behavior of modules should mimic the one of core as much as possible, so yes, for me it should go in `pureconfig.joda.configurable`.

melrief: We'll do

melrief: Yes, I'll do it

melrief:
@leifwickland @jcazevedo because there are some conflicts with master, I'll rebase this branch on master and force push it.

core/src/main/scala/pureconfig/configurable/package.scala
```diff
@@ -4,17 +4,31 @@ import scala.util.Try
...
+ *
+ * The result of calling one of the methods can be assigned to an `implicit val` so that `pureconfig` will be able to
+ * use it:
+ * {{{
+ *   implicit val localDateConfigConvert = makeLocalDateConfigConvert(DateTimeFormatter.ISO_TIME)
+ * }}}
+ *
+ * @example we cannot provide a [[ConfigConvert]] for [[java.time.LocalDate]] because traditionally there are many different
+ * [[java.time.format.DateTimeFormatter]] to parse a [[java.time.LocalDate]] from a [[java.lang.String]]. This package
+ * provides a method that takes in input a [[java.time.format.DateTimeFormatter]] and returns a [[ConfigConvert]] for
```
leifwickland: Nitpick: I think 
"... takes an input [[java.time.format.DateTimeFormatter]] ..."
would read more naturally.

core/src/main/scala/pureconfig/configurable/package.scala
```diff
@@ -4,17 +4,31 @@ import scala.util.Try
 import java.time.{ LocalDate, LocalDateTime, LocalTime }
 import java.time.format.DateTimeFormatter
 
+/**
+ * Provides methods that create [[ConfigConvert]] instances from a set of parameters used to configure the instances.
+ *
+ * The result of calling one of the methods can be assigned to an `implicit val` so that `pureconfig` will be able to
+ * use it:
+ * {{{
+ *   implicit val localDateConfigConvert = makeLocalDateConfigConvert(DateTimeFormatter.ISO_TIME)
```
leifwickland, line 13: I think it might be useful to try to explain where the user of the library needs to include this line of code, as is always the case with Scala's implicits.

melrief:
@leifwickland I've added a section of the README.md, you can see it [here](https://github.com/melrief/pureconfig/tree/add_local_datetime_instances#configurable-converters). I wasn't sure how to explain it properly, let me know if it's understandable.

leifwickland:
@melrief I think your addition to the README makes sense. I can't think of a way to improve it beside the nitpick I mentioned. 👍

melrief:
@jcazevedo you had a look at his PR before, what do you think of my changes?

jcazevedo (approved):
I think this is good to merge.

I'm not particularly fond of having these factory methods prefixed with `make`, but I'm OK with it as long as we're consistent with this naming convention in future configurable instances.

melrief:
@jcazevedo we can remove the `make` prefix for configurable instances, I'm not particularly fond of the prefix anyway. @leifwickland @ruippeixotog what do you think about it? I prefer to decide now for the future instead of having to refactor the configurable instances another time.

ruippeixotog:
I support the `<type>ConfigConvert` naming 👍

leifwickland:
I'm all for removing the `make` prefix.

melrief:
Removed, thanks guys! Waiting for CI to run the tests and then I think we can merge.

PR #69, opened 2016-12-17
Add CoproductHint for customization of sealed family conversions

### LINKED ISSUE #13: PureConfig fails to load coproducts when two entities have the same number of parameters
Minimal example

``` scala
> sealed trait A
> case object B extends A
> case object C extends A
> pureconfig.loadConfig[A](Map("foo"->"A"), "foo")
res4: util.Try[A] = Success(B)
```

The problem is that PureConfig is ignoring the labels which are important for coproducts.

ruippeixotog:
This pull request adds a new `CoproductHint` trait that allows easy customization of the serialization of sealed families. The general idea was taken from [spray-json-shapeless](https://github.com/fommil/spray-json-shapeless), with a handful of improvements and tweaks to better match the API patterns in pureconfig.

A `CoproductHint` decides, for a given `Config` and the name of a coproduct option (such as the name of a subclass of the sealed trait), if the received config should be deserialized as an instance of the specific option. In most cases, it uses disambiguation information, such as an extra `type` field, to encode and decode the concrete instance to create.

Two implementations of `CoproductHint` are added. The most common in my view, and the default in spray-json-shapeless, is a `FieldCoproductHint`, which writes configs with an extra field (`type` by default) and reads it to instantiate concrete models. The name of the field and the mapping between class names and field values can be customized as described in the Scaladoc. I also added `FirstSuccessCoproductHint`, which doesn't encode any extra information and defines the current pureconfig behavior.

I opted to change the default serialization of sealed families to use `FieldCoproductHint`. It's a breaking change, but the current behavior is arguably flawed (in fact, this fixes #13). People relying on the old behavior can easily put in scope an instance of `FirstSuccessCoproductHint`, though.

With this change, pureconfig stopped supporting pure `Coproduct` instances, as it now requires coproducts to be `LabelledGeneric` in order to retrieve the names of the various options. I don't think it is common for people to use `Coproducts`, particularly in configs; however, if you think otherwise I think I can rewrite this to keep support for non-labelled coproducts (at the cost of having some code duplicated, I guess).

### DIFF
--- a/core/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/core/src/main/scala/pureconfig/ConfigConvert.scala
@@ -84,7 +84,9 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
     stringConvert(ensureNonEmpty(ct)(_).flatMap(fromF), toF)
   }
 
-  private[pureconfig] trait WrappedDefaultValueConfigConvert[Wrapped, SubRepr <: HList, DefaultRepr <: HList] extends ConfigConvert[SubRepr] {
+  private[pureconfig] trait WrappedConfigConvert[Wrapped, SubRepr] extends ConfigConvert[SubRepr]
+
+  private[pureconfig] trait WrappedDefaultValueConfigConvert[Wrapped, SubRepr <: HList, DefaultRepr <: HList] extends WrappedConfigConvert[Wrapped, SubRepr] {
     final def from(config: ConfigValue): Try[SubRepr] =
       Failure(
         new UnsupportedOperationException("Cannot call 'from' on a WrappedDefaultValueConfigConvert."))
@@ -155,27 +157,42 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
   case class NoValidCoproductChoiceFound(config: ConfigValue)
     extends RuntimeException(s"No valid coproduct type choice found for configuration $config.")
 
-  implicit def cNilConfigConvert: ConfigConvert[CNil] = new ConfigConvert[CNil] {
+  implicit def cNilConfigConvert[Wrapped]: WrappedConfigConvert[Wrapped, CNil] = new WrappedConfigConvert[Wrapped, CNil] {
     override def from(config: ConfigValue): Try[CNil] =
       Failure(NoValidCoproductChoiceFound(config))
 
     override def to(t: CNil): ConfigValue = ConfigFactory.parseMap(Map().asJava).root()
   }
 
-  implicit def coproductConfigConvert[V, T <: Coproduct](
-    implicit vFieldConvert: Lazy[ConfigConvert[V]],
-    tConfigConvert: Lazy[ConfigConvert[T]]): ConfigConvert[V :+: T] =
-    new ConfigConvert[V :+: T] {
+  implicit def coproductConfigConvert[Wrapped, Name <: Symbol, V, T <: Coproduct](
+    implicit coproductHint: CoproductHint[Wrapped],
+    vName: Witness.Aux[Name],
+    vFieldConvert: Lazy[ConfigConvert[V]],
+    tConfigConvert: Lazy[WrappedConfigConvert[Wrapped, T]]): WrappedConfigConvert[Wrapped, FieldType[Name, V] :+: T] =
+    new WrappedConfigConvert[Wrapped, FieldType[Name, V]:+: T] {
 
-      override def from(config: ConfigValue): Try[V :+: T] = {
-        vFieldConvert.value.from(config)
-          .map(s => Inl[V, T](s))
-          .orElse(tConfigConvert.value.from(config).map(s => Inr[V, T](s)))
+      override def from(config: ConfigValue): Try[FieldType[Name, V] :+: T] = {
+        coproductHint.from(config, vName.value.name) match {
+          case Success(Some(hintConfig)) =>
+            vFieldConvert.value.from(hintConfig) match {
+              case Failure(_) if coproductHint.tryNextOnFail(vName.value.name) =>
+                tConfigConvert.value.from(config).map(s => Inr(s))
+
+              case vTry => vTry.map(v => Inl(field[Name](v)))
+            }
+
+          case Success(None) => tConfigConvert.value.from(config).map(s => Inr(s))
+          case Failure(ex) => Failure(ex)
+        }
       }
 
-      override def to(t: V :+: T): ConfigValue = t match {
-        case Inl(l) => vFieldConvert.value.to(l)
-        case Inr(r) => tConfigConvert.value.to(r)
+      override def to(t: FieldType[Name, V] :+: T): ConfigValue = t match {
+        case Inl(l) =>
+          // Writing a coproduct to a config can fail. Is it worth it to make `to` return a `Try`?
+          coproductHint.to(vFieldConvert.value.to(l), vName.value.name).get
+
+        case Inr(r) =>
+          tConfigConvert.value.to(r)
       }
     }
 
@@ -254,7 +271,7 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
   }
 
   // used for products
-  implicit def deriveInstanceWithLabelledGeneric[F, Repr <: HList, DefaultRepr <: HList](
+  implicit def deriveProductInstance[F, Repr <: HList, DefaultRepr <: HList](
     implicit gen: LabelledGeneric.Aux[F, Repr],
     default: Default.AsOptions.Aux[F, DefaultRepr],
     cc: Lazy[WrappedDefaultValueConfigConvert[F, Repr, DefaultRepr]]): ConfigConvert[F] = new ConfigConvert[F] {
@@ -269,9 +286,9 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
   }
 
   // used for coproducts
-  implicit def deriveInstanceWithGeneric[F, Repr <: Coproduct](
-    implicit gen: Generic.Aux[F, Repr],
-    cc: Lazy[ConfigConvert[Repr]]): ConfigConvert[F] = new ConfigConvert[F] {
+  implicit def deriveCoproductInstance[F, Repr <: Coproduct](
+    implicit gen: LabelledGeneric.Aux[F, Repr],
+    cc: Lazy[WrappedConfigConvert[F, Repr]]): ConfigConvert[F] = new ConfigConvert[F] {
     override def from(config: ConfigValue): Try[F] = {
       cc.value.from(config).map(gen.from)
     }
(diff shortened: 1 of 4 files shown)


melrief:
Ehi @ruippeixotog, thanks for the PR! For what I can see, this is a great improvement over the previous flawed implementation and a step in the right direction on how to handle coproducts. I'll need some days to review the patch properly.

ruippeixotog:
Sure, no problem!

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -155,27 +157,42 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
...
-    tConfigConvert: Lazy[ConfigConvert[T]]): ConfigConvert[V :+: T] =
-    new ConfigConvert[V :+: T] {
+  implicit def coproductConfigConvert[Wrapped, Name <: Symbol, V, T <: Coproduct](
+    implicit coproductHint: CoproductHint[Wrapped],
+    vName: Witness.Aux[Name],
+    vFieldConvert: Lazy[ConfigConvert[V]],
+    tConfigConvert: Lazy[WrappedConfigConvert[Wrapped, T]]): WrappedConfigConvert[Wrapped, FieldType[Name, V] :+: T] =
+    new WrappedConfigConvert[Wrapped, FieldType[Name, V]:+: T] {
+
+      override def from(config: ConfigValue): Try[FieldType[Name, V] :+: T] = {
```
jcazevedo: I think it's better to return a `Failure(CannotConvertNullException)` here if `config` is `null`, otherwise this can throw null pointer exceptions when there's a recursive structure using coproducts. For example:

```scala
sealed trait AnimalConfig
case class DogConfig(age: Int) extends AnimalConfig
case class CatConfig(age: Int) extends AnimalConfig
case class BirdConfig(canFly: Boolean) extends AnimalConfig
case class InnerAnimalConfig(animal: AnimalConfig) extends AnimalConfig

  type: inneranimalconfig
}""")

pureconfig.loadConfig[AnimalConfig](conf)
// throws java.lang.NullPointerException
```

core/src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -227,6 +225,51 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
...
+      implicit val hint = new FirstSuccessCoproductHint[AnimalConfig]
+
+      val conf = ConfigFactory.parseString("{ canFly = true }")
+      loadConfig[AnimalConfig](conf) should be(Success(BirdConfig(true)))
+
+      saveConfigAsPropertyFile[AnimalConfig](DogConfig(2), configFile, overrideOutputPath = true)
+      loadConfig[TypesafeConfig](configFile).map(_.hasPath("type")) should be(Success(false))
+    }
+  }
+
```
jcazevedo, line 270: Can you also add some tests where the implicit coproduct derivation fails (namely when `CollidingKeysException` is returned and the field of `FieldCoproductHint` is missing)?

jcazevedo (changes requested):
This is great, @ruippeixotog! Thanks!

core/src/test/scala/pureconfig/PureconfSuite.scala
ruippeixotog: Makes sense, I have just added them in 34f4cfb.

core/src/main/scala/pureconfig/ConfigConvert.scala
ruippeixotog: Do you think we should give the "power" to `CoproductHint` instances to read nulls and possibly accept them as some coproduct option, or do we disallow them right at the `ConfigConvert`? The use case could be choosing a coproduct option based on the presence of a field. The drawback is the extra validation every `CoproductHint` will have to do.

jcazevedo: I think a missing key should be handled consistently across the implicit derivations PureConfig provides. If we allow `CoproductHint` instances to read nulls, a missing key that points to a coproduct type can be handled differently by the coproduct `ConfigConvert`, and I think that decision should be left to the enclosing object's `ConfigConvert`. For instance, if I have:

```scala
case class Conf(animal: AnimalConfig)
pureconfig.loadConfig[Conf](ConfigFactory.empty)
```

I'm expecting to see a `KeyNotFoundException("animal")`, regardless of the `ConfigConvert` that is being implicitly provided for `AnimalConfig`. If the key is explicitly set to `null` in the configuration file, then we're operating on a `ConfigNull` and we can have a special treatment for it in the `CoproductHint`.

jcazevedo: I think that inner `ConfigConverts` shouldn't even have to know that `null` refers to a missing key and deal with it. I've opened #70 for your consideration, as I think it makes the handling of missing keys more consistent throughout PureConfig.

ruippeixotog: It's a design choice IMO, to allow or not allow `ConfigConvert` to read missing configs and return constants / default values. I think I agree with your view on #70 - I prefer to have nulls handled automatically for me (particularly when I define my own `ConfigConvert`), as you can always provide a custom `ConfigConvert` for the enclosing class in the rare cases you need custom null handling.

ruippeixotog: Added null handling in cb1435d.

jcazevedo (approved):

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -155,27 +157,44 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
...
+          case Failure(ex) => Failure(ex)
+        }
       }
 
-      override def to(t: V :+: T): ConfigValue = t match {
-        case Inl(l) => vFieldConvert.value.to(l)
-        case Inr(r) => tConfigConvert.value.to(r)
+      override def to(t: FieldType[Name, V] :+: T): ConfigValue = t match {
+        case Inl(l) =>
+          // Writing a coproduct to a config can fail. Is it worth it to make `to` return a `Try`?
```
melrief, line 193: I wonder this too. I always had the idea that `to` has to be total but I see your point here. Using `get` makes sense if you want to throw exception but personally I prefer to wrap `to` into `Try` and just let the user know that `to` can result in a `Failure`. Method signatures are good communication mechanism in particular when they point to possible failures.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -254,7 +273,7 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
   }
 
   // used for products
-  implicit def deriveInstanceWithLabelledGeneric[F, Repr <: HList, DefaultRepr <: HList](
+  implicit def deriveProductInstance[F, Repr <: HList, DefaultRepr <: HList](
```
melrief, line 276: I appreciate the new names, they make sense, but I would have prefer to have the change in another PR :sweat_smile: . I think for this time is fine but I'll try to keep PRs focused on one thing in future.

core/src/main/scala/pureconfig/CoproductHint.scala
```diff
@@ -0,0 +1,108 @@
...
+   *
+   * If `cv` is a config for the given class name, this method returns `Success(Some(v))`, where `v` is the config
+   * related to the specific class (possibly the same as `cv`). If it determines that `cv` is a config for a different
+   * class, it returns `Success(None)`. If `cv` is missing information for disambiguation or has a wrong type, a
+   * `Failure` is returned.
+   *
+   * @param cv the `ConfigValue` of the sealed family
+   * @param name the name of the class or coproduct option to try
+   * @return a `Try[Option[ConfigValue]]` as defined above.
+   */
```
melrief, line 28: Good documentation is good, thanks for this.

melrief (approved):
Only small comments and the PR looks really good and it's very well documented. I did a test and it works well too. Thanks for this! @leifwickland both me and @jcazevedo approve this PR, I think two approves are enough for merging but of course I won't do this without knowing if that's ok for you. Just let us know if you want to review the PR.

core/src/main/scala/pureconfig/ConfigConvert.scala
ruippeixotog: Yeah, I understand your preference and I also have that personal policy :) In this case I think you really had to change something as otherwise you'd end up with two methods, `deriveInstanceWithLabelledGeneric` and `deriveInstanceWithGeneric`, both using now `LabelledGeneric`. But if you can think of other, less "disruptive" names I can change it without a problem.

melrief: No, don't worry, for this one is fine.

core/src/main/scala/pureconfig/ConfigConvert.scala
ruippeixotog: I still think that `to` should be total, as you. The kind of errors that can make `to` fail are errors that should be able to be checked in compile time or at least when the `ConfigConvert` is instantiated.

The only reason I made this a `Try` is in [CoproductHint.scala#L86](https://github.com/ruippeixotog/pureconfig/blob/cb1435dd4f328973f8cb563e5f18fd29e596b435/core/src/main/scala/pureconfig/CoproductHint.scala#L86). When this happens, _all_ instances of the sealed family fail to serialize, not just part of them. Technically, pureconfig should be able to detect a conflicting key when deriving the `ConfigConvert` instance. It can be improved in a future pull request, I think.

It may be that there are other type of errors I am not thinking of that can cause `to` to fail, but right now we do not have such cases. Conceptually, it seems to me that the function is total - every instance should have a representing config.

melrief: Agreed

leifwickland:
@melrief I approve the PR. Sorry for the slow turnaround.
@ruippeixotog Thanks for doing this! After it's merged please consider sending an update for the README explaining how this feature works.

melrief:
@ruippeixotog thanks for the contribution(s)! I'll publish the next version of pureconfig `0.5.0` with this change and the fix to `FiniteDuration` tonight. @jcazevedo @leifwickland this is a breaking change and I think a minor version incremental must be done.

In the meanwhile, I would like to ask you if I can add you to the [CONTRIBUTORS](https://github.com/melrief/pureconfig/blob/master/CONTRIBUTORS) file and, if you agree, what email can I write.

I thought about this PR a lot and I think your contributions to the very core of `pureconfig` qualifies you as author of the library. If you want to continue to develop this library together with me, @leifwickland and @jcazevedo, I would like to promote you to author of the library. You did a great work with this PR!

ruippeixotog:
Sure, no problem at all - you can use ruippeixotog _at_ gmail.com. And thank you for the recognition and the opportunity too, I'd like to keep contributing to pureconfig! This turned out to be a very useful library for me, so I must also mention the great work done until now!

PR #70, opened 2016-12-19
Consistently handle keys not found

jcazevedo:
This PR adds a more consistent handling of missing keys, avoiding `ConfigConverts` to have to deal with `null` in cases where a key is not found. This also disallows custom `ConfigConvert` implementations to operate on missing keys, as I think it is error-prone. For example, the following was previously allowed:

```scala
case class Bar(v: Int)
case class Foo(bar: Bar)

implicit val conv = new ConfigConvert[Bar] {
  def from(cv: ConfigValue) = Success(Bar(4))
  def to(b: Bar) = ???
}

pureconfig.loadConfig[Foo](ConfigFactory.empty())
// Would return Success(Foo(Bar(4)))
```

In this case I was expecting a `KeyNotFoundException("bar")`, regardless of the implementation of `ConfigConvert[Bar]`.


melrief:
I like the PR but I have a question: do you think nobody will ever want to create a `ConfigConvert` that accepts also missing keys? This PR removes the possibility and I'm not sure this is what we want. `Option` is not a special data type and I can think about other datatypes for which "key not found" would result into one of the possible values. Do we want to have only `Option` as special datatype that can be encoded/decoded from/to `null` and not provide this possibility to the user?

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -122,11 +116,14 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
     override def fromConfigObject(co: ConfigObject, default: Option[V] :: U): Try[FieldType[K, V] :: T] = {
       val keyStr = mapping(key.value.toString().tail)
       for {
-        v <- improveFailure(
-          vFieldConvert.value.from(co.get(keyStr)) match {
-            case Failure(CannotConvertNullException) =>
+        v <- improveFailure[V](
+          (co.get(keyStr), vFieldConvert.value) match {
+            case (null, converter: OptionConfigConvert[_]) =>
```
leifwickland: It seems like the more open approach to this would be to introduce a trait, e.g. `AllowMissingConfigConvert` or `AllowNullConfigConvert`, which is used for the type check on `ConfigConvert.scala:121`.
Pureconfig would ship with `OptionConfigConvert` as the only child of that trait.

leifwickland (approved):
In general I like the direction you're heading with this. Thanks for the PR!

core/src/main/scala/pureconfig/ConfigConvert.scala
melrief: I like this idea, what do you think @jcazevedo ?
FYI an alternative could be to break the API again and pass an `Option[ConfigValue]` to `.from` to help the user understand that a key could be not found. I'm not sure about this approach because in most of the cases I expect a missing key to be an error, but it's a possibility.

jcazevedo: I like that idea, and I've implemented it in 971c2c7. I named the trait `AllowMissingKey`, though. I don't think it is necessary to put `ConfigConvert` in the name, as it will already be mixed into one, and I think `AllowMissingKey` is clearer than `AllowMissing` or `AllowNull`, but I'm open to change it.

jcazevedo: Even though handling `null` isn't ideal, I don't like having the `from` method receiving an `Option[ConfigValue]`, as most of the time custom implementations will only be handling `Some` values, but would still have to do something about `None`. 

One alternative would be for `ConfigConvert` to have a `onMissingKey: Option[T]` method that would be `None` by default. That would allow customizing the behavior for missing keys while not breaking the API and not having the `from` method called with `null`.

README.md
```diff
@@ -209,7 +209,36 @@ always read lower case. We can do:
 util.Try[String] = Success(foobar)
 ```
 
+## Handling missing keys
+
+The default behavior of `ConfigConverts` that are derived in PureConfig is to
```
leifwickland: Nitpick: I'd move the "s" outside the backticks, i.e. `ConfigConvert`s.
Ditto below in a few places.
Feel free to ignore.

README.md
```diff
@@ -209,7 +209,36 @@ always read lower case. We can do:
 util.Try[String] = Success(foobar)
 ```
 
+## Handling missing keys
```
leifwickland, line 212: Kudos on the excellent documentation.

leifwickland (approved):
Thanks for doing this, @jcazevedo!

README.md
jcazevedo: Fixed in 156ea57.

melrief:
I like this a lot @jcazevedo, thanks for the work. Imho ready to merge

melrief (approved):

melrief:
@ruippeixotog me and @leifwickland have approved the PR, if you can have a look and you agree, we can merge this.

ruippeixotog (approved):
LGTM too. I'm not particularly fond of magic marker traits to customize the core library behavior, but it is a better approach than having special treatment for `OptionConfigConvert`. Other than that, it's very nice to have `ConfigConvert` implementations not having to worry with missing configs!

melrief:
@jcazevedo do we want to release `0.5.1` right now with only this change or can it wait until #66 is merged?

jcazevedo:
Let's wait until #66 is merged. I'll have a look at it later today.

PR #72, opened 2017-01-02
Support loading 0 without units as a Duration and FiniteDuration

### LINKED ISSUE #71: "0" is not a valid FiniteDuration anymore
On pureconfig 0.3.x, "0" was accepted as a valid duration. It's the only case where the time unit does not matter, and Akka frequently uses that value in their configs. However, in pureconfig 0.4.0 "0" throws a parsing error.

Can you add again that special case? As a motivating use case, I frequently use values from akka-actor in my refefence.conf as defaults for my own domain configs.

jcazevedo:
This PR adds support for loading "0" (without units) as a `Duration` and `FiniteDuration`, something that was supported in PureConfig 0.3.x but throws a parsing error in 0.4.0, since it is not supported by `scala.concurrent.duration.Duration`'s apply method.

This fixes #71.

### DIFF
--- a/core/src/main/scala/pureconfig/DurationConvert.scala
+++ b/core/src/main/scala/pureconfig/DurationConvert.scala
@@ -12,7 +12,7 @@ private[pureconfig] object DurationConvert {
    * Convert a string to a Duration while trying to maintain compatibility with Typesafe's abbreviations.
    */
   def fromString[D](durationString: String, ct: ClassTag[D]): Try[Duration] = {
-    Try { Duration(justAMinute(itsGreekToMe(durationString))) }
+    Try { Duration(addZeroUnit(justAMinute(itsGreekToMe(durationString)))) }
       .recoverWith {
         case ex: NumberFormatException =>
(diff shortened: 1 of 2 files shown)


melrief (approved):
Nice fix @jcazevedo, thanks. This makes me think that maybe we should introduce more scalacheck tests, at least for basic types that we support. The test that you introduced could be refactored to be generator driven where the code you wrote in the forAlls would be in the generator.

Anyway, this looks correct to me and I'll merge it to master. Tests refactoring can wait another PR. Should we release a new version of pureconfig just with this fix or should we wait for the coproduct PR to be included? I honestly would like to have the coproduct PR included in the next release.

jcazevedo:
I agree. We should add some more scalacheck tests.

Regarding the publishing of a new version, I think it's OK to wait for the coproduct PR.

leifwickland:
This was my fault. Sorry about introducing the bug. Thanks for fixing it, @jcazevedo.

PR #74, opened 2017-01-05
Explain how to use coproduct hints in README

ruippeixotog:
This pull request adds a mini-tutorial for using `CoproductHints` to README.

I'm not sure of the best section to put this explanation. "Override behaviour for types" seemed the best section of the existing ones, but we can also have a dedicated section for this. Let me know your thoughts!

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ A boilerplate-free Scala library for loading configuration files
 - [Customizing naming conventions](#customizing-naming-conventions)
 - [Extends the library to support new types](#extend-the-library-to-support-new-types)
 - [Override behaviour for types](#override-behaviour-for-types)
+- [Override behaviour for sealed families](#override-behaviour-for-sealed-families)
 - [Example](#example)
 - [Whence the config files](#whence-the-config-files)
 - [Contribute](#contribute)
@@ -208,6 +209,8 @@ always read lower case. We can do:
 > ConfigConvert[String].from(ConfigValueFactory.fromAnyRef("FooBar"))
 util.Try[String] = Success(foobar)
 ```
+
+## Override behaviour for sealed families
 
 In order for `pureconfig` to disambiguate between different options of a sealed
 family of case classes, it must read and write additional information in
--- a/README.md
+++ b/README.md
@@ -209,7 +209,85 @@ always read lower case. We can do:
 util.Try[String] = Success(foobar)
 ```
 
+In order for `pureconfig` to disambiguate between different options of a sealed
+family of case classes, it must read and write additional information in
+configurations. By default it uses the additional field `type`, encoding the
+concrete class represented in the configuration:
 
+```scala
+sealed trait AnimalConf
+case class DogConf(age: Int) extends AnimalConf
+case class BirdConf(canFly: Boolean) extends AnimalConf
+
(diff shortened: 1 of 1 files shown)


jcazevedo:
I think it makes sense to isolate this mini-tutorial, either in a subsection of "Override behaviour for types" or in a section of its own.

leifwickland:
I concur with @jcazevedo about putting the mini-tutorial either in a subsection of "Override behaviour for types" or in a section of its own.

melrief:
I propose a section just for this, @ruippeixotog what do you think? The content is good imho.

melrief (changes requested):
I propose to create a section just for this. It will be easy to link in future for questions.

README.md
```diff
@@ -209,7 +209,85 @@ always read lower case. We can do:
...
+loadConfig[AnimalConf](parseString("""{ type: "Bird", canFly: true }"""))
+// returns Success(BirdConf(true))
+```
+
+With a `CoproductHint` you can even opt not to use any extra field at all. For
+example, if you encode enumerations using sealed traits, you can just write the
+name of the class:
+
+```scala
+import com.typesafe.config.{ConfigFactory,ConfigValue}
```
leifwickland, line 258: It occurs to me that the main PR for `CoproductHint` didn't include a unit test for the  "enum-y" case. It seems like it might be useful to drop this code into a test to ensure it keeps working.

leifwickland:
The "Supported Types" section of the README includes a "sealed families of case classes (ADTs)" bullet. Please add a link from there to the new section you create.

README.md
ruippeixotog: Yes, you're right. In fact, when I was writing this example it also occurred to me that I use that kind of convention so much in my enums that it probably should be included `pureconfig` as an alternative `CoproductHint` implementation instance, like a `EnumCoproductHint` - even if it's not a general purpose one. What do you think?

melrief (approved):

melrief:
I like the new section, thanks for the changes!

ruippeixotog:
@melrief, @jcazevedo, I have just moved the `CoproductHint` content to a new section. I also agree it is better this way.

If we end up going forward with #76 I would create a section "Override behaviour for case classes" or, equivalently, "Override behaviour for products" to complement this one. That section would most probably be a modified version of "Customizing naming conventions".

melrief:
@ruippeixotog thanks for opening #76, it's very interesting and I can see why `ConfigFieldMapping` can be seen as a special case of a hint for products. Imho it's a natural evolution of the current system.
About this PR, @jcazevedo what do you think? Imho this is ready to be merged and @leifwickland is on the same page.

jcazevedo:
I also think this is ready to be merged. There's only a tiny conflict to be fixed in the TOC of the README introduced when we merged #70.

ruippeixotog:
@jcazevedo I have just resolved the merge conflicts.

jcazevedo (approved):

melrief:
Thanks @ruippeixotog! Our README is already much better than the one of many github projects 😄

melrief:
Last famous words, the merge of this PR just broke the [README after "override behaviour for sealed families"](https://github.com/melrief/pureconfig#override-behaviour-for-sealed-families). I'll fix it right now with a quick PR.

melrief:
Fixed with e1d4679

Issue #77, opened 2017-01-09
Add a changelog


ruippeixotog:
It's always good to have a log of what changed between releases. The commit tree sometimes works, but particularly for external users it makes sense to have a compact list enumerating what can users expect after an upgrade - which bugs were fixed, what improvements were made and what are the breaking changes. This last one is particularly important; if there is not a way to make things backwards compatible, it makes sense for users to have a place where they at least know what changed.

Although they are not libraries, I take as good examples the changelogs for [terraform](https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md) and [marathon](https://github.com/mesosphere/marathon/blob/master/changelog.md). They are well organized and easily searchable. Based on those, I adopted a much more modest version [for one of my libraries](https://github.com/ruippeixotog/scala-scraper/blob/master/CHANGELOG.md).

What do you think of adding such a document? Adding it does not mean we shouldn't try to avoid breaking APIs, but it would give more information to users in case we decide to do it.

(@melrief @jcazevedo @leifwickland I'm opening these issues to get your general opinion, if you think it's a good idea I don't mind writing it :))

leifwickland:
I love change logs. I just hate writing them. ;-)
I think it would be a terrific idea to at one, particularly to document the breaking changes we've been introducing.

melrief:
I think changelogs are essential at this point. I like the changelog of @ruippeixotog example, I propose to follow that format. Now the main issue that I see is that we have a lot of legacy versions. Shall we collect info for the previous version or just start with the `0.5.1`?

leifwickland:
I suggest we just hit the versions which introduced breaking changes in the last few months and move forward. I think that's a reasonable compromise of appearing to be responsible with versioning while not exerting too much undue effort.

ruippeixotog:
Yeah, logging only breaking changes for old versions and starting to write all changes for future versions seems a nice compromise.

melrief:
Now that we all agree, I guess we have to decide who takes care of adding the document. I propose myself for obvious reasons and I also propose that we should not release `0.5.1` without completing the changelog.

ruippeixotog:
I can write the changes for this version and the last version probably. Versions older than that I'm not so sure of the changes.

melrief:
If you want to create the changelog file for the last two releases, I can fill the remaining parts with another PR. The last two versions are the important one for the changelog anyway.

PR #78, opened 2017-01-11
Add a first version of CHANGELOG

ruippeixotog:
Adds the list of changes for version 0.5.0 and for the next version. A first step for #77.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,5 +17,7 @@
       `type: "dogconf"` field in the config file;
     - The old behavior can be restored by putting an implicit instance of `FirstSuccessCoproductHint` in scope (the
       migration to the new format is strongly recommended though, as the previous one may lead to ambiguous behavior);
+    - More information about the default representation and on how to customize it can be seen in the
+      [README](https://github.com/melrief/pureconfig#override-behaviour-for-sealed-families).
 - Bug fixes
   - `0` is accepted again as a valid `Duration` in configs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,19 @@
+### 0.5.1 (unreleased)
+
+- New features
+  - More consistent handling of missing keys: if a config key is missing pureconfig always throws a
+    `KeyNotFoundException` now, unless the `ConfigConvert` extends the new `AllowMissingKey` trait.
+
+### 0.5.0 (Jan 3, 2017)
+
+- New features
+  - Sealed families are now converted to and from configs unambiguously by using an extra `type` field (customizable) in
+    their config representation; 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,19 @@
...
+    `KeyNotFoundException` now, unless the `ConfigConvert` extends the new `AllowMissingKey` trait.
+
+### 0.5.0 (Jan 3, 2017)
+
+- New features
+  - Sealed families are now converted to and from configs unambiguously by using an extra `type` field (customizable) in
+    their config representation; 
+  - New `CoproductHint` trait which allows customization of the derived `ConfigConvert` for sealed families;
+- Breaking changes
+  - The default config representation for sealed families has changed. Configs for such classes that were read
```
leifwickland: Suggestion: Provide a little more guidance on how the new feature works.
```
  - The default config representation for sealed families has changed. 
    - By default Pureconfig now expects to find a `type` field containing the 
      lowercase simple class name of the type to be read. For example for a family
      including `DogConf` and `CatConf`, Pureconfig expects to find a 
      `type: "dogconf"` member in the config file.
    - The old behavior can be restored by putting an implicit instance 
      of `FirstSuccessCoproductHint` in scope (the migration to the new format is strongly recommended though, as the
      previous one may lead to ambiguous behavior).
```

I think that second bullet point is correct. If not, please correct it.

leifwickland (approved):
Thanks for doing this, @ruippeixotog! 👍

melrief (approved):

CHANGELOG.md
leifwickland: Suggestion: Provide a link to the description of the new behavior in the README.

melrief: I also would like to add the links, people will probably check the CHANGELOG on github and being able to go to the section of the README describing the new feature can be very useful.

melrief: I usually prefer the CHANGELOG to be concise but I can see the importance of describing what's different, in particular with a Breaking changes section. What do you think @ruippeixotog, should we make the CHANGELOG a bit more descriptive about what's going to break? I personally support @leifwickland idea.

ruippeixotog: Yes, I agree with providing more information. In a bigger project there would be a section for logging changes and a separate migration guide; in this case, I don't think there is the need for that.

ruippeixotog: Added @leifwickland suggestion in e5340bd896ccebf090b4099f29be2faf2b77ca26.

ruippeixotog: Good idea. Added in 2420237.

melrief:
Thanks for the great work @ruippeixotog!

PR #79, opened 2017-01-11
Add functionality to parse collections from objects with numeric keys

froesef-zz:
I propose to make it possible to parse objects with numeric keys as lists. 
This functionality exists in typesafe config and can help when overwriting list items through the command line or properties files. 

As example a configuration like the following could be parsed:
```
a.0=1
a.1=2
```

### DIFF
--- a/core/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/core/src/main/scala/pureconfig/ConfigConvert.scala
@@ -235,6 +235,23 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
           }
 
           tryBuilder.map(_.result())
+        case o: ConfigObject =>
+          if (o.keySet().asScala.forall(key => Try(key.toInt).isSuccess)) {
+            val sortedEntryList = o.entrySet().asScala.toList
+              .map(entry => (entry.getKey.toInt, entry.getValue))
+              .sortBy(_._1)
+            val tryBuilder = sortedEntryList.foldLeft(Try(cbf())) {
+              case (tryResult, (_, configValue)) =>
+                for {
+                  result <- tryResult
+                  value <- configConvert.value.from(configValue)
+                } yield result += value
+            }
+
+            tryBuilder.map(_.result())
+          } else {
+            Failure(new IllegalArgumentException("Cannot parse this object as list! Expecting syntax like (a.0=0, a.1=1, ...) when loading 'a'"))
+          }
         case other =>
           Failure(WrongTypeException(other.valueType().toString))
       }
--- a/core/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/core/src/main/scala/pureconfig/ConfigConvert.scala
@@ -254,7 +254,7 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
 
     private def reportBadIndex(keyString: String): PartialFunction[Throwable, Failure[Nothing]] = {
       case NonFatal(e) =>
-        val message = s"Cannot interpet $keyString as a numeric index. Tried to read the object as an indexed sequence. " +
+        val message = s"Cannot interpet '$keyString' as a numeric index. Tried to read the object as an indexed sequence. " +
           "Expecting syntax like (a.0=0, a.1=1, ...) when loading a sequence called 'a'"
         Failure(new IllegalArgumentException(message))
     }
--- a/core/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/core/src/main/scala/pureconfig/ConfigConvert.scala
@@ -16,10 +16,13 @@ import scala.language.higherKinds
 import scala.reflect.ClassTag
 import scala.util.{ Failure, Success, Try }
 import java.net.URL
-import scala.concurrent.duration.{ Duration, FiniteDuration }
 
-import pureconfig.ConfigConvert.{ fromNonEmptyString, fromString, stringConvert, nonEmptyStringConvert }
+import scala.concurrent.duration.{ Duration, FiniteDuration }
+import pureconfig.ConfigConvert.{ fromNonEmptyString, fromString, nonEmptyStringConvert, stringConvert }
 import pureconfig.error.{ CannotConvertNullException, KeyNotFoundException, WrongTypeException, WrongTypeForKeyException }
+
+import scala.collection.mutable.Builder
+import scala.util.control.NonFatal
 
 /**
  * Trait for conversion between `T` and `ConfigValue`.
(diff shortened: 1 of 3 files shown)


core/src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -413,6 +413,29 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
     loadConfig[ConfWithListOfFoo](conf) shouldBe Success(expected)
   }
 
+  it should "be able to load a set of Ints from the system properties" in {
+    System.setProperty("pure.conf.intSet.0", "1")
```
leifwickland: Please use one of the other methods for loading configuration that doesn't modify global state, such as `ConfigFactory.parseString`. While I recognize that you made an effort to clear those settings, a failed test assertion would throw an exception, which would prevent the global state from being clear. Furthermore, the current approach is brittle, requiring a dev to make changes two places when modifying a test.

For example I think these lines could be written as:
```
val cf = ConfigFactory.parseString("""
  pure.conf: {
    intSet.0: 1
    intSet.1: 2  
  }""")
loadConfig[Set[Int]](cf, "pure.conf.intSet")
```

core/src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -740,6 +763,17 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
     loadConfig[ConfigWithDouble](conf).failure.exception.getMessage shouldEqual "Cannot read a Double from an empty string."
   }
 
+  "Converting from a wrong system properties list" should "complain about having the wrong sytem properties list syntax" in {
+    System.setProperty("pure.conf.intSet.0", "1")
+    System.setProperty("pure.conf.intSet.a", "2")
+    val expected = Set(1, 2)
```
leifwickland: This is unused.

core/src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -740,6 +763,17 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
...
+
+    assert(loadConfig[Set[Int]]("pure.conf.intSet").failure.exception.getMessage.contains("Cannot parse this object as list! Expecting syntax like"))
+
+    System.clearProperty("pure.conf.intSet.0")
+    System.clearProperty("pure.conf.intSet.a")
+  }
+
```
leifwickland, line 786: Please add a test which defines the behavior if an index is skipped. For example, what happens when the configuration looks like:

```
val cf = ConfigFactory.parseString("""
  pure.conf: {
    // Note there's no 0 index
    intSet.1: 2  
    intSet.2: 3
  }""")
```

froesef-zz:
I checked the behavior if one index is skipped as @leifwickland suggested.
The current behavior is to still accept the list as valid which is the same behavior that typesafe Config also has. I added a test.

core/src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -413,6 +413,40 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
...
+  it should "be able to load a list of Ints from the system properties in correct order when one element is missing" in {
+    val conf = ConfigFactory.parseString("""
+    pure.conf: {
+      intList.2: 3
+      intList.0: 1
+    }""")
+
+    val expected = List(1, 3)
```
leifwickland, line 446: Gosh, I find this behavior surprising. I'd expect that a missing index would result in either an error or  a `null` (though a may loathe them), in that index.

That isn't to say that this behavior is wrong; it just wouldn't be my first choice.

melrief: I think the behavior is smart in the sense that what the number is is not important because the number is not used as index in the structure. For what I can see, it could also be not a number: given a type `T`, if it has an `Ordering[T]` then you can use it as index because you just need it to sort the key-values.

melrief (approved):
I personally like this PR, it adds a functionality that is available in typesafe config without breaking anything. I think being able to create collections from indexed values make sense.

Because this changes the behavior of pureconfig, I would ask @leifwickland, @jcazevedo and @ruippeixotog (the authors of the library) to have a look at it.

core/src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -413,6 +413,40 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
     loadConfig[ConfWithListOfFoo](conf) shouldBe Success(expected)
   }
 
+  it should "be able to load a set of Ints from the system properties" in {
```
leifwickland: Thanks for updating all the tests to parse string instead of using system properties!
Nitpick: Please update all of the test descriptions to remove the reference to "system properties".

leifwickland (changes requested):

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -235,6 +235,23 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
           }
 
           tryBuilder.map(_.result())
+        case o: ConfigObject =>
```
leifwickland, line 233: I think we could eliminate some duplicated code and improve the error message when the index is non-numeric if we replaced the entire `from` method with something like this:

```
    override def from(config: ConfigValue): Try[F[T]] = {
      config match {
        case co: ConfigList =>
          traverseTry(co.asScala, cbf())(configConvert.value.from)
        case o: ConfigObject =>
          for {
            indexedEntries <- traverseTry(o.asScala, new collection.mutable.ListBuffer[(Int, ConfigValue)]) {
              case (keyString, value) => Try(keyString.toInt -> value).recoverWith(reportBadIndex(keyString))
            }
            sortedValues = indexedEntries.sortBy(_._1).map(_._2)
            result <- traverseTry(sortedValues, cbf())(configConvert.value.from)
          } yield result
        case other =>
          Failure(WrongTypeException(other.valueType().toString))
      }
    }

    // Akin to scalaz's `traverseU` but limited to operating on Try instead of all monads.
    private def traverseTry[A, B, C](i: TraversableOnce[A], builder: Builder[B, C])(f: A => Try[B]): Try[C] = {
      i.foldLeft(Try(builder)) {
        case (builderMaybe, a) => builderMaybe.flatMap { builder =>
          f(a).map(builder += _)
        }
      }.map(_.result)
    }

    private def reportBadIndex(keyString: String): PartialFunction[Throwable, Failure[Nothing]] = {
      case _: Throwable =>
        val message = s"Cannot interpet $keyString as a numeric index. Tried to read the object as an indexed sequence. " +
          "Expecting syntax like (a.0=0, a.1=1, ...) when loading a sequence called 'a'"
        Failure(new IllegalArgumentException(message))
    }
```

@melrief What do you think about that? Do you have a better name than `traverseTry`?

ruippeixotog (commented):
I won't be able to review the code of this PR in the upcoming days as I'll be on a trip. I agree with the change in itself, though.

Even though in Typesafe config README the feature isn't mentioned, it is discussed [here](https://github.com/typesafehub/config/issues/69) and featured [in their changelog](https://github.com/typesafehub/config/blob/f6680a5dad51d992139d45a84fad734f1778bf50/NEWS.md#101-may-19-2013). These links may provide more insights and cases to test here.

core/src/main/scala/pureconfig/ConfigConvert.scala
leifwickland: @froesef Do the changes above make sense to you? In my clone of your branch they pass all the unit tests except the one which verified the error message. If so would you be willing to incorporate them into your PR? 
I'm not sure if the other authors agree with me, but that's my personal preference.
Please feel free to improve upon my suggestion.

melrief: `traverseTry` is a nice idea considering that we use it all over the place. @froesef what do you think of it? It's similar to a common pattern used in functional programming. The only think is that I prefer to not catch `Throwable` and I would switch that to `Exception` at least.

melrief:
@ruippeixotog thanks, I think we should add the links to the CHANGELOG

core/src/main/scala/pureconfig/ConfigConvert.scala
leifwickland: @melrief 
Good point about `Throwable`. What I should've used is `NonFatal`. The updated function would look like:
```
    private def reportBadIndex(keyString: String): PartialFunction[Throwable, Failure[Nothing]] = {
      case scala.util.control.NonFatal(e) =>
        val message = s"Cannot interpet $keyString as a numeric index. Tried to read the object as an indexed sequence. " +
          "Expecting syntax like (a.0=0, a.1=1, ...) when loading a sequence called 'a'"
        Failure(new IllegalArgumentException(message, e))
    }
```

froesef-zz: I think it's a good idea to extract this as a function. I will incorporate it in to the PR. As for the change with ```Throwable```. The failure is caught with ```NonFatal``` in ```Try(keyString.toInt -> value)``` anyways. But it is probably good style to still match on it again.

As for the signature of the method. Should we change it to the following to not be using ```Nothing``` as type?
```private def reportBadIndex[U](keyString: String): PartialFunction[Throwable, Failure[U]]```

melrief: The two signatures are similar in this case but, practically speaking, if you use `Nothing` then you don't have to specify what `U` is and in the context of `reportBadIndex` we don't need to specify it. In my opinion it makes sense to use `Nothing` because `reportBadIndex` shouldn't care about what a successful values would be.

melrief:
Thanks @froesef for all the changes! @leifwickland I think we can merge this.

leifwickland (approved):
@melrief 
I made one more nitpicky suggestion. I'm OK merging now or waiting for that change.
@froesef Thanks for all your hard work on this and for putting up with my myriad suggestions.

melrief:
Beautiful, thanks @froesef for the PR! I would like to ask you if I can add you to the [CONTRIBUTORS](https://github.com/melrief/pureconfig/blob/master/CONTRIBUTORS) files. If you agree, please make a small PR in which you add a line to that file with your name, optionally the enclosed between < and > and your github username.

PR #82, opened 2017-01-12
Document default value behavior

### LINKED ISSUE #80: Support for case class default values?
Does _PureConfig_ support default arguments to case classes?

In other words, given the [example from the README](https://github.com/melrief/pureconfig#supported-types), does

```scala
> import pureconfig.loadConfig
> import com.typesafe.config.ConfigFactory.parseString
> sealed trait MyAdt
> case class AdtA(a: String) extends MyAdt
> case class AdtB(b: Int) extends MyAdt
> case class MyClass(int: Int = 13, adt: MyAdt, list: List[Double], map: Map[String, String], option: Option[String])
> val conf = parseString("""{ "adt": { "type": "adtb", "b": 1 }, "list":["1", "20%"], "map": { "key": "value" } }""")
```

return a `MyClass` instance with `int` set to 13?

leifwickland:
In response to #80, document the pureconfig's support for default values.
Closes #80

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ A boilerplate-free Scala library for loading configuration files
 - [Override behaviour for types](#override-behaviour-for-types)
 - [Override behaviour for sealed families](#override-behaviour-for-sealed-families)
 - [Handling missing keys](#handling-missing-keys)
+- [Default values](#default-values)
 - [Example](#example)
 - [Whence the config files](#whence-the-config-files)
 - [Contribute](#contribute)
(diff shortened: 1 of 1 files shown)


melrief (approved):
Concise, complete and well written, thanks for doing it @leifwickland! I'll wait for @jcazevedo to review this because he is the one that made the patch for defaults but in my opinion this could be merged as it is.

README.md
```diff
@@ -325,6 +326,36 @@ available `ConfigConvert` for that type with a `null` value:
 scala.util.Try[Foo] = Success(Foo(42))
 ```
 
+## Default values
+
+Pureconfig supports default values. If a `case class` has a default argument and the underlying configuration is missing a value for that field, then Pureconfig will happily create an instance of the `class`, loading the other values from the configuration.
+
+For exampe:
```
jcazevedo: "Exampe" should be "example".

jcazevedo (approved):
Other than the minor typo, this looks good to merge. Thanks @leifwickland for doing this!

README.md
leifwickland: Thanks for spotting the typo, @jcazevedo. Fixed.

PR #83, opened 2017-01-12
Upgrade to the latest sbt, 0.13.13.

leifwickland:


melrief:
Thanks @leifwickland! If tests pass, I'll merge this because I don't see why we shouldn't keep sbt up-to-date.

PR #84, opened 2017-01-16
Fix broken link

n4to4:


leifwickland:
Thanks for fixing this, @n4to4!

PR #88, opened 2017-01-17
Add java.time support

melrief:
This PR adds support for the entire [`java.time` package](https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html) via either configurable converters or standard converters when no format is required to parse/format a type. This PR follows #66 in style and should resolve the issue #65.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,9 @@ and percentage format ending with `%`), `Float` (also supporting percentage),
 is in this list
 - `Option` for optional values, i.e. value that can or cannot be in the configuration
 - `Map` with `String` keys and any value type that is in this list
-- `LocalDate`, `LocalTime`, `LocalDatetime` (see [Configurable converters](#configurable-converters))
+- everything in [`java.time`](https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html).
+ Because different formats are supported, converters for types that support multiple formats must be configured
+ before they can be used (see [Configurable converters](#configurable-converters))
 - typesafe `ConfigValue`, `ConfigObject` and `ConfigList`
 - case classes
 - sealed families of case classes (ADTs)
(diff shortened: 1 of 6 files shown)


melrief:
@leifwickland @jcazevedo @ruippeixotog can we have a quick look at this PR? I would like to merge it before we release the new version `0.5.1`. I just need one review, it's similar to the one that we already merged for `LocalDate/Time/DateTime`

jcazevedo (approved):
We should eventually refactor the `LowPriorityConfigConvertImplicits` trait and the `PureconfSuite`, as we're starting to have a lot of implicits and tests scattered there. 

This shouldn't block this feature, though, so I think this is good to merge as is.

melrief:
@jcazevedo I can't agree more with you, maybe we should do this before publishing the version after `0.5.1`

README.md
```diff
@@ -103,7 +103,9 @@ and percentage format ending with `%`), `Float` (also supporting percentage),
 is in this list
 - `Option` for optional values, i.e. value that can or cannot be in the configuration
 - `Map` with `String` keys and any value type that is in this list
-- `LocalDate`, `LocalTime`, `LocalDatetime` (see [Configurable converters](#configurable-converters))
+- everything in [`java.time`](https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html).
```
leifwickland, line 106: Nitpick: It seems like saying "everything" might be a bit of a stretch given types like `Clock` and `MonthDay` aren't covered. Perhaps "most types in ..." instead

leifwickland (approved):
Nice work, @melrief

README.md
melrief: [`MonthDay`](https://github.com/melrief/pureconfig/pull/88/files#diff-afed246583326bb75c3a166abab42fa1R35) is covered and `Clock` cannot be created from a configuration. I think it's honest to say that we cover the entire package.

Issue #76, opened 2017-01-06
`ConfigFieldMapping` and `ProductHint` concepts


ruippeixotog:
The current naming conventions currently only apply (and are only relevant) to products. In my view, a `ConfigFieldMapping` is a just a special case of a `ProductHint` - a way for developers to "advise" pureconfig on how to derive converters for certain kinds of products.

In the derivation code a `ConfigFieldMapping` is already handled just like a `CoproductHint` and a first version of `ProductHint` could start to have an interface just equal to the current `ConfigFieldMapping`. However, the more general concept would allow us to add in the future new ways to customize how products are converted. One example that comes to my mind is customizing how `Option` fields should be handled (currently they are a special case).

By separating the concept of `ConfigFieldMapping` from a `ProductHint`, we could start to use `ConfigFieldMapping` for both products and coproducts. The current `FieldCoproductHint` could take an instance of `ConfigFieldMapping` instead of having a `fieldValue` function, for example (that is actually something that we can start doing now, but the semantics would be a bit muddier IMO).

What do you think of this change?

ruippeixotog:
Another thing that would be useful in a `ProductHint`: how should unknown keys be treated? If we see in a config a key that has no matching field in the product, do we want to be strict and throw an error or do we want to ignore it? Like naming conventions, we probably want a default behavior most of the time and have the ability to customize case-by-case.

melrief:
I like this idea and I think we should consider it. In particular, I think the separation of `ConfigFieldMapping` from a `ProductHint` is good design.
I'm not sure how to handle unknown keys. I would like to let the user configure it but I'm not sure what should be default. I would go with the most conservative strategy, which is the strict strategy that you were talking about.
I personally like the idea that config keys with no matching fields are ignored because this allows to reduce or change the fields of the configuration class and still be able to use a configuration available. For instance, two versions of a program could use two different configuration schema but one single configuration could be provided with the fields of both schemas. I used this in the past and it worked. The issue with this is that it's easier to make mistakes in the configuration which means that probably we don't want it as default behavior. A more strict and safe approach is better imho.

jcazevedo:
I'm also supportive of the idea of a `ProductHint`. One other thing that we can configure through a `ProductHint`, besides the ones mentioned before, is whether or not we should apply default arguments of case classes when a field is missing from a config.

ruippeixotog:
[...]

PR #90, opened 2017-01-21
Product hints

### LINKED ISSUE #68: Default ConfigFieldMapping should be (CamelCase, KebabCase)
The current default mapping between config keys and case class fields is camel case to camel case. However, in Typesafe config the _de facto_ standard is to use dashes between words (see examples in [Typesafe config](https://github.com/typesafehub/config#uses-of-substitutions), [akka](http://doc.akka.io/docs/akka/snapshot/general/configuration.html#Listing_of_the_Reference_Configuration), [spray](https://github.com/spray/spray/blob/master/spray-can/src/main/resources/reference.conf), [kamon](https://github.com/kamon-io/Kamon/blob/master/kamon-core/src/main/resources/reference.conf) and many more). As such, it seems more natural to me for the default to be `ConfigFieldMapping[T](CamelCase, KebabCase)`.

What is your opinion on this? Is there a reason for the current default? If you agree with that change I can do a quick pull request.

ruippeixotog:
This is an implementation of a `ProductHint`, as discussed in #76. It supports:

- Defining the mapping between field names and config keys (with `ConfigFieldMapping`);
- Controlling whether default arguments should be applied or not;
- Controlling whether unknown keys are allowed or not.

For easier reviewing, I added all those features in different commits.

Some random details:

- As much as I'd like it, there is a major problem with enabling "strict mode" as default: configs that are loaded using `ConfigFactory.load` and are read from the root will always have unknown keys, as Typesafe config mixes in environment variables and system properties;
- I could have created an implicit conversion from `ConfigFieldMapping` to `ProductHint` in order not to break existing code using implicit `ConfigFieldMappings`. However, that will lead to confusion in the long run if we want to use mappings for other things, like coproduct fields. I personally prefer to have users deal with breaking changes if the alternative is to introduce hacks and somewhat magical behaviors that will hinder future improvements;
- I'd like to take this opportunity to close #68, if we're all ok with it. If we're already changing the API for `ConfigFieldMapping` we could make users also do the transition to a new naming convention, dealing with breaking code only once;
- I did not touch README and CHANGELOG in this pull request to avoid being repeatedly updating it as changes are requested. I prefer to do a separate pull request updating all documentation once this is closed.

### DIFF
--- a/core/src/main/scala/pureconfig/ProductHint.scala
+++ b/core/src/main/scala/pureconfig/ProductHint.scala
@@ -2,16 +2,19 @@ package pureconfig
 
 trait ProductHint[T] {
   def fieldMapping: ConfigFieldMapping
+  def useDefaultArgs: Boolean
 }
 
 private[pureconfig] case class ProductHintImpl[T](
-  fieldMapping: ConfigFieldMapping) extends ProductHint[T]
+  fieldMapping: ConfigFieldMapping,
+  useDefaultArgs: Boolean) extends ProductHint[T]
 
 object ProductHint {
 
   def apply[T](
-    fieldMapping: ConfigFieldMapping = ConfigFieldMapping(CamelCase, CamelCase)): ProductHint[T] =
-    ProductHintImpl[T](fieldMapping)
+    fieldMapping: ConfigFieldMapping = ConfigFieldMapping(CamelCase, CamelCase),
+    useDefaultArgs: Boolean = true): ProductHint[T] =
+    ProductHintImpl[T](fieldMapping, useDefaultArgs)
 
   implicit def default[T]: ProductHint[T] = apply()
 }
--- a/core/src/main/scala/pureconfig/ProductHint.scala
+++ b/core/src/main/scala/pureconfig/ProductHint.scala
@@ -8,9 +8,12 @@ package pureconfig
 trait ProductHint[T] {
 
   /**
-   * The mapping used to map field names to config keys
+   * Returns the key in the config object associated with a given case class field.
+   *
+   * @param fieldName the case class field
+   * @return the key in the config object associated with the given case class field.
    */
-  def fieldMapping: ConfigFieldMapping
+  def configKey(fieldName: String): String
 
   /**
    * A boolean indicating if the default arguments of the case class should be used when fields are missing
@@ -24,9 +27,12 @@ trait ProductHint[T] {
 }
 
 private[pureconfig] case class ProductHintImpl[T](
-  fieldMapping: ConfigFieldMapping,
-  useDefaultArgs: Boolean,
-  allowUnknownKeys: Boolean) extends ProductHint[T]
+    fieldMapping: ConfigFieldMapping,
+    useDefaultArgs: Boolean,
+    allowUnknownKeys: Boolean) extends ProductHint[T] {
+
+  def configKey(fieldName: String) = fieldMapping(fieldName)
+}
 
 object ProductHint {
 
--- a/core/src/main/scala/pureconfig/ProductHint.scala
+++ b/core/src/main/scala/pureconfig/ProductHint.scala
@@ -1,8 +1,25 @@
 package pureconfig
 
+/**
+ * A trait that can be implemented to customize how case classes are read from and written to a config.
+ *
+ * @tparam T the type of case class for which this hint applies
+ */
 trait ProductHint[T] {
+
+  /**
+   * The mapping used to map field names to config keys
+   */
   def fieldMapping: ConfigFieldMapping
+
+  /**
+   * A boolean indicating if the default arguments of the case class should be used when fields are missing
+   */
   def useDefaultArgs: Boolean
+
+  /**
+   * A boolean indicating if config keys that do not map to a case class field are allowed in config objects
+   */
   def allowUnknownKeys: Boolean
 }
 
--- a/core/src/main/scala/pureconfig/ProductHint.scala
+++ b/core/src/main/scala/pureconfig/ProductHint.scala
@@ -1,0 +1,17 @@
+package pureconfig
+
+trait ProductHint[T] {
+  def fieldMapping: ConfigFieldMapping
+}
+
+private[pureconfig] case class ProductHintImpl[T](
+  fieldMapping: ConfigFieldMapping) extends ProductHint[T]
+
+object ProductHint {
+
+  def apply[T](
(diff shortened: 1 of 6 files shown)


melrief (approved):
@ruippeixotog I had a look at this and I have no comments. Ready to merge but I would prefer if either @leifwickland or @jcazevedo can have a look at this and approve it as well. Thanks for doing this.

core/src/main/scala/pureconfig/ProductHint.scala
```diff
@@ -0,0 +1,40 @@
...
+ * A trait that can be implemented to customize how case classes are read from and written to a config.
+ *
+ * @tparam T the type of case class for which this hint applies
+ */
+trait ProductHint[T] {
+
+  /**
+   * The mapping used to map field names to config keys
+   */
+  def fieldMapping: ConfigFieldMapping
```
jcazevedo: Do you think it still makes sense for `ConfigFieldMapping` to exist as an entity? I think we could simplify the interface if `ProductHint` only had a `def configFieldName(typeFieldName: String): String` method. We should still provide helpers to create `ProductHint`s from two `NamingConvention`s, as is currently available for `ConfigFieldMapping`s.

jcazevedo:
I'm also in favor of closing #68. If we are to change the default, now is a good time to do it.

melrief:
I agree on fixing #68, please @ruippeixotog proceed

core/src/main/scala/pureconfig/ProductHint.scala
ruippeixotog: I agree that without its the implicit usage, `ConfigFieldMapping` is not much more than a `String => String`. Having a way to create such a function object - at any time, and not only via a `ProductHint` - still seems useful, though. First, you can use an independent `ConfigFieldMapping` to manually convert fields when you are implementing a custom `ConfigConvert`. Second, it is slightly more elegant to pass a function to the `ProductHint` constructor than passing two `NamingConvention`s without extra context.

How about we leave the `ConfigFieldMapping` object and make its `apply` methods simply return `String => String`? We could then accept any `String => String` function in hints and use the interface you proposed. The downside is less context - when a `ProductHint` receives a string function, how is it applied? Is it a function from config keys to fields or viceversa?

ruippeixotog:
Ok! Changed the default naming convention in 01ce8e7.

melrief:
Thanks @ruippeixotog. I'm waiting for one more review and then merge it, good job.

core/src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -619,24 +620,22 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
...
 
-    implicit val mapping = new ConfigFieldMapping[SampleConf] {
-      def apply(fieldName: String) = fieldName.toUpperCase
-    }
+    implicit val productHint = ProductHint[SampleConf](ConfigFieldMapping(_.toUpperCase))
 
     loadConfig[SampleConf](conf) shouldBe Success(SampleConf(2, "two"))
   }
 
   it should "allow customizing the field mapping with word delimiters" in {
```
leifwickland: Suggestion: Update this test's name to indicate that it is testing CamelCase instead of word delimiters.

leifwickland (approved):
Great work, @ruippeixotog!
Please consider sending a PR to update the README to reflect the new way of customizing field mappings and the ability to influence whether default values are accepted.

core/src/main/scala/pureconfig/ProductHint.scala
jcazevedo: I think we can leave the `ConfigFieldMapping` and the `apply` method of `ProductHint` as is, and delegate the call in the default `ProductHintImpl` to the `ConfigFieldMapping`, no? Something along the lines of (if the method in `ProductHint` would be called `configFieldName`):

```scala
private[pureconfig] case class ProductHintImpl[T](
    fieldMapping: ConfigFieldMapping,
    useDefaultArgs: Boolean,
    allowUnknownKeys: Boolean) extends ProductHint[T] {
  def configFieldName(typeFieldName: String) = fieldMapping(typeFieldName)
}
```

melrief:
I and @leifwickland approved this PR. @jcazevedo what about you?

core/src/main/scala/pureconfig/ProductHint.scala
ruippeixotog: Ok, I agree with that. @melrief and @leifwickland, do you agree too?

melrief:
@ruippeixotog merging #94 changed the style of `ConfigConvert` in master and that's why there is a conflict. It shouldn't be too difficult to fix it. I'll wait for the last review and merge this.

core/src/test/scala/pureconfig/PureconfSuite.scala
ruippeixotog: Well, the intention of this test is not to test specifically the camel case conversion - that is done extensively in `NamingConventionTest.scala`. This one is for testing that adding an appropriate hint allows different field mappings to be applied. However, I can replace "word delimiters" with "different naming conventions" and test more than one mapping. What do you think?

ruippeixotog:
@leifwickland, I intend to update the README and the CHANGELOG once this is merged :)

core/src/test/scala/pureconfig/PureconfSuite.scala
leifwickland: That seems great!

core/src/main/scala/pureconfig/ProductHint.scala
leifwickland: _shrug_ Either way.

jcazevedo:
> I and @leifwickland approved this PR. @jcazevedo what about you?

I'd rather not have the `ConfigFieldMapping` entity in the `ProductHint` trait, as suggested, but I'm OK with leaving this as it is right now if you guys prefer it.

core/src/test/scala/pureconfig/PureconfSuite.scala
ruippeixotog: Added in ae169e040ad5935fca410c27c9ec2f6496854230.

core/src/main/scala/pureconfig/ProductHint.scala
ruippeixotog: Changed it in c84305a3253abf7c04432038c6d8090923050d06. @melrief, let me know if you prefer the other way.

jcazevedo (approved):
LGTM

melrief:
Nice one, thanks!

PR #94, opened 2017-01-24
Upgrade sbt-scalariform version

jcazevedo:
This PR upgrades the `sbt-scalariform` version and moves its settings to the top-level `build.sbt`, as I think we want to be consistent on this across different modules.

I relied on the default configuration, except on trying to prevent [dangling parenthesis](https://github.com/scala-ide/scalariform#danglingcloseparenthesis) and [double indenting on class declaration](https://github.com/scala-ide/scalariform#doubleindentclassdeclaration). One thing that changed was that the `implicit` keyword is now placed on its own line. I'm not particularly fond of it, but changing the setting to disable it [has other implications](https://github.com/scala-ide/scalariform#alignparameters).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,6 @@
+import com.typesafe.sbt.SbtScalariform.ScalariformKeys
+import scalariform.formatter.preferences._
+
 lazy val core = (project in file("core")).
   settings(settings)
 
@@ -17,6 +20,14 @@ lazy val newerCompilerLintSwitches = Seq(
   "-Ywarn-numeric-widen" // In 2.10 this produces a some strange spurious error
 )
 
+lazy val formattingPreferences = FormattingPreferences()
+  .setPreference(DanglingCloseParenthesis, Prevent)
+  .setPreference(DoubleIndentClassDeclaration, true)
+
+lazy val formattingSettings = SbtScalariform.scalariformSettings ++ Seq(
+  ScalariformKeys.preferences in Compile := formattingPreferences,
+  ScalariformKeys.preferences in Test := formattingPreferences)
+
 lazy val settings = Seq(
   scalaVersion := "2.12.1",
   crossScalaVersions := Seq("2.10.5", "2.11.8", "2.12.1"),
@@ -30,5 +41,4 @@ lazy val settings = Seq(
     val required = "1.8"
     val current  = sys.props("java.specification.version")
     assert(current == required, s"Unsupported JDK: java.specification.version $current != $required")
-  }
-)
+  }) ++ formattingSettings
--- a/build.sbt
+++ b/build.sbt
@@ -23,6 +23,7 @@ lazy val newerCompilerLintSwitches = Seq(
 lazy val formattingPreferences = FormattingPreferences()
   .setPreference(DanglingCloseParenthesis, Prevent)
   .setPreference(DoubleIndentClassDeclaration, true)
+  .setPreference(SpacesAroundMultiImports, true)
 
 lazy val formattingSettings = SbtScalariform.scalariformSettings ++ Seq(
   ScalariformKeys.preferences in Compile := formattingPreferences,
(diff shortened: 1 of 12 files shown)


melrief (approved):

build.sbt
```diff
@@ -17,6 +20,14 @@ lazy val newerCompilerLintSwitches = Seq(
   "-Ywarn-numeric-widen" // In 2.10 this produces a some strange spurious error
 )
 
+lazy val formattingPreferences = FormattingPreferences()
+  .setPreference(DanglingCloseParenthesis, Prevent)
+  .setPreference(DoubleIndentClassDeclaration, true)
```
leifwickland, line 25: Could I interest you in adding
```
.setPreference(SpacesAroundMultiImports, true)
```
to stick closer to the scalariform defaults? (See https://github.com/sbt/sbt-scalariform/blob/master/src/main/scala/com/typesafe/sbt/SbtScalariform.scala#L43)

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -224,7 +226,8 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
   }
 
   // traversable of types with an instance of ConfigConvert
-  implicit def deriveTraversable[T, F[T] <: TraversableOnce[T]](implicit configConvert: Lazy[ConfigConvert[T]],
+  implicit def deriveTraversable[T, F[T] <: TraversableOnce[T]](implicit
```
leifwickland: Nitpick: Please manually move the second `implicit` down to the next line for consistency with other methods in the neighborhood.

build.sbt
```diff
@@ -30,5 +41,4 @@ lazy val settings = Seq(
     val required = "1.8"
     val current  = sys.props("java.specification.version")
     assert(current == required, s"Unsupported JDK: java.specification.version $current != $required")
-  }
-)
+  }) ++ formattingSettings
```
leifwickland, line 45: The `formattingSettings` aren't applied to the root project.
```
> scalariformPreferences
[info] core/compile:scalariformPreferences
[info]  FormattingPreferences(Map(DanglingCloseParenthesis -> Prevent, DoubleIndentClassDeclaration -> true))
[info] {.}/*:scalariformPreferences
[info]  FormattingPreferences(Map(SpacesAroundMultiImports -> true, DoubleIndentClassDeclaration -> true))
```

Do we care? I don't think so, but figured I'd ask.

leifwickland (approved):
Thanks for doing this.

ruippeixotog (approved):
LGTM

build.sbt
melrief: I'm not sure but I wonder if this is applies to the `build.sbt` file because in that case we need the settings to be applied also to root.

build.sbt
melrief: Fine for me

build.sbt
leifwickland: It doesn't seem like there's a particularly good way to include the build files per se in the reformatting set. The nearest I can find is something like:
http://stackoverflow.com/questions/19201509/how-to-format-the-sbt-build-files-with-scalariform-automatically
Even if we did that, then we'd end up making a new target to which the `settings` were applied, I think.

build.sbt
jcazevedo: Added it in 451acb0.

core/src/main/scala/pureconfig/ConfigConvert.scala
jcazevedo: Fixed in 451acb0.

build.sbt
jcazevedo: I don't mind not having the `formattingSettings` applied to the root project. Regarding the `build.sbt` and `/project` files, I don't think the extra boilerplate justifies it, as they're not modified that often, but I don't mind adding it if you think it's worth it.

melrief:
Thanks @jcazevedo!

PR #95, opened 2017-01-28
Update README and CHANGELOG with ProductHint sections

ruippeixotog:


melrief (approved):
We probably want to tell the user in the first line of the README about the breaking change on the  fields mapping as that will be the first issue with the new update

ruippeixotog:
I'm not fond of using READMEs as announcement boards - I like them as "timeless" descriptions of the project, possibly doubling as documentation/wiki for smaller projects like this one.

Users should be able to reach CHANGELOG with ease if they visit the project in GitHub with the intention to see what made their project break after an upgrade. Nevertheless, we can make that even more obvious if we include a section in the README called "Migrating from previous versions" (with an associated link in the TOC) just pointing to the CHANGELOG. What do you think?

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+    - Whether unknown keys are ignored or cause pureconfig to return a `Failure`
+      ([docs](https://github.com/melrief/pureconfig#unknown-keys)).
+- Breaking changes
+  - `ConfigFieldMapping` has no type parameters now;
+  - `ConfigFieldMapping` was replaced by `ProductHint` as the type of object to put in scope in order to customize
+    the derivation of `ConfigConvert` for case class. Old `ConfigFieldMapping` implicit instances in scope have no
+    effect now. The migration can be done by replacing code like
+    `implicit def mapping: ConfigFieldMapping[T] = <mapping>` with
+    `implicit def productHint: ProductHint[T] = ProductHint(<mapping>)`;
+  - The default field mapping changed from camel case config keys (e.g. `exampleKey`) to kebab case keys (e.g.
```
leifwickland: This probably ought to be the top item in the Breaking list for this release. It is more likely to affect users than the one above.
Pushing it down feels like feels like burying the lede.

leifwickland (commented):
@ruippeixotog @melrief I think I like the idea of
- a separate page for migrating from previous versions, or
- a migrating section for each breaking change in the change log
rather than including such directions in the README. 

It seems like it'd be sufficient to include text like the following in the "Add PureConfig to your project" section of README:
> For guidance on upgrading from previous versions please see (link to changelog OR link to migration guide)

My preference would be for the migration to go in the change log so that old information naturally ages out without being removed by pushing it down the page. On the other hand, other folks have expressed an interest in maintaining a terse change log, so I can see why wouldn't.

A possible third alternative that I've seen used in Cats is to include detailed migration suggestions in the FAQ and link to it from the change log.

README.md
```diff
@@ -149,24 +151,84 @@ loadConfig[Conf](conf)
...
-## Customizing naming conventions
+## Extend the library to support new types
+
+Not all types are supported automatically by `pureconfig`. For instance, classes
+that are not case classes are not supported out-of-the-box:
+
+```scala
+> class Foo(var bar: Int) { override def toString: String = s"Foo(${bar.toString})" }
+> loadConfig[Foo]
+could not find implicit value for parameter conv: pureconfig.ConfigConvert[Foo]
```
leifwickland, line 162: Nitpick: This line should be visually distinct in some way to indicate that it is a compiler error message rather than straight up code. Typically we do that with a leading `//`

README.md
```diff
@@ -149,24 +151,84 @@ loadConfig[Conf](conf)
...
+```
+
+`pureconfig` can be extended to support those types. To do so, an instance for the
```
leifwickland, line 165: Nitpick: This line should end in a colon, `:`.

README.md
```diff
@@ -149,24 +151,84 @@ loadConfig[Conf](conf)
...
+
+`pureconfig` has to assume some default conventions and behaviours when
+deriving `ConfigConvert` instances for case classes:
+
+- How do keys in config objects map to field names of the case class?
+- Are unknown keys allowed in the config object?
+- Should default values in case class fields be applied when its respective
+  config key is missing?
+  
+By default, `pureconfig` assumes that config keys are written in kebab case
```
leifwickland: Suggestion: I liked how you laid out the questions above as bullet points. I'd lay out the answers here in the say way, one per bullet point above.

leifwickland (changes requested):
Really nice work, @ruippeixotog! I made a few suggestions I'd like you to consider.

README.md
ruippeixotog: I agree with you but although it appears in the diff, I just moved this chunk of text from one place to another. The README is pretty rich in content, but lacks an uniform writing style - not only in this case but in several others that doesn't appear in the diff.

Do you mind if we leave this to another pull request? I don't mind scanning the README and making the examples more uniform.

README.md
ruippeixotog: Same thing as above; I can deal with this in another pull request.

CHANGELOG.md
ruippeixotog: Makes sense. Done in 27ce21d7a5c8cbb7352e7565ebc56fbe837d7280.

README.md
ruippeixotog: I have no strong feelings one way or another. Followed your suggestion on 34e977d751b6232477a24cf9773a7153f432b0a2.

README.md
leifwickland: Oh, no problem. I was just trying to bring those inconsistencies to light as I noticed them.

README.md
leifwickland: 👍

leifwickland (approved):

PR #99, opened 2017-01-30
Add support for parsing enumeratum enums

### LINKED ISSUE #97: Support for enumeratum
I'm using [Enumeratum](https://github.com/lloydmeta/enumeratum) with pureconfig, and I wrote a ConfigConvert instance for enumeratum enums. Would there be interest in including it in pureconfig?

I couldn't figure out the correct place to put the implementation if that was the case, but this is the code:

```scala
import enumeratum.{Enum, EnumEntry}
import pureconfig.ConfigConvert

import scala.util.Try

trait EnumeratumConfigConvert {
  implicit def enumeratumConfigConvert[A <: EnumEntry](implicit enum: Enum[A]): ConfigConvert[A] =
    ConfigConvert.stringConvert[A](name => Try(enum.withName(name)), _.toString)
}
```

aeons:
Moved version.sbt to root project, so all modules will follow the version of core.

I added a ConfigConvert instance for the base `EnumEntry` from enumeratum as well as the six different `ValueEnumEntry`(ie)s.

Fixes #97.


melrief:
Only the tests for scala 2.10 are failing and only for the new project. I'm not sure why, the code looks correct and the derivation is failing. Usually when this happens is because I haven't added the scalaMacrosParadise plugin to the project. Can you try to add it by adding `Dependencies.scalaMacrosParadise` to `libaryDependencies` of your module and see if tests are still failing (try to run `sbt +test` which should test for all Scala versions set in `build.sbt`)?

aeons:
For some reason I thought macro paradise was added to the build at another level than a library dependency. :)

Seems it works now.

melrief (changes requested):
Everything seems correct to me, LGTM. I put @leifwickland too for a review because he is probably interested in modules.

@aeons there is only one thing missing: you probably want to publish this as library so that people can use it. The `build.sbt` doesn't contain info about where and how to publish the library. You can have a look at the core [`build.sbt`](https://github.com/melrief/pureconfig/blob/master/core/build.sbt#L27) file on how we do it for the core module. You can copy-paste most of the code, except the `<developers>` section in which you should put only yourself. If you do this, then once the module is merged in master you can publish the module using [sonatype](https://oss.sonatype.org/).

aeons:
Alright, I'll have a look at that tomorrow.

modules/enumeratum/src/main/scala/pureconfig/module/enumeratum.scala
```diff
@@ -0,0 +1,30 @@
...
+import pureconfig.ConfigConvert
+
+import scala.util.Try
+
+package object enumeratum {
+  implicit def enumeratumConfigConvert[A <: EnumEntry](implicit enum: Enum[A]): ConfigConvert[A] =
+    ConfigConvert.stringConvert[A](name => Try(enum.withName(name)), _.toString)
+
+  implicit def enumeratumIntConfigConvert[A <: IntEnumEntry](implicit enum: IntEnum[A]): ConfigConvert[A] =
+    ConfigConvert.stringConvert[A](v => Try(enum.withValue(v.toInt)), _.toString)
```
leifwickland: The second argument to `stringConvert` is for converting the `A` back into a value which can be written to a config file. I don't think `toString` is going to be the correct conversion for that. I think you need something like `_.value.toString`.
Ditto for the various converter instances below.
(I could be totally off base. I'm just looking at Enumeratum for the first time now.)

modules/enumeratum/src/main/scala/pureconfig/module/enumeratum.scala
```diff
@@ -0,0 +1,30 @@
...
+    ConfigConvert.stringConvert[A](v => Try(enum.withValue(v.toShort)), _.toString)
+
+  implicit def enumeratumStringConfigConvert[A <: StringEnumEntry](implicit enum: StringEnum[A]): ConfigConvert[A] =
+    ConfigConvert.stringConvert[A](v => Try(enum.withValue(v)), _.toString)
+
+  implicit def enumeratumByteConfigConvert[A <: ByteEnumEntry](implicit enum: ByteEnum[A]): ConfigConvert[A] =
+    ConfigConvert.stringConvert[A](v => Try(enum.withValue(v.toByte)), _.toString)
+
+  implicit def enumeratumCharConfigConvert[A <: CharEnumEntry](implicit enum: CharEnum[A]): ConfigConvert[A] =
+    ConfigConvert.stringConvert[A](v => Try(enum.withValue(v.charAt(0))), _.toString)
```
leifwickland: It seems like perhaps this ought to verify that the string is exactly 1 character long before reading the first character. Providing a better error message than index out of bounds or silently dropping characters would be helpful, I think.

modules/enumeratum/src/main/scala/pureconfig/module/enumeratum.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import _root_.enumeratum.values._
+import _root_.enumeratum._
+import pureconfig.ConfigConvert
+
+import scala.util.Try
+
+package object enumeratum {
+  implicit def enumeratumConfigConvert[A <: EnumEntry](implicit enum: Enum[A]): ConfigConvert[A] =
+    ConfigConvert.stringConvert[A](name => Try(enum.withName(name)), _.toString)
```
leifwickland: Just a suggestion: I would consider importing `ConfigConvert.stringConvert` just to remove a little boilerplate from each of these lines.

modules/enumeratum/build.sbt
```diff
@@ -0,0 +1,7 @@
+name := "pureconfig-enumeratum"
+
+libraryDependencies ++= Seq(
+  "com.beachape" %% "enumeratum" % "1.5.6",
+  Dependencies.scalaMacrosParadise,
+  Dependencies.scalaTest
+)
```
leifwickland, line 18: Not sure if it'll help, but here's what I was planning for the `pureconfig-joda` module I've been working on:

```
name := "pureconfig-joda"

organization := "com.github.melrief"

homepage := Some(url("https://github.com/melrief/pureconfig"))

licenses := Seq("Mozilla Public License, version 2.0" -> url("https://www.mozilla.org/MPL/2.0/"))

resolvers ++= Seq(
  Resolver.sonatypeRepo("releases"),
  Resolver.sonatypeRepo("snapshots")
)

libraryDependencies ++= Seq(
  ("joda-time" % "joda-time" % Dependencies.Version.joda),
  ("org.joda" % "joda-convert" % Dependencies.Version.jodaConvert),
  Dependencies.scalaMacrosParadise,
  Dependencies.scalaTest,
  Dependencies.scalaCheck,
  Dependencies.scalaCheckShapeless
)

publishMavenStyle := true

publishTo := {
  val nexus = "https://oss.sonatype.org/"
  if (isSnapshot.value)
    Some("snapshots" at nexus + "content/repositories/snapshots")
  else
    Some("releases"  at nexus + "service/local/staging/deploy/maven2")
}

publishArtifact in Test := false

pomExtra := (
    <scm>
      <url>git@github.com:melrief/pureconfig.git</url>
      <connection>scm:git:git@github.com:melrief/pureconfig.git</connection>
    </scm>
    <developers>
      <developer>
        <id>melrief</id>
        <name>Mario Pastorelli</name>
        <url>https://github.com/melrief</url>
      </developer>
      <developer>
        <id>leifwickland</id>
        <name>Leif Wickland</name>
        <url>https://github.com/leifwickland</url>
      </developer>
    </developers>)

osgiSettings

OsgiKeys.exportPackage := Seq("pureconfig.module.joda.*")

OsgiKeys.privatePackage := Seq()

OsgiKeys.importPackage := Seq(s"""scala.*;version="[${scalaBinaryVersion.value}.0,${scalaBinaryVersion.value}.50)"""", "*")
```

@melrief Do you see anything in there you don't thing belongs?

leifwickland (changes requested):
Thanks for doing this, @aeons!
I made a few suggestions. Please let me know you think.

leifwickland:
@aeons I'm not saying I've done it correctly, but for comparison sake, I pushed my joda support as #100. If nothing else we probably ought to try to be consistent. :-)

modules/enumeratum/src/main/scala/pureconfig/module/enumeratum.scala
aeons: You're right, it's .entryName and .value.toString for the value enums.

aeons:
Should I be using nonEmptyStringConvert instead of stringConvert?

modules/enumeratum/build.sbt
melrief: Everything seems correct

melrief (approved):

melrief:
LGTM for me, nice one!

leifwickland:
@aeons Regarding using `nonEmptyStringConvert`, probably. I think it's safe to do so for all of the numeric enum types and will generally provide better error messages. However, I think that any of the enum types which allow the user to specify an arbitrary string could result in someone providing an empty string for one entry in the enum. While unlikely, it feels like we probably need to allow them to do that goofy thing. If you don't think that enabling an empty string for the enum entry's value is meritorious, then please use `nonEmptyStringConvert` in all cases for its improved error reporting.

aeons:
I guess that makes sense. So, everything except string value enum is nonEmptyString now.

String value enum allows for the empty string, and there's a test case specifically for that.

melrief:
Thanks @aeons for changing the code. I had a quick look and I approve the changes. I'll wait for @leifwickland to have a look and approve the changes.

leifwickland (approved):
LGTM! 👍  
Thanks again, @aeons

melrief:
Thanks @aeons! I'll publish the artifact right now, nice job!

jcazevedo:
Thanks @aeons for the first PureConfig module! 🎉

melrief:
Released the artifact, it will be available in the next minutes. Thanks again @aeons for the contribution!

melrief:
Announcement tweet https://twitter.com/mapastr/status/827548322634268672

PR #100, opened 2017-02-01
Support joda time analogously to pureconfig-core's support of java.time.

leifwickland:


leifwickland:
@melrief I'm particularly interested in if you think I'm doing the publishing bits wrong. I suspect if I say that this module is in `com.github.melrief` then that implies that I'd need permission for your org.
I've never published a library to sonatype previously, so please assume I'll need hand holding.

melrief:
I can either give you permission or publish it myself once this is merged. FYI the idea was to move pureconfig into its own organization (yes, pureconfig which now is not available anymore) and then allow the authors to publish there. Let's first merge the PRs and then we can think about publishing.

melrief (approved):
LGTM, nice job!

The only thing that I maybe I can propose it that you could reuse the `java.time` generators that we have in `core` instead of creating new ones and then just convert the generated values to the java counterpart. For instance, you can have `Arbitrary[java.time.LocalTime].map(t => new LocalTime(t.getHour, t.getMinute, t.getSecond, t.getMillis))`. It's not mandatory for merging this PR but you could skip a lot of code in this way and eventual fixes on the generators would fix both core and this module.

leifwickland:
@melrief 
Updated in to use the core's generators in: https://github.com/melrief/pureconfig/pull/100/commits/d8205e3db5f0c48f98392231f1601682d8571c97

leifwickland:
@melrief I updated the code to reuse the generators from core. It didn't end up as pretty as I'd hoped, but at least there's reuse.
Also, I hope I never have to work in a project which mixes joda and java.time. The naming mismatches are epically frustrating.

melrief:
I work everyday in projects with the two mixed :disappointed:.

LGTM, do you mind if I wait #99 before merging to not make the life of the new contributor more difficult?

leifwickland:
@melrief 
My condolences on the quagmire.
Thanks for taking a look.
Yes, please wait to merge until #99 is done. I'll deal with the conflicts then.

melrief:
@leifwickland I'm going to merge this as soon as you fix the conflicts.

leifwickland:
@melrief I resolved that conflict. I want to make sure you're OK with the change I just made in addition in `Dependencies` so that I could reuse the joda dependencies in the `joda` module.

melrief:
It's fine, I would expect modules to add dependencies versions in the main `Dependencies` file. That is how sbt is supposed to be used. Publishing the module now.

We main need something in the front README pointing to which modules are available...

leifwickland:
@melrief Totally agree about needing READMEs for the modules and links to them from the root README.

melrief:
Published, nice job @leifwickland !

PR #109, opened 2017-02-05
Make the writing style in README more uniform

### LINKED ISSUE #102: Move away from the monolithic README.md

ruippeixotog:
The title is self-explanatory.

For OCD purposes, we should set a standard for line-wrapping Markdown text in the README, as right now pretty much each section has its own column width. I actually advocate for not line wrapping text at all and instead relying on soft-wrapping in editors - the fact that English written text has no inherent structure other than phrases makes line wrapping very artificial to me. However, it seems most developers prefer to hard-wrap lines in Markdown, so we could just set a column width that we should try to use.

After this, I'll move on to splitting this in multiple documents in order to fix #102.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ An almost comprehensive example is:
 
 ```scala
 import com.typesafe.config.ConfigFactory.parseString
+import pureconfig.loadConfig
 
 sealed trait MyAdt
 case class AdtA(a: String) extends MyAdt
@@ -139,9 +140,11 @@ the package `pureconfig.configurable`. Once the output of a `pureconfig.configur
 PureConfig can start using that configured converter:
 
 ```scala
+import com.typesafe.config.ConfigFactory.parseString
 import java.time.LocalDate
 import java.time.format.DateTimeFormatter
 import pureconfig.configurable._
+import pureconfig.loadConfig
 
 case class Conf(date: LocalDate)
 
@@ -160,6 +163,9 @@ Not all types are supported automatically by PureConfig. For instance, classes
 that are not case classes are not supported out-of-the-box:
 
 ```scala
+import com.typesafe.config.ConfigFactory.parseString
+import pureconfig._
+
 class MyInt(var value: Int) {
   override def toString: String = s"MyInt($value)"
 }
@@ -194,6 +200,7 @@ in the configuration:
 
 ```scala
 import com.typesafe.config.ConfigValueFactory
+import pureconfig._
 
 ConfigConvert[String].from(ConfigValueFactory.fromAnyRef("FooBar"))
 // returns Success("FooBar")
@@ -203,6 +210,8 @@ Now let's say that we want to override this behaviour such that `String`s are
 always read lower case. We can do:
 
 ```scala
+import scala.util.Try
+
 implicit val overrideStrConvert = ConfigConvert.fromString(s => Try(s.toLowerCase))
 
 ConfigConvert[String].from(ConfigValueFactory.fromAnyRef("FooBar"))
@@ -243,6 +252,9 @@ example where the configuration file has all keys in upper case and we're
 loading it into a type whose fields are all in lower case:
 
 ```scala
+import com.typesafe.config.ConfigFactory.parseString
+import pureconfig._
+
 case class SampleConf(foo: Int, bar: String)
 
 implicit val productHint = ProductHint[SampleConf](new ConfigFieldMapping {
@@ -276,6 +288,8 @@ can make sure the following implicit is in scope before loading or writing
 configuration files:
 
 ```scala
+import pureconfig._
+
 implicit def hint[T] = ProductHint[T](ConfigFieldMapping(CamelCase, CamelCase))
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,127 +99,142 @@ val config: Try[YourConfClass] = loadConfig[YourConfClass]
...
-- case classes
-- sealed families of case classes (ADTs)
+  and percentage format ending with `%`), `Float` (also supporting percentage),
+  `Int`, `Long`, `Short`, `URL`, `Duration`, `FiniteDuration`;
+- all collections implementing the `TraversableOnce` trait where the type of
+  the elements is in this list;
+- `Option` for optional values, i.e. values that can or cannot be in the configuration;
+- `Map` with `String` keys and any value type that is in this list;
+- everything in [`java.time`](https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html) (must be
+  configured first - see [Configurable converters](#configurable-converters));
```
melrief, line 109: Not sure about "configured first" because some people may not understand what that means. Said that, I'm not sure how we can write this point...

README.md
```diff
@@ -99,127 +99,142 @@ val config: Try[YourConfClass] = loadConfig[YourConfClass]
...
-> import pureconfig.loadConfig
-> import com.typesafe.config.ConfigFactory.parseString
-> sealed trait MyAdt
-> case class AdtA(a: String) extends MyAdt
-> case class AdtB(b: Int) extends MyAdt
-> case class MyClass(int: Int, adt: MyAdt, list: List[Double], map: Map[String, String], option: Option[String])
-> val conf = parseString("""{ "int": 1, "adt": { "type": "adtb", "b": 1 }, "list":["1", "20%"], "map": { "key": "value" } }""")
-> loadConfig[MyClass](conf)
-res0: util.Try[MyClass] = Success(MyClass(1,AdtB(1),List(1.0, 0.2),Map(key -> value),None))
+import com.typesafe.config.ConfigFactory.parseString
```
melrief, line 117: Why removing the import of `pureconfig.loadConfig`? This example doesn't compile anymore.

README.md
```diff
@@ -99,127 +99,142 @@ val config: Try[YourConfClass] = loadConfig[YourConfClass]
...
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
 import pureconfig.configurable._
 
 case class Conf(date: LocalDate)
 
 implicit val localDateInstance = localDateConfigConvert(DateTimeFormatter.ISO_DATE)
 
-val conf = ConfigFactory.parseString(s"""{date:"2011-12-03"}""")
+val conf = parseString(s"""{ date: "2011-12-03" }""")
```
melrief, line 153: I think we should import the methods that we use, otherwise what is `parseString`? I'm thinking in particular about developers that don't know Typesafe config.

README.md
```diff
@@ -99,127 +99,142 @@ val config: Try[YourConfClass] = loadConfig[YourConfClass]
...
 
 loadConfig[Conf](conf)
-// returns Success(Conf(2016-01-01))
+// returns Success(Conf(LocalDate.parse("2011-12-03", DateTimeFormatter.ISO_DATE)))
```
melrief, line 156: Why replacing the result with an expression that will create the result? Because 2011-12-03 is different from 2016-01-01? We may want to fix that instead.

melrief (commented):
Nice job! I commented some parts that you have changed based on my personal style. I'm not saying to change those points, I just want to be sure that we are on the same page.
The most important comment so far is that I would like to have self-contained examples and that means that imports should always be explicitly imported, otherwise things like `parseString` require the developer to know Typesafe `Config`. I'm moderately strong about this point because it always make me struggle with libraries examples.

README.md
ruippeixotog: Regarding self-contained examples, I opted to treat all examples as a single console session, which means I didn't repeat `import` statements made before in the README. That's why I removed some of them and that's also why I used `parseString` instead of `ConfigFactory.parseString`. But I'm ok either way; we can "re-import" everything in each example, it makes examples more self-contained :)

README.md
ruippeixotog: In "returns" comments I opted to write an expression what can be used in the console to create the expected value. I could also write the `toString` of the resulting value, I'm ok either way!

This is the example where the two "standards" differ the most; in other cases, it's just a question of whether I quote strings or not, or whether I write `KeyNotFoundException("a")` or the full exception message.

README.md
ruippeixotog: I think people would follow the link after that, where the "configuring" part is explained in detail. As this is just a list of the supported types, I would like not to have extensive usage instructions here and instead point to the relevant sections in the README.

README.md
melrief: I see advantages in both styles and I don't see a good reason to not accept your style over mine. Let's use your style and forget about this comment.

README.md
melrief: 👍

README.md
melrief: I prefer having the imports all the time. I know it's more verbose and not nice for us that we know the library, but I'm specifically thinking about the new developer that just wants to try some examples. If you agree, do the change and then I'll merge the PR. I assume one review will be enough considering that this is juts a matter of style and a bigger review will happen with #102.

ruippeixotog: Fixed in 0de7277.

melrief:
Well done @ruippeixotog!

Issue #105, opened 2017-02-01
Collect all errors instead of stopping at the first one


jcazevedo:

melrief:
Done with #110

Issue #103, opened 2017-02-01
Move from Try[A] to Either[ConfigReadFailure, A]


melrief:
I'll use `Failure` instead of `Error` because Java uses `Error` to defined a subtype of `Throwables`. The word `Error` can be confusing. `Failure` is good because it would be used when a `ConfigConverter` **fails** to convert a configuration. It's not an error.

Issue #92, opened 2017-01-23
Split ConfigConverter into two parts: for reading and for writing


13h3r:
There are some cases where end-user just want to read config and do not need to write it. In such cases we do not need to force to implement writer part.

melrief:
I agree with this issue and I think it's important to split `ConfigConvert`. Right now, pureconfig required both the encoding and the decoding to be defined but then it uses only one of the two in each part of the library. From a design point of view this is a mistake. Also I think it's normal and to be expected from some users to just define the reading part.

The main issue that we have to address is how to manage this core change. There are two issues to address: breaking the public API and breaking core:

1.  the API doesn't have to break for this issue to be implemented if we keep `ConfigConvert`: we can create a `ConfigEncoder` and a `ConfigDecoder` to convert to/from a configuration and then make `ConfigConvert` extend both. In this way existing code should still work. If we agree on keeping `ConfigConvert` then this is easy to do.
2. the main issue is that we need to change almost everything in the core part of pureconfig. We need to split [`ConfigConvert.scala`](https://github.com/melrief/pureconfig/blob/51021e7436ec55cc6991cdd07c284ee2daee60d9/core/src/main/scala/pureconfig/ConfigConvert.scala#L31) into two files, `ConfigEncoder.scala` and `ConfigDecoder.scala` and then split every functionality in `ConfigConvert.scala` into the encoding and decoding functionality. It's not difficult to do but it's a lot of code to change and we should do it when nobody is working on a PR. I propose to fix the PRs for the next release and then implement this as very last PR

@leifwickland @jcazevedo @ruippeixotog let me know what do you think of this, I would like to do it as soon as possible and @13h3r seems interested in pushing a PR. He is just waiting for our comments.

ruippeixotog:
I agree with the split. I would even go so far as to say that `ConfigDecoder` is what most people will ever have to use. We can still provide a way to implicitly construct `ConfigConvert` from `ConfigReader` and `ConfigWriter` instances and keep compatibility with existing code (as well as convenience for code that both reads and writes configs).

[...]

Issue #85, opened 2017-01-16
Add the line to the error


melrief:
As a user pointed out in our gitter, it is possible to write some information together with an error about where the error happens. Typesafe config has [`ConfigValue::origin`](http://typesafehub.github.io/config/latest/api/com/typesafe/config/Config.html#origin--) which is supposed to be used exactly for this.

PR #110, opened 2017-02-05
switch from Try to Either

melrief:
This is the work that I've done so far on #103 and, as side-effect, on #105 . I opened it in case somebody wants to have a look but it's a WIP. `core` compiles and `core` tests are working but the transition needs some time because I have to check all the possible issues and problems arising.

I've kept intact the old API working with `Try` while I've developed a new one for `Either`, so everything that worked with PureConfig < 0.6.0 should work also with the next version.

I think I'll need two more days of work before this PR will be ready. Also, I won't include any work on the README because that will need a PR after the effort of @ruippeixotog.


melrief:
`core` does work but I'll have to work on the modules because using the old API is deprecated and our current scalac options transform deprecated warnings into errors. I'll fix it once `core` is in good state.

core/src/main/scala/pureconfig/DurationConvert.scala
```diff
@@ -11,13 +16,14 @@ private[pureconfig] object DurationConvert {
...
-      .recoverWith {
-        case ex: NumberFormatException =>
-          val err = s"Could not parse a ${ct.runtimeClass.getSimpleName} from '$durationString'. (try ns, us, ms, s, m, h, d)"
-          new util.Failure(new IllegalArgumentException(err, ex))
-      }
+  def fromString[D](durationString: String, ct: ClassTag[D]): Either[ConfigReaderFailure, Duration] = {
+    try {
+      Right(Duration(addZeroUnit(justAMinute(itsGreekToMe(durationString)))))
+    } catch {
+      case NonFatal(ex) =>
```
leifwickland: I think this could be made more consistent by calling `catchReadError`.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -0,0 +1,18 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package pureconfig.error
+
+import com.typesafe.config.ConfigValue
+
+sealed abstract class ConfigReaderFailure
+final case object CannotConvertNull extends ConfigReaderFailure
```
leifwickland, line 9: Does this not include a key because it's only used for the top level?

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package pureconfig.error
+
+case class ConfigReaderFailures(head: ConfigReaderFailure, tail: Seq[ConfigReaderFailure]) {
```
leifwickland: Suggestions:
- Explicitly state that this is intended to be a non-empty collection of failures.
- Make the `tail` a `List` or `Vector` so that it doesn't have surprising performance characteristics. (My preference is `List`)
- Replace `toSeq` with either `toList` or `toVector` based on the direction you go above.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -56,76 +56,139 @@ trait ConfigConvert[T] {
...
-      // Because we can't trust `fromF` or Typesafe Config not to throw, we wrap the
-      // evaluation in one more `Try` to prevent an unintentional exception from escaping.
-      // `Try.flatMap(f)` captures any non-fatal exceptions thrown by `f`.
-      Try(config.valueType match {
-        case ConfigValueType.STRING => config.unwrapped.toString
-        case _ => config.render(ConfigRenderOptions.concise)
-      }).flatMap(fromF)
+      // Because we can't trust Typesafe Config not to throw, we wrap the
+      // evaluation into a `try-catch` to prevent an unintentional exception from escaping.
+      try {
```
leifwickland, line 91: Idle musing: It's too bad `Try` doesn't have a method akin to scalaz's `\/.valueOr`. I think that'd clean this up.

melrief:
@leifwickland @jcazevedo @ruippeixotog I have a problem and I would like to have your opinion: right now the `ConfigReaderResult[T]` is an alias for [`Either[ConfigReaderFailures, T]`](https://github.com/melrief/pureconfig/pull/110/files#diff-728a2dd82313028848c6c6380835503aR62). This works well except that the only way to add a new methods to `ConfigReaderResult`, like [`get`](https://github.com/melrief/pureconfig/pull/110/files#diff-12918ae14aad73e551e678d26010a1deR166) aka `getOrThrow`, is to create an [implicit wrapper of `Either` and add the new methods](https://github.com/melrief/pureconfig/pull/110/files#diff-12918ae14aad73e551e678d26010a1deR160). This isn't super nice because it means that importing `pureconfig._` will pollute the current namespace with this implicit wrapper. The nastier side-effect that comes to my mind is that this trick makes possible to do `Right("foobar").get` because `Right("foobar")` has type `Either[Nothing, String]`. Also note that the wrapper is not a value class.

I'm not sure whether this is the best way to go. I can think about two possible alternatives:

1. change `ConfigReaderResult[T]` to be a case class containing `Either[ConfigReaderFailures, T]`. Something like
```scala
final case class ConfigReaderResult[T](toEither: Either[ConfigReaderFailures, T])
```
this has the advantage that it's easy to add methods to a class in our space but it also means that we lose the possibility to use directly the right API of `Either` and, worse, all the nice instances added by Scalaz and Cats. Basically we would have `result.toEither` all over the place in `ConfigConvert.scala` to go back to the better and more general type `Either`

2. Same as 1. but instead of using the wrapper everywhere, we just use it as result of the `loadConfig` methods in the PureConfig package object. In this way, both internally and with modules users could use `Either` everywhere but still we would be able to add methods to the result of `loadConfig`. The main drawback of this is that only the `loadConfig` results will have the enriched `Either`

I'm starting to think that 2. could be a better solution than 1. and what I've done until now with this PR. Using `Either` directly would be nice but I really want to add more methods specific to the result of a `ConfigConvert`. What do you guys think about this?

ruippeixotog:
By my own experience, aliasing a standard type just to get a more domain-specific name is a bad idea. It hides from users that they can use all standard `Either` methods on `ConfigReaderResult`, while gaining virtually nothing: no new methods apart from ones added with an implicit wrapper (which could also be added to `Either`) and no flexibility to change its interface later (if you later wanted to turn `ConfigReaderResult` into a case class you would have to forward all `Either` methods to the new class or risk breaking compatibility). 

Wrapping the `Either` in a `ConfigReaderResult` would be less bad, but I'm curious: what methods would you like to add that warrant the extra unwieldiness of the result type?

melrief:
@ruippeixotog this PR is done such that compatibility with `0.5.x` is preserved for `ConfigConvert` even if the underlying type has changed. The only thing that would be nice to add is the ability to call `.get` on the result of `loadConfig[T]` so that developers could easily get a `T` or throw an exception because, I think, the form `loadConfig[T](..).get` is so common that we should provide something like that. Differently from `Try`, `Either` doesn't have a proper `.get` method (there is a `.right.get` that throws a generic exception). I would like to add it so that all the developers using `.get` won't have to change that in the less readable `loadConfig[T](..) match { case Left(failures) => throw new Exception(failures.toString); case Right(t) => t}`. Waht do you think?

ruippeixotog:
I thought we agreed to provide a `loadConfigOrThrow` method for that specific use case. I think that solution is even better than calling `.get` on a result, as the method name explicitly describes the potential effects of the operation. It wouldn't be backwards compatible, sure... but it already isn't for people using the `Try` result for other purposes and the migration is pretty easy. For those reasons I think it is worth it :)

melrief:
Uh ops, I forgot about the `loadConfigOrThrow`. You are right about everything, I'll fix this.

About the alias, ok from my side, let's remove the alias. @leifwickland @jcazevedo do you agree in using `Either[ConfigReaderFailures, T]` everywhere instead of a type alias called `ConfigReaderResult[T]`?

leifwickland:
@melrief,
I agree with @ruippeixotog that aliasing rarely works out as well as I wish it would. While the `Either[...]` type is lengthy, it seems like the least bad option.

jcazevedo:
I also prefer not using the type alias.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -143,9 +145,9 @@ package object pureconfig {
    *
    * @param files Files ordered in decreasing priority containing part or all of a `Config`. Must not be empty.
    */
-  def loadConfigFromFiles[Config: ConfigConvert](files: Traversable[java.io.File]): Try[Config] = {
+  def loadConfigFromFiles[Config: ConfigConvert](files: Traversable[java.io.File]): ConfigReaderResult[Config] = {
     if (files.isEmpty) {
-      new scala.util.Failure(new IllegalArgumentException("The config files to load must not be empty."))
+      ConfigConvert.failWithThrowable[Config](new IllegalArgumentException("The config files to load must not be empty."))
```
jcazevedo: I think this can have a specialized `ConfigReaderFailure` instead of a `ThrowableFailure` with an `IllegalArgumentException`.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -56,76 +56,139 @@ trait ConfigConvert[T] {
...
+        val string = config.valueType match {
+          case ConfigValueType.STRING => config.unwrapped.toString
+          case _ => config.render(ConfigRenderOptions.concise)
+        }
+        eitherToResult(fromF(string))
+      } catch {
+        case NonFatal(t) => failWithThrowable(t)
```
jcazevedo, line 98: This might be out of the scope of this PR, but I believe we can provide more context here, namely about the type we were trying to convert to and possibly about the `ConfigValue` we tried to convert from.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -56,76 +56,139 @@ trait ConfigConvert[T] {
...
-      case CannotConvertNullException => Failure(KeyNotFoundException(keyStr))
-      case KeyNotFoundException(suffix) => Failure(KeyNotFoundException(keyStr + "." + suffix))
-      case WrongTypeException(typ) => Failure(WrongTypeForKeyException(typ, keyStr))
-      case WrongTypeForKeyException(typ, suffix) => Failure(WrongTypeForKeyException(typ, keyStr + "." + suffix))
+  private[pureconfig] def improveFailure[Z](failure: ConfigReaderFailure, keyStr: String): ConfigReaderFailure =
+    failure match {
+      case CannotConvertNull => KeyNotFound(keyStr)
+      case KeyNotFound(suffix) => KeyNotFound(keyStr + "." + suffix)
+      case WrongType(foundTyp, expectedTyp) => WrongTypeForKey(foundTyp, expectedTyp, keyStr)
+      case WrongTypeForKey(foundTyp, expectedTyp, suffix) => WrongTypeForKey(foundTyp, expectedTyp, keyStr + "." + suffix)
```
jcazevedo: We should also match `CollidingKeys` and `UnknownKey` and apply the parent to the current path here.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -135,26 +198,28 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
...
+            val defaultValue = if (hint.useDefaultArgs) default.head else None
+            defaultValue.fold(fail[V](CannotConvertNull))(Right[Nothing, V](_))
+          case (value, converter) =>
+            converter.from(value)
+        },
+        keyStr)
+      // for performance reasons only, we shouldn't clone the config object unless necessary
+      val tailCo = if (hint.allowUnknownKeys) co else co.withoutKey(keyStr)
+      val tailResult = tConfigConvert.value.fromWithDefault(tailCo, default.tail)
+      (headResult, tailResult) match {
```
jcazevedo: We can use the `combineResults` helper here.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -243,61 +309,59 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
...
-          for {
-            indexedEntries <- traverseTry(o.asScala, new collection.mutable.ListBuffer[(Int, ConfigValue)]) {
-              case (keyString, value) => Try(keyString.toInt -> value).recoverWith(reportBadIndex(keyString))
-            }
-            sortedValues = indexedEntries.sortBy(_._1).map(_._2)
-            result <- traverseTry(sortedValues, cbf())(configConvert.value.from)
-          } yield result
+          val z: Either[ConfigReaderFailures, List[(Int, T)]] = Right(List.empty[(Int, T)])
+          def keyValueReader(key: String, value: ConfigValue): ConfigReaderResult[(Int, T)] = {
+            val keyResult = catchReadError(key, _.toInt).left.flatMap(t => fail(CannotConvert(key, "Int",
```
jcazevedo: We should consider using a different failure here other than `CannotConvert`. If we do, we can provide more contextual information by concatenating the remaining path to the `key` through the `improveFailure` method.

jcazevedo: We should still catch `CannotConvert` failures in the `improveFailure` method in order to provide the path to the value we can't convert.

melrief:
@ruippeixotog @leifwickland @jcazevedo I've removed the type alias and introduced the `loadConfigOrThrow` methods. I still have to see @jcazevedo comments. I would like to have your opinion on this PR even if tests are failing because the test failing has nothing to do with the PR (apparently, in Scala 2.10 there is a problem with a very special `java.time.ZonedDateTime` and I'll have to spend sometime to understand why)

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
melrief: Yes, my idea was to use `KeyNotFound` to represent a `CannotConvertNull` with the missing key.

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
melrief: Done with 0ab3792

core/src/main/scala/pureconfig/ConfigConvert.scala
melrief: Also things like [`Either.catchOnly`](https://github.com/typelevel/cats/blob/7919a307cb8f139b0e4963535f14776bd562af01/core/src/main/scala/cats/syntax/either.scala#L262) could be very useful...

Funny fact is that with `Either.catchOnly` the type `Try` doesn't make much sense anymore...

core/src/main/scala/pureconfig/ConfigConvert.scala
melrief: I agree with you, let's do it in another PR.

core/src/main/scala/pureconfig/ConfigConvert.scala
melrief: Done with 39a9c1f. With this PR I'm noticing how important is to model the errors that we can get and how those errors are propagated. I can see in future more PRs that could, for instance, give the context of an exception thrown.

core/src/main/scala/pureconfig/ConfigConvert.scala
melrief: Good catch, done with b614a15

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
leifwickland: That distinction might be worth a comment to help differentiate the purpose of those two types.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -154,4 +261,14 @@ package object pureconfig {
       loadConfig[Config](resolvedTypesafeConfig)
     }
   }
+
+  /**
+   * @see [[loadConfigFromFiles]]
```
melrief, line 266: Maybe we should use the annotation see to avoid duplicating the documentation everywhere because every load method has two versions. Duplicating the documentation is not useful.

jcazevedo:
> (apparently, in Scala 2.10 there is a problem with a very special java.time.ZonedDateTime and I'll have to spend sometime to understand why)

I think it's this OpenJDK bug: https://bugs.openjdk.java.net/browse/JDK-8138664.

melrief:
@ruippeixotog @leifwickland @jcazevedo sorry for the delay on this PR but was taking some time to make it nicer and in the meanwhile we had support for a couple of modules.

The last work on this PR looks much better than before and I think the general API is nice and compatible with the previous one.

The major changes are:

1. The `ConfigConvert` is more consistent regarding method names. Methods matching `from[..]Reader[..]` take in input only one function, the reading function, while methods matching `from[..]Convert[..]` take in input both the reading and the writing functions
2. All the helper methods, the one starting with `from`, have three versions now:
    1. `String => Either[ConfigReaderFailure, T]`, e.g. [`fromStringReader`](https://github.com/melrief/pureconfig/pull/110/files#diff-728a2dd82313028848c6c6380835503aR144), which are the best ones because `Either` is very rich
    2. `String => Try[T]`, e.g. [`fromStringReaderTry`](https://github.com/melrief/pureconfig/pull/110/files#diff-728a2dd82313028848c6c6380835503aR149), which are similar to the old one but `Failure(e)` is encapsulated into a `CannotConvert` failure. They need a `ClassTag` for the name of `T`
    3. `String => Option[T]`, e.g. [`fromStringReaderOpt`](https://github.com/melrief/pureconfig/pull/110/files#diff-728a2dd82313028848c6c6380835503aR153), which support `Option` and convert `None` to a `CannotConvert` failure without message. `Option` is not very good for errors but there are cases in which it can be used. They need a `ClassTag` for the name of `T`
3. old methods have been deprecated because there is no easy way to get a `CannotConvert` without a `ClassTag`. I think it's the best option, also considering that we provide an alternative with the same semantic.
4. we have support for multiple errors, so this solved also #105 
5. we now have [`loadConfigOrThrow`](https://github.com/melrief/pureconfig/pull/110/files#diff-728a2dd82313028848c6c6380835503aR153) together with `loadConfig` to throw a nicely descriptive exception in case of failure

Otherwise there are many small changes but nothing too big. I've tried to improve error messages all over the place and some small tests are promising:

```scala
scala> val conf = ConfigFactory.parseString("""{ i: "foobar" }""")
scala> case class Conf(i: Int, z: String)
scala> loadConfigOrThrow[Conf](conf)
pureconfig.error.ConfigReaderException: Cannot convert configuration to a value of class Conf. Failures are:
  Keys not found:
    - 'z'
  Value conversion failures:
    - cannot convert 'foobar' to type 'Int' because java.lang.NumberFormatException: For input string: "foobar"
```

I think this is ready for another review when you guys have time.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -33,10 +32,10 @@ trait ConfigConvert[T] {
    * Convert the given configuration into an instance of `T` if possible.
    *
    * @param config The configuration from which load the config
-   * @return `Success` of `T` if the conversion is possible, `Failure` with the problem if the
-   *         conversion is not
+   * @return either a list of failures wrapped inside a `ConfigReaderFailures` or an object of type `Right[T]`
```
ruippeixotog: Minor nitpick: this should read either "wrapped inside a Left[ConfigReaderFailures]" or "an object of type T" - as it is it doesn't typecheck in my head 😄 . I prefer the latter one.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -56,76 +55,197 @@ trait ConfigConvert[T] {
...
   private[pureconfig] trait WrappedDefaultValueConfigConvert[Wrapped, SubRepr <: HList, DefaultRepr <: HList] extends WrappedConfigConvert[Wrapped, SubRepr] {
-    final def from(config: ConfigValue): Try[SubRepr] =
-      Failure(
-        new UnsupportedOperationException("Cannot call 'from' on a WrappedDefaultValueConfigConvert."))
-    def fromWithDefault(config: ConfigValue, default: DefaultRepr): Try[SubRepr] = config match {
-      case co: ConfigObject =>
-        fromConfigObject(co, default)
-      case other =>
-        Failure(WrongTypeException(config.valueType().toString))
+    final def from(config: ConfigValue): Either[ConfigReaderFailures, SubRepr] =
```
ruippeixotog: Why does `WrappedDefaultValueConfigConvert` need to implement `ConfigConvert`? It seems that it doesn't suit the interface and this is never used as a `ConfigConvert` anyway, so it should have its own interface.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -56,76 +55,197 @@ trait ConfigConvert[T] {
...
 
   implicit def hNilConfigConvert[Wrapped](
     implicit
     hint: ProductHint[Wrapped]): WrappedDefaultValueConfigConvert[Wrapped, HNil, HNil] = new WrappedDefaultValueConfigConvert[Wrapped, HNil, HNil] {
 
-    override def fromConfigObject(config: ConfigObject, default: HNil): Try[HNil] = {
-      if (!hint.allowUnknownKeys && !config.isEmpty) Failure(UnknownKeyException(config.keySet.iterator.next))
-      else Success(HNil)
+    override def fromConfigObject(config: ConfigObject, default: HNil): Either[ConfigReaderFailures, HNil] = {
+      if (!hint.allowUnknownKeys && !config.isEmpty) fail(UnknownKey(config.keySet.iterator.next))
```
ruippeixotog: Now that you can return multiple errors it makes sense to add an `UnknownKey` failure for each key in `config.keySet`, and not only for the first one.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -97,11 +99,116 @@ package object pureconfig {
...
+    }
+  }
+
+  /**
+   * Load a configuration of type `Config` from the standard configuration files
+   *
+   * @return the configuration
+   */
+  @throws[ConfigReaderException[_]]
+  def loadConfigOrThrow[Config](implicit conv: ConfigConvert[Config], ct: ClassTag[Config]): Config = {
```
ruippeixotog, line 120: It doesn't seem reasonable to require a `ClassTag` just for the purpose of printing a pretty error message. Have you considered using the `null` default implicit value trick, like:

```scala
def loadConfigOrThrow[Config](implicit conv: ConfigConvert[Config], ct: ClassTag[Config] = null): Config
```

If you define the method this way the compiler will provide a `ClassTag` when it can and you can still handle `null` class tags in a special way in `ConfigReaderException` - for example, by printing a generic header message instead of specifying the class name.

ruippeixotog (changes requested):
LGTM in general. It's a great enhancement to the API, good work! :)

Just one thing, I'm not very fond of how our failure summary is being printed. A thing we could improve in a future pull request is to have every failure contain a config key and group the failures by key at least in the exception message. In other words, instead of having:

```
  Keys not found:
    - 'z'
    - 'myobject.key'
  Value conversion failures:
    - cannot convert 'foobar' to type 'Int' because java.lang.NumberFormatException
```

I think it's more natural to have a list by key, such as:

```
foobar: cannot convert to type 'Int' because java.lang.NumberFormatException
myobject.key: required key not found
z: required key not found
```

If you agree with that alternative way of presenting the errors, I can work on it later.

core/src/main/scala/pureconfig/ConfigConvert.scala
jcazevedo: I think it was me who made `WrappedDefaultValueConfigConvert` implementing `ConfigConvert`, but, in retrospect, I agree with you that it doesn't make much sense. This `from` method is pretty much inaccessible code.

jcazevedo:
> I think it's more natural to have a list by key

I also prefer having the failures listed by key. From the perspective of an user fixing problems with a Config, most of the times just looking at the key is enough to identify the error.

core/src/main/scala/pureconfig/ConfigConvert.scala
melrief: Fixed with d509127

core/src/main/scala/pureconfig/ConfigConvert.scala
melrief: Fixed with 7281316, good catch!

core/src/main/scala/pureconfig/ConfigConvert.scala
melrief: Nice one, fixed with 71c03f8

core/src/main/scala/pureconfig/package.scala
melrief: I propose to do this in another PR because I'm now aware of all the implications of using `null` there. Let's open another issue, if needed.

melrief:
@ruippeixotog @jcazevedo fine by me with the errors but I'd like to move on with this PR. Error messages can be solved in another PR but the sooner this is in the new master, the easier will be to integrate new modules for external developers. The only thing that worries me right now is #92 because merging this will undo a lot of work done for that.

ruippeixotog (approved):
I agree with doing the remaining two suggestions on another pull request.

Before releasing the new version we shouldn't forget to update the documentation and the changelog :)

melrief:
@leifwickland @jcazevedo I've got one approval from @ruippeixotog but before I merge this core-breaking PR, I would like to have the green light also from you. Do you think you can find some time to review this PR?

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
```diff
@@ -0,0 +1,24 @@
...
+package pureconfig.error
+
+/**
+ * A non-empty list of ConfigReader failures
+ */
+case class ConfigReaderFailures(head: ConfigReaderFailure, tail: List[ConfigReaderFailure]) {
+
+  def toList: List[ConfigReaderFailure] = head +: tail
+
+  def +(failure: ConfigReaderFailure): ConfigReaderFailures =
```
leifwickland: Suggestion: I would go with `+:` for uniformity with the standard `List`, preserve the visual ordering semantics that of having the new head on the left, and to avoid the weirdness around string concatenation.
Feel free to disagree.

core/src/main/scala/pureconfig/package.scala
leifwickland: @melrief Did you want to do anything with this comment? I'm just asking so it doesn't get lost.

core/src/main/scala/pureconfig/DurationConvert.scala
```diff
@@ -11,13 +15,8 @@ private[pureconfig] object DurationConvert {
...
    */
-  def fromString[D](durationString: String, ct: ClassTag[D]): Try[Duration] = {
-    Try { Duration(addZeroUnit(justAMinute(itsGreekToMe(durationString)))) }
-      .recoverWith {
-        case ex: NumberFormatException =>
-          val err = s"Could not parse a ${ct.runtimeClass.getSimpleName} from '$durationString'. (try ns, us, ms, s, m, h, d)"
-          new util.Failure(new IllegalArgumentException(err, ex))
-      }
+  val fromString: String => Either[ConfigReaderFailure, Duration] = {
+    catchReadError(string => Duration(addZeroUnit(justAMinute(itsGreekToMe(string)))))
```
leifwickland: Could we preserve the improved error message in this case?
I think something like this would do it:
```
catchReadError{ string =>
  Try(Duration(addZeroUnit(justAMinute(itsGreekToMe(string)))))
  .recoverWith {		
    case ex: NumberFormatException =>		
      val err = s"Could not parse a ${ct.runtimeClass.getSimpleName} from '$durationString'. (try ns, us, ms, s, m, h, d)"		
     new util.Failure(new IllegalArgumentException(err, ex))		
  }.get  
}

leifwickland (approved):
👍 
I made a few suggestions. 
I'd like to restore the more descriptive error message for `Duration`s.

jcazevedo (approved):
LGTM! Thanks for the great improvement to the API.

I was delaying the start of #85 until we had the definitions of failures more or less settled, and I think it's OK to start now.

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
melrief: Good idea, done with 8d9a659

core/src/main/scala/pureconfig/DurationConvert.scala
melrief: I've improved the error message with 005753f. Note that I've decided to follow another route because the error message can be found inside the exception and the string is already in the `CannotConvert` failure. Basically we don't use `Try` or throw exceptions anymore and that should improve everything :smile:

melrief:
Merging as soon as the CI is green. Thanks everybody for reading this huge PR :smiley:

PR #111, opened 2017-02-07
Add support for Squants.

cranst0n:


modules/squants/build.sbt
```diff
@@ -0,0 +1,57 @@
...
+}
+
+publishArtifact in Test := false
+
+pomExtra := (
+    <scm>
+      <url>git@github.com:melrief/pureconfig.git</url>
+      <connection>scm:git:git@github.com:melrief/pureconfig.git</connection>
+    </scm>
+    <developers>
```
jcazevedo, line 38: You should add yourself as a developer here.

modules/squants/src/test/scala/pureconfig/module/squants/SquantsConvertTest.scala
```diff
@@ -0,0 +1,124 @@
...
+  checkDimension(time.Time)
+
+  it should "parse Money" in forAll { (m: Money) =>
+    checkConfig(SquantConfig(m))
+  }
+
+  case class SquantConfig[T](value: T)
+
+  def checkDimension[T <: Quantity[T]](dim: Dimension[T])(implicit tag: ClassTag[T], cc: ConfigConvert[T]): Unit = {
+    val quantityName = tag.toString.reverse.takeWhile(_ != '.').reverse
```
leifwickland: Nitpick: I think `tag.runtimeClass.getSimpleName` is a slightly more straightforward way to do this.

modules/squants/src/main/scala/pureconfig/module/squants.scala
```diff
@@ -0,0 +1,184 @@
...
+      case regex(value, Celsius.symbol) => Success(Celsius(value.toDouble))
+      case regex(value, "c") => Success(Celsius(value.toDouble))
+      case regex(value, "C") => Success(Celsius(value.toDouble))
+      case regex(value, Kelvin.symbol) => Success(Kelvin(value.toDouble))
+      case regex(value, "k") => Success(Kelvin(value.toDouble))
+      case regex(value, "K") => Success(Kelvin(value.toDouble))
+      case regex(value, Rankine.symbol) => Success(Rankine(value.toDouble))
+      case regex(value, "r") => Success(Rankine(value.toDouble))
+      case regex(value, "R") => Success(Rankine(value.toDouble))
+      case _ => Failure(QuantityParseException("Unable to parse Temperature", s))
```
leifwickland: Suggestion: 
What would you think about adding a case above this one something like:
```
case regex(_, unknownUnit) => Failure(new Exception(s"PureConfig library programmer error: The regex '$regex' found unit '$unknownUnit' which was not defined."))
```

I recognize this is intended to be temporary and is copied from squants, so feel free to disregard.

modules/squants/src/main/scala/pureconfig/module/squants.scala
```diff
@@ -0,0 +1,184 @@
...
+    ConfigConvert.nonEmptyStringConvert[Frequency](Frequency.apply, _.toString)
+  implicit val timeConfigConvert =
+    ConfigConvert.nonEmptyStringConvert[Time](Time.apply, _.toString)
+
+  // This is temporary until https://github.com/typelevel/squants/pull/183 is released (1.2.0 ???)
+  // Without it, ScalaCheck has a pretty easy time breaking the conversion
+  private[this] def parseTemperature(s: String): Try[Temperature] = {
+    val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)[ °]*(f|F|c|C|k|K|r|R)".r
+    s match {
+      case regex(value, Fahrenheit.symbol) => Success(Fahrenheit(value.toDouble))
```
leifwickland: Suggestion:
I'd be tempted to write this as the following to reduce duplication:
```
s match {
  case regex(value, unit) => unit match {
    case Fahrenheit.symbol | "f" | "F" => Success(Fahrenheit(value.toDouble))
    case Celsius.symbol | "c" | "F" => Success(Celsius(value.toDouble))
    // etc
  }
  case _ => Failure(QuantityParseException("Unable to parse Temperature", s))
}
```

I recognize this is intended to be temporary and is copied from squants, so feel free to disregard.

modules/squants/src/main/scala/pureconfig/module/squants.scala
```diff
@@ -0,0 +1,184 @@
...
+  // time
+  implicit val frequencyConfigConvert =
+    ConfigConvert.nonEmptyStringConvert[Frequency](Frequency.apply, _.toString)
+  implicit val timeConfigConvert =
+    ConfigConvert.nonEmptyStringConvert[Time](Time.apply, _.toString)
+
+  // This is temporary until https://github.com/typelevel/squants/pull/183 is released (1.2.0 ???)
+  // Without it, ScalaCheck has a pretty easy time breaking the conversion
+  private[this] def parseTemperature(s: String): Try[Temperature] = {
+    val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)[ °]*(f|F|c|C|k|K|r|R)".r
```
leifwickland: Suggestion:
It feels a bit odd to me that multiple degree symbols are allowed.
I would replace `[ °]*` with ` *°? *` to only allow a single degree symbol.

I recognize this is intended to be temporary and is copied from squants, so feel free to disregard.

modules/squants/build.sbt
```diff
@@ -0,0 +1,62 @@
...
+      <developer>
+        <id>melrief</id>
+        <name>Mario Pastorelli</name>
+        <url>https://github.com/melrief</url>
+      </developer>
+      <developer>
+        <id>leifwickland</id>
+        <name>Leif Wickland</name>
+        <url>https://github.com/leifwickland</url>
```
leifwickland: You ought to remove me. I don't deserve credit.

modules/squants/build.sbt
cranst0n: Added!

leifwickland (approved):
Awesome addition, @cranst0n!
Thank you!
I've made some suggestions. Take them or leave them.

cranst0n:
Thanks @leifwickland. I've updated based on your feedback.

modules/squants/src/main/scala/pureconfig/module/squants.scala
```diff
@@ -0,0 +1,179 @@
...
+  // Without it, ScalaCheck has a pretty easy time breaking the conversion
+  private[squants] def parseTemperature(s: String): Try[Temperature] = {
+    val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)*°? *(.{1})".r
+    s match {
+      case regex(value, unit) => unit match {
+        case Fahrenheit.symbol | "f" | "F" => Success(Fahrenheit(value.toDouble))
+        case Celsius.symbol | "c" | "C" => Success(Celsius(value.toDouble))
+        case Kelvin.symbol | "k" | "K" => Success(Kelvin(value.toDouble))
+        case Rankine.symbol | "r" | "R" => Success(Rankine(value.toDouble))
+        case unknownUnit => Failure(new Exception(s"PureConfig library programmer error: The regex '$regex' found unit '$unknownUnit' which was not defined."))
```
leifwickland: @cranst0n I may have confused you. I was thinking that this error would only occur if `regex` included a unit character which wasn't supported here. I was hoping this wouldn't be an error that a normal user of the library would see.
The way you've defined things now, an input of `13°` will hit this case, reporting `°` as an invalid unit. `13(space)` will also hit this case.

I think you should go back to a defined set of expected units in the regex, causing this case to only be hit if the two lists of units end up diverging.

cranst0n: @leifwickland Maybe I'm being thick here but I'm not sure exactly how you would get the behavior that you're looking for. If you match all supported units in the regex, something like
```scala
case regex(value, unit) => unit match {
  // cases covering F/C/K/R
}
case regex(_, unknownUnit) => ...
case _ => ...
```
The second top level case with `unknownUnit` will never hit. Or maybe I'm missing something on how scala's regex pattern matching works.

leifwickland: @cranst0n 
This is me being an old paranoid engineer assuming that somebody somewhere is going to screw this up. 
This case that I'm describing would only be hit if somebody changed the regex but forgot to add a matching case for the unit.
The whole point is that the regex and the cases have diverged.
Since there's not a nice way to make the compiler check for that, I was adding a runtime check to give the pureconfig author a hint how we messed up.
Belt and suspenders I call it.
Sorry I confused you. Feel free to ignore my suggestion.

modules/squants/src/main/scala/pureconfig/module/squants.scala
derekmorr: Would you open an issue with squants for this? I'm a maintainer on squants and would be happy to look at a PR or issue.

modules/squants/src/main/scala/pureconfig/module/squants.scala
derekmorr: Would you open an issue with squants for this? I'm a maintainer on squants and would be happy to look at a PR or issue.

modules/squants/src/main/scala/pureconfig/module/squants.scala
derekmorr: Would you open an issue with squants for this? I'm a maintainer on squants and would be happy to look at a PR or issue.

cranst0n:
@leifwickland I can appreciate a belt and suspenders approach. See my latest update. I've opened an issue here: typelevel/squants#189 for any further suggestions.

derekmorr:
I opened a PR in squants to fix the double-degree issue: https://github.com/typelevel/squants/pull/190

I adapted the regex from the version here. I'm using:

```"([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)*[ ]*°? *(f|F|c|C|k|K|r|R)".r```

File: modules/squants/src/main/scala/pureconfig/module/squants.scala [hunk 8]
```diff
@@ -0,0 +1,179 @@
...
+  implicit val timeConfigConvert =
+    ConfigConvert.nonEmptyStringConvert[Time](Time.apply, _.toString)
+
+  // This is temporary until https://github.com/typelevel/squants/pull/183 is released (1.2.0 ???)
+  // Without it, ScalaCheck has a pretty easy time breaking the conversion
+  private[squants] def parseTemperature(s: String): Try[Temperature] = {
+    val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)[ °]*(f|F|c|C|k|K|r|R)".r
+    s match {
+      case regex(value, unit) => unit match {
+        case Fahrenheit.symbol | "f" | "F" => Success(Fahrenheit(value.toDouble))
```
leifwickland: Now that I have looked at squants a bit more, it occurs to me that there's no reason to have `Fahrenheit.symbol` etc in the cases because they can't match. They all include the degree symbol, which the regex will not include the capturing group.

modules/squants/src/main/scala/pureconfig/module/squants.scala
```diff
@@ -0,0 +1,179 @@
...
+  // time
+  implicit val frequencyConfigConvert =
+    ConfigConvert.nonEmptyStringConvert[Frequency](Frequency.apply, _.toString)
+  implicit val timeConfigConvert =
+    ConfigConvert.nonEmptyStringConvert[Time](Time.apply, _.toString)
+
+  // This is temporary until https://github.com/typelevel/squants/pull/183 is released (1.2.0 ???)
+  // Without it, ScalaCheck has a pretty easy time breaking the conversion
+  private[squants] def parseTemperature(s: String): Try[Temperature] = {
+    val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)[ °]*(f|F|c|C|k|K|r|R)".r
```
leifwickland: It looks like @derekmorr is updating this in squants. If you want to follow along now, it'd be:
```
"([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)* *°? *(f|F|c|C|k|K|r|R)".r
```

modules/squants/src/main/scala/pureconfig/module/squants.scala
derekmorr: Right. This is the current version in squants - https://github.com/typelevel/squants/blob/6affa29e75c3d5109f8345e2263af9a24ee35fff/shared/src/main/scala/squants/thermal/Temperature.scala#L143 :

```scala
def apply(s: String): Try[Temperature] = {
    val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)* *°? *(f|F|c|C|k|K|r|R)".r
    s match {
      case regex(value, unit) => unit match {
        case "f" | "F" => Success(Fahrenheit(value.toDouble))
        case "c" | "C" => Success(Celsius(value.toDouble))
        case "k" | "K" => Success(Kelvin(value.toDouble))
        case "r" | "R" => Success(Rankine(value.toDouble))
      }
      case _ => Failure(QuantityParseException("Unable to parse Temperature", s))
    }
  }
```

This version is probably going to be merged later today.

derekmorr:
I don't know if you want to update the tests for this, but I added in a few more temperature-parsing tests into squants earlier today:

```scala
  they should "be flexible in parsing strings with regard to degree symbol and whitespace" in {
    Temperature("10.22 f").get should be (Fahrenheit(10.22))
    Temperature("10.22 °f").get should be (Fahrenheit(10.22))
    Temperature("10.22  °f").get should be (Fahrenheit(10.22))
    Temperature("10.22 ° f").get should be (Fahrenheit(10.22))
    Temperature("10.22 °   f").get should be (Fahrenheit(10.22))
    Temperature("10.22  °   f").get should be (Fahrenheit(10.22))
    Temperature("10.22°f").get should be (Fahrenheit(10.22))

    Temperature("10.22°°f").failed.get should be (QuantityParseException("Unable to parse Temperature", "10.22°°f"))
  }
```

Since you're duplicating the parsing code, it might be good to include the tests.

melrief:
@leifwickland @jcazevedo thanks for having a look at this PR, I didn't have time. @leifwickland already approved the PR but I'll wait also for the approval from @jcazevedo as he commented on this PR too (just one thing about the authors, but I'll wait for LGTM anyway). I would like to merge this before #103 so I can merge this in my PR and switch also this module to the new result type.

melrief (approved):
Nothing to comment, everything is well done and I like how you wrote the tests. I'll publish an artifact for `0.5.1` right now.

melrief:
I've published the package and announced it https://twitter.com/mapastr/status/830197666898444288

PR #115, opened 2017-02-08
parse UUIDs

### LINKED ISSUE #114: parse UUID
Pureconfig should parse UUIDs from config as a java.util.UUID.

derekmorr:
This fixes #114

### DIFF
--- a/core/src/test/scala/pureconfig/PureconfSuite.scala
+++ b/core/src/test/scala/pureconfig/PureconfSuite.scala
@@ -7,12 +7,14 @@ import java.io.PrintWriter
 import java.net.URL
 import java.nio.file.{ Files, Path }
 import java.time._
+import java.util.UUID
 import java.util.concurrent.TimeUnit
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable._
 import scala.concurrent.duration.{ Duration, FiniteDuration }
 import scala.util.{ Failure, Success, Try }
+
 import com.typesafe.config.{ ConfigFactory, Config => TypesafeConfig, _ }
 import org.joda.time.DateTime
 import org.joda.time.format.ISODateTimeFormat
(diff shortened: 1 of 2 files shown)


core/src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -592,6 +594,25 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with TryVal
...
+  case class ConfWithUUID(uuid: UUID)
+
+  it should "be able to read a config with a UUID" in {
+    val expected = "d25aed6a-ef6d-4c10-954c-02edc949aef1"
+    val config = loadConfig[ConfWithUUID](ConfigValueFactory.fromMap(Map("uuid" -> expected).asJava).toConfig)
+    config.toOption.value.uuid shouldBe UUID.fromString(expected)
+  }
+
+  it should "round trip a UUID" in {
+    saveAndLoadIsIdentity(ConfWithUUID(UUID.randomUUID()))
```
leifwickland: This would probably be better expressed as a `scalacheck` test like:
```
it should "round trip a UUID" in forAll { (uuid: UUID) =>
   saveAndLoadIsIdentity(ConfWithUUID(uuid))
}
```
I believe that the Scalacheck `Gen[UUID]` should already be in scope. If not, you may need to import something additional.

leifwickland (approved):
This seems like a solid addition to me.
Thanks, @derekmorr!
I've made a suggestion I'd like you to consider.

core/src/test/scala/pureconfig/PureconfSuite.scala
derekmorr: Looks good. Should I also change """it should "be able to read a config with a UUID"""" to use  a property-based test ?

leifwickland: @derekmorr Eh, I wouldn't. Other folks might. I feel like this test covers that case. I kind of like having one simple case with fixed data in addition to my property tests.

derekmorr: Ok, I pushed up a new revision with this change.

leifwickland:
@derekmorr LGTM! Thanks.

leifwickland:
@melrief @ruippeixotog @jcazevedo Would a couple of you be willing to give this a look before we call it ready to merge? 
I think it's solid, but would sure appreciate a second or third set of eyes.

jcazevedo (approved):
For me, this is good to merge as is. @derekmorr: Thanks for the PR!

melrief:
Thanks for the PR @derekmorr ! I would like to add you to the [CONTRIBUTORS](https://github.com/melrief/pureconfig/blob/master/CONTRIBUTORS) file. If you agree, please do another PR in which you just add one line with your name, optionally your email and your username. Thanks again!

ruippeixotog:
LGTM! I'll add this to the README and CHANGELOG later.

ruippeixotog (commented):
LGTM! I'll add this to the README and CHANGELOG later.

PR #123, opened 2017-02-11
parse java.nio.file.Path

### LINKED ISSUE #122: Support loading Path
The library should support loading java.nio.file.Path by parsing Strings from config.

derekmorr:
This resolves #122 .

Some notes:

The example docs used `java.nio.file.Path` as a basis for making a custom `Converter`. So I changed the example to use `javax.security.auth.kerberos.KerberosPrincipal` since it's also a String wrapper and seemed relevant to the email example.

In `PureconfSuite`, in `ConfWithPath`, I couldn't name the field `path` because that is overwritten by `loadConfig()` -- it's the name of a built-in variable that takes precedence.


jcazevedo (commented):
LGTM overall.

I don't like the use of `KerberosPrincipal` in the examples, though. It doesn't seem to me the most adequate type to represent what we're assuming to be an email address.

leifwickland:
@jcazevedo Do you have a good alternative type to use to represent an email address? I couldn't find one in the standard java library.

leifwickland (approved):
LGTM! 👍 
Thanks, @derekmorr.

derekmorr:
I'm happy to change the example to something else. Not sure what would be appropriate though.  I could pick another class from the Java standard library, but that seems like whack-a-mole -- as the project expands it will probably add support for most of the stabdard library types that can be trivially parsed from strings. I wonder if it would be better to create a custom String wrapper class to use for the example.

derekmorr:
I realized that the sample code in `example/` wasn't updated to reflect the changes in `example.md`. I'll hold off pushing an update until we decide if `KerberosPrincipal` is staying, but this PR shouldn't be merged until the example code is updated.

jcazevedo:
I don't have a good alternative. Maybe creating a custom String wrapper (that isn't a `case class`) is better for the example. We can leave the `KerberosPrincipal` if you guys prefer it, though.

melrief:
@derekmorr I agree that `KerberosPrincipal` is not as super nice but I also agree that eventually all the classes in the Java stdlib will be covered. The only thing that I have in mind is a custom class that is not a case class. What do you think?

derekmorr:
That sounds fine. I'll update the PR.

derekmorr:
I updated the PR with a basic `Email` class wrapping a String and updated the docs to reflect this.

I had to leave in the implicit converter for `Path` because the example project pull in 0.5.1 which doesn't natively support `Path` yet. 

I'm getting compiler errors on 2.10:
```
pureconfig/example/src/main/scala/pureconfig/example/conf/Main.scala:17: could not find implicit value for parameter conv: pureconfig.ConfigConvert[pureconfig.example.conf.Config]
```

not sure if that's a blocker.

melrief:
Thanks @derekmorr for the contribution!

About the error, don't worry: the example was intended for Scala 2.11 and it doesn't import macro paradise, which is necessary to use pureconfig in Scal 2.10. We can fix that another day.

PR #126, opened 2017-02-14
add macro paradise; set crossScalaVersions; enable compile flags

derekmorr:
This updates `build.sbt` in the example project to:

* add macro paradise
* explicitly set crossScalaVersions
* enable more stringent compiler flags.

This resolves #125 .

PR #128, opened 2017-02-14
support parsing URIs

### LINKED ISSUE #127: Parse java.net.URI
We currently support parsing URLs, but Java also supports URIs. Since not all URLs are URIs, we should support parsing both.

derekmorr:
This would resolve #127


derekmorr:
Err, this branch was built wrong and includes changes for #126 . I'll close, rebase, and push again.

PR #129, opened 2017-02-14
add support for java.net.URI

### LINKED ISSUE #127: Parse java.net.URI

derekmorr:
This would resolve #127


leifwickland (approved):
LGTM 👍 
Thanks, @derekmorr!

melrief:
Thanks @derekmorr!

PR #131, opened 2017-02-14
support KerberosPrincipal

### LINKED ISSUE #130: Support parsing KerberosPrincipal
In the discussion for #123 we discussed using `KerberosPrincipal` in the example but decided against it. So I decided to add support for it to core, since it's a trivial string converter.

derekmorr:
This would close #130

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ Currently supported types for fields are:
   configured first - see [Configurable converters](#configurable-converters));
 - [`java.util.UUID`](https://docs.oracle.com/javase/8/docs/api/java/util/UUID.html);
 - [`java.nio.file.Path`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html)
+- [`javax.security.auth.kerberos`](https://docs.oracle.com/javase/8/docs/api/javax/security/auth/kerberos/KerberosPrincipal.html)
 - Typesafe `ConfigValue`, `ConfigObject` and `ConfigList`;
 - case classes;
 - sealed families of case classes (ADTs).
--- a/core/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/core/src/main/scala/pureconfig/ConfigConvert.scala
@@ -18,6 +18,7 @@ import java.net.URL
 import java.nio.file.{ Path, Paths }
 import java.time._
 import java.util.UUID
+import javax.security.auth.kerberos.KerberosPrincipal
 
 import scala.concurrent.duration.{ Duration, FiniteDuration }
 
@@ -390,6 +391,7 @@ trait LowPriorityConfigConvertImplicits {
   implicit val readURL = stringConvert[URL](s => Try(new URL(s)), _.toString)
   implicit val readUUID = stringConvert[UUID](s => Try(UUID.fromString(s)), _.toString)
   implicit val readPath = stringConvert[Path](s => Try(Paths.get(s)), _.toString)
+  implicit val readKerberosPrincipal = stringConvert[KerberosPrincipal](s => Try(new KerberosPrincipal(s)), _.toString)
 
   implicit val readConfig: ConfigConvert[Config] = new ConfigConvert[Config] {
     override def from(config: ConfigValue): Try[Config] = config match {
--- a/core/src/test/scala/pureconfig/PureconfSuite.scala
+++ b/core/src/test/scala/pureconfig/PureconfSuite.scala
@@ -9,6 +9,7 @@ import java.nio.file.{ Files, Path, Paths }
 import java.time._
 import java.util.UUID
 import java.util.concurrent.TimeUnit
+import javax.security.auth.kerberos.KerberosPrincipal
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable._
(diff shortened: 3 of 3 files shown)


ruippeixotog:
Thanks for the pull request @derekmorr! To be honest though, I don't think `KerberosPrincipal` is general-purpose enough to warrant being in the pureconfig core.

I'm a bit skeptical about adding this to the PureConfig repo at all, as it seems adequate only to very specific, rare use cases - there is a broad range of classes included in the JDK and most projects use only a very small part of them, so we shouldn't strive to support them all in this repository. However, if `javax.security` has enough models commonly loaded from configs, we could create a `pureconfig-security` submodule with converters for that package.

melrief:
@ruippeixotog there isn't an issue in supporting all the classes in the standard jdk, even if not widely used. I'm leaning toward accepting this PR. Do you have a good reason to not include as many classes as possible?

ruippeixotog:
I can think of a few reasons and I'll try to expose them here, but for me the bottom line is that polluting scopes with objects the user doesn't need is a bad design principle - being included in the JDK _per se_ does not make a class special to me on that regard.

The main reason why we currently separate PureConfig in modules instead of adding joda-time, enumeratum and others as simple library dependencies, I think, is not polluting the users' classpath with classes they don't use. From performance issues in JVM to namespace polluting, we should strive to minimize the class payload we force our users to have.

In my view, the fact that the class comes with the JDK don't change that. According to [this source](http://stackoverflow.com/questions/3112882/how-many-classes-are-there-in-java-standard-edition), JDK 8 has a whopping 4240 classes (I'm aware not all of them are "models", but it gives an idea of how large it is). The JDK is far from a "core" library containing only building blocks for more high-level libraries - it has a whole range of `javax` extensions (of which `KerberosPrincipal` is a part of), AWT and other enterprise-directed classes. We must not forget each implementation of `ConfigConvert` we add is compiled to an anonymous class (even in 2.12, as our `ConfigConvert` is not a SAM), which effectively makes the policy to support all JDK classes costly.

Other aspect I can think of is on the implicit resolution. It bothers me that importing `pureconfig._` can bring hundreds of implicit instances in scope if we start accepting all JDK classes in the core. Despite the obvious namespace pollution, I'm a bit afraid it can make the compilation slower.

For those reasons, I'm against adding all JDK classes to the `pureconfig` core module - we should strive to use modules as we see fit for these cases. Maybe @jcazevedo and @leifwickland can give their opinion on this, too :)

jcazevedo:
I'm not against adding this to PureConfig, but I agree with @ruippeixotog that it shouldn't be part of the `core` module. 

I think we can make an effort to support all models in the standard Java API (things in the `java.*` package) sensible to be loaded from a configuration directly in the `core` module, but try to scope things in the standard Java extensions (things in the `javax.*` package) in more specific modules.

derekmorr:
Even if the JDK does have 4,000-some classes, realistically we wouldn't be supporting anywhere near all of them. There's no reason to think we'd be reading AWT- or Swing-related classes from a config file, for example (to say nothing of `org.omg.CORBA`,` javax.xml`, ` javax.sound`, `javax.transaction`, etc). So I think the class-count argument is overstated. 

Ultimately, we're talking about supporting immutable value classes, which are fairly rare in the JDK. And further, we'd only support a subset of those which are relevant to PureConfig (do we really think anyone will read a `javax.xml.ws.Service` out of a config file?)

I'm in favor of supporting `KerberosPrincipal` only because Kerberos is so prevalent in enterprise environments, and it's a straight-forward String wrapper type.

I've casually inspected javax.*, and there only appear to be a few classes that we'd realistically support:

- `javax.security.auth.kerberos.KerberosPrincipal`
- `javax.security.auth.x500.X500Principal`
- `org.ietf.jgss.Oid`
- maybe some of the `javax.crypto.*` and `javax.net.ssl.*` classes if only to verify that a String is valid (for example that a String containing an encryption algorithm name is valid).

leifwickland:
I agree that muddying the core with a proliferation of supported types is a bad idea. 
However, I'm sympathetic to @derekmorr's argument. I suspect his point about there being few immutable value classes is spot on.
In the end, I very much doubt that it would have much adverse impact.
My vote is to accept the PR.

melrief:
Sorry @derekmorr to make you wait, I wanted to work on the release `0.6.0` and I decided to postpone the decision about this PR. This is the first time that half of the authors disagree with the other half. I made it clear with the authors that when there isn't a clear majority in accepting something into `core`, then we should not add it to the `core`. I'll apply the rule for this PR. This means that we won't accept the PR like this. But I still would like to have support for `javax` somewhere. I think there is an agreement among the authors about creating a `pureconfig-javax` module containing the immutable data structures from the java stdlib extensions. If you want, you could adapt this PR and move the support for `KerberosPrincipal` there. For what it matters, I think you had good points in your comments but creating a submodule won't invalidate any of those points. Let me know what do you think.

derekmorr:
Sounds good. I'll have some time this weekend and will work up a revised PR.

ruippeixotog:
Sorry for not having followed up on this discussion, I was on holidays :) I just want to leave a few comments for further consideration, if you want:

> here's no reason to think we'd be reading AWT- or Swing-related classes from a config file, for example

Why not? In a context of configurable GUI applications I can really see a lot of AWT classes that could be read from a file - starting by all graphic primitives, such as `Point`, `Rectangle`, `Color` and `Stroke`.

> Ultimately, we're talking about supporting immutable value classes, which are fairly rare in the JDK.

Is there any reason why pureconfig built-in support should be limited to immutable classes? At its core, configurations are just immutable representations of domain objects - but that doesn't mean it isn't useful to read a mutable class from a configuration. In fact, it could be argued that adding support for common Java collections from `java.util.*` is more useful than classes from `javax` used by a niche of users.

> I'm in favor of supporting KerberosPrincipal only because Kerberos is so prevalent in enterprise environments, and it's a straight-forward String wrapper type. (...) I've casually inspected javax.*, and there only appear to be a few classes that we'd realistically support:

I'll be honest: I never used or even heard of any of the classes you mentioned. Don't get me wrong - I really believe they ought to be useful for you and for a lot more people out there, but there are also many projects that don't ever deal with these enterprise concepts and mechanisms. The same can be said of AWT and `javax.sound`: while you and me won't ever put those classes in a config in the kind of projects we work at, I can imagine ways in which that would be desirable for others.

That is the main reason I strongly support creating modules for these kind of use cases instead of creating a monolith dependency with all classes one could ever think of using in the JDK - I'm positively sure that there aren't just "a few classes that we'd realistically support" :)

PR #134, opened 2017-02-17
coproducthint use ConfigReaderFailures

melrief:
Fix #133

PR #138, opened 2017-02-18
Changelog 0.6.0

melrief:


ruippeixotog (changes requested):
The changelog is still missing the new converter for `java.net.URI` (#129) and the change from `Try` to `Either` in `CoproductHint` (#134). Other than that, LGTM!

melrief:
@ruippeixotog added both, thanks. I have one question about best practices with the changelog: do we have to link the issue involved? We are not doing it right now but it could be useful for the user to track what the point means.

ruippeixotog:
Yeah, we're not enforcing it now but it could be a good idea to reference either the pull request (always present) or the issue (not always present, but possibly more informative) relative to each change.

ruippeixotog (approved):

melrief:
I won't merge this yet because we need the last change from @jcazevedo with #85 and the release date.

leifwickland (approved):
Thanks, @melrief!

PR #139, opened 2017-02-20
Add location (filename and line number) to errors

### LINKED ISSUE #85: Add the line to the error

jcazevedo:
This PR adds the location (filename and line number) to `ConfigReaderFailure`. The location is an `Option` since it might not always be available (we don't always load configurations from a file and some errors can't be pinned to a specific location).

This PR attempts to fix #85.

This is still WIP since I'd like to add more tests and improve the documentation.


jcazevedo:
I've improved the documentation of the sealed family of `ConfigReaderFailure`s and added a description to the README on how to provide locations to failures.

@melrief and @leifwickland: I think this is ready to be reviewed.

melrief:
I will review it tomorrow, thanks for the work @jcazevedo !

core/docs/error-handling.md
```diff
@@ -0,0 +1,23 @@
+## Error handling
+
+PureConfig provides some semantically rich case classes to describe failures
```
leifwickland: Suggestion: I'd reword this sentence as:

> The `pureconfig.error` package provides semantically rich case classes to describe failures reading from a `ConfigValue`.

Feel free to prefer your wording.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -5,14 +5,152 @@ package pureconfig.error
...
-final case class WrongTypeForKey(typ: String, expectedTyp: String, key: String) extends ConfigReaderFailure
-final case class ThrowableFailure(throwable: Throwable) extends ConfigReaderFailure
-final case class EmptyStringFound(typ: String) extends ConfigReaderFailure
-final case class NoValidCoproductChoiceFound(value: ConfigValue) extends ConfigReaderFailure
\ No newline at end of file
+/**
+ * The physical location of a ConfigValue, represented by a file name and a line
+ * number
+ *
+ * @param filename the complete pathname to the file that contains a given
```
leifwickland, line 12: If the config file is inside of a JAR does this become some useful value like:
`path/to/the.jar#resource/path.conf`?

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -5,14 +5,152 @@ package pureconfig.error
...
+ *                 ConfigValue
+ * @param lineNumber the line number (starting at 0), where the given
+ *                   ConfigValue definition starts
+ */
+case class ConfigValueLocation(filename: String, lineNumber: Int) {
+  def description: String = s"(file:$filename: $lineNumber)"
```
leifwickland: Just a suggestion: 
I think I'd rather this be 
```
s"($filename:$lineNumber)"
```
which seems like a more traditional format for an error report.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -5,14 +5,152 @@ package pureconfig.error
...
+   *         possible to derive a location. It is not possible to derive
+   *         locations from ConfigValues that are not in files or for
+   *         ConfigValues that are null.
+   */
+  def apply(cv: ConfigValue): Option[ConfigValueLocation] =
+    Option(cv).flatMap { v =>
+      val origin = v.origin()
+      if (origin.filename != null && origin.lineNumber != -1)
+        // I believe lineNumber starts at 1 in ConfigOrigin
+        Some(ConfigValueLocation(origin.filename, origin.lineNumber - 1))
```
leifwickland: If ConfigOrigin's line numbering starts at 1 and you're subtracting 1, does that mean you're trying to get zero-based line numbering?
If so I object. I would prefer one-based line numbering.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -5,14 +5,152 @@ package pureconfig.error
...
+ */
+sealed abstract class ConfigReaderFailure {
+  /**
+   * The optional location of the ConfigReaderFailure
+   */
+  def location: Option[ConfigValueLocation]
+}
+
+/**
+ * A failure representing the inability of converting a null value. Since a null
```
leifwickland: Nitpick: Here and below in a place or two, I'd suggest the phrase "the inability to convert" instead of "the inability of converting".

core/src/test/scala/pureconfig/ConfigReaderFailureLocationSuite.scala
```diff
@@ -0,0 +1,60 @@
...
+      """java.lang.NumberFormatException: For input string: "hello"""",
+      Some(ConfigValueLocation(workingDir + file, 8))))
+  }
+
+  it should "show proper error location when loading from multiple files" in {
+    import pureconfig.syntax._
+    case class Conf(a: Int, b: String, c: Int)
+
+    val workingDir = getClass.getResource("/").getFile
+    val file1 = "conf/error2/a.conf"
```
leifwickland: Suggestion: Rename the `error2` and `error1` parts of the paths these test files are in to something like `errorLocationReporting` to try to tie them to their purpose.

core/src/test/scala/pureconfig/ConfigReaderFailureLocationSuite.scala
```diff
@@ -0,0 +1,60 @@
...
+    val file2 = "conf/error2/b.conf"
+    val conf = ConfigFactory.load(file1).withFallback(ConfigFactory.load(file2)).root()
+
+    val failures = conf.get("conf").to[Conf].left.value.toList
+    failures should have size 1
+    failures should contain(CannotConvert(
+      "string",
+      "Int",
+      """java.lang.NumberFormatException: For input string: "string"""",
+      Some(ConfigValueLocation(workingDir + file2, 1))))
```
leifwickland: That seems like it should be line 2, not 1, to me.

leifwickland (changes requested):
I left a few suggestions.
I'd sure line numbering for error reporting to be one-based.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: I agree. I'm not sure what I was thinking. 1-based makes much more sense. I'll fix it later today.

core/docs/error-handling.md
jcazevedo: I've reworded it as per your suggestion in 402a1d8.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: Yes. It will be something like `path/to/the.jar!/path.conf`.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: I've modified it in e1f4694. I don't have a strong preference either way, so I'm OK to go with the more traditional format.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: Fixed it in a8248b7.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: Modified the wording as per your suggestion in 402a1d8.

core/src/test/scala/pureconfig/ConfigReaderFailureLocationSuite.scala
jcazevedo: I've scoped the test files used by this suite under the `configFailureLocation` folder in 3c08eb3.

core/src/test/scala/pureconfig/ConfigReaderFailureLocationSuite.scala
jcazevedo: Yes, you're right. This had to do with the line numbering not being one-based. I've fixed it in a8248b7.

jcazevedo:
@leifwickland: Thanks for your review! I believe I've addressed all your comments.

leifwickland (approved):
Awesome work, @jcazevedo! This is a terrific feature.
Thanks for considering my suggestions.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -5,14 +5,151 @@ package pureconfig.error
...
+        None
+    }
+}
+
+/**
+ * A representation of a failure that might be raised from reading a
+ * ConfigValue. The failure contains an optional location of the ConfigValue
+ * that raised the error.
+ */
+sealed abstract class ConfigReaderFailure {
```
melrief, line 49: @jcazevedo thanks a lot for adding documentation to the failures!

melrief (approved):
Amazing job @jcazevedo !

Issue #135, opened 2017-02-17
Add toOrThrow methods to pureconfig.syntax


melrief:
In the new pureconfig we have `loadConfig` that returns `Either` and `loadConfigOrThrow` that returns the result or throws exception. We should add methods to `pureconfig.syntax` to support the same schema, with `to` that returns `Either` and `toOrThrow` that returns the result or throws exception.

stsatlantis:
Working on it

Issue #86, opened 2017-01-16
Add number of bytes class and ConfigConvert


melrief:
Typesafe config supports [`getBytes`](http://typesafehub.github.io/config/latest/api/com/typesafe/config/Config.html#getBytes-java.lang.String-) to read the number of bytes from a human friendly string such as `10MB`. The spec for this format is [here](https://github.com/typesafehub/config/blob/master/HOCON.md#size-in-bytes-format). It would be nice to add this to pureconfig. The way I think we should do this is slightly different from what typesafe config does: I would add a class Bytes and create a `ConfigConvert` for it. The obvious issue with this is that it would be different from what you have in Typesafe config.

leifwickland:
@melrief I really like this the idea. My initial take is that the `Bytes` class might be similar in design to the scala`Duration` class, tracking a scalar and a unit.

I think my inclination might be to call that class `ByteSize` rather than`Bytes` in order to avoid conflicts who have a byte array wrapper called `Bytes`.

melrief:
@leifwickland you are right and I can think of one library that uses the name `Bytes` for binary format encoding and decoding. `ByteSize` is better.

derekmorr:
Is this still needed now that the squants module was merged? Squants has an `Information` type that supports bytes and both the decimal (MB, GB, ...) and binary (MiB, GiB, ...) suffixes.

OTOH, it might be nice to have a converter that doesn't depend on squants.

jcazevedo:
I prefer not introducing a custom type in PureConfig's core. I think we can support reading strings the method `getBytes` allows into numeric types (`Float`, `Double`, `Short`, `Int` and `Long`), but otherwise rely on external libraries for more semantically rich types (such as Squants, as @derekmorr suggested).

melrief:
@derekmorr @jcazevedo you have good points and I think I prefer to not having it in core too after reading them. I should think less about typesafe config and more about what we'd like pureconfig core to be. Closing this.

@stsatlantis this was assigned to you but I haven't received any update here or on gitter. I hope it's fine for you too if we close this, sorry for the confusion.

stsatlantis:
Np I forgot to send the PR but np I've learned a lot from it :)

PR #143, opened 2017-02-25
Format fix in README.md

stsatlantis:
Format fix in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ val config: Either[ConfigReaderFailures,YourConfClass] = loadConfig[YourConfClas
 Currently supported types for fields are:
 - `String`, `Boolean`, `Double` (standard
   and percentage format ending with `%`), `Float` (also supporting percentage),
-  `Int`, `Long`, `Short`, `URL`, `URI`, Duration`, `FiniteDuration`;
+  `Int`, `Long`, `Short`, `URL`, `URI`, `Duration`, `FiniteDuration`;
 - all collections implementing the `TraversableOnce` trait where the type of
   the elements is in this list;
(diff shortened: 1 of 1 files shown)


melrief:
Thanks @stsatlantis for fixing the README! I'd like to add you to the library [CONTRIBUTORS](https://github.com/melrief/pureconfig/blob/master/CONTRIBUTORS) list. If you want to be added, please make a PR with a new line in the file with your name, email address and nickname. Thanks again.

stsatlantis:
Thanks for the offer. I would like to do something meaningful before been called a contributor. :)
I'm reading the issues and tring to find one suitable for my level of knowledge. :)

melrief:
Sure! Among the current issues I can think about two small issues plus one not yet there that would be nice to have solved:
1. #135 which consists in adding a new extension method to [PimpedConfigValue](https://github.com/melrief/pureconfig/blob/master/core/src/main/scala/pureconfig/syntax/package.scala#L11) and to [PimpedConfig](https://github.com/melrief/pureconfig/blob/master/core/src/main/scala/pureconfig/syntax/package.scala#L15) which should behave like `loadConfigOrThrow`
2. #86, which consists in adding a type representing the number of bytes and a converter from String following the syntax used in typesafe config. This requires some work on the parser I think, otherwise it's trivial.
3. support for any class in the standard library of Scala or Java that is still missing is nice and welcome. An example here is a personal mistake: I forgot to add support for `java.time.Duration` when I pushed for the other data structures. It's a relatively simple PR, you can follow the converter for scala `Duration` that is already in core. If you decide to proceed with this, please create a small issue because currently it's missing.

PR #150, opened 2017-03-01
Resolve issue #135

stsatlantis:
Resolves issue #135

### DIFF
--- a/core/src/main/scala/pureconfig/syntax/package.scala
+++ b/core/src/main/scala/pureconfig/syntax/package.scala
@@ -23,7 +23,6 @@ package object syntax {
   }
 
 
-
   implicit class PimpedConfig(val conf: TypesafeConfig) extends AnyVal {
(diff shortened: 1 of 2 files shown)


melrief (approved):
Thanks for the contribution! I'd like @jcazevedo to have a look at this too because he is the main implementor of the `syntax` package but then we can merge it.

jcazevedo (changes requested):
Can you also add a test for the positive case, i.e. when `toOrThrow` succeeds and doesn't throw an exception?

stsatlantis:
Extra test added let me know if you want more complex tests

jcazevedo:
Thanks for this, @stsatlantis! There's no need for more complex tests.

We would like to add you to the [CONTRIBUTORS](https://github.com/melrief/pureconfig/blob/master/CONTRIBUTORS) of the library. If you agree, please make another PR that just adds one line to the file with your name, the email and your username.

jcazevedo (approved):

melrief:
Nice!

Issue #154, opened 2017-03-02
Polish the failure summary


jcazevedo:
The failure summary is currently grouping failures by type. It is more natural to group it by key. We can also integrate it better with the failure location (file and line number).

PR #156, opened 2017-03-07
Normalize ConfigReaderFailures

### LINKED ISSUE #153: Normalize ConfigReaderFailures
We have various `ConfigReaderFailure`s, but some are missing important information. Namely, we can add the `key` field to `NoValidCoproductChoiceFound`, `EmptyStringFound` and `ThrowableFailure`. The pattern of having a failure with and without key is also common practice, so we can make an effort to abstract it better.

jcazevedo:
This PR attempts to normalize `ConfigReaderFailure`s by forcing them to have an optional path to the value that raised the failure, as well as a `withImprovedContext` method that allows them to be enhanced with more context (the key and optional physical location of the parent node).

This attempts to fix #153.

I also want to improve the test suite in regards to failures, but I'd rather do it after tackling #154, which should come after this PR.


core/src/main/scala/pureconfig/syntax/package.scala
```diff
@@ -19,12 +19,11 @@ package object syntax {
 
   implicit class PimpedConfigValue(val conf: ConfigValue) extends AnyVal {
     def to[T](implicit configConvert: ConfigConvert[T]): Either[ConfigReaderFailures, T] = configConvert.from(conf)
-    def toOrThrow[T](implicit configConvert: ConfigConvert[T], cl:ClassTag[T]): T = getResultOrThrow(configConvert.from(conf))(cl)
+    def toOrThrow[T](implicit configConvert: ConfigConvert[T], cl: ClassTag[T]): T = getResultOrThrow(configConvert.from(conf))(cl)
```
melrief: We have an issue with code non-formatted when it's submitted, we should fix this eventually.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -232,22 +232,11 @@ object ConfigConvert extends LowPriorityConfigConvertImplicits {
...
-      case e: ConfigReaderFailure => e
-    }
-
   private[pureconfig] def improveFailures[Z](result: Either[ConfigReaderFailures, Z], keyStr: String, location: Option[ConfigValueLocation]): Either[ConfigReaderFailures, Z] =
     result.left.map {
       case ConfigReaderFailures(head, tail) =>
-        val headImproved = improveFailure[Z](head, keyStr, location)
-        val tailImproved = tail.map(improveFailure[Z](_, keyStr, location))
+        val headImproved = head.withImprovedContext(keyStr, location)
+        val tailImproved = tail.map(_.withImprovedContext(keyStr, location))
```
melrief, line 239: Much better, thank you!

melrief (approved):
I like the changes, it's more clean now. I'd like to wait for @leifwickland review too but this looks good to me.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -69,8 +84,12 @@ final case object CannotConvertNull extends ConfigReaderFailure {
  * @param because the reason why the conversion was not possible
  * @param location an optional location of the ConfigValue that raised the
  *                 failure
+ * @param path an optional path to the value that couldn't be converted
```
leifwickland, line 101: I wish there were a good way to disambiguate "path" from the typical filesystem sense in which it's used. I don't have any good ideas. Do you?
(Ditto on 56.)

leifwickland (approved):
Thanks, @jcazevedo

ruippeixotog (approved):
LGTM!

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: Unfortunately, I don't have a better idea. I ended up calling it path because that's the name Typesafe Config uses, but I agree it can be ambiguous.

PR #157, opened 2017-03-08
initial support for javax (KerberosPrincipal)

derekmorr:
This is the javax module, initially only supporting KerberosPrincipal.


melrief (approved):
Thank you!

PR #159, opened 2017-03-09
Add X500 Principal to javax module

derekmorr:
I don't know how common X500Principal is, but it's a straightforward String-wrapper value class.


modules/javax/src/main/scala/pureconfig/module/package.scala
```diff
@@ -13,4 +14,7 @@ import pureconfig.ConfigConvert.fromStringConvertTry
 package object javax {
   implicit val readKerberosPrincipal: ConfigConvert[KerberosPrincipal] =
     fromStringConvertTry[KerberosPrincipal](s => Try(new KerberosPrincipal(s)), _.toString)
+
+  implicit val readX500Principal: ConfigConvert[X500Principal] =
+    fromStringConvertTry[X500Principal](s => Try(new X500Principal(s)), _.toString)
```
melrief: Looks good to me.

FYI, while the use of `fromStringConvertTry(s => Try(f(s)), _.toString)` is correct, `core` provides another pattern that doesn't require to instantiate a `Try`, which is `fromStringConvert(catchReadError(f), _toString)`. In theory, the latter should be faster because [`catchReadError`](https://github.com/melrief/pureconfig/blob/ae08e571b62a4e8714702fc44ceb74a100df60a8/core/src/main/scala/pureconfig/ConfigConvert.scala#L107) uses `try(f) catch { case NonFatal(e) => .. }` instead of `Try(f)`. `core` uses it [for default instances](https://github.com/melrief/pureconfig/blob/ae08e571b62a4e8714702fc44ceb74a100df60a8/core/src/main/scala/pureconfig/ConfigConvert.scala#L502).

melrief (approved):
Looks good to me, it's fine if we publish this in the next release of pureconfig, which should be `0.6.1`?

modules/javax/src/main/scala/pureconfig/module/package.scala
derekmorr: Thanks. I'll update the code.

derekmorr:
Sure, I'm ok waiting. No rush - the javax module is pretty barebones right now.

leifwickland (approved):
Thanks, @derekmorr!

melrief:
Thanks!

Issue #148, opened 2017-03-01
Unresolved placeholders cause thrown exception


megri:
`loadConfig` doesn't catch this in the `ConfigReaderFailure` but instead throws it directly.

melrief:
Hey @megri, thanks for opening the issue. I guess we forgot to check `ConfigFactory.load`. After reading the [documentation](https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigFactory.html#load-com.typesafe.config.Config-) I think I know why: there is no documentation about the exception that can be thrown by those methods and I was assuming it was safe to call. I should have known that a method that doesn't return either errors or the config was unsafe.

The good news is that it should be easy to fix. The only issue that I have is that I don't know which exceptions can we should catch. Maybe we can catch `NonFatal` and see how it goes.

leifwickland:
I agree with catching NonFatal as you suggest, @melrief

melrief:
I propose to create a failure only for this kind of error

leifwickland:
Does "this type of error" mean the resolution failure or a catchall for
exceptions pureconfig didn't expect? I think we need the latter given that
the type signature seems to promise that non fatal errors will not escape.
I also like the idea of a specific error for that resolution failure.

melrief:
I was talking about confining the exceptions thrown by `ConfigFactory.load` inside a failure class that is reserved just for that method. In my mind, that's not a `CannotConvert` failure but a `CannotLoadConfig` (or something like that).

megri:
I think the exceptions thrown are the ones found in the ConfigLoader class. The strictest/cleanest path would be to map those to an ADT

leifwickland:
I totally agree that it's a positive to map known errors to the failure ADT. However, I think we probably should go a step beyond that and catch unknown errors too, dropping them in an appropriately vaguely named member of the ADT.

melrief:
@megri I'll fix this for the next release, thanks for reporting it.

megri:
Thanks for fixing it :)

PR #161, opened 2017-03-09
Catch typesafeconfig exceptions

melrief:
Solves #148 

This PR is an attempt to catch and wrap inside failures the exception thrown by the typesafe config parser and loader. Changes:

- add `ConfigFactoryWrapper` as safer version of the typesafe `ConfigFactory`
- change the `pureconfig.load*` methods to use the `ConfigFactoryWrapper` methods instead of the `ConfigFactory` ones
- change `pureconfig.loadConfigFromFiles` to work on `Path` instead of `File`. Not sure why I accepted a PR working on something different but we are using `Path` everywhere and I'd like to be consistent

p.s. I've noticed in this and another PR that there are some changes on the styles of the files. Something is not working properly but I'm not sure what. I'll leave the style changes in the PR because we should have well formatted code...

### DIFF
--- a/core/src/main/scala/pureconfig/backend/TypesafeConfigWrapper.scala
+++ b/core/src/main/scala/pureconfig/backend/TypesafeConfigWrapper.scala
@@ -27,7 +27,7 @@ object ConfigFactoryWrapper {
 
   /** Utility methods that parse a file and then calls `ConfigFactory.load` */
   def loadFile(path: Path): Either[ConfigReaderFailures, Config] =
-    parseFile(path).flatMap(rawConfig => unsafeToEither(ConfigFactory.load(rawConfig)))
+    parseFile(path).right.flatMap(rawConfig => unsafeToEither(ConfigFactory.load(rawConfig)))
 
   private def unsafeToEither[A](f: => A): Either[ConfigReaderFailures, A] = {
     try (Right(f)) catch {
--- a/core/src/main/scala/pureconfig/backend/TypesafeConfigWrapper.scala
+++ b/core/src/main/scala/pureconfig/backend/TypesafeConfigWrapper.scala
@@ -1,0 +1,38 @@
+package pureconfig.backend
+
+import java.nio.file.Path
+
+import com.typesafe.config.{ Config, ConfigFactory }
+import pureconfig.error.{ ConfigReaderFailures, ThrowableFailure }
+
+import scala.util.control.NonFatal
+
+/**
+ * A wrapper of [[com.typesafe.config.ConfigFactory]] whose methods return `Either` instead
+ * of throwing exceptions
+ */
+object ConfigFactoryWrapper {
+
+  /** @see [[ConfigFactory.invalidateCaches()]] */
+  def invalidateCaches(): Either[ConfigReaderFailures, Unit] =
+    unsafeToEither(ConfigFactory.invalidateCaches())
+
+  /** @see [[ConfigFactory.load()]] */
+  def load(): Either[ConfigReaderFailures, Config] =
+    unsafeToEither(ConfigFactory.load())
+
+  /** @see [[ConfigFactory.parseFile()]] */
+  def parseFile(path: Path): Either[ConfigReaderFailures, Config] =
+    unsafeToEither(ConfigFactory.parseFile(path.toFile))
+
+  /** Utility methods that parse a file and then calls `ConfigFactory.load` */
+  def loadFile(path: Path): Either[ConfigReaderFailures, Config] =
+    parseFile(path).flatMap(rawConfig => unsafeToEither(ConfigFactory.load(rawConfig)))
+
+  private def unsafeToEither[A](f: => A): Either[ConfigReaderFailures, A] = {
+    try (Right(f)) catch {
+      case NonFatal(e) =>
+        Left(ConfigReaderFailures(ThrowableFailure(e, None)))
+    }
+  }
+}
--- a/core/src/main/scala/pureconfig/package.scala
+++ b/core/src/main/scala/pureconfig/package.scala
@@ -24,8 +24,11 @@ package object pureconfig {
    *         isn't possible
    */
   def loadConfig[Config](implicit conv: ConfigConvert[Config]): Either[ConfigReaderFailures, Config] = {
-    invalidateCaches()
-    load().right.flatMap(loadConfig[Config](_)(conv))
+    for {
+      _ <- invalidateCaches().right
+      rawConfig <- load().right
+      config <- loadConfig[Config](rawConfig)(conv).right
+    } yield config
   }
 
   /**
@@ -37,8 +40,11 @@ package object pureconfig {
    *         isn't possible
    */
   def loadConfig[Config](namespace: String)(implicit conv: ConfigConvert[Config]): Either[ConfigReaderFailures, Config] = {
-    invalidateCaches()
-    load().right.flatMap(rawConfig => improveFailures[Config](loadConfig[Config](rawConfig.getConfig(namespace))(conv), namespace, None))
+    for {
+      _ <- invalidateCaches().right
+      rawConfig <- load().right
+      config <- improveFailures[Config](loadConfig[Config](rawConfig.getConfig(namespace))(conv), namespace, None).right
+    } yield config
   }
 
   /**
@@ -50,8 +56,11 @@ package object pureconfig {
    *         isn't possible
    */
   def loadConfig[Config](path: Path)(implicit conv: ConfigConvert[Config]): Either[ConfigReaderFailures, Config] = {
-    invalidateCaches()
-    loadFile(path).right.flatMap(loadConfig[Config](_)(conv))
+    for {
+      _ <- invalidateCaches().right
+      rawConfig <- loadFile(path).right
+      config <- loadConfig[Config](rawConfig)(conv).right
+    } yield config
   }
 
   /**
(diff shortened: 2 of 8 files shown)


core/src/main/scala/pureconfig/package.scala
```diff
@@ -252,15 +258,18 @@ package object pureconfig {
    *
    * @param files Files ordered in decreasing priority containing part or all of a `Config`. Must not be empty.
    */
-  def loadConfigFromFiles[Config: ConfigConvert](files: Traversable[java.io.File]): Either[ConfigReaderFailures, Config] = {
+  def loadConfigFromFiles[Config: ConfigConvert](files: Traversable[Path]): Either[ConfigReaderFailures, Config] = {
```
leifwickland, line 297: Sorry. This is my fault originally.

core/src/main/scala/pureconfig/backend/TypesafeConfigWrapper.scala
```diff
@@ -0,0 +1,38 @@
...
+    unsafeToEither(ConfigFactory.parseFile(path.toFile))
+
+  /** Utility methods that parse a file and then calls `ConfigFactory.load` */
+  def loadFile(path: Path): Either[ConfigReaderFailures, Config] =
+    parseFile(path).right.flatMap(rawConfig => unsafeToEither(ConfigFactory.load(rawConfig)))
+
+  private def unsafeToEither[A](f: => A): Either[ConfigReaderFailures, A] = {
+    try (Right(f)) catch {
+      case NonFatal(e) =>
+        Left(ConfigReaderFailures(ThrowableFailure(e, None)))
```
leifwickland: Thanks for doing this.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -23,8 +24,8 @@ package object pureconfig {
    *         isn't possible
    */
   def loadConfig[Config](implicit conv: ConfigConvert[Config]): Either[ConfigReaderFailures, Config] = {
-    ConfigFactory.invalidateCaches()
-    loadConfig[Config](ConfigFactory.load())(conv)
+    invalidateCaches()
```
leifwickland: If this fails the method will continue loading the configuration mixed with the stale cache. 
Previously if this failed by throwing an exception, the method would fail.
I don' think you intended to change the behavior.
It seems like we need to chain `invalidateCaches()` with the following load call to propagate errors correctly.
You roughly want the `*>` from scalaz. It might be worth implementing something like that here for concision.
Ditto multiple places below.

leifwickland (changes requested):
I don't think you intended the behavior change around `invalidateCache`

core/src/main/scala/pureconfig/package.scala
melrief: What a mistake! I've done this very late and I missed this obvious mistake.

melrief:
@leifwickland thanks for the review! I've addressed the issue with `invalidateCaches`, let me know what do you think.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -143,9 +166,13 @@ package object pureconfig {
...
-    ConfigFactory.invalidateCaches()
-    val config = ConfigFactory.load(ConfigFactory.parseFile(path.toFile))
-    getResultOrThrow[Config](loadConfig[Config](config)(conv))
+    val errorOrConfig =
+      for {
+        _ <- invalidateCaches().right
+        rawConfig <- loadFile(path).right
+        config <- loadConfig[Config](rawConfig)(conv).right
+      } yield config
+    getResultOrThrow[Config](errorOrConfig)
```
leifwickland, line 175: Do you recall why we didn't implement this as:
```
getResultOrThrow[Config](loadConfig(path)(conv))
```

Ditto for the other `loadConfigOrThrow` methods.

leifwickland (approved):

core/src/main/scala/pureconfig/package.scala
melrief: No but it's equivalent, let me fix it

melrief: Fixed with 612b620

leifwickland: Thanks!
It seems like we'd want to reuse the `loadConfig` implementations for for the other `loadConfigOrThrow` methods.

melrief: I was thinking the same and I was wondering why we didn't implemented the `loadConfigOrThrow` by using `loadConfig`. Then I noticed that the location is filled differently, e.g.

```scala
  def loadConfig[Config](namespace: String)(implicit conv: ConfigConvert[Config]): Either[ConfigReaderFailures, Config] = {
    for {
      _ <- invalidateCaches().right
      rawConfig <- load().right
      config <- improveFailures[Config](
        loadConfig[Config](rawConfig.getConfig(namespace))(conv), namespace, None).right
    } yield config
  }
```

but

```scala
def loadConfigOrThrow[Config](namespace: String)(implicit conv: ConfigConvert[Config], ct: ClassTag[Config]): Config = {
    val errorOrConfig =
      for {
        _ <- invalidateCaches().right
        rawConfig <- load().right.map(_.getConfig(namespace)).right
        config <- improveFailures[Config](
          loadConfig[Config](rawConfig)(conv), namespace, ConfigValueLocation(rawConfig.root())).right
      } yield config
    getResultOrThrow[Config](errorOrConfig)
  }
```

I'm not sure why the first set the position of the error to `None` while the second to `ConfigValueLocation(rawConfig.root())` but this is preventing me from reusing the first in the second. @jcazevedo do you know why the two methods are different for the error location?

leifwickland: @melrief If you can't get resolution quickly, I'd be tempted to just revert 612b620 and pick up the thread in a separate issue focused on deduplication.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -143,9 +166,7 @@ package object pureconfig {
    */
   @throws[ConfigReaderException[_]]
   def loadConfigOrThrow[Config](path: Path)(implicit conv: ConfigConvert[Config], ct: ClassTag[Config]): Config = {
```
leifwickland, line 168: I can't figure out where the `ClassTag` is getting used. `loadConfig` doesn't require it. I think it can go.

I think that's also true of all (or most) of the other `loadConfigOrThrow` methods.

melrief: The `ClassTag` is propagated all the way down to the `ConfigReaderException`, which uses it to know the runtime name of the class for the error message.

core/src/main/scala/pureconfig/package.scala
melrief: I `git revert`ed the change with fadbf26 and I've opened #162 to work on the code reuse issue. Let's move the discussion in that issue.

jcazevedo: > I'm not sure why the first set the position of the error to None while the second to ConfigValueLocation(rawConfig.root()) but this is preventing me from reusing the first in the second. @jcazevedo do you know why the two methods are different for the error location?

It looks like a mistake. I think the `location` on `loadConfig` should be obtained via `ConfigValueLocation(rawConfig.root())` as well.

melrief: Then @leifwickland is right and we should always reuse `loadConfig` in `loadConfigOrThrow`. Let's tackle #162 in the near future and fix this inconsistency in the API.

jcazevedo: Having had a better look at it, this is definitely a mistake. We should call `ConfigValueLocation(rawConfig.root())` [here](https://github.com/melrief/pureconfig/pull/161/files#diff-12918ae14aad73e551e678d26010a1deR46) and [here](https://github.com/melrief/pureconfig/pull/161/files#diff-12918ae14aad73e551e678d26010a1deR79) (and reuse this code on the `loadConfigOrThrow` methods).

I don't know if you prefer doing this in the context of this PR or in a separate one.

core/src/main/scala/pureconfig/backend/TypesafeConfigWrapper.scala
```diff
@@ -0,0 +1,38 @@
...
+import com.typesafe.config.{ Config, ConfigFactory }
+import pureconfig.error.{ ConfigReaderFailures, ThrowableFailure }
+
+import scala.util.control.NonFatal
+
+/**
+ * A wrapper of [[com.typesafe.config.ConfigFactory]] whose methods return `Either` instead
+ * of throwing exceptions
+ */
+object ConfigFactoryWrapper {
```
jcazevedo: I prefer having this file named `ConfigFactoryWrapper`, to better reflect its contents.

core/src/main/scala/pureconfig/backend/TypesafeConfigWrapper.scala
jcazevedo: I think we can do a little better here. If we catch [`ConfigException`](https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigException.html), we have access to a `ConfigOrigin` which we can then use to create a `ConfigValueLocation`. This would add more context to our errors. Also, we can map some types of `ConfigException` (I see at least [`ConfigException.Parse`](https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigException.Parse.html) and [`ConfigException.NotResolved`](https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigException.NotResolved.html)) which can be mapped to more semantically rich failures of PureConfig, other than just a `ThrowableFailure`.

jcazevedo (changes requested):

core/src/main/scala/pureconfig/package.scala
melrief: Separate one but in this release. Let's do the review of this PR to manage exceptions from typesafe API and use the PR that solves #162 to fix the API. I prefer to have two commits in the git history tbh.

melrief:
@jcazevedo with 3a7ca16 I've changed the name of the file and added a custom failure called `CannotParse` for typesafe config parsing errors. I'd like to not add too many customs failures here because I think goal should be to be generic over the backend. Our failures should reflect a generic backend error rather than a specific one. Parsing errors, for instance, are generic but `NotResolved` seems specific to typesafe config.
I'm open to objections on this but I really like the idea ofallowing to choose the backend for parsing the configuration. It won't be easy but it's doable.
Let me know what do you think.

jcazevedo:
> I'd like to not add too many customs failures here because I think goal should be to be generic over the backend. Our failures should reflect a generic backend error rather than a specific one. Parsing errors, for instance, are generic but NotResolved seems specific to typesafe config.

It makes perfect sense. 

This looks good to merge as it is. Thanks!

jcazevedo (approved):

PR #163, opened 2017-03-12
first pass at tut integration

### LINKED ISSUE #96: Generate README and other documentation via `tut` to ensure example code compiles
I propose that we migrate our README and any other applicable documentation to being generated by [`tut`](https://github.com/tpolecat/tut). I recently worked on another project which made the transition and it was overwhelmingly positive. (See https://github.com/rubicon-project/rubiz/commit/d16b4996094703a33f8cf385292ad852113ed3b7)

The major benefits of tut are:
- The compiler ensures that examples compile and
- The runtime ensures that sample outputs are correct, rather than being copy-pasted into a comment.

The downsides of using tut are:
- Contributors need to remember to edit `src/main/tut/README`, etc, instead of the normal file they'd expect.
- Separating the PR for a big change to the code from the PR to update the documentation is no long possible.

If you guys are on board with that idea I'll take a swing at migrating the README, when the docs in a quiescence period between PRs.

derekmorr:
This would resolve #96 .

This PR isn't ready to merge yet. In particular, the enumeratum README doesn't build (it's missing an import), the tutSettings are crudely bolted into build.sbt, and I haven't touched core/docs yet.

I've added these commands to sbt:

- `tut` will rebuild the README.md's for core and all modules
- `core/tut` only rebuild's core's README (the main project README)
- `_module_/tut` only rebuilds _module_'s README (e.g., `enum/tut`, `joda/tut`)

I've added `src/main/tut` to core and each module. There were some minor stylistic changes made to each README, and some small syntax adjustments made in enum to get it to build. I also had to add a sample implementation of `YourConfig` to core's README to get it to build.
 
Tut can style code blocks in two ways: Just adding syntax highlighting, or adding the REPL output. I chose the former for the code sample, as it reduces REPL noise and makes the samples easier to copy-and-paste. But this does mean it doesn't show the output from the `loadConfig[Foo](conf)` calls. We can edit that if reviewers prefer.

I haven't added tut into the Travis CI configuration. I think we should do that in a separate PR once we get all the READMEs building.


derekmorr:
I pushed a new commit that switches to `tut:book` for some code samples -- that way we get the output from calling `loadConfig`. I rearranged some of the imports to reduce REPL noise in the output.

modules/enumeratum/README.md
```diff
@@ -42,7 +47,6 @@ val conf = parseString("""{
   end: SHOUT_GOOD_BYE
 }""")
 loadConfig[GreetingConf](conf)
```
leifwickland: I don't understand why the result of `loadConfig` didn't get printed here. It seems like it should've. Any ideas?

modules/joda/README.md
```diff
@@ -18,28 +18,39 @@ libraryDependencies += "com.github.melrief" %% "pureconfig-joda" % "0.6.0"
...
+```scala
 val conf = parseString("""{
   apollo: "1969-07-20T20:18:00.000Z"
   pluto: "2021-01-20T06:59:59.999Z"
 }""")
+// conf: com.typesafe.config.Config = Config(SimpleConfigObject({"apollo":"1969-07-20T20:18:00.000Z","pluto":"2021-01-20T06:59:59.999Z"}))
+
 loadConfig[GreatDatesConfig](conf)
-// Success(GreatDatesConfig(1969-07-20T14:18:00.000-06:00,2021-01-19T23:59:59.999-07:00))
+// res1: Either[pureconfig.error.ConfigReaderFailures,GreatDatesConfig] = Right(GreatDatesConfig(1969-07-20T16:18:00.000-04:00,2021-01-20T01:59:59.999-05:00))
```
leifwickland: Nitpick: It's sort of noisome that these dates are going to be rewritten each time the documentation is generated by someone with a different local time zone. Can you think of a way to avoid that? I've got nothing.

leifwickland:
@derekmorr @melrief 
Do you think we ought to exercise `sbt tut` in some fashion from Travis to have a better indication if your documentation has gotten out of sync with our code base?
Do you have an appreciation for the problems that generating new `.md` files will cause due to a dirty git tree.

leifwickland (commented):
@derekmorr Thanks so much for doing this! 
In general, it looks exactly like I'd hoped.
I made a few suggestions.

modules/enumeratum/README.md
derekmorr: The enumeratum module doesn't build with tut right now -- it complains about a missing import. I'm guessing that's why there's no output.

modules/joda/README.md
derekmorr: Is there a way to set Joda to output in UTC? I can look into it -- I'm guessing Joda has an equivalnt of ZonedDateTime.

derekmorr:
I suggest we hold off running `sbt tut` in Travis until we have things working, otherwise we're going to get lots of failed builds. I should be up-to-date with master.

Not sure I understand your second question.

leifwickland:
@derekmorr 
> Do you have an appreciation for the problems that generating new .md files will cause due to a dirty git tree.

Sorry that was confusing. I was trying to ask if we ran `tut` on travis and it caused the git tree there to be dirty, would that would cause some other sort of problem with updating the tree for the next run.

I agree that we don't want to enable `tut` on Travis until we sort out the problems. I was really trying to ask a longer term question than that. Again, sorry I wasn't clear.

derekmorr:
Ah, I understand now. I don't think it will cause problems. We've been running `sbt tut` on Travis with  squants for a few weeks without incident.

derekmorr:
Got the enumeratum readme to compile. Not sure why the example is failing in tut. It works for me if I copy/paste into the REPL:

```scala
> enumeratum/console
[info] Starting scala interpreter...
[info] 
Welcome to Scala 2.12.1 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).
Type in expressions for evaluation. Or try :help.

scala> :paste
// Entering paste mode (ctrl-D to finish)

import pureconfig.loadConfig
import pureconfig.module.enumeratum._
import com.typesafe.config.ConfigFactory.parseString
import enumeratum._
import enumeratum.EnumEntry._

sealed trait Greeting extends EnumEntry with Snakecase

object Greeting extends Enum[Greeting] {
  val values = findValues
  case object Hello extends Greeting
  case object GoodBye extends Greeting
  case object ShoutGoodBye extends Greeting with Uppercase
}
case class GreetingConf(start: Greeting, end: Greeting)
val conf = parseString("""{
  start: hello
  end: SHOUT_GOOD_BYE
}""")
loadConfig[GreetingConf](conf)

// Exiting paste mode, now interpreting.

import pureconfig.loadConfig
import pureconfig.module.enumeratum._
import com.typesafe.config.ConfigFactory.parseString
import enumeratum._
import enumeratum.EnumEntry._
defined trait Greeting
defined object Greeting
defined class GreetingConf
conf: com.typesafe.config.Config = Config(SimpleConfigObject({"end":"SHOUT_GOOD_BYE","start":"hello"}))
res0: Either[pureconfig.error.ConfigReaderFailures,GreetingConf] = Right(GreetingConf(Hello,ShoutGoodBye))
```

modules/joda/README.md
derekmorr: This is fixed as of 2c14d93.

derekmorr:
This is really perplexing -- tut isn't finding the ` implicit def enumeratumConfigConvert[A <: EnumEntry](implicit enum: Enum[A], ct: ClassTag[A]): ConfigConvert[A]` in the enumeratum package object. But copy/pasting the code from the README into the REPL works. I've tried adding instrumentation to the code, and it's failing in `ConfigConvert.fromConfigObject`.

At this point, I suspect it's something simple I'm just overlooking. I'd appreciate a fresh set of eyes if you have some time.

leifwickland:
@derekmorr I've taken a couple tries at this and haven't made any progress. I'm seeing the same thing you are and it's baffling. I'll try to make time to give it another look.

derekmorr:
Thanks for looking at it. I'm honestly at a loss to explain it. I'll be at the Typelevel Summit/NEScala next week, and I'll try to ask folks if they've seen this before.

In the meantime, I'll start working on the `docs/` directory.

derekmorr:
Enumeratum issue was resolved by leif in gitter.

jcazevedo:
> Do you think we ought to exercise sbt tut in some fashion from Travis to have a better indication if your documentation has gotten out of sync with our code base?

Do you suggest just running the `tut` command to see if it runs successfully or compare the files it generates with the ones committed? I think we need something like the latter to be sure the documentation is in sync with the codebase.

jcazevedo:
@derekmorr: I like the direction you're going with this. Thanks!

ruippeixotog:
So far, this looks very good :)

I think we should make Travis verify that the generated docs are synced with tut sources sooner rather than later. I can easily see external contributors, unaware we're using `tut`, modifying the generated docs directly. But I agree that it should be done in a future pull request, after this is stable.

ruippeixotog:
@jcazevedo I think it would suffice to run `sbt tut` and check that there are no git changes. Seems hacky, but I can't think of another reasonable solution.

jcazevedo:
@ruippeixotog: I also can't think of something better than that. 

We can even do a similar check for code formatting issues. We've accepted PRs not formatted according to our Scalariform settings and we can be warned of that when the `compile` task produces git changes.

derekmorr:
I'm wondering if we should split this into multiple PRs. This PR is already at 15 changed files. If I add docs to it, it'll only get bigger. Should we review and merge this first, then do docs? Or would folks prefer to have everything in one PR?

derekmorr:
In terms of diff'ing the markdown files vs tut's output, you have to be careful that the output is always the same. Some example code doesn't have stable output -- for example, printing the time in the user's current timezone. Or in `override-behavior-for-sealed-families.md`, the two `animalConfHint` examples print their hex locations by default:

```scala
implicit val animalConfHint = new FieldCoproductHint[AnimalConf]("kind")
// animalConfHint: pureconfig.FieldCoproductHint[AnimalConf] = pureconfig.FieldCoproductHint@2558dc9a
```

We can work around those if we're careful about using `tut:silent` vs `tut:book`, but it'll be fragile.

melrief:
If the PR is getting too big and there is a clear logical split between what you have done before and what you are going to do next, I propose to split in two. I'll wait for two of the authors to review and approve the changes and then we can merge this. @derekmorr let us know when the code is ready for review.

derekmorr:
I think this PR is ready for review.

ruippeixotog:
Hmm, well pointed out... It would be nice to have stable output, even if we had to modify the examples to use only stable values and/or tweak the tut tags. Moreover, changing the pureconfig source code can lead to examples stopping to compile or changing their output, so it is even more important to check the docs consistency on every Travis build later.

ruippeixotog (changes requested):
@derekmorr, can you modify the examples and/or silence some commands so that the tut output becomes stable? Other than that, the PR looks (very) good to me :)

derekmorr:
If you want to take it to an extreme you can add asserts() into the docs in a `tut:silent` block to ensure the code samples produce the correct results. That could be a future PR.

derekmorr:
@ruippeixotog I believe all of the output in the main README and the modules' READMEs is stable. The only issue was the timezone output that was fixed in 2c14d93. When I do repeated runs of `sbt tut`, the markdown files have the same hash each time.

ruippeixotog:
Oh, never mind. I forgot that the subpages like `override-behavior-for-sealed-families.md` are still static markdowns in this PR.

ruippeixotog (approved):

jcazevedo (approved):

melrief:
@ruippeixotog @jcazevedo thanks for the reviews!

@leifwickland I think you had a look at this PR already, do you approve it too?

leifwickland (approved):
👍 
Thanks, @derekmorr!

derekmorr:
Ack! I just had the plugins.sbt merged! Didn't have the README merged yet!

PR #166, opened 2017-03-18
fix example project for 0.6.0

### LINKED ISSUE #165: Example project is broken
The example project has several bugs:

1. It won't load in sbt. PR #158 added CrossVersion.patch which requires sbt 0.13.13, but the example project uses 0.13.8.
1. The Main function doesn't compile since `loadConfig` now returns an `Either` (as of #110 ). Also, Main can be simplified call `loadConfigOrThrow`.
1. The sample `ConfigConvert` instances call `fromString` which is deprecated in 0.6.0.
1. It no longer needs the `PathConvert` instance as of 0.6.0.

derekmorr:
This fixes #165

### DIFF
--- a/example/project/build.properties
+++ b/example/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version = 0.13.8
+sbt.version = 0.13.13
--- a/example/src/main/scala/pureconfig/example/conf/Main.scala
+++ b/example/src/main/scala/pureconfig/example/conf/Main.scala
@@ -36,10 +36,7 @@ object Main extends App {
   case class EmailConfig(host: String, port: Int, message: String, recipients: Set[Email], sender: Email)
 
 
-  val config = loadConfig[Config] match {
-    case Failure(f) => throw f
-    case Success(conf) => conf
-  }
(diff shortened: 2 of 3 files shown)


derekmorr:
I wonder if we should modify `.travis.yml` to build the example project as part of CI to detect these issues sooner. If interested, I can modify this PR to do that.

melrief:
Not having the example built by CI is bad, it would be great to have a system to test the changes to example. Imho what we could do is to provide a small script that can be run to both build pureconfig and the example and then tell TravisCI to build the repo using that script, instead of sbt directly. I don't know the limitation of this method but you just have to run twice sbt, one for pureconfig and the other for the example, and fail if one of the two fail. See the chapter ["Customizing the Build"](https://docs.travis-ci.com/user/customizing-the-build#Customizing-the-Build-Step).

@derekmorr do you mind if we do this in another PR? If yes, then I think we can merge this PR and open another one. Let me know.

derekmorr:
Sure, I'm fine doing that in another PR.

melrief:
Thanks!

PR #167, opened 2017-03-18
remove extraneous import in javax build.sbt

derekmorr:
At some point my IDE inserted a bogus import into javax's build.sbt. I should have caught this earlier, apologies.


melrief:
Np, it happens

PR #168, opened 2017-03-18
build the example project as part of CI

derekmorr:
This will build the example project as part of CI to ensure it stays working.

I've been experimenting with this in my fork of pureconfig. 

Here's a failed Travis job with the example project (before #166 was merged): https://travis-ci.org/derekmorr/pureconfig/jobs/212421774

If I apply the fixes in #166  to the code, it now builds: https://travis-ci.org/derekmorr/pureconfig/jobs/212423114

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,6 @@ scala:
     - 2.12.1
 jdk:
     - oraclejdk8
+script:
+  - sbt ++$TRAVIS_SCALA_VERSION test
+  - (cd example ; sbt ++$TRAVIS_SCALA_VERSION test)


melrief (approved):

melrief:
@derekmorr @leifwickland @jcazevedo @leifwickland just a quick note about this PR: as pointed out by the [TravisCI doc](https://docs.travis-ci.com/user/customizing-the-build#Customizing-the-Build-Step), this patch will always execute both the builds even if building pureconfig will fail. I would prefer the other solution, which is to concatenate the two bash commands via `&&` so that if pureconfig tests fails, then the example is not even run. I'm not too strong on this one, what's your opinions?

derekmorr:
That's fine with me.

jcazevedo:
Ideally, we would have a different job in the CI server for each build (and different checks for GitHub PRs). Unfortunately, I don't think that's possible in Travis CI.

Still, I think I prefer running both builds, even if building PureConfig fails. Potentially, we can get more information about the build of the example project without having to wait for the PureConfig build to be fixed.

jcazevedo (approved):

Issue #151, opened 2017-03-02
Split PureconfSuite in different suites


jcazevedo:
`PureconfSuite` is currently huge and spans different topics. It would be better to split it in different suites so that tests are better scoped.

melrief:
Solved with #169

PR #169, opened 2017-03-18
Refactor tests

melrief:
This is the first PR to partially solve #151 and in the meanwhile improve the tests suites. This isn't just reorganization of the tests but a full refactoring. It's a WIP and I'm submitting this to have a honest opinion about this style for the tests. The style that I'm trying to achieve can be seen in the new [`ConfigConvertSuite`](https://github.com/melrief/pureconfig/compare/refactor_tests?expand=1#diff-dc06a933e4eed042d7baefae349af772R21).

Note that in this PR there are also some notes in the old `PureconfSuite` which I use to know which aspect of pureconfig the tests is testing (e.g. `ConfigConvert` default instances, override, hints...). They will be removed eventually, when we solve #151.

The rationale behind this refactoring is:

1. logically split PureconfSuite into other smaller suites. In this PR, `ConfigConvert` and the `pureconfig.syntax` tests are moved to their own but ideally each aspect should go in its own suite. I'd like to have, for instance, one suite just for products and one for coproducts
2. test `ConfigConvert` instances without writing files and reading with `loadConfig`. Initially the idea was to test both `ConfigConvert` and the API together in one single test but after sometime I decided that this mix of functionalities tested is not helpful. `ConfigConvert` instances should be tested in their own suite while the API should be tested in another suite. A side-effect of this separation is that testing `ConfigConvert` doesn't require to write tests anymore and it's much faster.
3. make easy to create new tests for datatypes with `Arbitrary` instances. When possible use `Arbitrary` to test most of the values. I don't think you can avoid using scalacheck for `ConfigConvert` without writing tons of tests.
4. add some utilities for testing to remove boilerplate and increase readability. This is subjective but I think each test should be written in no more than couple of lines and only the "important" parts should be written. People write more tests if defining a test is very easy :). For instance, checking that a `ConfigConvert` should fail for a certain `ConfigValue` with a certain failure should be very easy.
5. test titles should be uniform among the tests

There are some drawbacks in this solution but I still think it's worth it.

Let me know what do you think and whether I should continue or switch to a more light refactoring. The next steps are to create the suites for products and coproducts and then for the API (`loadConfig` and friends).


core/src/test/scala/scalacheck/arbitrary/package.scala
```diff
@@ -0,0 +1,19 @@
+package scalacheck
```
ruippeixotog: I'd prefer to scope this inside `pureconfig`. Using `scalacheck` may mislead people into thinking this is a built-in package from scalacheck.

core/src/test/scala/scalacheck/gen/package.scala
```diff
@@ -0,0 +1,59 @@
+package scalacheck
```
ruippeixotog: Same thing here.

ruippeixotog (commented):
I fully agree with your rationale on the refactoring. I think reasons 2, 3 and 4 are particularly important to fulfill in order to keep our tests simple and very focused - that allows us to easily pinpoint failures when they happen.

I like, and think it is a good rule of thumb, to have one suite for each source file. If we think the logic in one source file is so complex that it needs tests in multiple files, maybe we should consider splitting the source file into multiple ones instead. Of course nothing is written in stone, it's just food for thought :) In the specific case of `ConfigConvert`, #171 may lead us into separating tests at least for basic and derived converters.

core/src/test/scala/pureconfig/ApiSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  it should s"be able to load a realistic configuration file" in {
+    case class DriverConf(cores: Int, maxResultSize: String, memory: String)
+    case class ExecutorConf(memory: String, extraJavaOptions: String)
+    case class SparkAppConf(name: String)
+    case class SparkLocalConf(dir: String)
+    case class SparkNetwork(timeout: FiniteDuration)
+    case class SparkConf(master: String, app: SparkAppConf, local: SparkLocalConf, driver: DriverConf, executor: ExecutorConf, extraListeners: Seq[String], network: SparkNetwork)
+    case class SparkRootConf(spark: SparkConf)
+    val configFile = createTempFile("""
```
leifwickland: Suggestion: Drop the `"""` to the line after the `(`so that there isn't so much indentation to the `|`s.

core/src/test/scala/pureconfig/BasicConvertersSuite.scala
```diff
@@ -0,0 +1,204 @@
...
+  it should "be able to supersede the default Duration ConfigConvert with a locally defined ConfigConvert" in {
+    val expected = Duration(220, TimeUnit.DAYS)
+    implicit val readDurationBadly = new ConfigConvert[Duration] {
+      override def from(config: ConfigValue): Either[ConfigReaderFailures, Duration] = Right(expected)
+      override def to(t: Duration): ConfigValue = throw new Exception("Not Implemented")
+    }
+    loadConfig(ConfigValueFactory.fromMap(Map("i" -> "42 h").asJava).toConfig)(ConfigConvert[ConfWithDuration]).right.value shouldBe ConfWithDuration(expected)
+  }
+
+  // ** override instance
```
leifwickland: Do you want all this dead code?

core/src/test/scala/pureconfig/PathUtils.scala
```diff
@@ -0,0 +1,22 @@
+package pureconfig
+
+import java.nio.charset.StandardCharsets
+import java.nio.file.{ Files, Path, Paths }
+
+object PathUtils {
+
+  def createTempFile(content: String): Path = {
```
leifwickland, line 8: Suggestion: Consider renaming to `writeToTempFile`.

leifwickland (approved):
👍 
I left a few comments, but feel free to ignore them. I don't see any deal breakers.

core/src/test/scala/pureconfig/gen/package.scala
```diff
@@ -0,0 +1,66 @@
...
+  val genDuration: Gen[Duration] =
+    Gen.frequency(
+      1 -> Gen.oneOf(Duration.Inf, Duration.MinusInf, Duration.Undefined),
+      99 -> genFiniteDuration)
+
+  val genInstant: Gen[Instant] =
+    Gen.choose(Instant.MIN.getEpochSecond, Instant.MAX.getEpochSecond).map(Instant.ofEpochSecond)
+
+  val genZoneId: Gen[ZoneId] =
+    Gen.oneOf(ZoneId.SHORT_IDS.values().asScala.toSeq.distinct).map(ZoneId.of)
```
ruippeixotog: Instead of obtaining all short IDs with `ZoneId.SHORT_IDS.values().asScala.toSeq.distinct` you should use `ZoneId.getAvailableZoneIds.asScala`.

core/src/test/scala/pureconfig/gen/package.scala
```diff
@@ -0,0 +1,66 @@
...
+    Gen.nonEmptyListOf(Gen.alphaNumStr).map(parts => parts.map(str => Paths.get(str)).reduce(_ resolve _))
+
+  val genPercentage: Gen[Percentage] =
+    Gen.choose[Int](0, 100).map(Percentage.apply)
+
+  val genJodaDateTime: Gen[org.joda.time.DateTime] =
+    for {
+      dateTime <- ConfigurableSuite.localDateTimeArbitrary.arbitrary
+    } yield new org.joda.time.DateTime(dateTime.getYear, dateTime.getMonthValue, dateTime.getDayOfMonth,
+      dateTime.getHour, dateTime.getMinute, dateTime.getSecond, org.joda.time.DateTimeZone.UTC)
```
ruippeixotog, line 54: Shouldn't we use arbitrary time zones instead of UTC only?

core/src/test/scala/pureconfig/ApiSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+    config.spark.network.timeout should be(FiniteDuration(45, TimeUnit.SECONDS))
+  }
+
+  "loadConfigFromFiles" should "load a complete configuration from a single file" in {
+    case class Conf(b: Boolean, d: Double)
+    val files = listResourcesFromNames("/conf/loadConfigFromFiles/priority2.conf")
+    loadConfigFromFiles[Conf](files) shouldBe Right(Conf(false, 0.001D))
+  }
+
+  "loadConfigWithFallBack" should "fallback if no config keys are found" in {
```
ruippeixotog: Minor nitpick: `loadConfigWithFallBack` should read `loadConfigWithFallback`.

core/src/test/scala/pureconfig/ProductConvertersSuite.scala
```diff
@@ -0,0 +1,131 @@
...
+    val conf5 = ConfigFactory.parseMap(Map("a" -> 2, "d.e" -> 5, "d.g" -> 6).asJava).root()
+    ConfigConvert[Conf].from(conf5).right.value shouldBe Conf(2, "default", 42, InnerConf(5, 6))
+
+    val conf6 = ConfigFactory.parseMap(Map("a" -> 2, "d" -> "notAnInnerConf").asJava).root()
+    val failures = ConfigConvert[Conf].from(conf6).left.value.toList
+    failures should have size 1
+    failures.head shouldBe a[WrongType]
+  }
+
+  it should "not use default arguments if specified through a product hint" in {
```
ruippeixotog: This test should be in `ProductHintSuite`.

core/src/test/scala/pureconfig/ProductConvertersSuite.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  /* A configuration with only simple values and `Option` */
+  case class FlatConfig(b: Boolean, d: Double, f: Float, i: Int, l: Long, s: String, o: Option[String])
+
+  /* A configuration with a field of a type that is unknown to `ConfigConvert` */
+  case class ConfigWithUnknownType(d: org.joda.time.DateTime)
+
+  // arbitrary
+
+  implicitly[Arbitrary[FlatConfig]]
```
ruippeixotog: Why are these `implicitly` calls here?

core/src/test/scala/pureconfig/ProductConvertersSuite.scala
```diff
@@ -0,0 +1,131 @@
...
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{ EitherValues, FlatSpec, Matchers }
+import pureconfig.ConfigConvert.{ catchReadError, fromStringConvert, fromStringReader }
+import pureconfig.arbitrary._
+import pureconfig.error.{ ConfigReaderFailures, KeyNotFound, WrongType }
+
+import scala.collection.JavaConverters._
+import scala.collection.immutable.Seq
+
+class ProductConvertersSuite extends FlatSpec with ConfigConvertChecks with Matchers with EitherValues with GeneratorDrivenPropertyChecks {
```
ruippeixotog: Minor nitpick: this is called `ProductConvertersSuite`, while the one for coproducts is `CoproductConverterSuite`. Can we use either `Converter` or `Converters` in both of them?

core/src/test/scala/pureconfig/CoproductConverterSuite.scala
```diff
@@ -0,0 +1,90 @@
+package pureconfig
+
+import com.typesafe.config.{ ConfigFactory, ConfigObject, ConfigValueFactory }
+import org.scalacheck.{ Arbitrary, Gen }
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{ EitherValues, FlatSpec, Matchers }
+import pureconfig.error.{ ConfigReaderException, KeyNotFound }
+
+class CoproductConverterSuite extends FlatSpec with ConfigConvertChecks with Matchers with EitherValues with GeneratorDrivenPropertyChecks {
```
ruippeixotog: Similar to what you've done with products, don't you want to split the checks for the default coproduct converter and the ones for converters with hints?

ruippeixotog (changes requested):
Great refactor, the tests are surely more organized now! Just take a quick look at the comments below.

core/src/test/scala/pureconfig/ApiSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+    case class Conf(d: Double, i: Int, s: String)
+    loadConfig[Conf] shouldBe Right(Conf(0D, 0, "default"))
+  }
+
+  it should "loadConfig from reference.conf with a namespace" in {
+    case class Conf(f: Float)
+    loadConfig[Conf](namespace = "foo") shouldBe Right(Conf(3.0F))
+  }
+
+  it should "loadConfig from a typelevel Config" in {
```
jcazevedo: This should read `...from a Typesafe Config`.

core/src/test/scala/pureconfig/data/instances/package.scala
```diff
@@ -0,0 +1,14 @@
...
+
+package object instances {
+
+  private val failConfigReadPercentage =
+    (s: String) => (l: Option[ConfigValueLocation]) =>
+      Left(CannotConvert(s, "Percentage", "Percentage is a dummy type, you should not read it", l, None))
+
+  implicit val percentageConfigWriter =
+    ConfigConvert.fromNonEmptyStringConvert[Percentage](failConfigReadPercentage, percentage => s"${percentage.value} %")
+}
```
jcazevedo: I suggest moving this `ConfigConvert` definition to a companion object of the `Percentage` class. It took me a while to figure out where you were redefining the default `case class` derivation logic for this to work with `checkArbitrary2`. Also, it removes the need for the explicit import in `BasicConvertersSuite`.

core/src/test/scala/pureconfig/ApiSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+    loadConfig[Conf](path = path) shouldBe Right(Conf("str", true))
+  }
+
+  it should "loadConfig from a configuration file with a namespace" in {
+    case class Conf(s: String, b: Boolean)
+    val path = createTempFile("""foo.bar { b: true, s: "str" }""")
+    loadConfig[Conf](path = path, namespace = "foo.bar") shouldBe Right(Conf("str", true))
+  }
+
+  it should s"be able to load a realistic configuration file" in {
```
derekmorr: Is the "s" prefix necessary here?

derekmorr:
I wonder if we should factor out some of the common mixins into a shares base class. There are several instances of `class FooSuite extends FlatSpec with ConfigConvertChecks with Matchers with EitherValues with GeneratorDrivenPropertyChecks` 

Perhaps make:

`class BaseSuite extends FlatSpec with ConfigConvertChecks with Matchers with EitherValues with GeneratorDrivenPropertyChecks`

And then have suites descend from that? ScalaTest encourages that - http://www.scalatest.org/user_guide/defining_base_classes

core/src/test/scala/pureconfig/gen/package.scala
```diff
@@ -0,0 +1,66 @@
...
+import java.util.UUID
+
+import org.scalacheck.{ Arbitrary, Gen }
+import pureconfig.configurable.ConfigurableSuite
+import pureconfig.data._
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration.{ Duration, FiniteDuration }
+
+package object gen {
```
derekmorr, line 13: 47 degrees has a useful library of scalacheck generators which includes date/time generators. Should we consider using that instead of making our own ? 

Docs: https://47deg.github.io/scalacheck-toolbox/docs/datetime/index.html

core/src/test/scala/pureconfig/ProductHintSuite.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  behavior of "ProductHint"
+
+  case class ConfWithCamelCaseInner(thisIsAnInt: Int, thisIsAnotherInt: Int)
+  case class ConfWithCamelCase(camelCaseInt: Int, camelCaseString: String, camelCaseConf: ConfWithCamelCaseInner)
+
+  val confWithCamelCase = ConfWithCamelCase(1, "foobar", ConfWithCamelCaseInner(2, 3))
+
+  /** return all the keys in a `ConfigObject` */
+  def allKeys(configObject: ConfigObject): Set[String] = {
```
jcazevedo, line 19: If you convert `ConfigObject` to `Config`, you can list all entries of the config through the `entrySet()` method, so I think this can be simplified to:

```scala
configObject.toConfig().entrySet().asScala.flatMap(_.getKey.split('.')).toSet
```

jcazevedo (changes requested):

core/src/test/scala/pureconfig/gen/package.scala
```diff
@@ -0,0 +1,66 @@
...
+  val genYear: Gen[Year] =
+    Gen.choose(Year.MIN_VALUE, Year.MAX_VALUE).map(Year.of)
+
+  val genHost: Gen[String] =
+    Gen.alphaNumStr
+
+  val genPort: Gen[Int] =
+    Gen.chooseNum(0, 65535)
+
+  val genUUID: Gen[UUID] =
```
derekmorr: Scalacheck's Gen includes a uuid generator. Is this implementation needed?
https://www.scalacheck.org/files/scalacheck_2.11-1.13.4-api/index.html#org.scalacheck.Gen$@uuid:org.scalacheck.Gen[java.util.UUID]

core/src/test/scala/pureconfig/gen/package.scala
```diff
@@ -0,0 +1,66 @@
...
+    for {
+      years <- Arbitrary.arbInt.arbitrary
+      months <- Arbitrary.arbInt.arbitrary
+      days <- Arbitrary.arbInt.arbitrary
+    } yield Period.of(years, months, days)
+
+  val genYear: Gen[Year] =
+    Gen.choose(Year.MIN_VALUE, Year.MAX_VALUE).map(Year.of)
+
+  val genHost: Gen[String] =
```
derekmorr: Is this generating hostnames? If so, it can produce invalid data if we're being strict (RFC 1123).

core/src/test/scala/pureconfig/gen/package.scala
melrief: Just `ZonedDateTime` and `Duration` classes are supported, I think. I'm not sure if we want to add a dependency for just two arbitrary. OT: maybe we should push a PR with the generators to their project?

core/src/test/scala/pureconfig/ProductHintSuite.scala
melrief: Thanks!

core/src/test/scala/pureconfig/gen/package.scala
melrief: Done. Do you know why there is a generator and not an `Arbitrary`?

core/src/test/scala/pureconfig/gen/package.scala
melrief: It's not used, I forgot it together with `genPort`...

core/src/test/scala/pureconfig/ApiSuite.scala
melrief: Done

core/src/test/scala/pureconfig/BasicConvertersSuite.scala
melrief: Removed

core/src/test/scala/pureconfig/gen/package.scala
derekmorr: I've wondered that myself.

core/src/test/scala/pureconfig/PathUtils.scala
melrief: I prefer `createTempFile` because, while it indeed writes the content of the file, it also create the file and return the `Path`. The fact that it writes is only one of the steps :).

core/src/test/scala/pureconfig/gen/package.scala
melrief: This generator is just used for a testing instance and the point of the test is not to test that the instance is correct. I would keep it as it is, it's more then enough for the tests we have.

core/src/test/scala/pureconfig/ApiSuite.scala
melrief: Fixed

core/src/test/scala/pureconfig/ProductConvertersSuite.scala
melrief: `Converters` :), fixed

core/src/test/scala/pureconfig/ProductConvertersSuite.scala
melrief: Removed

core/src/test/scala/pureconfig/ProductConvertersSuite.scala
melrief: Moved, thanks!

core/src/test/scala/pureconfig/CoproductConverterSuite.scala
melrief: Right, done! Thanks.

core/src/test/scala/pureconfig/ApiSuite.scala
melrief: Removed

melrief:
@derekmorr good idea to create the `BaseSuite`, done!

ruippeixotog (approved):

derekmorr (approved):

jcazevedo (approved):

PR #171, opened 2017-03-19
Split ConfigConvert object into multiple traits

ruippeixotog:
In this PR `ConfigConvert` is split into a `ConvertHelpers` trait with all helpers such as `fromStringReader`, a `BasicConverters` with `ConfigConvert` instances for primitive objects (previously `LowPriorityConfigConvertImplicits`) and a `DerivedConversions` with higher-level converters.

The way it is now it is very difficult to refactor `ConfigConvert`, as it is such a big source file and different kinds of definitions and objects are somewhat mixed. This separates them logically and helps us later in #92, where it will be simpler to keep the reading and writing parts in sync.

This change should not break source compatibility.


melrief (approved):
I had a look at the structure and not at the source code because I assume you haven't changed anything there.
I like the names you picked and the fact that everything is a trait with a companion object extending the trait itself. Readable and clean. Good job.

core/src/main/scala/pureconfig/ConvertHelpers.scala
```diff
@@ -0,0 +1,161 @@
...
+  @deprecated(message = "The usage of Try has been deprecated. Please use fromStringConvert instead", since = "0.6.0")
+  def stringConvert[T](fromF: String => Try[T], toF: T => String): ConfigConvert[T] =
+    fromStringConvert[T](fromF andThen tryToEither, toF)
+
+  def fromStringConvert[T](fromF: String => Option[ConfigValueLocation] => Either[ConfigReaderFailure, T], toF: T => String): ConfigConvert[T] = new ConfigConvert[T] {
+    override def from(config: ConfigValue): Either[ConfigReaderFailures, T] = stringToEitherConvert(fromF)(config)
+    override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(toF(t))
+  }
+
+  def fromStringConvertTry[T](fromF: String => Try[T], toF: T => String)(implicit ct: ClassTag[T]): ConfigConvert[T] = {
```
leifwickland, line 136: Not strictly related to your change:
Should this method be deprecated like the rest of the methods which take `Try`?

core/src/main/scala/pureconfig/BasicConverters.scala
```diff
@@ -0,0 +1,106 @@
...
+import com.typesafe.config._
+import pureconfig.error._
+
+/**
+ * Trait containing `ConfigConvert` instances for primitive types and simple classes in Java and Scala standard
+ * libraries.
+ */
+trait BasicConverters extends ConvertHelpers {
+
+  // primitive type converters
```
leifwickland: (Please pardon the following grumpiness. I'm just expressing a preference. If my suggestion seems crazy, don't worry about it.)


I kind of loathe comments used like this to demarcate blocks of similar things. I'd rather see types used to group similar things.

My preference would be for something more like:

```
trait PrimativeConverters {
  ...
}

trait UriAndPathConverters {
  ...
}

trait JavaTimeConverters {
  ...
}

// etc

trait BasicConverters 
  extends PrimativeConverters 
  with UriAndPathConverters
  with JavaTimeConverters
  with ... {
...
}

object BasicConverters extends BasicConverters
```

core/src/main/scala/pureconfig/DerivedConverters.scala
```diff
@@ -0,0 +1,268 @@
...
+          fail(WrongType(other.valueType().toString, "ConfigObject", ConfigValueLocation(other), None))
+      }
+    }
+
+    override def to(keyVals: Map[String, T]): ConfigValue = {
+      ConfigValueFactory.fromMap(keyVals.mapValues(configConvert.value.to).asJava)
+    }
+  }
+
+  // used for products
```
leifwickland, line 237: (I know you didn't add this comment. I'm just realizing it's here and it annoys me. Feel free to ignore me.)
Nitpick: This comment doesn't feel useful. It doesn't tell me anything the name doesn't. It feels like either the comment ought to be expanded with some additional insight or deleted entirely.

core/src/main/scala/pureconfig/DerivedConverters.scala
```diff
@@ -0,0 +1,268 @@
...
+    override def from(config: ConfigValue): Either[ConfigReaderFailures, F] = {
+      cc.value.fromWithDefault(config, default()).right.map(gen.from)
+    }
+
+    override def to(t: F): ConfigValue = {
+      cc.value.to(gen.to(t))
+    }
+  }
+
+  // used for coproducts
```
leifwickland, line 253: (I know you didn't add this comment. I'm just realizing it's here and it annoys me. Feel free to ignore me.)
Ditto the above nitpick about the comment.

core/src/main/scala/pureconfig/DerivedConverters.scala
```diff
@@ -0,0 +1,268 @@
...
+            case Left(failures) => throw new ConfigReaderException[FieldType[Name, V] :+: T](failures)
+            case Right(r) => r
+          }
+
+        case Inr(r) =>
+          tConfigConvert.value.to(r)
+      }
+    }
+
+  // For Option[T] we use a special config converter
```
leifwickland, line 151: (I know you didn't add this comment. I'm just realizing it's here and it annoys me. Feel free to ignore me.)
Please expand the comment to indicate why or remove it. The comment isn't telling me anything the name doesn't.

core/src/main/scala/pureconfig/ConvertHelpers.scala
ruippeixotog: I think this method was introduced at the time of the change from `Try` to `Either`. I believe it's OK and useful to have helper methods taking `Try` functions, as long as they are clearly marked as such and options using `Either` are also provided.

core/src/main/scala/pureconfig/BasicConverters.scala
ruippeixotog: I don't mind changing it, I'm ok either way :) @melrief and @jcazevedo, do you have any preference about this?

leifwickland (approved):
@ruippeixotog Thanks for making these improvements. 
I made a few suggestions. Feel free to take or leave them.

core/src/main/scala/pureconfig/ConvertHelpers.scala
leifwickland: OK! Thanks for explaining.

ruippeixotog:
@leifwickland I took note of your comments about useless comments in the source code. I agree with you on all of them, but if you don't mind I'll do that in the next PR. The reason is simple: I already have local changes on top of this PR moving a lot of code around and it's way easier for me to remove the comments on my latest commit instead of rebasing :)

leifwickland:
@ruippeixotog I heartily approve of dealing with the useless comments in another PR.

core/src/main/scala/pureconfig/BasicConverters.scala
leifwickland: If you'll permit the logical fallacy of appeal to authority, I'll refer you to how Cats did something similar to this here:
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/instances/anyval.scala#L4

jcazevedo: I'm OK either way.

melrief: +1 on the proposal from @leifwickland. Cats style of having one trait per-type is a bit too much for me but grouping is helpful.

ruippeixotog: I also do that on my on code, it's just that my threshold of "number of instances" for triggering that split is slightly higher :D But I understand the advantages and agree with the split.

Done in 866c95f.

core/src/main/scala/pureconfig/ConvertHelpers.scala
melrief: The reason is that the error representation in `Try` and even more in `Option` are very primitive and carry less information compared to `Either[ConfigReaderFailure]`, which means that it can be difficult for the user to convert them into an `Either`. We provide helpers that try to get the best out of those types.

PR #175, opened 2017-03-20
add Derek Morr to the AUTHORS and the pom of core

melrief:


derekmorr:
This looks good to me.

PR #178, opened 2017-03-21
Tut docs

derekmorr:
This is an initial version of tut-support for `docs/`. 

Running `docs/tut` from sbt will build docs.

Changes:

- I moved docs from `core/docs` into `docs/` because the main README.md is already in `core/`. The `core/tut` already builds that readme and copies it into `/`. I don't think it's possible to have multiple sources and targets defined for tut inside a single project. So, the docs tut files are in `docs/src/main/tut` and the output is written to `docs/`

- One of the docs doesn't have stable output: `docs/error-handling.md`. The `KeyNotFound` example includes the full path to the conf file: `KeyNotFound(xpto,Some(ConfigValueLocation(` **/Users/dvm105/pureconfig/** `docs/target/scala-2.12/classes/application.conf,11)))`. I'm not sure how to remove this.


docs/error-handling.md
```diff
@@ -0,0 +1,26 @@
...
+import com.typesafe.config._
+import pureconfig.error._
+
+val cv = ConfigFactory.load.root().get("conf")
+```
+
+We can try to load a missing key and get a useful error:
+```scala
+KeyNotFound("xpto", ConfigValueLocation(cv))
+// res1: pureconfig.error.KeyNotFound = KeyNotFound(xpto,Some(ConfigValueLocation(/Users/dvm105/pureconfig/docs/target/scala-2.12/classes/application.conf,11)))
```
leifwickland: Perhaps what we could do to make this have a stable output here is to just print a substring of the filename and the line number. For example:
```
val notFound =  KeyNotFound("xpto", ConfigValueLocation(cv))
notFound.filename.takeRight(24)
notFound.lineNumber
```

derekmorr:
Just thinking out loud here -- should `ConfigValueLocation`'s filename be a `java.nio.file.Path` ? If so, we would have more control of output. We could just call `filename.getFileName` to get `application.conf`. Or we could get fancy and relativize the path, like:

```scala
scala> import java.nio.file.{Path, Paths}
import java.nio.file.{Path, Paths}

scala> val appConfPath = Paths.get("/Users/dvm105/pureconfig/docs/target/scala-2.12/classes/application.conf")
appConfPath: java.nio.file.Path = /Users/dvm105/pureconfig/docs/target/scala-2.12/classes/application.conf

scala> appConfPath.getFileName
res0: java.nio.file.Path = application.conf

scala> System.getProperty("user.dir")
res1: String = /Users/dvm105/pureconfig

scala> val pc = Paths.get(System.getProperty("user.dir"))
pc: java.nio.file.Path = /Users/dvm105/pureconfig

scala> pc.relativize(appConfPath)
res2: java.nio.file.Path = docs/target/scala-2.12/classes/application.conf
```

leifwickland (approved):

docs/error-handling.md
```diff
@@ -0,0 +1,26 @@
...
+```scala
+import com.typesafe.config._
+import pureconfig.error._
+
+val cv = ConfigFactory.load.root().get("conf")
+```
+
+We can try to load a missing key and get a useful error:
+```scala
+KeyNotFound("xpto", ConfigValueLocation(cv))
```
leifwickland, line 24: @derekmorr 
Much like you suggested in your comment about filename being a path, I was pondering the possibility that we could make the output stable by only printing the relative path of the file. Something like:
```
val notFound = KeyNotFound("xpto", ConfigValueLocation(cv)).location.get
notFound.filename.takeRight(24)
notFound.lineNumber
```

jcazevedo:
> Just thinking out loud here -- should `ConfigValueLocation`'s filename be a `java.nio.file.Path`?

I think that makes sense.

leifwickland:
@derekmorr Regarding making the error location's `filename` a `Path`, I think we might need to prefer a `URL` instead given that you have no guarantee that the config "files" were in fact files. They could've been elements within the `.jar` file, which I think means that they "filename" will end up being an invalid path.
The source of that information is this class from typesafe config:
https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigOrigin.html

I would bet Havoc had a good reason for leaving the filename as a string instead of a `Path`.

In general, I think perhaps we want the `filename`-as-`Path` discussion as a separate PR. I think thehacky approach to making the output of `error-handling.md` stable that I suggested might be the most expedient path to done for this one.

jcazevedo:
I've opened #180 to propose the change to `Path`. 

@leifwickland your point is valid. I'll make some tests and see if the change to `URL` covers all cases.

melrief (approved):
Nice!

PR #179, opened 2017-03-21
Split `ConfigConvert` into readers and writers

### LINKED ISSUE #92: Split ConfigConverter into two parts: for reading and for writing

ruippeixotog:
In this PR `ConfigConvert` is redefined as an union of two new traits: `ConfigReader` for reading objects from configs and `ConfigWriter` to write objects to configs.

`ConfigConvert` continues to exist and I think it still plays an important role in pureconfig. Most of the times users will be able to define both operations for a type, and for organization purposes it's good to define them together. However, with PR users will be able to define only one of the operations and still use operations such as `loadConfig` and `toConfig`, which is particularly useful for some particular models that only have a one-way conversion.

I opted to split the `BasicConverters` and `DerivedConverters` in readers and writers in order to mix into the companion objects of `ConfigReader` and `ConfigWriter` only the implicits that mattered to each one. However, it definitely doesn't have to be the case and I think we should leave the `ConfigConvert` definitions in pureconfig modules as they are. A `ConfigConvert` can be seen both as a `ConfigReader as a `ConfigWriter`, and in the modules we don't have any advantage in splitting them.

This change, when seen together with #171, is source-compatible with the latest version 0.6.0. If we do a release without this PR it may be possible for users to start using `BasicConverters` and `DerivedConverters` directly and this PR causes them to disappear.

Closes #92.


ruippeixotog:
This PR is a WIP as it is still missing tests and documentation. I'd like to get some feedback from you before starting writing documentation. I'll be writing tests meanwhile.

melrief (dismissed):
I'm not sure why  this is still WIP, I've read the PR and I think it's nice. I have no comments.

ruippeixotog:
This PR is a WIP as it is still missing tests and documentation. However, as the tests for `ConfigConvert` still check both parts of the conversion, I don't think anything more is necessary for now. If you agree with the PR as it is now, I'll update the README and then we can merge this.

melrief:
@ruippeixotog I'm concerned about possible conflicts with #178 , do you think we should wait for that to be merged before we change the documentation?

ruippeixotog:
Good question... I guess it's better if this is merged without the documentation then and I'll do another PR later, after #178. That way no one will have much trouble dealing with merge conflicts.

melrief:
Agreed, I'll wait for another approve and then merge

core/src/main/scala/pureconfig/ConvertHelpers.scala
```diff
@@ -94,36 +102,68 @@ trait ConvertHelpers {
     override def to(t: T): ConfigValue = ConfigValueFactory.fromAnyRef(t)
   }
 
-  def fromStringReader[T](fromF: String => Option[ConfigValueLocation] => Either[ConfigReaderFailure, T]): ConfigConvert[T] = new ConfigConvert[T] {
+  def fromStringReader[T](fromF: String => Option[ConfigValueLocation] => Either[ConfigReaderFailure, T]): ConfigReader[T] = new ConfigReader[T] {
```
melrief: Do you think we should move reader/writer helpers in the `Reader`/`Writer` companion object? I think it could be nice to have `ConfigReader.fromString` instead of `ConverterHelper.fromStringReader`. You don't need the `Reader` at the end of `fromStringReader` if the object in which the method is defined is a `Reader` (I hope I explained myself). What do you think?

ruippeixotog: Yes, that's a good idea! I'll move all the methods, but I'll leave forwarder deprecated methods in `ConfigConvert` in order not to break compatibility here, ok?

melrief: Yes

ruippeixotog:
@melrief besides the renaming you suggested, I ended up doing a greater number of changes in order to keep the code clean and with consistent names in 3a9823b.

`ConvertHelpers` doesn't have factory methods anymore, not even `ConfigConvert` ones. It ended up with nothing more than... helpers. All factory methods were moved to the adequate companion object. I prefer to have all constructors directly in the companion object and not in a separate trait, and I also like that the companion objects are not polluted anymore with methods not relating to them directly.

I tried to match the factory method name to the type of entity to build. `ConfigReader` has the `fromString` factory method family, with all try/option and non-empty variants. `ConfigWriter` has `forPrimitive`, `toDefaultString` and `toString` factory methods. `ConfigConvert` has a `viaString` factory method family like `ConfigReader`.

I left a large number of deprecated constructors in `ConfigConvert` in order not to break existing code. I hope I didn't forget any method, but everything should be there.

Let me know what you think of it now!

melrief:
@ruippeixotog thanks for the changes, the new API is much nicer in my opinion! I don't have time to do the review but tomorrow I'll have a look in deep.

melrief (approved):
Nice, I like the new API for `ConfigReader` and `ConfigWriter` even more. I'm waiting for another review.

core/src/main/scala/pureconfig/ConfigConvert.scala
```diff
@@ -6,30 +6,116 @@
...
-   * Converts a type `T` to a `ConfigValue`.
-   *
-   * @param t The instance of `T` to convert
-   * @return The `ConfigValue` obtained from the `T` instance
-   */
-  def to(t: T): ConfigValue
-}
+trait ConfigConvert[T] extends ConfigReader[T] with ConfigWriter[T]
+
+object ConfigConvert extends ConvertHelpers {
```
leifwickland: Suggestion: Add a class-level comment like:
```
/**
 * This object provides methods to create [ConfigConvert] instances.
 */
```

core/src/main/scala/pureconfig/ConfigReader.scala
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Convert the given configuration into an instance of `T` if possible.
+   *
+   * @param config The configuration from which load the config
+   * @return either a list of failures or an object of type `T`
+   */
+  def from(config: ConfigValue): Either[ConfigReaderFailures, T]
+}
+
+object ConfigReader extends BasicReaders with DerivedReaders {
```
leifwickland: Suggestion: Add a class-level comment like:
```
/**
 * This object provides methods to create [ConfigReader] instances.
 */
```

core/src/main/scala/pureconfig/ConfigWriter.scala
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * Converts a type `T` to a `ConfigValue`.
+   *
+   * @param t The instance of `T` to convert
+   * @return The `ConfigValue` obtained from the `T` instance
+   */
+  def to(t: T): ConfigValue
+}
+
+object ConfigWriter extends BasicWriters with DerivedWriters {
```
leifwickland: Suggestion: Add a class-level comment like:
```
/**
 * This object provides methods to create [ConfigWriter] instances.
 */
```

core/src/main/scala/pureconfig/DerivedWriters.scala
```diff
@@ -0,0 +1,113 @@
...
+    implicit
+    key: Witness.Aux[K],
+    vFieldConvert: Lazy[ConfigWriter[V]],
+    tConfigWriter: Lazy[WrappedConfigWriter[Wrapped, T]],
+    hint: ProductHint[Wrapped]): WrappedConfigWriter[Wrapped, FieldType[K, V] :: T] = new WrappedConfigWriter[Wrapped, FieldType[K, V] :: T] {
+
+    override def to(t: FieldType[K, V] :: T): ConfigValue = {
+      val keyStr = hint.configKey(key.value.toString().tail)
+      val rem = tConfigWriter.value.to(t.tail)
+      // TODO check that all keys are unique
```
leifwickland, line 33: The immortal `TODO` lives on.

core/src/main/scala/pureconfig/DerivedWriters.scala
```diff
@@ -0,0 +1,113 @@
...
+    implicit
+    coproductHint: CoproductHint[Wrapped],
+    vName: Witness.Aux[Name],
+    vFieldConvert: Lazy[ConfigWriter[V]],
+    tConfigWriter: Lazy[WrappedConfigWriter[Wrapped, T]]): WrappedConfigWriter[Wrapped, FieldType[Name, V] :+: T] =
+    new WrappedConfigWriter[Wrapped, FieldType[Name, V] :+: T] {
+
+      override def to(t: FieldType[Name, V] :+: T): ConfigValue = t match {
+        case Inl(l) =>
+          // Writing a coproduct to a config can fail. Is it worth it to make `to` return a `Try`?
```
leifwickland, line 63: Is this the next great zombie comment?

leifwickland (approved):
Kudos, @ruippeixotog! This was a huge change that seems to be have executed extremely well.
I offered a couple suggestions, but this LGTM. 👍

core/src/main/scala/pureconfig/DerivedWriters.scala
melrief: I think we should address old TODOs in another PR because this one is dealing with enough stuff :-)

derekmorr (approved):

core/src/main/scala/pureconfig/ConfigConvert.scala
ruippeixotog: Added in 5845cac.

core/src/main/scala/pureconfig/ConfigReader.scala
ruippeixotog: Added in 5845cac.

core/src/main/scala/pureconfig/ConfigWriter.scala
ruippeixotog: Added in 5845cac.

core/src/main/scala/pureconfig/DerivedWriters.scala
ruippeixotog: I agree with @leifwickland. The comment is still relevant and unrelated to this PR.

core/src/main/scala/pureconfig/DerivedWriters.scala
ruippeixotog: I believe the comment is still relevant - it mentions a question for which there is still no clear answer. Though it may not be the best place to make TODOs and discussion topics visible, I don't mind having comments like this around for as long as it takes.

ruippeixotog:
I have just fixed the merge conflicts. Unless you want to wait for @jcazevedo to review this, this is ready to merge, @melrief.

@leifwickland I believe I replied to all your comments. Please let me know if you see anything else :)

jcazevedo (approved):
This looks good to merge as is! Sorry for not reviewing it earlier.

PR #180, opened 2017-03-21
Modify ConfigValueLocation to use URL instead of String

jcazevedo:
As suggested by @derekmorr in #178, this PR modifies `ConfigValueLocation` to use `Path` instead of `String` to encode the location of failures.

### DIFF
--- a/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
+++ b/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
@@ -4,6 +4,7 @@
 package pureconfig.error
 
 import com.typesafe.config.{ ConfigOrigin, ConfigValue }
+import java.nio.file.{ Path, Paths }
 
 /**
  * The physical location of a ConfigValue, represented by a file name and a line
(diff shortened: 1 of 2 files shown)


derekmorr:
This looks good to me, but @leifwickland raised some concerns in https://github.com/melrief/pureconfig/pull/178#issuecomment-288207971

jcazevedo:
I've modified `ConfigValueLocation` to use `URL` instead of `Path`, in order to be able to encode the location of configuration files inside `.jar` files.

derekmorr:
`Path` should be able to handle files in a jar file, but arguable using `URL` is better since `ConfigFactory` can load files from` URL`s.

melrief:
I prefer `URL` because a `URL` can identify any resource in the system and can be used for locations that are not necessarily files.

melrief (approved):
Looks good to me

jcazevedo:
Since `ConfigOrigin` has `url` defined in cases where `filename` is `null`, I also think we're better off with `URL`.

melrief:
I approved this, let's wait for another approval and merge.

derekmorr:
Approved.

PR #181, opened 2017-03-22
Polish the failure summary

### LINKED ISSUE #154: Polish the failure summary

jcazevedo:
This PR attempts to polish the failure summary, i.e. the message that is printed by the exception the user may decide to throw when attempting to derive an object from a `ConfigValue`. It does so by grouping and sorting errors by the keys they originated from. 

Incidentally, this fixes an issue with the keys for errors of Map derivations not being properly concatenated (the derivation was missing a call to `improveFailures`), makes the output of the `WrongType` failure more uniform, by always referring to the appropriate `ConfigValueType`s, and removes the `WrongTypeException` class, which was not being used, and was most likely forgotten.

This fixes #154.


core/src/main/scala/pureconfig/BasicConverters.scala
```diff
@@ -90,15 +90,15 @@ trait TypesafeConfigConverters extends ConvertHelpers {
   implicit val configConfigConvert: ConfigConvert[Config] = new ConfigConvert[Config] {
     override def from(config: ConfigValue): Either[ConfigReaderFailures, Config] = config match {
       case co: ConfigObject => Right(co.toConfig)
-      case other => fail(WrongType(other.valueType().toString, "ConfigObject", ConfigValueLocation(config), None))
+      case other => fail(WrongType(s"${other.valueType}", s"${ConfigValueType.OBJECT}", ConfigValueLocation(config), None))
```
melrief: I'm maybe too picky but when I prefer to use the standard `.toString` instead of string interpolation for one value. Same for all the other occurrences.

melrief (approved):

core/src/main/scala/pureconfig/BasicConverters.scala
jcazevedo: Sure. I've changed it in 9c20d08.

core/src/main/scala/pureconfig/BasicConverters.scala
```diff
@@ -90,15 +90,15 @@ trait TypesafeConfigConverters extends ConvertHelpers {
   implicit val configConfigConvert: ConfigConvert[Config] = new ConfigConvert[Config] {
     override def from(config: ConfigValue): Either[ConfigReaderFailures, Config] = config match {
       case co: ConfigObject => Right(co.toConfig)
-      case other => fail(WrongType(other.valueType().toString, "ConfigObject", ConfigValueLocation(config), None))
+      case other => fail(WrongType(other.valueType.toString, ConfigValueType.OBJECT.toString, ConfigValueLocation(config), None))
```
ruippeixotog: I see you started using `ConfigValueType` instances to refer to expected types in `WrongType`. In order to make `WrongType` more type-safe, why don't you make `expectedType` a `List[ConfigValueType]`? That way you make sure the expected types passed are always valid.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -72,6 +72,16 @@ sealed abstract class ConfigReaderFailure {
...
 
   /**
+   * A human-readable description of the failure.
+   */
+  def description: String
+
+  /**
+   * A human-readable description of the failure, along with its location.
+   */
+  def descriptionWithLocation: String = location.fold(description)(_.description + " " + description)
```
ruippeixotog: I preferred if this was a private function local to `ConfigReaderException`. In my view at least, `ConfigReaderFailure` should be as decoupled from presentation as possible, while all logic for turning "failure" domain models into nice error messages should be in `ConfigReaderException` whenever possible. What do you think?

core/src/main/scala/pureconfig/error/ConfigReaderException.scala
```diff
@@ -12,42 +12,30 @@ final case class ConfigReaderException[T](failures: ConfigReaderFailures)(implic
     val linesBuffer = mutable.Buffer.empty[String]
     linesBuffer += s"Cannot convert configuration to a value of class ${ct.runtimeClass.getName}. Failures are:"
```
ruippeixotog: Suggestion: "value of class" is a bit redundant here. It would be nicer if it read "Cannot convert configuration to a ${ct.runtimeClass.getName}", IMO.

core/src/main/scala/pureconfig/error/ConfigReaderException.scala
```diff
@@ -12,42 +12,30 @@ final case class ConfigReaderException[T](failures: ConfigReaderFailures)(implic
...
 
-    if (otherFound.nonEmpty) {
-      linesBuffer += "  Other failures:"
-      for (failure <- otherFound) {
-        val desc = failure.location.fold("")(_.description)
-        linesBuffer += s"    - $desc $failure"
-      }
+    failuresWithPath.foreach {
+      case (p, failures) =>
+        linesBuffer += s"  with '$p':"
```
ruippeixotog: Suggestion: "at" seems more natural than "with" here.

ruippeixotog (changes requested):
LGTM, and kudos for adding nice specs for this! I requested some changes but none of them are major blockers, feel free to leave the PR as it is if you prefer.

core/src/main/scala/pureconfig/BasicConverters.scala
jcazevedo: Makes sense. I modified `WrongType` to only accept `ConfigValueType`s in 6079331.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: I agree. Moved the `descriptionWithLocation` method to the `ConfigReaderException` companion object in 53ede88.

core/src/main/scala/pureconfig/error/ConfigReaderException.scala
jcazevedo: Fixed it in 668d6cb.

core/src/main/scala/pureconfig/error/ConfigReaderException.scala
jcazevedo: I struggled a bit with words there. Originally I was using "in", but then it didn't make much sense for missing keys, because they're not **in** there (they're missing!). I ended up with "with", but I agree "at" sounds better. Replaced it in 482f56f.

ruippeixotog (approved):

melrief:
Thanks @jcazevedo !

PR #183, opened 2017-03-22
fix ConfigConvert[Option[_]].to for None

melrief:
Doing this before #151 because I'd like to have a commit in the git history

### DIFF
--- a/core/src/test/scala/pureconfig/PureconfSuite.scala
+++ b/core/src/test/scala/pureconfig/PureconfSuite.scala
@@ -51,7 +51,7 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with Either
     }
   }
 
-  "ConfigConvert" should s"be able to convert from and to None" in {
+  "ConfigConvert" should "be able to convert from and to None" in {
     val cc = ConfigConvert[Option[Int]]
     cc.from(cc.to(None)) shouldBe Right(None)
   }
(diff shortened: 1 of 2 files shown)


jcazevedo (approved):

core/src/test/scala/pureconfig/PureconfSuite.scala
```diff
@@ -51,6 +51,11 @@ class PureconfSuite extends FlatSpec with Matchers with OptionValues with Either
     }
   }
 
+  "ConfigConvert" should s"be able to convert from and to None" in {
```
derekmorr: minor nit: We shouldn't need the "s" here, but it's harmless to leave it in.

derekmorr:
Looks good to me. I made one very small comment, which you can feel free to ignore.

core/src/test/scala/pureconfig/PureconfSuite.scala
melrief: Fixed, thanks for pointing out when I add the `s` without reason. I think I'm doing it a lot and it's completely involuntary.

melrief:
Ok then, I think we can merge. @derekmorr in the comments for PRs there is a button to approve or request changes for a PR, if you want. It's easier for the contributor to see the list of reviewers and what they think in that way because it's just on the right of the PR when you open it.

derekmorr (approved):

PR #185, opened 2017-03-23
run tut as part of CI

derekmorr:
This is the last part of tut-ify'ing the docs -- running tut in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ scala:
 jdk:
     - oraclejdk8
 script:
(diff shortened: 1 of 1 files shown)


melrief:
Waiting for CI and then merge. There is no need for two reviews for a PR that adds one word :smiley:

jcazevedo:
As discussed in #163, I think we can do a little better by checking if running `tut` produces git changes, as that way we can be confident that the documentation is in sync with the codebase.

derekmorr:
We need to make the output stable first. There's one example that isn't yet. I'l push up a PR later today to fix that.

PR #186, opened 2017-03-23
make error-handling output stable

derekmorr:
This fixes the last bit of tut output that wasn't stable between runs.

I relativized the path using Paths. This may be over complicated.

### DIFF
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -10,7 +10,7 @@ using the `ConfigReaderFailure` sealed family of case classes, you can use the
 `ConfigValueLocation.apply(cv: ConfigValue)` method to automatically create an
 optional location from a `ConfigValue`.
 
-Given this setup:
+If we load a config and try to load a missing key:
 
 ```scala
 import com.typesafe.config._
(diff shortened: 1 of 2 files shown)


docs/error-handling.md
```diff
@@ -15,12 +15,20 @@ Given this setup:
...
 
 val cv = ConfigFactory.load.root().get("conf")
 ```
 
 We can try to load a missing key and get a useful error:
 ```scala
-KeyNotFound("xpto", ConfigValueLocation(cv))
-// res1: pureconfig.error.KeyNotFound = KeyNotFound(xpto,Some(ConfigValueLocation(/Users/dvm105/pureconfig/docs/target/scala-2.12/classes/application.conf,11)))
+val notFound = KeyNotFound("xpto", ConfigValueLocation(cv)).location.get
+// notFound: pureconfig.error.ConfigValueLocation = ConfigValueLocation(file:/home/derek/pureconfig/docs/target/scala-2.12/classes/application.conf,11)
```
leifwickland: `tut` is printing this line, which I think will cause the output to be unstable due to including the full path. Perhaps this line ought to be in a `tut:silent` block.

leifwickland (changes requested):
@derekmorr I think this is a fine approach, but I don't think it accomplishes the goal. I hope my suggestion helps.

docs/error-handling.md
derekmorr: Yah, that'll teach me to push PR during a conference talk.

leifwickland (approved):

melrief (approved):

PR #188, opened 2017-03-28
initial version of akka module

derekmorr:
As discussed on gitter today, here's a module that loads Akka `Timeout` values.

This still fails on 2.10 b/c SBT tries to build for 2.10 (and Akka 2.4 isn't published for 2.10).


melrief (approved):
LGTM, no comments. Now that we have a module for `akka`, I wonder how many other classes we could support.

This is a module and you don't need a second review if you want to. If you agree, I can merge it but we will have to wait for the next version of pureconfig to publish the artifact.

derekmorr:
This will fail the builds on Travis bc it doesn't compile on Scala 2.10. I need to figure out how not to run it on 2.10.

ruippeixotog:
@derekmorr I have fixed the SBT build to work in the configuration we want, but I cannot push to this branch as it's in your personal remote. Instead, I pushed the fix to the branch `derekmorr-akka-module` in `pureconfig/pureconfig`. You should be able to pull those commits and do a fast-forward merge to yours. That way we won't lose the history of this PR or any of the commits.

PR #189, opened 2017-03-28
update github reference to pureconfig project

derekmorr:
This updates the metadata in various `build.sbt`'s to our new Github project.


ruippeixotog (commented):
I agree with this change, but we should be careful updating the organization.

Users must know the next version of pureconfig will be under the `com.github.pureconfig` organization in Maven. Updating the README will make new users use the latest pureconfig, but people already using pureconfig (in particular people like me, which run a command to list dependencies that can be updated) will take a long time realizing they're not on the latest version. Do you have a suggestion on how we should approach this?

core/build.sbt
```diff
@@ -2,9 +2,9 @@ import Dependencies._
 
 name := "pureconfig"
 
-organization := "com.github.melrief"
+organization := "com.github.pureconfig"
```
melrief, line 5: I wonder if it still make sense to have `com.github`

melrief:
I could mirror the artifact on `com.github.melrief` if you want, so existing users will be able to use the new versions even with the old organization. I just have to publish twice the artifact for a while. It's not the simplest thing to do but it's doable...

melrief (approved):
LGTM but let's first answer to @ruippeixotog and decide how to deal with the artifact organization.

jcazevedo:
> I could mirror the artifact on `com.github.melrief` if you want, so existing users will be able to use the new versions even with the old organization. I just have to publish twice the artifact for a while. It's not the simplest thing to do but it's doable...

I think that's hard to manage, and we're just delaying the problem until we stop publishing to the `com.github.melrief` organization. 

One thing we can do is publish one last version to the `com.github.melrief` organization in which we deprecate all classes and methods and we point users to the new organization in the deprecation message. It's a bit of an hack, but we're sure to get people's attention that way.

derekmorr:
We should probably update the README and post messages on Twitter (do we have a twitter feed) about the name change. Maybe the typelevel twitter account could post a message too.

We should also update Scaladex, assuming it's not done automatically.

core/build.sbt
ruippeixotog: In order to be accepted for publishing at Sonatype, the organization name has to follow a DNS-like name and you have to have some "degree" of control over it. I personally prefer that way too, both with Maven dependencies and Java packages - it is a good measure to ensure that the names are unique.

melrief:
I like @jcazevedo, I can do it when we release the next version of pureconfig. I can create a branch where everything is deprecated.

ruippeixotog:
I also agree with @jcazevedo's suggestion. Despite being a hack, it will get the attention of everyone, particularly the attention of people that trust [sbt-updates](https://github.com/rtimush/sbt-updates) to tell them exactly what to update, like me.

derekmorr:
I wonder if we also need to update the reference in `example/build.sbt` : 

```scala
libraryDependencies ++= Seq(
  "com.github.melrief" %% "pureconfig" % "0.6.0",
```

to point to `com.github.pureconfig`

melrief:
After, we first need to publish the artifact. Otherwise the tests will fail when compiling the example.

Issue #194, opened 2017-03-30
Created a "deprecated" branch for com.github.melrief organization


melrief:
As proposed by @jcazevedo in #189, we should publish the `pureconfig` artifact with the old organization `com.github.melrief` one last time with all the classes and methods deprecated and pointing to the new artifact.

melrief:
Branch [`deprecated_artifact`](https://github.com/pureconfig/pureconfig/tree/deprecated_artifact) change all the methods and classes to be deprecated.

PR #198, opened 2017-03-31
set version to 0.7.0

melrief:


ruippeixotog (commented):
LGTM, only two questions:

- Regarding your last commit, shouldn't the example project use always the latest version of pureconfig, to ensure it serves as an example for new users?
- Will we update the organization to `com.github.pureconfig` in this version? If so, we should have that change ready before merging this, in order to minimize the time the README will point to a non-published version.

melrief:
1. after last issue with the example, which wasn't compiling anymore, @derekmorr changed the testing on travis to [also compile the example](https://github.com/pureconfig/pureconfig/blob/master/.travis.yml#L10). Because the example uses the artifact published on maven, we must publish the artifact for the last version before we can update the example. Which means that in this case we must first publish pureconfig 0.7.0 and then change the example to use it. It's a bit more complex than previously but in this way we can be sure that the example compiles against the last artifact published, which is the one used by our users.
2. we will. I want to incorporate first all the changes, then make a branch with all classes and methods deprecated (#194), then publish quietly the artifact from that branch, then merge #189 and finally publish the new official artifact. Sorry to not be clear about the release process that I had in mind for 0.7.0.

leifwickland (approved):

PR #202, opened 2017-04-01
fix scaladoc errors and add sbt-api-mappings plugin

### LINKED ISSUE #200: Fix "Could not find any member to link for" in core scaladoc
Specifically

```
> packageDoc
[info] Main Scala API documentation to /home/mariop/prog/pureconfig/core/target/scala-2.12/api...
model contains 72 documentable templates
[error] /home/mariop/prog/pureconfig/core/src/main/scala/pureconfig/BasicReaders.scala:66: Could not find any member to link for "scala.concurrent.duration.Duration".
[error] /**
[error] ^
[error] /home/mariop/prog/pureconfig/core/src/main/scala/pureconfig/BasicWriters.scala:54: Could not find any member to link for "Duration".
[error] /**
[error] ^
[error] /home/mariop/prog/pureconfig/core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala:10: Could not find any member to link for "com.typesafe.config.ConfigFactory".
[error] /**
[error] ^
[error] /home/mariop/prog/pureconfig/core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala:16: Could not find any member to link for "ConfigFactory.invalidateCaches()".
[error]   /** @see [[ConfigFactory.invalidateCaches()]] */
[error]   ^
[error] /home/mariop/prog/pureconfig/core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala:20: Could not find any member to link for "ConfigFactory.load()".
[error]   /** @see [[ConfigFactory.load()]] */
[error]   ^
[error] /home/mariop/prog/pureconfig/core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala:24: Could not find any member to link for "ConfigFactory.parseFile()".
[error]   /** @see [[ConfigFactory.parseFile()]] */
[error]   ^
[error] 6 errors found
[error] (core/compile:doc) Scaladoc generation failed
[error] Total time: 4 s, completed Apr 1, 2017 3:28:03 PM

```

melrief:
Partially fixes #200. I couldn't find a way to fix the javadoc links while for Scala API I just used sbt-api-mappings which works well and could be used for other scaladoc links. For javadocs, I think there isn't a clear solution so I just removed the links. It's not elegant but the links that I'm talking about are in `ConfigFactoryWrapper`.

Without this we cannot publish `0.7.0` because the doc generation doesn't work.


project/plugins.sbt
```diff
@@ -4,3 +4,4 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.7.0")
 addSbtPlugin("org.tpolecat" % "tut-plugin" % "0.4.8")
 addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.1.0")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.0")
+addSbtPlugin("com.thoughtworks.sbt-api-mappings" % "sbt-api-mappings" % "latest.release")
```
derekmorr: I'm not sure we need this plugin. If we set `autoAPIMappings := true` in `build.sbt` in the `settings` val, it has the same effect, doesn't it?

derekmorr:
This looks good to me.

It's certainly no reason to delay 0.7.0, but for the future, there should be a way to get the links to TypeSafe Config working by adjusting `build.sbt` - this has a work-around (I haven't tried it, and it looks involved) - https://stackoverflow.com/questions/19786841/can-i-use-sbts-apimappings-setting-for-managed-dependencies/35673212#35673212

Alternatively, we could submit a patch to the Thoughtworks plugin to add TypeSafeConfig Javadoc URL.

derekmorr (approved):

derekmorr:
Having said all of that, I did leave a comment that we can remove the thoughtworks plugin and just use `autoapimappings` to fix the issue too.

project/plugins.sbt
ruippeixotog: I think you still need to point each library to their public scaladoc URL. In some projects I did something like this in order to make my scaladoc point to akka docs:

```scala
autoAPIMappings := true
apiMappings := findManagedDependency("com.typesafe.akka", "akka-actor").value.map {
  case (module, jar) => jar -> url(s"http://doc.akka.io/api/akka/${module.revision}/")
}.toMap
```

The plugin added by @melrief seems to do just that for a handful of common libraries.

ruippeixotog (approved):
I also believe we can achieve what the plugins does just by adding some settings to `build.sbt`, with the added advantage that by doing that we have better control of what libraries we support.

That said, at this moment complete and "linked" scaladoc has minimal priority - we aren't even publishing them to anywhere... For me, it isn't a blocker for 0.7.0 in any way.

project/plugins.sbt
derekmorr: I removed the thoughtworks plugin, added `autoAPIMappings`, and that fixed the errors about the scala std lib classes. I didn't have to add anything else. We would have do this for the typesafe config libraries.

melrief:
Removed the plugin because we don't need it

PR #207, opened 2017-04-21
Add combinators to ConfigReader, ConfigWriter and ConfigConvert

### LINKED ISSUE #93: Make Converters composable

ruippeixotog:
Adds some combinators to config reading/writing classes so they are easier to compose.

I ended up not needing anything from cats; my previous concerns with F-bounded polymorphism didn't apply to this specific situation. Now that we have these methods defined we should probably create a `pureconfig-cats` module (and a `pureconfig-scalaz`, if anybody asks for it) with type classes for our three classes.

Closes #93.

### DIFF
--- a/core/src/test/scala/pureconfig/ConfigReaderSuite.scala
+++ b/core/src/test/scala/pureconfig/ConfigReaderSuite.scala
@@ -2,7 +2,7 @@ package pureconfig
 
 import com.typesafe.config.{ ConfigFactory, ConfigValue, ConfigValueFactory }
 import org.scalacheck.{ Arbitrary, Gen }
-import pureconfig.error.ConfigReaderFailures
+import pureconfig.error.{ CannotConvertNull, ConfigReaderFailures }
 
 class ConfigReaderSuite extends BaseSuite {
   val intReader = ConfigReader[Int]
@@ -17,12 +17,20 @@ class ConfigReaderSuite extends BaseSuite {
     Gen.frequency(95 -> Gen.chooseNum(Int.MinValue, Int.MaxValue), 5 -> Gen.alphaStr)
       .map(ConfigValueFactory.fromAnyRef)
 
+  val genReaderFailure: Gen[ConfigReaderFailures] =
+    Gen.const(ConfigReaderFailures(CannotConvertNull))
+
   implicit val arbConfig = Arbitrary(genConfig)
+  implicit val arbReaderFailure = Arbitrary(genReaderFailure)
 
   behavior of "ConfigReader"
 
   it should "have a correct map method" in forAll { (conf: ConfigValue, f: Int => String) =>
     intReader.map(f).from(conf) === intReader.from(conf).right.map(f)
+  }
+
+  it should "have a correct emap method" in forAll { (conf: ConfigValue, f: Int => Either[ConfigReaderFailures, String]) =>
+    intReader.emap(f).from(conf) === intReader.from(conf).right.flatMap(f)
   }
 
   it should "have a correct flatMap method" in forAll { conf: ConfigValue =>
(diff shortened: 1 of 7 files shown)


core/src/test/scala/pureconfig/ConfigReaderSuite.scala
```diff
@@ -0,0 +1,70 @@
...
+    val g = { n: Int => intSummedReader(n) }
+    intReader.flatMap(g).from(conf) === intReader.from(conf).right.flatMap(g(_).from(conf))
+  }
+
+  it should "have a correct zip method" in forAll { conf: ConfigValue =>
+    def zip[A, B](r1: ConfigReader[A], r2: ConfigReader[B]): Either[ConfigReaderFailures, (A, B)] = {
+      (r1.from(conf), r2.from(conf)) match {
+        case (Right(a), Right(b)) => Right((a, b))
+        case (Left(fa), _) => Left(fa)
+        case (_, Left(fb)) => Left(fb)
```
ruippeixotog: Right now, `zip` short-circuits in case the first reader fails. We could also use always both readers and combine their failures if both fail. What do you prefer?

core/src/test/scala/pureconfig/ConfigReaderSuite.scala
```diff
@@ -0,0 +1,70 @@
...
+  it should "have a correct flatMap method" in forAll { conf: ConfigValue =>
+    val g = { n: Int => intSummedReader(n) }
+    intReader.flatMap(g).from(conf) === intReader.from(conf).right.flatMap(g(_).from(conf))
+  }
+
+  it should "have a correct zip method" in forAll { conf: ConfigValue =>
+    def zip[A, B](r1: ConfigReader[A], r2: ConfigReader[B]): Either[ConfigReaderFailures, (A, B)] = {
+      (r1.from(conf), r2.from(conf)) match {
+        case (Right(a), Right(b)) => Right((a, b))
+        case (Left(fa), _) => Left(fa)
```
ruippeixotog: Right now, `zip` short-circuits in case the first reader fails. We could also use always both readers and combine their failures if both fail. What do you prefer?

melrief: I prefer to run boths and combine their failures as we do for products in general. The only advantage of short-circuiting is that it runs faster but for us speed is not an issue.

melrief (approved):
Great PR, thanks!

I have two general comments:
1. you have added the `fromFunction` functions to `ConfigReader` and `ConfigWriter` to create new instances from a provided function, I would consider adding a `fromFunctions` to `ConfigConvert` to do the same.
2. I've noticed that you didn't add `ConfigReader.emap` to map over the underlying result type. I propose to add it as simplified version of `flatMap`, something like:

```scala
def emap[B](f: A => Either[ConfigReaderFailures, B]): ConfigReader[B] =
   fromFunction[B] { cv: ConfigValue => from(cv).right.flatMap(f) }
```

Both points are optional and I approve the PR.

melrief:
Thanks for `emap`, LGTM

ruippeixotog:
@melrief I have just changed the behavior of `zip` in 02e5bda, I agree that it's better to show errors from both sides.

I also did some minor changes in the specs in 29ecd77. It seems that `===` doesn't behave in the way I thought (I'm a specs2 user), so I replaced it with `shouldEqual`. I also increased the number of scalacheck iterations to 100 in those files, as I think 10 is not enough for most purposes.

jcazevedo (approved):
LGTM!

Thanks for this and also for changing the name of the type parameters from the Java style to the Scala style. We should strive to replace the remaining usages of the Java style and to make the usage of the Scala style consistent in future PRs.

PR #209, opened 2017-04-21
Provide an EnumCoproductHint for sealed traits of case objects

### LINKED ISSUE #208: Provide a CoproductHint to serialize sealed families of case objects
One very common use case I have with pureconfig is to serialize enums (encoded as sealed families of case objects) as simple strings with the names of the options. I think that it is a case common enough for us to provide a `CoproductHint`.

ruippeixotog:
Closes #208.

### DIFF
--- a/core/src/main/scala/pureconfig/CoproductHint.scala
+++ b/core/src/main/scala/pureconfig/CoproductHint.scala
@@ -59,7 +59,7 @@ trait CoproductHint[T] {
 class FieldCoproductHint[T](key: String) extends CoproductHint[T] {
 
   /**
-   * Returns the field name for a class or coproduct option name.
+   * Returns the field value for a class or coproduct option name.
    *
    * @param name the name of the class or coproduct option
    * @return the field value associated with the given class or coproduct option name.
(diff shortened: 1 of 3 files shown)


core/src/main/scala/pureconfig/CoproductHint.scala
```diff
@@ -92,6 +92,37 @@ class FieldCoproductHint[T](key: String) extends CoproductHint[T] {
...
+ */
+class EnumCoproductHint[T] extends CoproductHint[T] {
+
+  /**
+   * Returns the field value for a class or coproduct option name.
+   *
+   * @param name the name of the class or coproduct option
+   * @return the field value associated with the given class or coproduct option name.
+   */
+  protected def fieldValue(name: String): String = name.toLowerCase
```
melrief, line 108: The reason why this is protected is to give devs a hint at the fact that they can subclass this class and change how the `fieldValue` is calculated from a `name`?

core/src/main/scala/pureconfig/CoproductHint.scala
```diff
@@ -92,6 +92,37 @@ class FieldCoproductHint[T](key: String) extends CoproductHint[T] {
 }
 
 /**
+ * Hint applicable to sealed families of case objects where objects are written and read as strings. Trying to read or
```
melrief: I would add that the `string` used is the object name, e.g. `object Foo` will be read as `foo` with this. It's not super clear how objects are written and read as strings.

core/src/test/scala/pureconfig/CoproductHintSuite.scala
```diff
@@ -12,37 +12,68 @@ class CoproductHintSuite extends BaseSuite {
...
+
+    implicit val hint = new EnumCoproductHint[Color] {
+      override def fieldValue(name: String) = ConfigFieldMapping(CamelCase, KebabCase)(name)
+    }
+
+    implicit val badHint = new EnumCoproductHint[AnimalConfig]
+
+    it should "read values as expected when using an EnumCoproductHint" in {
+      val conf = ConfigValueFactory.fromAnyRef("rainy-blue")
+      ConfigConvert[Color].from(conf) shouldEqual Right(RainyBlue)
```
melrief, line 54: Nice!

melrief (approved):
LGTM, good to have this!

core/src/main/scala/pureconfig/CoproductHint.scala
ruippeixotog: Yes, it follows the same structure as `FieldCoproductHint`, where the translation between class names and config values is customizable.

Just as we did with `ProductHint` (following [the discussion here](https://github.com/pureconfig/pureconfig/pull/90#discussion_r97449022)), we could provide a constructor for both of these hints that receives a `ConfigFieldMapping`. I'd rather keep a consistent interface between hints here and leave a possible refactoring or new constructor for another PR.

core/src/main/scala/pureconfig/CoproductHint.scala
ruippeixotog: I agree, I clarified it in 1fab2b3.

melrief:
LGTM, waiting for a second review

melrief (approved):

melrief:
coveralls seems unhappy with the test coverage. It's not a blocker for the PR but maybe we can make a few more tests to cover the new code?

ruippeixotog:
If you see the list of changes in coverage, most of them are not related to this PR (or at least I cannot see what might have affected them), which is a bit strange. Maybe some of our tests are non-deterministic?

Regarding the lines related to this PR that are uncovered, so you want me to add tests for things like the description of the `ConfigReaderFailure`? I can test the exception message with a regex, if you think it's worth it.

melrief:
No @ruippeixotog, it's not worth it. I'm trying to understand what's going on.

jcazevedo (approved):

PR #212, opened 2017-04-27
Allow `Duration` to be `Undefined`.

### LINKED ISSUE #184: DurationConvert encodes Duration.Undefined as Duration.Inf
Not sure why this is happening but this test fails

```scala
it should "fromS(fromD(Duration.Undefined)) == Duration.Undefined" in {
  fromS(fromD(Duration.Undefined)) shouldEqual Right(Duration.Undefined)
}
```

the error is

```scala
[info] - should fromS(fromD(Duration.Undefined)) == Duration.Undefined *** FAILED ***
[info]   Right(Duration.Inf) did not equal Right(Duration.Undefined) (DurationConvertTest.scala:86)
```

leifwickland:
Closes #184.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 - New features
   - `ConfigReader`, `ConfigWriter` and `ConfigConvert` now have combinators such as `map`, `flatMap` and `contramap`,
     making them easier to compose.
+- Bug fixes
+  - `Duration.Undefined` is correctly handled when reading an writing configurations. [[#184](https://github.com/melrief/pureconfig/issues/184)].
 
 ### 0.7.0 (Apr 2, 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
   - `ConfigReader`, `ConfigWriter` and `ConfigConvert` now have combinators such as `map`, `flatMap` and `contramap`,
     making them easier to compose.
 - Bug fixes
-  - `Duration.Undefined` is correctly handled when reading an writing configurations. [[#184](https://github.com/melrief/pureconfig/issues/184)].
+  - `Duration.Undefined` is correctly handled when reading and writing configurations. [[#184](https://github.com/melrief/pureconfig/issues/184)]
 
 ### 0.7.0 (Apr 2, 2017)
 
--- a/core/src/main/scala/pureconfig/DurationConvert.scala
+++ b/core/src/main/scala/pureconfig/DurationConvert.scala
@@ -15,7 +15,7 @@ private[pureconfig] object DurationConvert {
    * Convert a string to a Duration while trying to maintain compatibility with Typesafe's abbreviations.
    */
   val fromString: String => Option[ConfigValueLocation] => Either[ConfigReaderFailure, Duration] = { string => location =>
-    if (string == Duration.Undefined.toString) Right(Duration.Undefined)
+    if (string == UndefinedDuration) Right(Duration.Undefined)
     else try {
       Right(Duration(addZeroUnit(justAMinute(itsGreekToMe(string)))))
     } catch {
@@ -44,12 +44,16 @@ private[pureconfig] object DurationConvert {
   def fromDuration(d: Duration): String = {
     d match {
       case f: FiniteDuration => fromFiniteDuration(f)
-      // We must do an `eq` instead of `==` comparison because Undefined is intentionally != itself.
-      case i: Duration.Infinite if i eq Duration.Undefined => Duration.Undefined.toString
-      case i: Duration.Infinite if i == Duration.MinusInf => "MinusInf"
-      case i: Duration.Infinite => "Inf"
+      case Duration.Inf => "Inf"
+      case Duration.MinusInf => "MinusInf"
+      // We must do an `eq` instead of `==` comparison because `Undefined` is intentionally != itself.
+      case d if d eq Duration.Undefined => UndefinedDuration
     }
   }
+
+  /// We need our own constant for `Duration.Undefined` because that value's `toString` is `Duration.Undefined`
+  /// which is inconsistent with the `Inf` and `Minus` `toString` provided by other special `Duration`s.
+  private final val UndefinedDuration = "Undefined"
 
   /**
    * Format a FiniteDuration as a string with a suitable time unit using units TypesafeConfig understands.
--- a/core/src/main/scala/pureconfig/DurationConvert.scala
+++ b/core/src/main/scala/pureconfig/DurationConvert.scala
@@ -15,7 +15,8 @@ private[pureconfig] object DurationConvert {
    * Convert a string to a Duration while trying to maintain compatibility with Typesafe's abbreviations.
    */
   val fromString: String => Option[ConfigValueLocation] => Either[ConfigReaderFailure, Duration] = { string => location =>
-    try {
+    if (string == Duration.Undefined.toString) Right(Duration.Undefined)
+    else try {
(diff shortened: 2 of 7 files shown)


derekmorr:
This LGTM, but I wonder if we should add a note to docs about the duration string syntax.

melrief (approved):
Good job!

core/src/main/scala/pureconfig/DurationConvert.scala
```diff
@@ -42,9 +43,11 @@ private[pureconfig] object DurationConvert {
    */
   def fromDuration(d: Duration): String = {
     d match {
+      case f: FiniteDuration => fromFiniteDuration(f)
+      // We must do an `eq` instead of `==` comparison because Undefined is intentionally != itself.
+      case i: Duration.Infinite if i eq Duration.Undefined => Duration.Undefined.toString
       case i: Duration.Infinite if i == Duration.MinusInf => "MinusInf"
       case i: Duration.Infinite => "Inf"
```
jcazevedo: I think we should be consistent here: either always use the `toString` return value, or encode `Duration.Undefined` as `"Undefined"`.

leifwickland:
@derekmorr I agree that it'd be nice to document, but I'm not sure where to document that. Do you have a suggestion? The best idea I've got is simply to add another page that details all the quirks of our duration converter.

core/src/main/scala/pureconfig/DurationConvert.scala
leifwickland: @jcazevedo I don't understand where you see an inconsistency. Are you saying that that the `if eq Duration.Undefined` is inconsistent because it's not `i.toString == Duration.Undefined.toString`?
I agree that the string `"Duration.Undefined"` in a configuration is ugly, but it seems somewhat less likely to be misleading than `undefined` which has the potential to present problems in the future if we ended up with a JSON backend for pureconfig.

jcazevedo: I was talking about the way we write those durations. We are currently writing `Duration.Undefined` as `"Duration.Undefined"`, `Duration.MinusInf` as `"MinusInf"` and `Duration.Inf` as `"Inf"`. I think it's preferable to go with either `"Undefined"`, `"MinusInf"` and `"Inf"`; or `"Duration.Undefined"`, `"Duration.MinusInf"` and `"Duration.Inf"`.

core/src/main/scala/pureconfig/DurationConvert.scala
```diff
@@ -42,9 +43,11 @@ private[pureconfig] object DurationConvert {
    */
   def fromDuration(d: Duration): String = {
     d match {
+      case f: FiniteDuration => fromFiniteDuration(f)
+      // We must do an `eq` instead of `==` comparison because Undefined is intentionally != itself.
+      case i: Duration.Infinite if i eq Duration.Undefined => Duration.Undefined.toString
```
ruippeixotog: Minor suggestion: the `Duration.Infinite` explicit typing in this whole `match` is redundant - you can just write `case d if d eq Duration.Undefined`, `case Duration.MinusInf` and `case Duration.Inf`. With that, you cover all three subclasses of the `Duration.Infinite` sealed trait.

core/src/main/scala/pureconfig/DurationConvert.scala
ruippeixotog: I also believe that we should encode `Duration.MinusInf`, `Duration.Inf` and `Duration.Undefined` in configs using the same convention. I have a preference for `MinusInf`, `Inf` and `Undefined`, but I'd be ok with encoding them with the leading `Duration.`.

core/src/test/scala/pureconfig/DurationConvertSuite.scala
```diff
@@ -82,4 +82,7 @@ class DurationConvertSuite extends FlatSpec with Matchers with EitherValues {
         ex.head shouldBe a[CannotConvert]
     }
   }
+  it should "fromS(fromD(Duration.Undefined)) == Duration.Undefined" in {
```
ruippeixotog: Minor nitpick: can you put the spec title in english, just for consistency with the other specs in pureconfig? Some suggestions are `it should "handle correctly Duration.Undefined values"` or `it should convert correctly Duration.Undefined to and from a config`.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 - New features
   - `ConfigReader`, `ConfigWriter` and `ConfigConvert` now have combinators such as `map`, `flatMap` and `contramap`,
     making them easier to compose.
+- Bug fixes
+  - `Duration.Undefined` is correctly handled when reading an writing configurations. [[#184](https://github.com/melrief/pureconfig/issues/184)].
```
ruippeixotog: I believe there is an extra `.` here :)

ruippeixotog (commented):
LGTM overall, thank you :) I added some comments with minor issues and suggestions.

Values like `Duration.Undefined` are always strange to deal with, unfortunately... The solution you implemented seems the best one, though.

core/src/main/scala/pureconfig/DurationConvert.scala
leifwickland: @ruippeixotog Excellent point. Thank you. Done.
I had been wanted to do something better, but wasn't seeing the solution.

core/src/main/scala/pureconfig/DurationConvert.scala
leifwickland: It sounds like folks would prefer to standardize the naming of the special values. I'll remove the `"Duration."` prefix, using a local constant of `"Undefined"` for the salient value.

CHANGELOG.md
leifwickland: Yup. And I misspelled "and." Thanks.

core/src/test/scala/pureconfig/DurationConvertSuite.scala
leifwickland: Oh, sure. I agree that's better I was just copying the test that was provided in #184.

leifwickland:
@derekmorr In response to your comment, I added some docs for Duration. I'm not saying they're good, but they're a start. Please let me know what you think.

leifwickland:
@jcazevedo @ruippeixotog I think I addressed the changes you requested. I'll assume you're happy unless I hear otherwise.

leifwickland:
@melrief Unless somebody objects the something here by the end of Monday, please go ahead and merge then.

docs/src/main/tut/support-for-duration.md
```diff
@@ -0,0 +1,25 @@
...
+- `us`, `µs`, `micro`,  `micros`, `microsecond` or `microseconds`
+- `ns`, `nano`, `nanos`, `nanosecond`, or `nanoseconds`
+
+All durations must have a unit except for `0` which may be specified as a bare scalar.
+
+### Duration
+
+For [`Duration`](http://scala-lang.org/api/current/scala/concurrent/duration/Duration.html) PureConfig supports the features mentioned above plus the following for the various flavors of [`Infinite`](http://scala-lang.org/api/current/scala/concurrent/duration/Duration$$Infinite.html) duration.
+
+- `Inf`, `PlusInf`, or "+Inf"`
```
ruippeixotog: "+Inf" and "-Inf" are not being formatted correctly here.

ruippeixotog (approved):
Thank you for the changes :) The documentation you added looks very good, it explains very well what should be expected of pureconfig regarding `Duration`.

docs/src/main/tut/support-for-duration.md
leifwickland: @ruippeixotog Thanks for noticing! Fixed.

PR #213, opened 2017-04-30
Fix link to "configurable converters"

gseitz:
Just a minor change that fixes the link to "configurable converters" to point to the markdown file (instead of being an anchor link).


derekmorr:
Thanks for the fix. The main README.MD is generated from the version in `core/src/main/tut`. Could you update the PR to put the fix in that file as well? Other than that, this looks good.

gseitz:
I changed the link in `core/src/main/tut` now as well.

derekmorr (approved):

derekmorr:
Thanks for the fix. I've merged it. We've been updating and restructuring the docs over the past few months, and we must have missed this. Thanks for catching it.

melrief:
Thanks for the fix, @gseitz!

PR #214, opened 2017-05-06
Avoid shapeless conflicts in spark doc

rberenguel:
A very trivial doc to explain how to use shading with `sbt-assembly` (since I guess it's happening to more people and seems hard to Google for). I tried to keep the overall style in the section titles, but I'm not sure though that this warrants a different file, maybe it could be under some of the other headings.


melrief:
Thanks @rberenguel for the PR! PureConfig uses [`tut`](https://github.com/tpolecat/tut) to generate the documentation so that we can validate it. Canyou move your new section inside the [documentation source](https://github.com/pureconfig/pureconfig/tree/master/docs/src/main/tut) and then build the documentation?

@derekmorr do you have time to have a look at this too?

README.md
```diff
@@ -29,6 +29,7 @@ A boilerplate-free Scala library for loading configuration files.
 - [Handling missing keys](docs/handling-missing-keys.md)
 - [Example](docs/example.md)
 - [Whence the config files](docs/whence-the-config-files.md)
+- [Playing nice with Spark's Shapeless](docs/playing-nice-with-sparks-shapeless.md)
```
derekmorr: This should also be added to `core/src/main/tut/README.md` or it will be overwritten when the docs are next generated.

derekmorr (commented):
Since these changes are self-contained to the main README, I think they just need to go in `core/src/main/tut/README.md`. Tut can't check the SBT DSL, so there's no need to use the `tut` annotation on the examples.

README.md
rberenguel: @derekmorr I think you were commenting while I was committing :), but I'm not sure (I moved this to a FAQ section in the main `README.md`)

rberenguel: Oh, sorry, missed the `core` in your comment. Fixing the real "main" `README.md`

melrief (approved):

derekmorr (approved):

PR #215, opened 2017-05-07
Make Travis check if generated docs didn't change after running tut

ruippeixotog:
I don't remember why we didn't this before, but the output of `sbt tut` seems deterministic and consistent with the current README in my laptop. We can make Travis check if this still holds after each PR so that people don't forget to update documentation and write the updates in tut sources.

EDIT: I took the opportunity to make Travis cache library dependencies and SBT boot files, since our builds are taking 15-20 minutes to run :/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,10 @@ jdk:
 script:
   - sbt ++$TRAVIS_SCALA_VERSION coverage test tut
   - (cd example ; sbt ++$TRAVIS_SCALA_VERSION test)
+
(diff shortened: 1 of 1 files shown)


derekmorr (approved):

derekmorr:
LGTM. The Scala REPL output isn't deterministic from version to version so we should only do this for one version (which I see you did).

melrief (approved):

melrief:
LGTM and merge

PR #216, opened 2017-05-07
Value classes config reader writer

### LINKED ISSUE #206: ConfigReaders and ConfigWriters for value classes
In the Circe issues pjan posted a simple `Decoder`/`Encoder` with the issue [469](https://github.com/circe/circe/issues/469) for value classes which I think we should have in pureconfig. The solution is simple: it uses an evidence that the class to read/write `T` is a subclass of `AnyVal`, a reader for another class `U` and a [`Unwrapped.Aux`](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/unwrapped.scala) from `T` to `U`:

```scala
implicit def decodeAnyVal[T, U](
      implicit
      ev: T <:< AnyVal,
      unwrapped: Unwrapped.Aux[T, U],
      reader: ConfigReader[U]): ConfigReader[T] =
      new ConfigReader[T] {
        def from(value: ConfigValue): Either[ConfigReaderFailures, T] =
           reader.from(value).right.map(unwrapped.wrap)
      }
```

Ant it works:

```scala
@ final case class FooInt(value: Int) extends AnyVal 
defined class FooInt
@ ConfigReader[FooInt].from(ConfigValueFactory.fromAnyRef(1) ) 
res7: Either[ConfigReaderFailures, FooInt] = Right(FooInt(1))
```

melrief:
Closes #206 

I've added readers and writers for value classes that "skip" the wrappers when they read or write. In this way, if you have a `final case class Port(value: Int) extends AnyVal`, you can read and write it as `Int` instead of product with one field of type `Int`.

This change has some consequences on the architecture of the traits `DerivedReaders` and `DerivedWriters` because just putting the decoders derivation for value classes inside the previous traits was causing conflicts between them and the product decoders. The solution that I've adopted is to split the traits in two and make the trait containing the new decoders extend the old decoders trait. In this way, the new decoders have higher priority during the implicit resolution.

A side-change that I've done is to rename `check` and `checkString` utility functions for testing into `checkRead` and `checkReadString` and change their implicit parameter to `ConfigReader`, because those two methods just test the reading of a type `T`. I also added `checkWrite` for the inverse test.

I haven't added a section for value classes but I mention them in the README and in the CHANGELOG. I'm not sure that what I've written is good enough, please have a look at that too and tell me what you think.

### DIFF
--- a/core/src/main/scala/pureconfig/DerivedReaders.scala
+++ b/core/src/main/scala/pureconfig/DerivedReaders.scala
@@ -20,9 +20,26 @@ import shapeless.labelled._
 trait AllowMissingKey { self: ConfigReader[_] => }
 
 /**
+ * Trait extending [[DerivedReaders1]] that contains `ConfigReader` instances for `AnyVal`.
+ *
+ * This trait exists to give priority to the `AnyVal` derivation over the generic product derivation.
+ */
+trait DerivedReaders extends DerivedReaders1 {
+  implicit def deriveAnyVal[T, U](
+    implicit
+    ev: T <:< AnyVal,
+    unwrapped: Unwrapped.Aux[T, U],
+    reader: ConfigReader[U]): ConfigReader[T] =
+    new ConfigReader[T] {
+      def from(value: ConfigValue): Either[ConfigReaderFailures, T] =
+        reader.from(value).right.map(unwrapped.wrap)
+    }
+}
+
+/**
  * Trait containing `ConfigReader` instances for collection, product and coproduct types.
  */
-trait DerivedReaders {
+trait DerivedReaders1 {
 
   private[pureconfig] trait WrappedConfigReader[Wrapped, SubRepr] extends ConfigReader[SubRepr]
 
@@ -198,6 +215,7 @@ trait DerivedReaders {
       cc.value.from(config).right.map(gen.from)
     }
   }
+
 }
 
 object DerivedReaders extends DerivedReaders
--- a/core/src/test/scala/pureconfig/ConfigConvertChecks.scala
+++ b/core/src/test/scala/pureconfig/ConfigConvertChecks.scala
@@ -20,7 +20,7 @@ trait ConfigConvertChecks { this: FlatSpec with Matchers with GeneratorDrivenPro
    * Note that this method doesn't check all the values but only the values that can be created by `Arbitrary[T]` and
    * only the `ConfigValue` created by `ConfigConvert[T].to`. While `Arbitrary[T]` is usually comprehensive,
    * `ConfigConvert[T].from` could support different kind of values that `ConfigConvert[T].to` doesn't produce
-   * because, for instance, multiple representation of `t: T` are possible. Use [[check()]] for those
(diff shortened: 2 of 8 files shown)


core/src/test/scala/pureconfig/ConfigConvertChecks.scala
```diff
@@ -55,28 +55,40 @@ trait ConfigConvertChecks { this: FlatSpec with Matchers with GeneratorDrivenPro
     }
 
   /**
-   * For each pair of value of type `T` and `ConfigValue`, check that `ConfigConvert[T].from`
+   * For each pair of value of type `T` and `ConfigValue`, check that `ConfigReader[T].from`
    * successfully converts the latter into to former. Useful to test specific values
    */
-  def check[T](valuesToReprs: (T, ConfigValue)*)(implicit cc: ConfigConvert[T], tag: ClassTag[T]): Unit =
+  def checkRead[T](valuesToReprs: (T, ConfigValue)*)(implicit cr: ConfigReader[T], tag: ClassTag[T]): Unit =
```
ruippeixotog, line 61: Nice change!

core/src/main/scala/pureconfig/DerivedReaders.scala
```diff
@@ -20,9 +20,26 @@ import shapeless.labelled._
 trait AllowMissingKey { self: ConfigReader[_] => }
 
 /**
+ * Trait extending [[DerivedReaders1]] that contains `ConfigReader` instances for `AnyVal`.
+ *
+ * This trait exists to give priority to the `AnyVal` derivation over the generic product derivation.
+ */
+trait DerivedReaders extends DerivedReaders1 {
```
ruippeixotog, line 27: You can also remove the ambiguity by requiring a `F <:!< AnyVal` evidence in `deriveProductInstance`. Have you considered it? I am not sure of which way I prefer.

ruippeixotog (dismissed):

core/src/main/scala/pureconfig/DerivedReaders.scala
melrief: I'm actually not sure what are the implications of using one method instead of the other. I know that a lot of projects working with typeclasses use my approach to disambiguate and I went for it. I don't see an issue except that now we have two traits.

ruippeixotog (approved):
LGTM

core/src/main/scala/pureconfig/DerivedReaders.scala
jcazevedo: Is the `<:!<` constraint provided by some library we require in the core module, or a trick we would implement? If it's the latter, then I prefer the two traits solution.

jcazevedo (approved):

core/src/main/scala/pureconfig/DerivedReaders.scala
ruippeixotog: `<:!<` is provided by shapeless. However, I also ended up preferring the two traits solution - I gave the suggestion so that we know there is an alternative.

PR #217, opened 2017-05-09
Support Java duration. Naming choices are up for discussion

rberenguel:
I've seen in the Scala source that jDuration might be more standard, but seems to be “less classy”). Also, since both Durations have the same SimpleName had to change the namer in the test suite, not sure if this is the best choice. I removed the limitation on the generated duration, though, java.time.Duration seems to handle that specific number just fine

### DIFF
--- a/core/src/test/scala/pureconfig/ConfigConvertChecks.scala
+++ b/core/src/test/scala/pureconfig/ConfigConvertChecks.scala
@@ -24,7 +24,7 @@ trait ConfigConvertChecks { this: FlatSpec with Matchers with GeneratorDrivenPro
    * representations.
    */
   def checkArbitrary[T](implicit cc: ConfigConvert[T], arb: Arbitrary[T], tag: ClassTag[T]): Unit =
-    it should s"read an arbitrary ${tag.runtimeClass.getSimpleName}" in forAll {
+    it should s"read an arbitrary ${tag.runtimeClass.getCanonicalName}" in forAll {
       (t: T) =>
         val result = cc.from(cc.to(t))
         result shouldBe a[Right[_, _]]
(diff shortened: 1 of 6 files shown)


rberenguel:
Why is coveralls taking ages to report back?

core/src/test/scala/pureconfig/ConfigConvertChecks.scala
```diff
@@ -78,7 +78,7 @@ trait ConfigConvertChecks { this: FlatSpec with Matchers with GeneratorDrivenPro
    */
   def checkFailure[T, E <: ConfigReaderFailure](values: ConfigValue*)(implicit cr: ConfigConvert[T], tag: ClassTag[T], eTag: ClassTag[E]): Unit =
     for (value <- values) {
-      it should s"fail when it tries to read a value of type ${tag.runtimeClass.getSimpleName} " +
+      it should s"fail when it tries to read a value of type ${tag.runtimeClass.getCanonicalName} " +
```
melrief, line 81: I'm ok with this change if everybody else is ok

melrief (approved):
Good job @rberenguel! The code looks follows very well the repository standard and the tests are well done. Thank you very much. I'll wait for another review and then merge.

derekmorr (approved):

core/src/test/scala/pureconfig/ConfigConvertChecks.scala
ruippeixotog: I'm ok with it, in this case I suppose it helps to distinguish a Scala Duration from a Java one.

ruippeixotog (approved):
LGTM, nice job!

rberenguel:
Thanks @ruippeixotog @derekmorr @melrief :)

ruippeixotog:
@rberenguel, there seems to be one problem with this PR, unfortunately :( In my computer I'm having some non-deterministic errors in the conversion of `java.time.Duration`:

```
[info] - should read an arbitrary java.time.Duration *** FAILED ***
[info]   TestFailedException was thrown during property evaluation.
[info]     Message: PT-1525073H-41M0.327123929S did not equal PT-1525073H-41M-0.327123929S
[info]     Location: (ConfigConvertChecks.scala:31)
[info]       arg0 = PT-1525073H-41M-0.327123929S
[info]     )
```

```
[info] - should read an arbitrary java.time.Duration *** FAILED ***
[info]   TestFailedException was thrown during property evaluation.
[info]     Message: PT-256047H-2M0.448787857S did not equal PT-256047H-2M-0.448787857S
[info]     Location: (ConfigConvertChecks.scala:31)
[info]       arg0 = PT-256047H-2M-0.448787857S
[info]     )
```

Do you know what might that be?

ruippeixotog:
Just happened in [Travis, too](https://travis-ci.org/pureconfig/pureconfig/jobs/230512293).

rberenguel:
@ruippeixotog yikes! I ran it three times, probabilistic testing ftw :( I'll check right after dinner, curious to see what the issue might be

In all three cases the smallest sign is reversed, smells of overflow...

PR #219, opened 2017-05-09
Test support for Duration.Undefined

ruippeixotog:
Since #184 is already closed, we can enable this again.


rberenguel:
What was the problem with `Duration.Undefined`?

ruippeixotog:
Simply put, it wasn't supported by the underlying Typesafe config `Duration` parser and it has `NaN`-like semantics, leading us to overlook some implementation details. You can read more at #184 :)

rberenguel (approved):
Not sure my approval gives much, but given that `Duration.Undefined` is supported, testing for it makes total sense

jcazevedo (approved):

PR #221, opened 2017-05-10
Glad to be able to help!

rberenguel:


rberenguel:
Damn!!
```
[info] - should read an arbitrary java.time.Duration *** FAILED ***
[info]   TestFailedException was thrown during property evaluation.
[info]     Message: PT-1260257119798114H-28M0.249993099S did not equal PT-1260257119798114H-28M-0.249993099S
[info]     Location: (ConfigConvertChecks.scala:31)
[info]       arg0 = PT-1260257119798114H-28M-0.249993099S
[info]     )
```

rberenguel:
I can't figure out how ScalaTest figures out equality, but the Java documentation about Duration says:

```
A program may produce unpredictable results if it attempts to distinguish two references to equal values of a value-based class, whether directly via reference equality or indirectly via an appeal to synchronization, identity hashing, serialization, or any other identity-sensitive mechanism. Use of such identity-sensitive operations on instances of value-based classes may have unpredictable effects and should be avoided.
```

Might be related? A hint in this direction:

```scala
@ val d1 = Duration.parse("PT-1260257119798114H-28M-0.249993099S")
d1: Duration = PT-1260257119798114H-28M0.249993099S
@ val d2 = Duration.parse("PT-1260257119798114H-28M0.249993099S")
d2: Duration = PT-1260257119798114H-28M0.249993099S
@ d1.equals(d2)
res3: Boolean = true
@ d2.equals(d1)
res4: Boolean = true
```

melrief (approved):

ruippeixotog:
One thing regarding scalacheck tests: I don't know if this is the default for scalatest with scalacheck, but the specs are only running 10 iterations per property. This is way too low, and may surely help getting non-deterministic tests. In some tests I wrote previously I [overrode that number](https://github.com/pureconfig/pureconfig/blob/master/core/src/test/scala/pureconfig/ConfigReaderSuite.scala#L8), but I think it would be better if we increased it globally.

ruippeixotog (approved):

melrief:
I think the issue is [JDK-8054978](https://bugs.openjdk.java.net/browse/JDK-8054978). The solution is to exclude from the generator values that are affected by this version.

melrief:
Welcome among the contributors @rberenguel!

PR #222, opened 2017-05-11
fix Gen[java.time.Duration] and increase num of tests

melrief:


rberenguel (approved):
That's some digging into the issue!

ruippeixotog (approved):
LGTM!

jcazevedo (approved):

melrief:
The issue with scala Duration is still there. I will fix it tomorrow in the same PR, I'm getting tired of this issues.

PR #223, opened 2017-05-14
first version of cats module; supports NonEmptyList and NonEmptyVector

### LINKED ISSUE #101: Supports for cats data structures
cats exposes some [interesting structures](https://github.com/typelevel/cats/tree/5a9717a4499265275bc5c3c6e4f2e3ac3d1e17f5/core/src/main/scala/cats/data) that can be converted from/to a configuration value. For instance, I use `NonEmptyList` often. We should create a module for pureconfig.

derekmorr:
This would fix #101.

I wonder if the `fromNonEmpty` and `nonEmptyTo` functions should be moved into `core` as other modules might want to reuse them.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,6 +10,10 @@ lazy val docs = (project in file("docs")).
   settings(settings, publishArtifact := false).
   dependsOn(core)
 
+lazy val cats = (project in file("modules/cats")).
+  settings(settings, publishArtifact := false).
+  dependsOn(core)
+
 lazy val enumeratum = (project in file("modules/enumeratum")).
   settings(settings).
   dependsOn(core)
--- a/build.sbt
+++ b/build.sbt
@@ -11,7 +11,7 @@ lazy val docs = (project in file("docs")).
   dependsOn(core)
 
 lazy val cats = (project in file("modules/cats")).
-  settings(settings, publishArtifact := false).
+  settings(settings).
   dependsOn(core)
 
 lazy val enumeratum = (project in file("modules/enumeratum")).
--- a/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
+++ b/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
@@ -217,6 +217,21 @@ final case class EmptyStringFound(typ: String, location: Option[ConfigValueLocat
 }
 
 /**
+ * A failure representing an unexpected empty traversable
+ *
+ * @param typ the type that was attempted to be converted to from an empty string
+ * @param location an optional location of the ConfigValue that raised the
+ *                 failure
+ * @param path an optional path to the value which was an unexpected empty string
+ */
+final case class EmptyTraversableFound(typ: String, location: Option[ConfigValueLocation], path: Option[String]) extends ConfigReaderFailure {
+  def description = s"Empty collection found when trying to convert to $typ."
+
+  def withImprovedContext(parentKey: String, parentLocation: Option[ConfigValueLocation]) =
+    this.copy(location = location orElse parentLocation, path = path.map(parentKey + "." + _) orElse Some(parentKey))
+}
+
+/**
  * A failure representing an unexpected non-empty object when using `EnumCoproductHint` to write a config.
  *
  * @param typ the type for which a non-empty object was attempted to be written
--- a/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
+++ b/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
@@ -61,7 +61,7 @@ object ConfigValueLocation {
  * ConfigValue. The failure contains an optional location of the ConfigValue
  * that raised the error.
  */
-sealed abstract class ConfigReaderFailure {
+abstract class ConfigReaderFailure {
   /**
    * The optional location of the ConfigReaderFailure.
    */
@@ -211,21 +211,6 @@ final case class ThrowableFailure(throwable: Throwable, location: Option[ConfigV
  */
 final case class EmptyStringFound(typ: String, location: Option[ConfigValueLocation], path: Option[String]) extends ConfigReaderFailure {
   def description = s"Empty string found when trying to convert to $typ."
-
-  def withImprovedContext(parentKey: String, parentLocation: Option[ConfigValueLocation]) =
-    this.copy(location = location orElse parentLocation, path = path.map(parentKey + "." + _) orElse Some(parentKey))
-}
-
-/**
- * A failure representing an unexpected empty traversable
- *
- * @param typ the type that was attempted to be converted to from an empty string
- * @param location an optional location of the ConfigValue that raised the
- *                 failure
- * @param path an optional path to the value which was an unexpected empty string
- */
-final case class EmptyTraversableFound(typ: String, location: Option[ConfigValueLocation], path: Option[String]) extends ConfigReaderFailure {
-  def description = s"Empty collection found when trying to convert to $typ."
 
   def withImprovedContext(parentKey: String, parentLocation: Option[ConfigValueLocation]) =
     this.copy(location = location orElse parentLocation, path = path.map(parentKey + "." + _) orElse Some(parentKey))
--- a/modules/cats/src/main/scala/pureconfig/module/package.scala
+++ b/modules/cats/src/main/scala/pureconfig/module/package.scala
@@ -8,7 +8,8 @@ import com.typesafe.config.ConfigValue
 import pureconfig.{ ConfigReader, ConfigWriter }
 import pureconfig.ConfigReader.{ fromFunction => fromFunctionReader }
 import pureconfig.ConfigWriter.{ fromFunction => fromFunctionWriter }
-import pureconfig.error.{ ConfigReaderFailures, ConfigValueLocation, EmptyTraversableFound }
+import pureconfig.error.{ ConfigReaderFailures, ConfigValueLocation }
+import pureconfig.module.Cats.EmptyTraversableFound
 
 /**
  * [[ConfigReader]] and [[ConfigWriter]] instances for cats data structures.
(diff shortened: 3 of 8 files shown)


core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -217,6 +217,21 @@ final case class EmptyStringFound(typ: String, location: Option[ConfigValueLocat
...
 
 /**
+ * A failure representing an unexpected empty traversable
+ *
+ * @param typ the type that was attempted to be converted to from an empty string
+ * @param location an optional location of the ConfigValue that raised the
+ *                 failure
+ * @param path an optional path to the value which was an unexpected empty string
+ */
+final case class EmptyTraversableFound(typ: String, location: Option[ConfigValueLocation], path: Option[String]) extends ConfigReaderFailure {
```
ruippeixotog: Right now, `ConfigReaderFailure` is `sealed`, which forces us to put all implementations in the same file. I don't believe it will be reasonable to continue adding new failures as the number of modules grow, each one with their (rightfully) specific errors. Moreover, users may want to have their own failures and right now they have to settle with one of the existing ones.

I know this one is not particularly specific to cats, but unless `fromNonEmpty` is moved to the core, I think this should be defined in `pureconfig-cats` too. Nevertheless, we should "unseal" `ConfigReaderFailure` and allow custom implementations anywhere.

modules/cats/src/main/scala/pureconfig/module/package.scala
```diff
@@ -0,0 +1,37 @@
...
+import pureconfig.ConfigReader.{ fromFunction => fromFunctionReader }
+import pureconfig.ConfigWriter.{ fromFunction => fromFunctionWriter }
+import pureconfig.error.{ ConfigReaderFailures, ConfigValueLocation, EmptyTraversableFound }
+
+/**
+ * [[ConfigReader]] and [[ConfigWriter]] instances for cats data structures.
+ */
+package object cats {
+
+  private[pureconfig] def fromNonEmpty[F[_], G[_], T](fromFT: F[T] => Option[G[T]])(configValue: ConfigValue)(implicit ct: ClassTag[F[T]], fReader: ConfigReader[F[T]]): Either[ConfigReaderFailures, G[T]] =
```
ruippeixotog, line 19: Given the very specific signature, which was written to fit `NonEmptyList.fromList` and `NonEmptyVector.fromVector`, I'd leave this here until there was a concrete need for it in the core or by an user. But I'm ok either way :)

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
```diff
@@ -0,0 +1,23 @@
...
+import cats.data.{ NonEmptyList, NonEmptyVector }
+import com.typesafe.config.ConfigFactory
+import org.scalatest._
+import pureconfig.syntax._
+
+class CatsSuite extends FlatSpec with Matchers with EitherValues {
+
+  case class Numbers(numbers: NonEmptyList[Int])
+
+  it should "be able to read a config with a NonEmptyList" in {
```
ruippeixotog, line 14: Can you add tests for checking that reading empty lists and vectors return a `EmptyTraversableFound` failure?

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
```diff
@@ -0,0 +1,23 @@
...
+    val config = ConfigFactory.parseString(s"""{ numbers: [1,2,3] }""")
+    config.to[Numbers].right.value shouldEqual Numbers(NonEmptyList(1, List(2, 3)))
```
ruippeixotog: Just a minor nitpick: can you change this to `config.to[Numbers] shouldEqual Right(Numbers(NonEmptyList(1, List(2, 3))))`? That way, if an unexpected error occurs the test will complete with a pretty message from `shouldEqual` instead of throwing an exception.

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  case class NumVec(numbers: NonEmptyVector[Int])
+
+  it should "be able to read a config with an NonEmptyVector" in {
+    val config = ConfigFactory.parseString(s"""{ numbers: [1,2,3] }""")
+    config.to[NumVec].right.value shouldEqual NumVec(NonEmptyVector(1, Vector(2, 3)))
```
ruippeixotog: Same thing here.

modules/cats/src/main/tut/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+In addition to [core pureconfig](https://github.com/melrief/pureconfig), you'll need:
+
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-cats" % "0.7.1"
+```
+
+## Example
+
+To load a `NonEmptyList[Int]` into a configuration, we need a class to hold our configuration:
```
ruippeixotog, line 15: I'm worried that this may give users the impression that they always have to wrap `NonEmptyList` into a class, even if the list is already part of a bigger hierarchy. The only reason why `parseString("[1,2,3]")`doesn't work is because a `Config` (returned by `parseString`) can only hold a config object. Maybe we can explain this here?

build.sbt
```diff
@@ -10,6 +10,10 @@ lazy val docs = (project in file("docs")).
   settings(settings, publishArtifact := false).
   dependsOn(core)
 
+lazy val cats = (project in file("modules/cats")).
+  settings(settings, publishArtifact := false).
```
ruippeixotog: Since we want to publish this module to Maven, `publishArtifact := false` should not be here.

ruippeixotog (changes requested):

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
derekmorr: It actually doesn't throw an exception, since it's using ScalaTest's `EitherValues` - you get an error message like, "The Either on which right.value was invoked was not defined as a Right." But I'll change it.

FYI, we also use `EitherValues` in many other test suites in core and modules, so if this is something we're moving away from, we should file an issue.

derekmorr:
I've addressed your initial concerns, aside from the README, which I'll get to later today.

modules/cats/src/main/tut/README.md
derekmorr: I'm not sure I understand the issue here. Every other module's README shows reading HOCON into a case class.

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
ruippeixotog: You're right, I didn't know about `EitherValues` and I assumed `.value` was an alias for `.get` on a `RightProjection`, and not an extension method provided by ScalaTest.

Personally, I still prefer the less magical equality comparison between the `Either` and a `Right(expected)`, but I have no issue with the way it was :)

modules/cats/src/main/tut/README.md
ruippeixotog: My issue is not so much with using a case class to hold a config (which is pretty much mandatory) as with the particular example and wording being used. Most modules make the detail I mentioned clearer because the holding case classes in their examples have more than one field than one field and they don't imply that "we need a class" to "load a `NonEmptyList[Int]` into a configuration". If you see `joda`, for example, they mention an "object to receive our configuration" with two `DateTime`s and talk about reading "a GreatDatesConfig", not a `DateTime` directly. At least for me, it makes it clearer that those non-root classes like `DateTime` can be used normally in case class fields.

I don't know if I explained myself clearly. I'm also aware I'm being pedantic here, so please take this as a suggestion and ignore it if you don't think this is an issue.

ruippeixotog (approved):

modules/cats/src/main/tut/README.md
melrief: I think the examples are well done and similar to the other modules 😄 . In the examples we want to show one field of the new type provided inside a configuration because every real-world usage of our library will end up having a "root" case class object. I understand this can be a bit unclear but I don't think there is a good example covering everything and if I have to choose I prefer to choose the example with a root config class and one field.

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  it should "be able to read a config with a NonEmptyList" in {
+    val config = parseString(s"""{ numbers: [1,2,3] }""")
+    config.to[Numbers] shouldEqual Right(Numbers(NonEmptyList(1, List(2, 3))))
+  }
+
+  it should "return an EmptyTraversableFound when reading empty lists into NonEmptyList" in {
+    val config = parseString("{ numbers: [] }")
+    config.to[Numbers] shouldEqual
+      Left(ConfigReaderFailures(EmptyTraversableFound("scala.collection.immutable.List", None, Some("numbers")), Nil))
```
melrief, line 22: Nice @derekmorr, having this new error will make easier to use the library. Eventually all the errors from our library should be easy to read and easy to solve.

melrief (approved):
LGTM, nice job @derekmorr !

modules/cats/src/main/tut/README.md
```diff
@@ -0,0 +1,41 @@
+# Cats module for PureConfig
+
+Adds support for selected [cats](http://typelevel.org/cats/) classes to PureConfig.
+
+## Add pureconfig-cats to your project
+
+In addition to [core pureconfig](https://github.com/melrief/pureconfig), you'll need:
```
jcazevedo: This should now point to `https://github.com/pureconfig/pureconfig`.

jcazevedo (approved):

modules/cats/src/main/tut/README.md
derekmorr: ok this is fixed in the latest commit.

PR #225, opened 2017-05-16
I hope this fixes test failures with arbitrary Durations.

leifwickland:
The key was to parse the value as a `Long` first and then as a `Double` if that failed.

Sadly there are some automatic formatting changes included in this diff.

(Note I created this PR against @melrief's branch. I'm not sure if that was the best thing to do, but...)


ruippeixotog (approved):
I have just run the three `checkArbitrary` for durations with `PropertyCheckConfiguration(minSuccessful = 500000)` and it all want well. LGTM!

PR #226, opened 2017-05-17
support for Pattern and Regex

derekmorr:
Support for `java.util.regex.Pattern` and `scala.util.matching.Regex`.

I had to add a new function to core's test suite, `checkReadF` which applies a function to the `Right` value. This is because neither `Pattern` or `Regex` implement structural equality:

```scala
scala> import java.util.regex.Pattern
import java.util.regex.Pattern

scala> val p1 = Pattern.compile("a|b")
p1: java.util.regex.Pattern = a|b

scala> val p2 = Pattern.compile("a|b")
p2: java.util.regex.Pattern = a|b

scala> p1 == p2
res0: Boolean = false

scala> p1.pattern == p2.pattern
res1: Boolean = true
```

So I can't compare two `Pattern`s directly, but if I map them to Strings, via `.pattern()`, I can.

I'm open to suggestions on how to do this more elegantly.

### DIFF
--- a/core/src/main/scala/pureconfig/BasicReaders.scala
+++ b/core/src/main/scala/pureconfig/BasicReaders.scala
@@ -50,8 +50,8 @@ trait UriAndPathReaders {
  */
 trait RegexReaders {
 
-  implicit val patternReader = ConfigReader.fromNonEmptyString[Pattern](catchReadError(Pattern.compile))
-  implicit val regexReader = ConfigReader.fromNonEmptyString[Regex](catchReadError(new Regex(_)))
+  implicit val patternReader = ConfigReader.fromString[Pattern](catchReadError(Pattern.compile))
+  implicit val regexReader = ConfigReader.fromString[Regex](catchReadError(new Regex(_)))
 }
 
 /**
--- a/core/src/main/scala/pureconfig/BasicReaders.scala
+++ b/core/src/main/scala/pureconfig/BasicReaders.scala
@@ -5,8 +5,10 @@ import java.nio.file.{ Path, Paths }
 import java.time._
 import java.time.{ Duration => JavaDuration }
 import java.util.UUID
+import java.util.regex.Pattern
 
 import scala.concurrent.duration.{ Duration, FiniteDuration }
+import scala.util.matching.Regex
 
 import com.typesafe.config._
 import pureconfig.ConvertHelpers._
@@ -41,6 +43,15 @@ trait UriAndPathReaders {
   implicit val uuidConfigReader = ConfigReader.fromNonEmptyString[UUID](catchReadError(UUID.fromString))
   implicit val pathConfigReader = ConfigReader.fromString[Path](catchReadError(Paths.get(_)))
   implicit val uriConfigReader = ConfigReader.fromString[URI](catchReadError(new URI(_)))
+}
+
+/**
+ * Trait containing `ConfigReader` instances for classes related to regular expressions.
+ */
+trait RegexReaders {
+
+  implicit val patternReader = ConfigReader.fromNonEmptyString[Pattern](catchReadError(Pattern.compile))
+  implicit val regexReader = ConfigReader.fromNonEmptyString[Regex](catchReadError(new Regex(_)))
 }
 
 /**
@@ -125,6 +136,7 @@ trait TypesafeConfigReaders {
 trait BasicReaders
   extends PrimitiveReaders
   with UriAndPathReaders
+  with RegexReaders
   with JavaTimeReaders
   with DurationReaders
   with TypesafeConfigReaders
--- a/core/src/test/scala/pureconfig/ConfigConvertChecks.scala
+++ b/core/src/test/scala/pureconfig/ConfigConvertChecks.scala
@@ -5,7 +5,6 @@ import org.scalacheck.Arbitrary
 import org.scalatest.prop.GeneratorDrivenPropertyChecks
 import org.scalatest.{ EitherValues, FlatSpec, Matchers }
 import pureconfig.error.ConfigReaderFailure
-
 import scala.reflect.ClassTag
 
 /**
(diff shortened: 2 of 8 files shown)


ruippeixotog:
Scalatest matchers do not depend directly on the `equals` method of each object. Scalatest has pluggable equality checking - if you provide a custom instance of `Equality[Pattern]`, you can do what you need:

```scala
  it should "compare two patterns correctly" in {
    implicit val patternEquality = new Equality[Pattern] {
      def areEqual(a: Pattern, b: Any) = b match {
        case bp: Pattern => a.pattern == bp.pattern
        case _ => false
      }
    }

    Pattern.compile("a|b") shouldEqual Pattern.compile("a|b")
  }
```

If we make our `checkRead` receive an implicit `Equality[T]` and put the `patternEquality` above somewhere in scope, everything will work as expected and our core test suite will become more integrated with Scalatest :)

derekmorr:
Ah, yes, of course. Not sure why that slipped my mind. I've implemented custom `Equality` instances before. I'll update the PR.

derekmorr:
I've pushed up a new version which passes implicit `Equality[T]` to the functions in ConfigConvertChecks.

core/src/test/scala/pureconfig/ConfigConvertChecks.scala
```diff
@@ -23,7 +23,7 @@ trait ConfigConvertChecks { this: FlatSpec with Matchers with GeneratorDrivenPro
    * because, for instance, multiple representation of `t: T` are possible. Use [[checkRead()]] for those
    * representations.
    */
-  def checkArbitrary[T](implicit cc: ConfigConvert[T], arb: Arbitrary[T], tag: ClassTag[T]): Unit =
+  def checkArbitrary[T](implicit cc: ConfigConvert[T], arb: Arbitrary[T], tag: ClassTag[T], equality: Equality[T]): Unit =
     it should s"read an arbitrary ${tag.runtimeClass.getCanonicalName}" in forAll {
       (t: T) =>
         val result = cc.from(cc.to(t))
```
derekmorr, line 29: I'm not sure we need line 30 here. Calling `result.right.value` already checks if the result is a `Right` and `EitherValues`'s `.value` helper will fail the test if it's a `Left`.

core/src/test/scala/pureconfig/ConfigConvertChecks.scala
```diff
@@ -23,7 +23,7 @@ trait ConfigConvertChecks { this: FlatSpec with Matchers with GeneratorDrivenPro
    * because, for instance, multiple representation of `t: T` are possible. Use [[checkRead()]] for those
    * representations.
    */
-  def checkArbitrary[T](implicit cc: ConfigConvert[T], arb: Arbitrary[T], tag: ClassTag[T]): Unit =
+  def checkArbitrary[T](implicit cc: ConfigConvert[T], arb: Arbitrary[T], tag: ClassTag[T], equality: Equality[T]): Unit =
```
melrief, line 26: Nice improvement!

melrief (approved):

core/src/main/scala/pureconfig/BasicReaders.scala
```diff
@@ -44,6 +46,15 @@ trait UriAndPathReaders {
 }
 
 /**
+ * Trait containing `ConfigReader` instances for classes related to regular expressions.
+ */
+trait RegexReaders {
+
+  implicit val patternReader = ConfigReader.fromNonEmptyString[Pattern](catchReadError(Pattern.compile))
```
ruippeixotog: Shouldn't we allow empty strings to be parsed as regexes? It may not be an overly useful regex, but it's still a valid one. Also, as `patternWriter` can write an empty string, a round-trip test would require `patternReader` to read them, too.

ruippeixotog (changes requested):
Other than the comment below, LGTM!

core/src/main/scala/pureconfig/BasicReaders.scala
derekmorr: Ok, this is fixed in the latest commit.

ruippeixotog (approved):

core/src/test/scala/pureconfig/ConfigConvertChecks.scala
leifwickland: @derekmorr Yeah, I wondered about the same thing when I was in this neighborhood recently. My vote would be to remove it. 

If we want to get separate line numbers reported for left vs right as compared to the equality failing, I'd do:
```
val value = result.right.value
value shouldEqual t
```

core/src/test/scala/pureconfig/equality/package.scala
```diff
@@ -0,0 +1,26 @@
...
+import java.util.regex.Pattern
+
+import scala.util.matching.Regex
+
+import org.scalactic.Equality
+import org.scalactic.TypeCheckedTripleEquals._
+
+package object equality {
+
+  implicit val patternEquality = new Equality[Pattern] {
```
leifwickland: Total nitpick: I'd normally mark these `final` and capitalize the first letter. I don't know that we have a consistent pattern on that in Pureconfig. 
Feel free to ignore me.

core/src/test/scala/pureconfig/equality/package.scala
```diff
@@ -0,0 +1,26 @@
+package pureconfig
+
+import java.util.regex.Pattern
+
+import scala.util.matching.Regex
+
+import org.scalactic.Equality
+import org.scalactic.TypeCheckedTripleEquals._
+
+package object equality {
```
leifwickland, line 10: 👍  Thanks for adding this.

leifwickland (approved):
Nice work, @derekmorr! 
I made some suggestions. Feel free to take 'em or leave 'em.

derekmorr:
@leifwickland Your nits should be addressed in the latest push.

leifwickland:
LGTM

PR #228, opened 2017-05-18
check scaladoc in travis

### LINKED ISSUE #201: Add scaladoc creation test to CI
The lack of `sbt packageDoc` in the building script for travis allows for broken documentation to be merged in master, e.g. in #200. This issue is for adding a check on the scaladoc.

derekmorr:
This fixes #201

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ scala:
 jdk:
     - oraclejdk8
 script:
-  - sbt ++$TRAVIS_SCALA_VERSION coverage test tut
+  - sbt ++$TRAVIS_SCALA_VERSION coverage test tut doc
   - (cd example ; sbt ++$TRAVIS_SCALA_VERSION test)
 
   # check if there are no changes after `tut` runs


.travis.yml
```diff
@@ -6,7 +6,7 @@ scala:
 jdk:
     - oraclejdk8
 script:
-  - sbt ++$TRAVIS_SCALA_VERSION coverage test tut
+  - sbt ++$TRAVIS_SCALA_VERSION coverage test tut doc
```
leifwickland, line 9: It kind of seems like you'd only want to generate docs for 2.12 (well, the most recent version) if possible, rather than for all versions.
I have no idea if that's possible in Travis.

ruippeixotog: I think there's no problem in compiling docs for all versions. Scaladoc features may not change that much, but I guess we still want to ensure our docs are compatible with all Scala versions.

That said, we can generate docs for 2.12.x only just as we check the tut output:

```
  - if [[ $TRAVIS_SCALA_VERSION =~ '^2\.12.*' ]]; then
      sbt ++$TRAVIS_SCALA_VERSION doc;
      fi
```

ruippeixotog (approved):

leifwickland (approved):

PR #231, opened 2017-05-20
support for Java Enums

### LINKED ISSUE #230: support for java Enum
If possible, I'd like pureconfig to support reading arbitrary java `Enum`s.

derekmorr:
This PR adds support for loading arbitrary Java Enums. For example, we could load `java.time.DayOfWeek` or `java.lang.Character.UnicodeScript`, etc. It would close #230.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
     The readers and writers of the inner type are used instead of the ones for products.
   - `ConfigReader` and `ConfigWriter` instances for `Pattern` and `Regex`.
   - `ConfigReader` and `ConfigWriter` for `java.io.File`
+  - `ConfigReader` and `ConfigWriter` for arbitrary Java Enums
 - Bug fixes
   - `Duration.Undefined` is correctly handled when reading and writing configurations. [[#184](https://github.com/melrief/pureconfig/issues/184)]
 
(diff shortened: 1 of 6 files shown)


melrief (approved):

derekmorr:
I wonder if this should be moved into a module. It adds a dependency on scala.reflect in core. I'm not sure of the long-term plans for scala.reflect; I've seen some discussion of scala.meta replacing it. Isolating this in a module reduces future churn for pureconfig's core.

Update: I see we already use `scala.reflect.ClassTag` in many other places in core, so I'm less inclined to move it to a module.

melrief:
@derekmorr the use of `scala.reflect.ClassTag` is because we need the class name and that's the only reliable way that I know to find it. If we find another way to get the class name, we could drop `scala.reflect` from our codebase.

ruippeixotog (approved):
LGTM!

`ClassTag`, despite being part of the `scala.reflect` package, is part of the standard Scala library - we are not introducing an extra dependency here and we already use it in other places. It would be different if we had to depend on [scala-reflect](http://docs.scala-lang.org/overviews/reflection/overview.html), which requires a separate Maven dependency and that's the library that will probably be superseded by scala.meta.

PR #232, opened 2017-05-20
Add cats type class instances for `ConfigReader`, `ConfigWriter` and `ConfigConvert`

### LINKED ISSUE #224: Support for cats type classes
Instances of cats type classes (like `Monad` and `Contravariant`) for `ConfigReader`, `ConfigWriter` and `ConfigConvert`, would be useful for people that want to use cats `syntax` package and need appropriate instances. We should probably provide them in the `pureconfig-cats` module.

ruippeixotog:
Closes #224.


derekmorr:
This looks good. I like that the data structures and typeclass instances are in separate package objects, so they can be imported independently.

One small suggestion - I think we should add some section headers to the README to make the two different use-cases stand out.

ruippeixotog:
A side note about the Scaladoc, which made f4d72903d666f6ec6abfdeaa3f7de39163297e6e in this PR fail the tests: we currently cannot link to core pureconfig classes from a module, as SBT/Scaladoc cannot get a stable URL to link to (Scaladoc from the various modules are not aggregated in the end, they stay together with the respective JAR files).

The only reason we seemed to be able to do this before is that Scaladoc apparently does not generate documentation in package objects of empty packages ([SI-8124](https://issues.scala-lang.org/browse/SI-8124)). If we want documentation on those, we have to turn them into normal objects in the parent package (which is actually an organization I prefer, since most are not really "packages" but rather a collection of type class instances).

ruippeixotog:
Also, I added the headers you mentioned to README, @derekmorr.

derekmorr:
I think the module reorganization is an issue we can discuss in a separate thread. For now, this LGTM. Are there other instances we want to add?

derekmorr (approved):

modules/cats/src/main/tut/README.md
```diff
@@ -39,3 +42,38 @@ then load the config:
...
+### Using cats type class instances for readers and writers
+
+In order to to put in scope the `cats` type classes for our readers and writers and extend the latter with the extra
+operations provided by `cats`, we need some extra imports:
+
+```tut:silent
+import cats._
+import cats.implicits._
+import pureconfig.module.cats.instances._
+import pureconfig.syntax._
```
melrief: I'm a bit confused by the imports and maybe you can explain me why we need all four

1. `cats._` is for the typeclasses, no issue here
2. `cats.implicits._` should be `cats.syntax.all`? Do we need the instances for Scala standard library structures?
3. `pureconfig.module.cats.instances._` is the content of this PR and adds the instances for `ConfigReader`, `ConfigWriter` and `ConfigConvert`. No issue here
4. `pureconfig.syntax._` is needed? I see that we are using `Some("abc").toConfig` but that's not required. You can write `someWriter.to(Some("abc"))`, right?

melrief (approved):
LGTM, one small comment on the README.

modules/cats/src/main/tut/README.md
ruippeixotog: You're right in all of them. I replaced `cats.implicits._` by `cats.syntax.all._` and replaced the `toConfig` by the `writer.to(...)` syntax.

jcazevedo (approved):

PR #235, opened 2017-05-21
scala-xml module

derekmorr:
This adds a module for loading XML. Admittedly, loading XML from HOCON is an unusual use-case. This was more inspired by refined's XML support.

I'm not sure it's worth merging, but I offer it for what it's worth.


melrief (approved):

leifwickland (approved):
@derekmorr I hope I never need this, but if I do I'll be glad it exists.

PR #237, opened 2017-05-22
Distinguish parsing failures from conversion failures in ConfigReaderException

### LINKED ISSUE #190: Print parsing errors in a special way in ConfigReaderException
When reviewing #181 I noticed that both parsing errors and errors in the root file are printed in the same way - as neither of them has a config path defined, they are shown to be just "in the configuration". But in reality, they correspond to different kinds of "in the configuration". While, say, finding a list instead of an object in the root of the configuration is a localized error (it happens in the root path), a parsing error is a different thing altogether - logically, it happens on a previous phase than all the other "schema" failures currently in pureconfig.

We should distinguish these type of failures. I don't think it is worth it to distinguish them using different types, given the necessary increase in the complexity of the model. However, I'd like to see distinct error messages in those cases when a `ConfigReaderException` is thrown.

One way to easily solve it is to set the `path` of a non-`CannotParse` failure as `Some("")` when it happens at the root, distinguishing them from errors with undefined or unknown paths with `path == None`. `ConfigReaderException` would then deal with it in a special way (for example, printing "at the root" instead of "at ${path}".

jcazevedo:
This PR proposes a distinction between parsing failures and conversion failures, mainly to make it clearer in the `ConfigReaderException` message what errors happened at the root of the configuration and what errors were raised due to an inability to parse the configuration.

This fixes #190.

### DIFF
--- a/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
+++ b/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
@@ -58,24 +58,19 @@ object ConfigValueLocation {
 
 /**
  * A representation of a failure that might be raised from reading a
- * ConfigValue. The failure contains an optional location of the ConfigValue
- * that raised the error.
+ * ConfigValue. The failure contains an optional physical location of the
+ * configuration that raised the failure.
  */
 abstract class ConfigReaderFailure {
-  /**
-   * The optional location of the ConfigReaderFailure.
-   */
-  def location: Option[ConfigValueLocation]
-
-  /**
-   * The optional path to the `ConfigValue` that raised the failure.
-   */
-  def path: Option[String]
-
   /**
    * A human-readable description of the failure.
    */
   def description: String
+
+  /**
+   * The optional location of the ConfigReaderFailure.
+   */
+  def location: Option[ConfigValueLocation]
 
   /**
    * Improves the context of this failure with the key to the parent node and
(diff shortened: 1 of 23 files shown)


core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -58,24 +58,19 @@ object ConfigValueLocation {
 
 /**
  * A representation of a failure that might be raised from reading a
- * ConfigValue. The failure contains an optional location of the ConfigValue
- * that raised the error.
+ * ConfigValue. The failure contains an optional physical location of the
```
ruippeixotog: Minor suggestion: "physical location" is a bit strange. Consider changing to "file system location".

ruippeixotog (approved):
LGTM!

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: Changed it to "file system location" in 6d3a463.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -106,14 +114,14 @@ final case object CannotConvertNull extends ConfigReaderFailure {
  * @param because the reason why the conversion was not possible
  * @param location an optional location of the ConfigValue that raised the
  *                 failure
- * @param path an optional path to the value that couldn't be converted
+ * @param path thet path to the value that couldn't be converted
```
melrief: Typo `s/thet/the`

melrief (approved):
LGTM, just fix the typo

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: Thanks! Fixed the typo in ea16b58.

derekmorr (approved):

Issue #147, opened 2017-02-26
Improve the error message when the FieldMapping is wrong


melrief:
The field mapping is a great feature of pureconfig but has the side-effect of making more difficult to understand certain problems. For people coming from pureconfig < `0.6.0`, the default mapping is CamelCase->CamelCase and the errors about keys not found make no sense. A solution could be that, given a certain set of keys not found, pureconfig could try to see if there are matching keys with other field mapping, e.g. `foo-bar` instead of `fooBar`. This isn't easy to implement because we need to be able to access to all the other possible field mapping and I'm not sure how this could be encoded. The main drawback of this is that the error message would take longer to be generated but I would say it's worth it.

jcazevedo:
Fixed in #238.

PR #238, opened 2017-05-22
Improve the error message in case of a misconfigured ProductHint

jcazevedo:
This PR proposes an improvement to the error message displayed when there's a suspicion of a misconfigured `ProductHint`, fixing #147.

In order to detect a possibly misconfigured `ProductHint`, I build a set of "candidate" keys (i.e. keys in the configuration that might map to the desired field value). A "candidate" key is a key whose lowercase form contains a subsequence of the desired field value considering only alphanumeric characters.


core/src/main/scala/pureconfig/DerivedReaders.scala
```diff
@@ -75,14 +75,15 @@ trait DerivedReaders1 {
     hint: ProductHint[Wrapped]): WrappedDefaultValue[Wrapped, FieldType[K, V] :: T, Option[V] :: U] = new WrappedDefaultValue[Wrapped, FieldType[K, V] :: T, Option[V] :: U] {
 
     override def fromConfigObject(co: ConfigObject, default: Option[V] :: U): Either[ConfigReaderFailures, FieldType[K, V] :: T] = {
-      val keyStr = hint.configKey(key.value.toString().tail)
+      val fieldName = key.value.toString().tail
```
ruippeixotog: I know that this is not new code, but this is not the best way to convert a `Symbol` to a `String`. We are calling its `toString` and dropping the single quote, while we could access its string value directly by doing `key.value.name`.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -87,15 +91,39 @@ abstract class ConfigReaderFailure {
...
+      true
+    else if (s2.isEmpty())
+      false
+    else if (s1.head == s2.head)
+      isSubsequence(s1.tail, s2.tail)
+    else
+      isSubsequence(s1, s2.tail)
+  }
+
+  def apply(fieldName: String, confObj: ConfigObject): CannotConvertNull = {
```
ruippeixotog: Minor suggestion: consider receiving here an `Iterable[String]` or even a `TraversableOnce[String]` with all available keys instead of a `ConfigObject`. While we always have access to a `ConfigObject` in the places we use it, users might want to create their own `ConfigReader` instances and that way we wouldn't be so restrictive.

ruippeixotog (changes requested):
LGTM overall!

core/src/main/scala/pureconfig/DerivedReaders.scala
jcazevedo: Fixed it in 22ee67f.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: Sure. Modified it in 039a256.

ruippeixotog (approved):

melrief (approved):

PR #241, opened 2017-05-24
syntactic sugar for Play Configuration

derekmorr:
I've been using pureconfig in a Play application lately. Play provides its own wrapper about typesafeconfig's `Config` object -- [Configuration](https://www.playframework.com/documentation/2.5.x/api/scala/index.html#play.api.Configuration).

Rather than calling `configuration.underlying.to[MyClass]`, I wrote a quick wrapper around `Configuration` adding the same syntactic sugar that `pureconfig.syntax._` provides.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,6 +29,12 @@ lazy val joda = (project in file("modules/joda")).
   dependsOn(core).
   dependsOn(core % "test->test") // In order to reuse the date/time related scalacheck generators.
 
+// play 2.5 is only published for Scala 2.11
+lazy val play = (project in file("modules/play")).
+  settings(settings).
+  settings(crossScalaVersions ~= { oldVersions => oldVersions.filter(_.startsWith("2.11")) }).
+  dependsOn(core)
+
 lazy val squants = (project in file("modules/squants")).
   settings(settings).
   dependsOn(core)
--- a/build.sbt
+++ b/build.sbt
@@ -35,6 +35,10 @@ lazy val play = (project in file("modules/play")).
   settings(crossScalaVersions ~= { oldVersions => oldVersions.filter(_.startsWith("2.11")) }).
   dependsOn(core)
 
+lazy val scalaxml = (project in file("modules/scala-xml")).
+  settings(settings).
+  dependsOn(core)
+
 lazy val squants = (project in file("modules/squants")).
   settings(settings).
   dependsOn(core)
--- a/modules/play/README.md
+++ b/modules/play/README.md
@@ -1,0 +1,36 @@
+# Play Configuration module for PureConfig
+
+Adds syntactic sugar to Play's 2.5's 
+[Configuration](https://www.playframework.com/documentation/2.5.x/api/scala/index.html#play.api.Configuration) class. 
+This is similar to the syntactic sugar added to TypeSafe Config's `Config` class by `pureconfig.syntax._`
+
+## Add pureconfig-play to your project
+
+In addition to [core pureconfig](https://github.com/melrief/pureconfig), you'll need:
+
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-play" % "0.7.1"
+```
+
+## Example
+
+To load a configuration data into a case class using the Play Configuration class:
+
+
+```scala
+import play.api.Configuration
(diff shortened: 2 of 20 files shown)


derekmorr:
It looks like Play 2.6 will support 2.11 and 2.12, so I wonder if we should rename this module to play25 to simplify Scala version support.

build.sbt
```diff
@@ -29,6 +29,12 @@ lazy val joda = (project in file("modules/joda")).
   dependsOn(core).
   dependsOn(core % "test->test") // In order to reuse the date/time related scalacheck generators.
 
+// play 2.5 is only published for Scala 2.11
+lazy val play = (project in file("modules/play")).
+  settings(settings).
+  settings(crossScalaVersions ~= { oldVersions => oldVersions.filter(_.startsWith("2.11")) }).
```
ruippeixotog, line 35: The `scalaVersion` key is still set to `2.12.2` here. You should change this line to something like:

```scala
  settings(
    scalaVersion := "2.11.11",
    crossScalaVersions ~= { oldVersions => oldVersions.filter(_.startsWith("2.11")) })
```

ruippeixotog (commented):
I'm a bit concerned about our "usefulness" threshold for creating new modules. As it is now, this module just provides syntactic sugar for writing `conf.to[A]` instead of `conf.underlying.to[A]` in a Play config. As an user, I would never include a new Maven dependency just to save a few characters (that may not be the case for everyone, of course). As a maintainer, adding new modules right now does not introduce that much complexity into the SBT project, but I wonder if that will become more and more burdensome as the number of modules we accept in the same repository as the core increases.

Regarding the module naming, I think this module would be called `play`, regardless of the Scala versions we support. When Play 2.6 is released we can release the next version of `play` for both Scala 2.11 and Scala 2.12.

Overall I'm ok with accepting this, although I'm not too excited by it.

derekmorr:
Regarding naming, Play doesn't have great cross-version compatibility, so I don't know if a single module would work. 

I have several Play apps where I need to load config in multiple places, and I ended up writing something similar to this to use internally. I figured, why not extract it and push it upstream. Different users will have different thresholds for including dependencies. I admit it's not the most useful module.

modules/play/src/main/scala/pureconfig/module/package.scala
```diff
@@ -0,0 +1,20 @@
...
+import pureconfig.ConfigReader
+import pureconfig.error.ConfigReaderFailures
+import pureconfig.syntax._
+
+/**
+ * Augmented syntax for Play's Configuration class
+ */
+package object play {
+
+  implicit class PimpedPlayConfig(val playConf: Configuration) extends AnyVal {
```
leifwickland, line 15: There are miscellaneous complaints in the community about using the termed "pimped." I'd recommend something like `PlayConfigOps` or `RichPlayConfig` or `PlayConfigSyntax` to avoid it.

modules/play/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+case class MyConfig(name: String)
+```
+
+We can read a `MyConfig` like:
+```scala
+val conf = Configuration.from(Map("name" -> "Susan"))
+// conf: play.api.Configuration = Configuration(Config(SimpleConfigObject({"name":"Susan"})))
+
+conf.to[MyConfig]
```
leifwickland, line 32: That is really slick. Nice work!

leifwickland:
Remember to add a link from the README to this module, please.
https://github.com/pureconfig/pureconfig#internal-modules

It looks like we're a bit behind on doing that for other modules. Would you be willing to add the others when you add this, please?

leifwickland:
@ruippeixotog I'm OK with accepting this module despite it being of somewhat questionable usefulness. I figure that commonly the two most important factors when picking a library are
- Ease of getting started, which I think PureConfig has killed with good documentation
- Breadth of functionality, which PureConfig's liberal modulization allows us to support.

If someday we're without a maintainer for a module, I won't feel too bad about axing it if the community fails to produce another.

leifwickland (approved):
LGTM

ruippeixotog:
Closing due to inactivity. Feel free to reopen it if you still need it :)

PR #245, opened 2017-05-26
Make ConfigConvertWrapper return a failure if a Path cannot be read

### LINKED ISSUE #244: parseFile and loadFile doesn't fail if the file doesn't exist

ruippeixotog:
Fixes #244.

### DIFF
--- a/core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
+++ b/core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
@@ -2,10 +2,10 @@ package pureconfig.backend
 
 import java.nio.file.Path
 
-import com.typesafe.config.{ Config, ConfigException, ConfigFactory }
-import pureconfig.error.{ CannotParse, ConfigReaderFailures, ConfigValueLocation, ThrowableFailure }
-
 import scala.util.control.NonFatal
+
+import com.typesafe.config.{ Config, ConfigException, ConfigFactory }
+import pureconfig.error._
 
 /**
  * A wrapper of `com.typesafe.config.ConfigFactory` whose methods return [[scala.Either]] instead
@@ -22,15 +22,17 @@ object ConfigFactoryWrapper {
     unsafeToEither(ConfigFactory.load())
 
   /** @see `com.typesafe.config.ConfigFactory.parseFile()` */
-  def parseFile(path: Path): Either[ConfigReaderFailures, Config] =
-    unsafeToEither(ConfigFactory.parseFile(path.toFile))
+  def parseFile(path: Path): Either[ConfigReaderFailures, Config] = {
+    if (!path.toFile.isFile || !path.toFile.canRead) Left(ConfigReaderFailures(CannotReadFile(path)))
+    else unsafeToEither(ConfigFactory.parseFile(path.toFile))
+  }
 
   /** Utility methods that parse a file and then calls `ConfigFactory.load` */
   def loadFile(path: Path): Either[ConfigReaderFailures, Config] =
     parseFile(path).right.flatMap(rawConfig => unsafeToEither(ConfigFactory.load(rawConfig)))
 
   private def unsafeToEither[A](f: => A): Either[ConfigReaderFailures, A] = {
-    try (Right(f)) catch {
+    try Right(f) catch {
       case e: ConfigException.Parse =>
         Left(ConfigReaderFailures(CannotParse(e.getLocalizedMessage, ConfigValueLocation(e.origin()))))
       case e: ConfigException =>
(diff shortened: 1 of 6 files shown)


core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
```diff
@@ -22,15 +22,17 @@ object ConfigFactoryWrapper {
     unsafeToEither(ConfigFactory.load())
 
   /** @see `com.typesafe.config.ConfigFactory.parseFile()` */
-  def parseFile(path: Path): Either[ConfigReaderFailures, Config] =
-    unsafeToEither(ConfigFactory.parseFile(path.toFile))
+  def parseFile(path: Path): Either[ConfigReaderFailures, Config] = {
+    if (!path.toFile.isFile || !path.toFile.canRead) Left(ConfigReaderFailures(CannotReadFile(path)))
```
ruippeixotog: This check catches many common cases, but it does not ensure that the file can indeed be read. The only sure way to know that is reading the file ourselves and pass its contents to `ConfigFactory.parseString`, but then we would need to care with other things, like dealing with charsets. Do you think this check is enough, at least for now?

derekmorr: I wonder if it would be better to use the Java 7 NIO APIs in the conditional:

Files.isReadable(path) && Files.isRegularFile(path)

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -298,6 +299,16 @@ final case class NoValidCoproductChoiceFound(value: ConfigValue, location: Optio
 }
 
 /**
+ * A failure representing the inability to read a requested file.
+ */
+final case class CannotReadFile(path: Path) extends ConfigReaderFailure {
```
derekmorr, line 304: Is it worth distinguishing between `path` not being a file vs not being able to read the file?

core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
ruippeixotog: Sure, I can change it to use NIO since we are already working with a `Path`. Out of curiosity, is there any functional difference in these checks between the two APIs?

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
ruippeixotog: Yeah, I thought about that too. And there are more reasons why you might not be able to read a file, such as the case where a file is locked by the OS.

In this PR I didn't focus in giving the user information about exactly why the file cannot be read, but I think that would be better passed along to the user by making a general failure like `CannotReadFile` hold a `reason` string than by creating a different `Failure` for each type of IO error.

core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
derekmorr: I don't believe there's a functional difference, but if we're already using NIO, it makes sense to keep using it. It also avoids some extra allocations.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
derekmorr: Makes sense.

core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
ruippeixotog: Sure. Changed it to java.nio in 7503363.

core/src/test/scala/pureconfig/ConfigReaderMatchers.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.scalatest._
+import org.scalatest.matchers.Matcher
+import pureconfig.error.{ ConfigReaderFailure, ConfigReaderFailures }
+
+trait ConfigReaderMatchers { this: FlatSpec with Matchers =>
+
+  def failWith(failure: ConfigReaderFailure): Matcher[Either[ConfigReaderFailures, Any]] =
+    be(Left(ConfigReaderFailures(failure, Nil)))
+
+  def failWithType[Failure <: ConfigReaderFailure: ClassTag]: Matcher[Either[ConfigReaderFailures, Any]] =
```
leifwickland, line 14: This is really expressive. Nice work!

leifwickland (approved):
LGTM. 👍

derekmorr (approved):

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
ruippeixotog: I have just opened #248 for me to deal with this later.

PR #247, opened 2017-05-27
Make loadConfig return a failure when a namespace doesn't exist

### LINKED ISSUE #218: loadConfig throws ConfigException$Missing if the namespace doesn't exists
This should return a `Failure`, not throw exception.

```scala
scala> loadConfig[Int](namespace = "foo.bar")
com.typesafe.config.ConfigException$Missing: No configuration setting found for key 'foo'
	at com.typesafe.config.impl.SimpleConfig.findKeyOrNull(SimpleConfig.java:152)
	at com.typesafe.config.impl.SimpleConfig.findKey(SimpleConfig.java:145)
	at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:172)
	at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:184)
	at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:189)
	at com.typesafe.config.impl.SimpleConfig.getObject(SimpleConfig.java:264)
	at com.typesafe.config.impl.SimpleConfig.getConfig(SimpleConfig.java:270)
	at com.typesafe.config.impl.SimpleConfig.getConfig(SimpleConfig.java:37)
	at pureconfig.package$$anonfun$loadConfig$2$$anonfun$apply$3.apply(package.scala:45)
	at pureconfig.package$$anonfun$loadConfig$2$$anonfun$apply$3.apply(package.scala:45)
	at scala.util.Either$RightProjection.map(Either.scala:536)
	at pureconfig.package$$anonfun$loadConfig$2.apply(package.scala:45)
	at pureconfig.package$$anonfun$loadConfig$2.apply(package.scala:44)
	at scala.util.Either$RightProjection.flatMap(Either.scala:523)
	at pureconfig.package$.loadConfig(package.scala:44)
```

I'm starting to think that we should wrap every part of typesafe config that we use in a safer wrapper.

ruippeixotog:
Depends on #245 and fixes #218.

### DIFF
--- a/core/src/main/scala/pureconfig/package.scala
+++ b/core/src/main/scala/pureconfig/package.scala
@@ -7,11 +7,10 @@
 import java.io.{ OutputStream, PrintStream }
 import java.nio.file.{ Files, Path }
 
-import com.typesafe.config.{ Config => TypesafeConfig }
-import pureconfig.error.{ ConfigReaderException, ConfigReaderFailures, ConfigValueLocation }
+import com.typesafe.config.{ ConfigException, Config => TypesafeConfig }
+import pureconfig.error.{ ConfigReaderException, ConfigReaderFailures, ConfigValueLocation, KeyNotFound }
 import pureconfig.ConfigConvert.improveFailures
 import pureconfig.backend.ConfigFactoryWrapper.{ invalidateCaches, load, loadFile, parseFile }
-
 import scala.reflect.ClassTag
 
 package object pureconfig {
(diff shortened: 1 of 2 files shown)


derekmorr (approved):

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -298,6 +299,16 @@ final case class NoValidCoproductChoiceFound(value: ConfigValue, location: Optio
 }
 
 /**
+ * A failure representing the inability to read a requested file.
+ */
+final case class CannotReadFile(path: Path) extends ConfigReaderFailure {
+  val location = None
+
+  def description = s"Unable to read file: ${path.toString}"
```
melrief: I would prefer to say why the file is not readable. The message could be `"Unable to read file ${path.toString} because it doesn't exist"`

melrief (changes requested):
LGTM except the error message that should be more precise

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
ruippeixotog: The addition of `CannotReadFile` was already in #245 and there was a discussion [there](https://github.com/pureconfig/pureconfig/pull/245#discussion_r118805343) about this topic. I agree with you, but I think it's a different issue and PR ;)

ruippeixotog: I have just opened #248 for me to deal with this later.

melrief (approved):

PR #252, opened 2017-05-29
Simplify SBT build

ruippeixotog:
This PR simplifies some things in our current SBT build. Some highlights:

- Extracted the boilerplate of declaring a module in the main `build.sbt` to a function, making the declaration of a new module an one-liner;
- Moved many settings which were replicated in all the modules and should always be the same (such as `organization` and `homepage`) to the main `build.sbt`;
- Simplified the logic of applying different lint flags for different Scala versions;
- Updated tut to 0.5.2, which makes tut an autoplugin.


build.sbt
```diff
@@ -1,110 +1,99 @@
...
+  settings(commonSettings, publishArtifact := false).
   dependsOn(core)
 
-lazy val cats = (project in file("modules/cats")).
-  settings(settings).
-  dependsOn(core)
+def module(proj: Project) = proj.
+  enablePlugins(TutPlugin).
+  dependsOn(core).
+  dependsOn(core % "test->test"). // In order to reuse the scalacheck generators
```
derekmorr, line 18: Won't this bring in core's test dependencies into scope for every module? Before, only the joda had them.

build.sbt
```diff
@@ -1,110 +1,99 @@
...
 lazy val docs = (project in file("docs")).
-  settings(settings, publishArtifact := false).
+  enablePlugins(TutPlugin).
+  settings(commonSettings, publishArtifact := false).
   dependsOn(core)
 
-lazy val cats = (project in file("modules/cats")).
-  settings(settings).
-  dependsOn(core)
+def module(proj: Project) = proj.
```
melrief, line 15: Nice!

melrief (approved):

derekmorr (changes requested):

build.sbt
melrief: Yes but we already discussed of sharing the testing part of PureConfig. Before there was no need but now we may want to have all modules to use the same format for tests as `core`, which means that each module should consider using [`this`](https://github.com/pureconfig/pureconfig/blob/master/core/src/test/scala/pureconfig/ConfigConvertChecks.scala).

melrief: Otherwise we can't add a boolean to `module` that decides whether we include `core "test" -> "test` or not. I'm not too strong about this.

ruippeixotog: Yes, that's the reason why. As I and @melrief discussed before, a more elegant solution would pass by either a `testkit` project with the test helpers, which besides being used internally could be used by other pureconfig users for their own `ConfigConvert` instances, or a private `tests` project with the tests for the core and all modules. Meanwhile, this is the solution we adopted, and every module should have access to those helpers.

derekmorr: Ok that's fine. I hadn't see the discussion.

derekmorr (approved):

PR #253, opened 2017-05-29
fix deriveAnyVal to derive only Generic

melrief:
Sneaky bug, it took sometime to understand what is going on. Basically without the `Generic` constraint, `deriveAnyVal` was deriving an instance of `ConfigReader`/`ConfigWriter` for any `T` extending `AnyVal` and having a `ConfigReader`/`ConfigWriter`. The issue is this [shapeless code](https://github.com/milessabin/shapeless/blob/39f73305e3cd2746948c6ef2dd1e065fe08cb1e1/core/src/main/scala/shapeless/unwrapped.scala#L70) which creates and instance of `Unwrapped` for any `T`. Given the fact that literally anything has an `Unwrapped` instance, anything else creating a real `ConfigReader` for `AnyVal`, like Refined, was conflicting and scalac gave up the implicit resolution with a diverging implicit expansion issue.

I've tried it with refined master and it works.

One thing that is still not nice is that if there in implicit resolution conflict, Scalac doesn't tell you at `loadConfig` level and you need to specifically ask for that type `ConfigReader` instance. Silent errors are the worst and I wonder if there is a flag to force printing this issue.

Note that I've set the version to `0.7.1.1` because I think it's not worth it increasing the point version. Let me know if it's fine for you.

### DIFF
--- a/core/src/main/scala/pureconfig/DerivedWriters.scala
+++ b/core/src/main/scala/pureconfig/DerivedWriters.scala
@@ -12,6 +12,7 @@ trait DerivedWriters extends DerivedWriters1 {
   implicit def deriveAnyVal[T, U](
     implicit
     ev: T <:< AnyVal,
+    generic: Generic[T],
     unwrapped: Unwrapped.Aux[T, U],
     writer: ConfigWriter[U]): ConfigWriter[T] =
     new ConfigWriter[T] {
--- a/core/src/main/scala/pureconfig/DerivedReaders.scala
+++ b/core/src/main/scala/pureconfig/DerivedReaders.scala
@@ -28,6 +28,7 @@ trait DerivedReaders extends DerivedReaders1 {
   implicit def deriveAnyVal[T, U](
     implicit
     ev: T <:< AnyVal,
+    generic: Generic[T],
     unwrapped: Unwrapped.Aux[T, U],
     reader: ConfigReader[U]): ConfigReader[T] =
     new ConfigReader[T] {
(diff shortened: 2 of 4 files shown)


derekmorr:
This LGTM, but I'd probably push the version to 0.7.2.

derekmorr (approved):

ruippeixotog (dismissed):
LGTM. Regarding the version, I prefer `0.7.2` - most of the dependencies I work with follow semantic versioning, which works only with three levels. Also, if you are going to try sbt-release again, I think you should not increase manually the version :)

core/src/main/scala/pureconfig/DerivedWriters.scala
```diff
@@ -12,6 +12,7 @@ trait DerivedWriters extends DerivedWriters1 {
   implicit def deriveAnyVal[T, U](
     implicit
     ev: T <:< AnyVal,
+    generic: Generic[T],
```
ruippeixotog, line 15: I did not know shapeless generated `Generic` instances for classes that are not case classes. Does this happen only because classes like `IntWrapper` are `AnyVal` instances?

melrief:
```scala
@ import shapeless._ 
import shapeless._
@ final class Foo(val int: Int) extends AnyVal 
defined class Foo
@ Generic[Foo] 
res2: Generic[Foo]{type Repr = Int :: shapeless.HNil} = $sess.cmd2$anon$macro$3$1@54ff310
@ res2.from(1 :: HNil).int 
res4: Int = 1
```

melrief: The important part is that we don't derive `ConfigReader`s and `ConfigWriter`s for classes that have no generic instance because `Unwrapped` there is defined as the identity.

PR #254, opened 2017-05-29
Release 0.7.2

melrief:


derekmorr (approved):

ruippeixotog (approved):

jcazevedo (approved):

derekmorr:
Minor nit:

In the "Add PureConfig to your project", should we update the scala versions:

```
In the sbt configuration file use Scala 2.10, 2.11 or 2.12:

scalaVersion := "2.12.1" // or "2.11.8", "2.10.5"
```

Shouldn't this be 2.12.2, 2.11.11 and 2.10.6 ?

ruippeixotog:
I don't think so. Scala 2.12.x binaries (as all as 2.11.x and 2.10.x) are binary compatible between them; the specific version with which we compiled pureconfig should not matter to users of the library, as they do not have to use the same version as us.

derekmorr:
Yes, I'm aware that end-users can use 2.12.x, but we should list the current versions.

ruippeixotog:
Oh, you're right, I was looking at the wrong line. Disregard what I said, please :)

That is not related to this PR though, so @melrief can decide if he wants to merge this now and leave that to a later PR or fix it in this one.

melrief:
Merged, let's do another PR.

PR #258, opened 2017-06-02
Arbitrary precision number support

derekmorr:
This isn't ready to merge yet because the tests intermittently fail. I sporadically get errors like:

 1.164824982791316359504978122363828E+219 did not equal 
 1.164824982791316359504978122363828010365105106538157160E+219

-4.496425998968164481481607314628320E-186 did not equal 
-4.4964259989681644814816073146283195719766919765873115220714445693225814848E-186

I'm not happy with the hacky way I had to make `Gen` and `Arbitrary` instances for the java math types. Scalacheck only supports the scala big number types, and only has `Arbitrary` instances for them.

I'll look into cleaning this up so it can be merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,6 @@
   - `ConfigFieldMapping` now has a `withOverrides` method that allows users to easily define exceptional cases to an
     existing mapping;
   - `ConfigReader` and `ConfigWriter` for `java.math.BigDecimal`, `java.math.BigInteger`, 
-    `scala.math.BigDecimal`, and `scala.math.BigInt`
-
 - Bug fixes
   - A breaking change introduced in v0.7.1 where `loadConfigFromFiles` stopped allowing missing files was reverted.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### 0.7.3 (unreleased)
+
+- Bug fixes
+  - A breaking change introduced in v0.7.1 where `loadConfigFromFiles` stopped allowing missing files was reverted.
+
 ### 0.7.2 (May 29, 2017)
 
 - Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 ### 0.7.3 (unreleased)
 
+- New features
+  - `ConfigFieldMapping` now has a `withOverrides` method that allows users to easily define exceptional cases to an
+    existing mapping;
 - Bug fixes
   - A breaking change introduced in v0.7.1 where `loadConfigFromFiles` stopped allowing missing files was reverted.
-- New features
-  - `ConfigFieldMapping` now has a `withExceptions` method that allows users to easily define exceptions to an existing
-    mapping.
 
 ### 0.7.2 (May 29, 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - Bug fixes
   - A breaking change introduced in v0.7.1 where `loadConfigFromFiles` stopped allowing missing files was reverted.
+- New features
+  - `ConfigFieldMapping` now has a `withExceptions` method that allows users to easily define exceptions to an existing
+    mapping.
 
 ### 0.7.2 (May 29, 2017)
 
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+### 0.X.Y (unreleased)
+
+- New features
+  - `ConfigReader` and `ConfigWriter` for `java.math.BigDecimal`, `java.math.BigINteger`,
```
ruippeixotog: There's a typo at `BigINteger` here.

core/src/test/scala/pureconfig/gen/package.scala
```diff
@@ -66,4 +67,14 @@ package object gen {
       dateTime <- ConfigurableSuite.localDateTimeArbitrary.arbitrary
     } yield new org.joda.time.DateTime(dateTime.getYear, dateTime.getMonthValue, dateTime.getDayOfMonth,
       dateTime.getHour, dateTime.getMinute, dateTime.getSecond, org.joda.time.DateTimeZone.UTC)
+
+  val genJavaBigDecimal: Gen[JavaBigDecimal] =
+    for {
+      bigDecimal <- Arbitrary.arbBigDecimal.arbitrary
+    } yield bigDecimal.bigDecimal
+
+  val genBigInt: Gen[BigInteger] =
```
ruippeixotog: Consider changing this to the more compact and idiomatic `Arbitrary.arbitrary[BigInt].map(_.bigInteger)`, it you prefer.

core/src/test/scala/pureconfig/gen/package.scala
```diff
@@ -66,4 +67,14 @@ package object gen {
       dateTime <- ConfigurableSuite.localDateTimeArbitrary.arbitrary
     } yield new org.joda.time.DateTime(dateTime.getYear, dateTime.getMonthValue, dateTime.getDayOfMonth,
       dateTime.getHour, dateTime.getMinute, dateTime.getSecond, org.joda.time.DateTimeZone.UTC)
+
+  val genJavaBigDecimal: Gen[JavaBigDecimal] =
+    for {
```
ruippeixotog: Consider changing this to the more compact and idiomatic `Arbitrary.arbitrary[BigDecimal].map(_.bigDecimal)`, it you prefer.

ruippeixotog (commented):
LGTM overall. Regarding the `Arbitrary` for the Java numeric types I think you did it the correct way; I have just suggested a better code style below.

Regarding the loss of precision, it's a bit strange that we are losing precision somewhere in our process... `BigInteger` and `BigDecimal` should be resistant to that. Did you see it happen only in the Java ones or does it also happen with the Scala wrappers?

We can always do tolerant equality with [TolerantNumerics](http://doc.scalatest.org/2.2.4/#org.scalactic.TolerantNumerics), but I don't think it's the right approach for these specific `Numeric` types.

derekmorr:
I've isolated the bugs to 2.10. It looks like defaults for math context changed from 2.10 to 2.11. I'll look into it more.

derekmorr:
Scala 2.10 had a bug which truncated `BigDecimal` when parsed from a `String`. This meant that a round-trip test would never succeed at least without hacking the test to duplicate the truncation bug.

I solved this by making a local copy of Scala 2.11's fixed BigDecimal code. I think this is the lesser of two evils. Alternatively, we could try to make a 2.10-specific version of `NumericReaders`.

core/src/main/scala/pureconfig/BasicReaders.scala
```diff
@@ -114,6 +116,35 @@ trait DurationReaders {
...
 /**
+ * Trait containing `ConfigReader` instances for Java and Scala arbitrary-precision numeric types.
+ */
+trait NumericReaders {
+
+  // BigDecimal String parsing copied from Scala 2.11.
+  // The Scala 2.10 bigDecimal class is hard-coded to use 128-bits of prevision when parsing BigDecimal from String.
+  // This causes truncation for large values.
+  // Scala 2.11+ fixed this bug by adapting precision to accommodate all digits in the string.
+  private def exact(repr: JavaBigDecimal): BigDecimal = {
```
ruippeixotog, line 127: You defined this here, but you forgot to use it in the readers :)

derekmorr: Yeah, I shouldn't have pushed that late at night. I wonder why if even compiled. Anyway, it's fixed.

ruippeixotog (approved):
LGTM!

jcazevedo (approved):
Looks good to me! Thanks for this!

melrief (approved):
Nice job!

Issue #204, opened 2017-04-04
Custom key name


david-bouyssie:
If I understood correctly what is described din the field-mappings page (https://github.com/pureconfig/pureconfig/blob/master/docs/override-behaviour-for-case-classes.md#field-mappings), pureconfig allows to define a mapping between a key naming convention and another one for Scala fields.

However what it if keys are named with a mixed naming convention ?
I think it could be nice to annotate case class fields to provide a specific key name like:
```scala
case class Dummy(
  @TypeSafeConfigKey("bla_bla") blaBla: Foo,
  @TypeSafeConfigKey("yo-yo") yoYo: Bar
)
```

Thanks!

jcazevedo:
I don't like annotating domain classes with (de)serialization information as I think it pollutes the model with unrelated logic.

It is possible to do what you want by defining a custom `ProductHint` for the desired types:

```scala
case class Dummy(blaBla: Int, yoYo: Int)

class CustomProductHint[T](mappings: Map[String, String]) extends ProductHint[T] {
  private val default = ProductHint.default[T]
  def configKey(fieldName: String) = mappings.getOrElse(fieldName, default.configKey(fieldName))
  def useDefaultArgs = default.useDefaultArgs
  def allowUnknownKeys = default.allowUnknownKeys
}

implicit val ph = new CustomProductHint[Dummy](Map("blaBla" -> "bla_bla", "yoYo" -> "yo-yo"))

Dummy(1, 2).toConfig
// returns SimpleConfigObject({"bla_bla":1,"yo-yo":2})
```

Nonetheless, I think we can consider adding helpers to create a `ProductHint` from a map of field names to typesafe config keys, as it is a common use case.

david-bouyssie:
Very nice and unobtrusive solution!

Thanks ;-)

melrief:
@jcazevedo the idea of adding helpers is not bad at all, it could be used for special cases where the user needs to control field mapping. Do we want to create an issue for this?

ruippeixotog:
I have just opened #256 and I'm going to take care of it in the next days :)

PR #259, opened 2017-06-03
Add a withOverrides method to ConfigFieldMapping

### LINKED ISSUE #256: Add an easy way to define exceptions to field mappings
The motivation for this is in #204. Sometimes, we want to define a custom field mapping for a specific key of a case class. This is particularly common when the field has numbers - for example, with the default field mapping a field named `s3BackupLocation` is mapped to the config key `s-3-backup-location`, while most of us would expect it to be `s3-backup-location`.

ruippeixotog:
Closes #256 and #204.

### DIFF
--- a/core/src/main/scala/pureconfig/ConfigFieldMapping.scala
+++ b/core/src/main/scala/pureconfig/ConfigFieldMapping.scala
@@ -14,10 +14,8 @@ trait ConfigFieldMapping extends (String => String) {
    * @return a `ConfigFieldMapping` that maps fields using `exceptions` if the field is present there and otherwise
    *         uses this mapping.
    */
-  def withExceptions(exceptions: (String, String)*) = {
-    val map = exceptions.toMap
-    ConfigFieldMapping { field => map.getOrElse(field, this.apply(field)) }
-  }
+  def withExceptions(exceptions: (String, String)*) =
+    ConfigFieldMapping(exceptions.toMap.withDefault(apply))
 }
 
 object ConfigFieldMapping {
--- a/core/src/main/scala/pureconfig/ConfigFieldMapping.scala
+++ b/core/src/main/scala/pureconfig/ConfigFieldMapping.scala
@@ -8,14 +8,14 @@ trait ConfigFieldMapping extends (String => String) {
   def apply(fieldName: String): String
 
   /**
-   * Returns a `ConfigFieldMapping` that uses this mapping with some exceptions.
+   * Returns a `ConfigFieldMapping` that uses this mapping with some overrides.
    *
-   * @param exceptions the exceptions for this mapping as pairs (field, configKey)
-   * @return a `ConfigFieldMapping` that maps fields using `exceptions` if the field is present there and otherwise
+   * @param overrides the overrides for this mapping as pairs (field, configKey)
+   * @return a `ConfigFieldMapping` that maps fields using `overrides` if the field is present there and otherwise
    *         uses this mapping.
    */
-  def withExceptions(exceptions: (String, String)*) =
-    ConfigFieldMapping(exceptions.toMap.withDefault(apply))
+  def withOverrides(overrides: (String, String)*) =
+    ConfigFieldMapping(overrides.toMap.withDefault(apply))
 }
 
 object ConfigFieldMapping {
--- a/core/src/main/scala/pureconfig/ConfigFieldMapping.scala
+++ b/core/src/main/scala/pureconfig/ConfigFieldMapping.scala
@@ -1,10 +1,27 @@
 package pureconfig
 
+/**
+ * A mapping between case class fields and their respective keys in the config.
+ */
 trait ConfigFieldMapping extends (String => String) {
+
   def apply(fieldName: String): String
+
+  /**
+   * Returns a `ConfigFieldMapping` that uses this mapping with some exceptions.
+   *
+   * @param exceptions the exceptions for this mapping as pairs (field, configKey)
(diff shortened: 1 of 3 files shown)


melrief (approved):

melrief:
Not much to comment, nice and useful addition. Are you planning to add a README part about this somewhere in another PR or you think it's not worth it? I would like to have this somewhere in the README.

ruippeixotog:
I'll add this to the README later, it is surely useful to have it there.

jcazevedo (approved):
Looks good to merge without reservations. It's a very helpful addition to the library.

core/src/main/scala/pureconfig/ConfigFieldMapping.scala
```diff
@@ -1,10 +1,27 @@
...
+
+  /**
+   * Returns a `ConfigFieldMapping` that uses this mapping with some exceptions.
+   *
+   * @param exceptions the exceptions for this mapping as pairs (field, configKey)
+   * @return a `ConfigFieldMapping` that maps fields using `exceptions` if the field is present there and otherwise
+   *         uses this mapping.
+   */
+  def withExceptions(exceptions: (String, String)*) = {
+    val map = exceptions.toMap
```
leifwickland: A slightly terser take:
```
ConfigFieldMapping(exceptions.toMap.withDefault(apply))
```

leifwickland (approved):
This is a very nice addition. Thanks, @ruippeixotog.
I made a small suggestion. Take it or leave it.

core/src/main/scala/pureconfig/ConfigFieldMapping.scala
ruippeixotog: Nicely done, I didn't think of using `withDefault` here :) Changed the method to your suggestion in 213551c628eb0752cc294c6463e25634cbcb7589.

core/src/main/scala/pureconfig/ConfigFieldMapping.scala
```diff
@@ -1,10 +1,27 @@
...
+ * A mapping between case class fields and their respective keys in the config.
+ */
 trait ConfigFieldMapping extends (String => String) {
+
   def apply(fieldName: String): String
+
+  /**
+   * Returns a `ConfigFieldMapping` that uses this mapping with some exceptions.
+   *
+   * @param exceptions the exceptions for this mapping as pairs (field, configKey)
```
leifwickland: @ruippeixotog It might be worth adding a comment explaining that any override will be applied globally throughout the configuration, e.g. if a `("foo", "bar")` exception is provided it will affect all fields named `foo` regardless of which object they're in and will affect all `bar` config keys regardless of the path leading up to the `.bar` part.

ruippeixotog: I think the distinction between `ConfigFieldMapping` and a `ProductHint` provides that information. A `ConfigFieldMapping` has no type parameter and has no effect as an implicit. On the other hand, a `ProductHint` has a type parameter restricting the hint to certain types (possibly one or all). Different product hints may be configured with different mappings, so "any override will be applied globally throughout the configuration" isn't entirely accurate. Do you still think it would be better to improve the documentation?

derekmorr: This may seem nitpicky, but I wonder about the naming. `Exceptions` makes me think this will throw Exceptions. Maybe something like name overrides?

leifwickland: @ruippeixotog I agree that I was wrong when I said "globally throughout the configuration." I didn't realize the scope that the `ConfigFieldMapping` was applied within was a `ProductHint`. (I probably should, but I didn't.)

Probably nothing needs to be added. I'm skeptical I could inform without confusing.

Sorry.

ruippeixotog: @leifwickland I understand the confusion, in older pureconfig versions `ConfigFieldMapping` was the object to provide implicitly to configure product derivation :) Since `ProductHint`s were added, `ConfigFieldMapping` was made only one of the multiple things you can configure via a `ProductHint`.

@derekmorr yeah, I wonder about that too. I don't think that the current name is perfect, but I'm not sure if `withOverrides` isn't less explicit/correct in the context of a mapping... I ended up using `withExceptions`. @leifwickland, @melrief, @jcazevedo, what do you think?

leifwickland: @ruippeixotog I'd probably go with `withOverrides`.

jcazevedo: I prefer `withExceptions`. They're effectively exceptions to the general mapping rule. I think any confusion users might have that this will throw exceptions can be dismissed by looking at the method's documentation.

ruippeixotog: @melrief, what do you think about this? We need a fifth opinion here :)

melrief: I prefer `overrides`. While I agree on the meaning of exception in this case, I also think the name can be confusing. I like override because it can be seen as "override the mapping behavior for the given fields"

ruippeixotog: All right, renamed to `withOverrides` in 274b3de3591a8ed9a948374af268d0f3aafbee0f! :)

PR #260, opened 2017-06-04
Add a reason to CannotReadFile and catch all read errors

### LINKED ISSUE #248: Improve failure description of CannotReadFile

ruippeixotog:
This PR improves the handling of file reading errors by preventing `ConfigFactory` from silently ignoring missing files in `parseFile` and dealing with the resulting `ConfigException.IO` exceptions in a special way.

Incidentally, this also makes pureconfig turn all reading IO exceptions into `CannotReadFile` failures, instead of wrapping them into generic `ThrowableFailure` failures as before.

Closes #248.

### DIFF
--- a/core/src/test/scala/pureconfig/ConfigReaderMatchers.scala
+++ b/core/src/test/scala/pureconfig/ConfigReaderMatchers.scala
@@ -3,7 +3,7 @@ package pureconfig
 import scala.reflect.ClassTag
 
 import org.scalatest._
-import org.scalatest.matchers.Matcher
+import org.scalatest.matchers.{ MatchResult, Matcher }
 import pureconfig.error.{ ConfigReaderFailure, ConfigReaderFailures }
 
 trait ConfigReaderMatchers { this: FlatSpec with Matchers =>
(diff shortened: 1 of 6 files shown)


jcazevedo (changes requested):
The implementation looks good to me! Thanks for this!

We should also update `ConfigReaderException` to add representation for `CannotReadFile` failures. We previously only had `CannotParse` as a failure without a path, but `CannotReadFile` also falls under that category.

ruippeixotog:
@jcazevedo you're right, I was not aware that we had special support for `CannotParse`. I generalized the construction of the exception message in cbb5d8b.

jcazevedo (approved):

core/src/test/scala/pureconfig/ConfigReaderMatchers.scala
```diff
@@ -13,4 +13,12 @@ trait ConfigReaderMatchers { this: FlatSpec with Matchers =>
 
   def failWithType[Failure <: ConfigReaderFailure: ClassTag]: Matcher[Either[ConfigReaderFailures, Any]] =
     matchPattern { case Left(ConfigReaderFailures(_: Failure, Nil)) => }
+
+  def failLike(pf: PartialFunction[ConfigReaderFailure, MatchResult]) =
+    new Matcher[Either[ConfigReaderFailures, Any]] with Inside with PartialFunctionValues {
+
+      def apply(left: Either[ConfigReaderFailures, Any]): MatchResult = {
+        inside(left) { case Left(ConfigReaderFailures(failure, Nil)) => pf.valueAt(failure) }
```
melrief, line 21: Nice!

melrief (approved):

PR #262, opened 2017-06-04
Restore the previous behavior of `loadConfigFromFiles`

### LINKED ISSUE #261: Restore the previous behavior of `loadConfigFromFiles`
PR #245 broke the contract of `loadConfigFromFiles`:

```scala
  /**
   * Loads `files` in order, allowing values in later files to backstop missing values from prior, and converts them into a `Config`.
   *
   * This is a convenience method which enables having default configuration which backstops local configuration.
   *
   * Note: If an element of `files` references a file which doesn't exist or can't be read, it will silently be ignored.
   *
   */
  def loadConfigFromFiles[Config: ConfigReader](files: Traversable[Path]): Either[ConfigReaderFailures, Config]
```

We should restore it, as the description mentions a valid use case for the previous behavior that could have been followed by many users.

ruippeixotog:
Closes #261.

### DIFF
--- a/core/src/main/scala/pureconfig/package.scala
+++ b/core/src/main/scala/pureconfig/package.scala
@@ -278,7 +278,7 @@ package object pureconfig {
       files
         .map(parseFile)
         .foldLeft[Either[ConfigReaderFailures, Seq[TypesafeConfig]]](Right(Seq())) {
-          case (c1, Left(ConfigReaderFailures(_: CannotReadFile, Nil))) => c1
+          case (c1, Left(failures)) if failures.toList.exists(_.isInstanceOf[CannotReadFile]) => c1
           case (c1, c2) => ConfigConvert.combineResults(c1, c2)(_ :+ _)
         }
         .right.map(_.reduce(_.withFallback(_)).resolve)
--- a/core/src/main/scala/pureconfig/package.scala
+++ b/core/src/main/scala/pureconfig/package.scala
@@ -8,7 +8,7 @@ import java.io.{ OutputStream, PrintStream }
 import java.nio.file.{ Files, Path }
 
 import com.typesafe.config.{ ConfigException, Config => TypesafeConfig }
-import pureconfig.error.{ ConfigReaderException, ConfigReaderFailures, ConfigValueLocation, KeyNotFound }
+import pureconfig.error._
 import pureconfig.ConfigConvert.improveFailures
 import pureconfig.backend.ConfigFactoryWrapper.{ invalidateCaches, load, loadFile, parseFile }
 import scala.reflect.ClassTag
@@ -278,8 +278,8 @@ package object pureconfig {
       files
         .map(parseFile)
         .foldLeft[Either[ConfigReaderFailures, Seq[TypesafeConfig]]](Right(Seq())) {
-          case (c1, c2) =>
-            ConfigConvert.combineResults(c1, c2)(_ :+ _)
+          case (c1, Left(ConfigReaderFailures(_: CannotReadFile, Nil))) => c1
+          case (c1, c2) => ConfigConvert.combineResults(c1, c2)(_ :+ _)
         }
         .right.map(_.reduce(_.withFallback(_)).resolve)
         .right.flatMap(loadConfig[Config])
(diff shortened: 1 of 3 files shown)


core/src/test/scala/pureconfig/ApiSuite.scala
```diff
@@ -128,9 +121,16 @@ class ApiSuite extends BaseSuite {
     loadConfigFromFiles[Conf](files) should failWithType[ThrowableFailure]
   }
 
-  it should "fail if any of the files does not exist" in {
-    case class Conf(f: Float)
-    val files = listResourcesFromNames("/conf/loadConfigFromFiles/priority1.conf") :+ nonExistingPath
-    loadConfigFromFiles[Conf](files) should failWithType[CannotReadFile]
+  it should "ignore files that doesn't exist" in {
```
leifwickland: Nitpick: `s/doesn't/don't/`

core/src/test/scala/pureconfig/ApiSuite.scala
```diff
@@ -128,9 +121,16 @@ class ApiSuite extends BaseSuite {
...
+    case class Conf(b: Boolean, d: Double)
+    val files = listResourcesFromNames("/conf/loadConfigFromFiles/priority2.conf") :+ nonExistingPath
+    loadConfigFromFiles[Conf](files) shouldBe Right(Conf(false, 0.001D))
+  }
+
+  "loadConfigWithFallback" should "fallback if no config keys are found" in {
+    case class Conf(f: Float, o: Option[Int], d: Double)
+    val priority1Conf = ConfigFactory.load("conf/loadConfigFromFiles/priority1.conf")
+    // first wo are in priority1.conf, the d is in reference.conf
```
leifwickland: What's `wo` in the comment?

core/src/main/scala/pureconfig/package.scala
```diff
@@ -278,8 +278,8 @@ package object pureconfig {
       files
         .map(parseFile)
         .foldLeft[Either[ConfigReaderFailures, Seq[TypesafeConfig]]](Right(Seq())) {
-          case (c1, c2) =>
-            ConfigConvert.combineResults(c1, c2)(_ :+ _)
+          case (c1, Left(ConfigReaderFailures(_: CannotReadFile, Nil))) => c1
```
leifwickland: Do you think we should look for `CannotRedFile` anywhere in the list of failures rather than only exactly a list of that one failure? It seems a bit more robust to me, if highly paranoid.

leifwickland (changes requested):
@ruippeixotog Thanks for doing this. Would you be willing to respond to the questions I asked, please?

leifwickland:
Would you be willing to add a line to the changelog mentioning the bug fix, please?

core/src/test/scala/pureconfig/ApiSuite.scala
ruippeixotog: Fixed in 2ab02930ff0733fd5c593241fad84bf3e8f02654.

core/src/test/scala/pureconfig/ApiSuite.scala
ruippeixotog: Good question... it was copy-pasted from [here](https://github.com/pureconfig/pureconfig/pull/262/files#diff-6e4d5f4a2619b22fc36712e8d1c9ac09L115). I think it means "two". I clarified that comment in 2ab02930ff0733fd5c593241fad84bf3e8f02654.

core/src/main/scala/pureconfig/package.scala
ruippeixotog: I cannot think of any case where a `CannotReadFile` can be accompanied by another failure, but logically speaking it makes sense to check all failures. I changed the behavior to what you suggested in ea17803c2f79049885867f4961ee054496630755.

ruippeixotog:
I have added this fix to the CHANGELOG in 49c82112ea38f90fd0259905d7b5abe6ee475ef4.

@melrief, it may make sense for us to release the next version soon after this PR is merged, since it's a fix for a breaking change introduced in v0.7.1.

core/src/test/scala/pureconfig/ApiSuite.scala
leifwickland: The amusing bit that the original typo is almost certainly mine. :-)
Thanks for fixing that.

leifwickland (approved):
@ruippeixotog Thanks for fixing this!
LGTM 👍

derekmorr (approved):

Issue #263, opened 2017-06-08
Unexpected behavior with Option


kellen:
I'm seeing some inconsistent and unexpected behavior with `Option`.

When a field is optional in a config class, it defaults to `None`, so this works:

```
import com.typesafe.config.ConfigFactory
import pureconfig._

val emptyConfig = ConfigFactory.parseString("")

case class A(a: Option[Int])
assert(A(None) == loadConfigOrThrow[A](emptyConfig))
```

If, however, you move the `Option` into the parameters for `loadConfigOrThrow`, no such defaulting occurs, and the load fails with an exception:

```
case class B(a: Int)
assert(None == loadConfigOrThrow[Option[B]](emptyConfig, "b")) 
// fails with exception
```

Also, defaults are not respected:

```
case class C(a: Option[Int] = Some(10))
assert(C(Some(10)) == loadConfigOrThrow[C](emptyConfig))
// fails, a == None
```

For case `B`, it appears a top-level check would work, something like: 
```
def handleOption[T: ConfigReader : ClassTag](
  config: Config, 
  path: String
): Either[ConfigReaderFailures, T] = {
  val reader = implicitly[ConfigReader[T]]
  val isOption = classOf[Option[_]] == implicitly[ClassTag[T]].runtimeClass
  reader.from(
    if(isOption && !config.hasPath(path)) null else config.getValue(path)
  )
}
```

ruippeixotog:
Hi @kellen! Thanks for submitting this. They are both interesting corner cases - I guess we didn't give much thought to them before.

First of all, note that pureconfig does not treat `Option[T]` as a special type - there are no explicit checks if a type is an option to allow missing keys. There is an `AllowMissingKeys` trait that any `ConfigConvert` instance can extend in order to get full control of what to return in such a case.

That said, case `B` is pushing the boundaries of what a namespace can and should do. When `loadConfigOrThrow` is called with a namespace, pureconfig looks first for the sub-config to read, and only after that is the appropriate `ConfigConvert` used. When a namespace is specified, it is implied to be mandatory; the type that a user requests does not change that. While I'm tending to prefer the current behavior, I agree that this need to be well documented.

On the other hand, the case with `C` seems more clearly a bug. The default values of the case class should have precedence over the "presence" status of the key, in my opinion.

kellen:
[...]

PR #264, opened 2017-06-08
Prefer default values every time a key is null

ruippeixotog:
With regard to #263, this PR makes the derived product readers prefer default values every time a `null` key is found, even when the reader is an instance of `AllowMissingKeys`.

While this takes some power away from `AllowMissingKeys` readers and it would be acceptable to define the behavior to be the other way around, I think that this behavior is more useful for `Option` and for any null-enabled reader I can think of.

### DIFF
--- a/core/src/test/scala/pureconfig/ProductConvertersSuite.scala
+++ b/core/src/test/scala/pureconfig/ProductConvertersSuite.scala
@@ -112,5 +112,8 @@ class ProductConvertersSuite extends BaseSuite {
 
     val conf7 = ConfigFactory.parseMap(Map("a" -> 2, "c" -> 50, "e" -> 1).asJava).root()
     ConfigConvert[Conf].from(conf7).right.value shouldBe Conf(2, "default", 50, InnerConf(43, 44), Some(1))
+
+    val conf8 = ConfigFactory.parseMap(Map("a" -> 2, "c" -> 50, "e" -> null).asJava).root()
+    ConfigConvert[Conf].from(conf8).right.value shouldBe Conf(2, "default", 50, InnerConf(43, 44), None)
   }
 }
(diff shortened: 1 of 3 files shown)


core/src/test/scala/pureconfig/ProductConvertersSuite.scala
```diff
@@ -103,11 +103,14 @@ class ProductConvertersSuite extends BaseSuite {
...
+    ConfigConvert[Conf].from(conf5).right.value shouldBe Conf(2, "default", 42, InnerConf(5, 6), Some(45))
 
     val conf6 = ConfigFactory.parseMap(Map("a" -> 2, "d" -> "notAnInnerConf").asJava).root()
     val failures = ConfigConvert[Conf].from(conf6).left.value.toList
     failures should have size 1
     failures.head shouldBe a[WrongType]
+
+    val conf7 = ConfigFactory.parseMap(Map("a" -> 2, "c" -> 50, "e" -> 1).asJava).root()
+    ConfigConvert[Conf].from(conf7).right.value shouldBe Conf(2, "default", 50, InnerConf(43, 44), Some(1))
   }
```
leifwickland: Do we have a test that proves the config can set a `None` overriding a default `Some`? I didn't see one. Something like:
```
    val conf8 = ConfigFactory.parseMap(Map("a" -> 2, "c" -> 50, "e" -> null).asJava).root()
    ConfigConvert[Conf].from(conf8).right.value shouldBe Conf(2, "default", 50, InnerConf(43, 44), None)
```

If we don't have one, please consider adding one.

leifwickland (approved):
The behavior the tests describe looks right to me. And the behavior that Kellen wants seems correct to me.

I don't understand the guts of the change well enough to have an opinion there.

melrief (approved):

core/src/test/scala/pureconfig/ProductConvertersSuite.scala
ruippeixotog: Good idea. I added that test in a2bc6733650155397d8ef3d618ae51eb6593990c.

PR #275, opened 2017-06-24
Allow reading config values with a namespace to any value with a ConfigReader

### LINKED ISSUE #263: Unexpected behavior with Option

ruippeixotog:
Replaces the algorithm for loading configs with namespaces as described in #263. It also fixes some bugs:
- `loadConfig` no longer throws exceptions when the namespace does not represent a config object;
- there were some cases where `improveFailures` was not being called after a `ConfigReader` was used. Now all code paths pass through `improveFailures`.

~~Since we want to navigate through the config one hop at a time, I needed a way to parse the string namespace into a proper path. The process is not straightforward, as Typesafe Config accepts special chars and escape sequences in their paths. It employs internally a [parser](https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/impl/Path.java) with a full fledged [tokenizer](https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/impl/Tokenizer.java), both of which are unfortunately package private. After deliberating a bit, I decided that it would be worth to use the object anyway, as I prefer to rely on their internal code (the library is not updated that much) than to replicate their logic or stop being fully compatible with their paths. Let me know if you disagree with this, any suggestions are welcome!~~ (thanks for finding a solution to this, @jcazevedo) 

Closes #263.


jcazevedo (changes requested):
Looks good to me! The improved test suite should help clarify the behaviour of `loadConfig` with a namespace.

I think you can use [`ConfigUtil`](https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigUtil.html) to access the internal parser for namespaces, thus avoiding direct usage of package private code.

ruippeixotog:
@jcazevedo I didn't know about `ConfigUtil`, great! I changed the wrapper to use it in f8e0a6f (the wrapper is still useful because the Typesafe Config parser doesn't deal with empty lists and path expressions, which I conveniently use to denote the root).

jcazevedo (approved):
Looks good to me. Thanks for this!

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 ### 0.7.3 (unreleased)
 
 - New features
+  - `loadConfig` methods now allow loading any type from a config when using a namespace, and not only types represented
+    by config objects;
   - `ConfigFieldMapping` now has a `withOverrides` method that allows users to easily define exceptional cases to an
     existing mapping;
-  - `ConfigReader` and `ConfigWriter` for `java.math.BigDecimal`, `java.math.BigInteger`, 
+  - `ConfigReader` and `ConfigWriter` for `java.math.BigDecimal` and `java.math.BigInteger`.
```
melrief, line 8: You are good in writing changelogs 👍

core/src/main/scala/pureconfig/package.scala
```diff
@@ -7,35 +7,63 @@
...
+            case childCv =>
+              getValue(childCv, remaining, key :: curr)
+          }
+
+        case _ => fail(WrongType(
+          cv.valueType, Set(ConfigValueType.OBJECT), ConfigValueLocation(cv.origin()), joinPath(curr.reverse)))
+      }
+    }
+
+    // we're not expecting any exception here, this `try` is just for extra safety
```
melrief, line 43: Good idea

melrief (approved):
Great PR, in particular the tests are a semi-specification of the different errors that we expect. Thanks for doing this.

PR #276, opened 2017-06-24
Add an extension method to convert ConfigReaderFailures into a NonEmptyList

ruippeixotog:


derekmorr (approved):

derekmorr:
LGTM. I was toying with this is too, since ConfigReaderFailures is isomorphic to NonEmptyList.

melrief (changes requested):
LGTM but you may want to add the change to the CHANGELOG

ruippeixotog:
@melrief this was added to `pureconfig-cats` and not to the core, and our "policy" has been not to track changes in the modules in the CHANGELOG until now.

melrief (approved):
@ruippeixotog right but we had a discussion about adding the CHANGELOG for the modules. I'll approve this as it is but I think we should consider adding a CHANGELOG to the module pureconfig-cats with the changes for the next version.

ruippeixotog:
I agree that we should have a change log for the modules, we just need to decide if we want a separate CHANGELOG file for each module or a single one for everything :)

Issue #205, opened 2017-04-20
[error] knownDirectSubclasses of ... observed before subclass ... registered


drewhk:
I have a simple ADT setup:

```
sealed trait KeyStoreLocation
  final case class Resource(path: String) extends KeyStoreLocation
  final case class File(path: String)     extends KeyStoreLocation
```

This type is used as a member of another case-class that I load via pureconfig. It seemed to work, but then after a while it stopped compiling:

```
[error] knownDirectSubclasses of KeyStoreLocation observed before subclass Resource registered
[error] knownDirectSubclasses of KeyStoreLocation observed before subclass File registered
```

What is weird that I have not changed any of the classes involved here! It seems like a compilation order issue to me, but I don't know how to fix it.

tkroman:
The workaround is to manually define the problematic implicit (the one this macro is trying to create). I've done this with upickle, which provides a way to do this semi-automatically, where you don't need to write the actual code but just specify the subclass parts for the root implicit (http://www.lihaoyi.com/upickle-pprint/upickle/#ManualSealedTraitPicklers). You could check if something like this might work for your case.

drewhk:
I assumed also that it can work but I was not able to figure out how I can manually create the implicit.

melrief:
Hey Drewhk, thanks for opening the issue. I'm trying to reproduce your error but I'm unable to trigger it.

The steps that I've followed are:

I opened a test REPL with

```
sbt core/test:console
```

I created the sealed trait:

```scala
@ {
    sealed trait KeyStoreLocation
    final case class File(path: String)     extends KeyStoreLocation
    final case class Resource(path: String) extends KeyStoreLocation
  }
```

Then checked if the `ConfigReader` was available, which it is:

```scala
@ pureconfig.ConfigReader[KeyStoreLocation] 
res3: pureconfig.ConfigReader[KeyStoreLocation] = pureconfig.DerivedReaders$$anon$8@14101287
```

Then I defined a class with a field of type `KeyStoreLocation` and checked again the `ConfigReader`, with no problems:

```scala
@ case class Foo(k: KeyStoreLocation) 
@ pureconfig.ConfigReader[Foo] 
res5: pureconfig.ConfigReader[Foo] = pureconfig.DerivedReaders$$anon$7@50d81ad9
```

Then tried to load the configuration via `loadConfig`, which worked:

```scala
@ import com.typesafe.config.ConfigFactory 
@ import pureconfig._ 
[...]

PR #278, opened 2017-06-26
Add failing test for #205

sirthias:
Adds a test illustrating #205.

If the three case objects are moved out of the `CarMaker` companion everything works as expected.

If they are in the companion compilation fails with these errors:

```
[error] knownDirectSubclasses of CarMaker observed before subclass BMW registered
[error] .../pureconfig/core/src/test/scala/pureconfig/CoproductConvertersSuite.scala:27: could not find implicit value for parameter conv: pureconfig.ConfigConvert[pureconfig.CarMaker]
[error]     ConfigConvert[CarMaker].from(conf.root()) shouldEqual Right(BMW)
[error]                  ^
[error] knownDirectSubclasses of CarMaker observed before subclass Mercedes registered
[error] knownDirectSubclasses of CarMaker observed before subclass Tesla registered
```

### DIFF
--- a/core/src/test/scala/pureconfig/CoproductConvertersSuite.scala
+++ b/core/src/test/scala/pureconfig/CoproductConvertersSuite.scala
@@ -21,6 +21,12 @@ class CoproductConvertersSuite extends BaseSuite {
     ConfigConvert[AnimalConfig].from(conf.root()) shouldEqual Right(DogConfig(2))
   }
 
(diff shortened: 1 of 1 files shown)


ruippeixotog:
This seems directly related to https://github.com/scala/bug/issues/10330, as both Circe and PureConfig depend on shapeless to to derive instances for sealed traits (which in turn relies on `knownDirectSubclasses`).

PR #279, opened 2017-06-27
Better implicit not found messages during compilation

ruippeixotog:
This pull request is a WIP to show you the current status of my solution for improving the error messages when an implicit is missing. As I mentioned earlier, better error messages are enabled by wrapping the implicit for which we want them in a new `Derivation` type class, much like we do with `Lazy`.

The `Derivation` must be the root implicit asked for; otherwise, the implicit not found error will be the default one provided by the compiler. During a `Derivation`, every "inner" `Derivation` found by the compiler during the implicit search will be a step showed in the "stack trace" of implicits tracked and shown if an error occurs. [DerivationSuite.scala](https://github.com/pureconfig/pureconfig/blob/better-implicit-not-found-msgs/core/src/test/scala-2.11%2B/pureconfig/DerivationSuite.scala) shows the current style of the errors.

For the sake of the reviewers and also for the sake of the future @ruippeixotog, I have commented [Derivation.scala](https://github.com/pureconfig/pureconfig/blob/better-implicit-not-found-msgs/macros/src/main/scala-2.11%2B/pureconfig/Derivation.scala) very thoroughly. Please take a look and let me know if you have any questions or suggestions about the inner workings of the macro.

As almost all macro code I've seen in Scala, there are messy parts. In particular:
- The idea to set the implicit not found error was taken from `Lazy`. While it accesses internal macro code, I do not expect the code to break without warning as shapeless also depends on it;
- ~~Mutable state and type casting on the path dependent types is commonly seen in Scala macros. At least the type casting can be avoided with some clever containers, but I do not think it's critical to change for now;~~ (not needed in the new algorithm)
- ~~What worries me most is my hack to `inferImplicitValue`, whose current behavior I do not understand until now. I have some ideas to avoid using `cc.universe.analyzer.inferImplicit`, which I'll try soon.~~ Fixed in 2.12.3 with https://github.com/scala/scala/pull/5971. Since we still need to support 2.11.x and 2.12.0 to 2.12.2, I'll leave it as it is and keep an eye on new Scala versions.

I have some ideas to allow customizing the messages according to the concrete structure being derived (product, coproduct, collection...), but I think it's better to leave that to a second development stage.

Finally, since `Derivation` ended up being a general-purpose macro suitable for the derivation of any type-class, I'm considering maintaining this macro as a project separate from pureconfig, maybe even asking if shapeless wants to take this. That way more people would be able to benefit from it and maybe it would get more eyes looking for bugs and improvements.


jcazevedo:
This is amazing! Thanks for your work on this! 🎉 

I might need some time to properly review this, but I agree that you should propose this to Shapeless, as it is very helpful for code that relies on automatic derivation of type classes.

melrief (commented):
I've read the entire PR but I have to admit that while I understand the derivation abstract idea, some of the macro implementation details are difficult to understand. By looking at the tests and by trying manually, this works really really well. It's also very nice to have our own `illTyped`. Good job :).
I may need a bit more time to understand the code. @ruippeixotog, sorry to make you wait too much. I'll come back with the approval as soon as I'm confident it works and why. It will be in the next days for sure.

ruippeixotog:
Thank you for taking the time to review this, @melrief :) There's no rush, the macro isn't even ready for merging yet. If you have any question about the implementation, by all means ask! Macro code is somewhat difficult to follow and debug.

Meanwhile, I'm working on https://github.com/scala/scala/pull/5971 in order to get `inferImplicitValue` to work for this use case. The problem turns out to be that the compiler is wrongly declaring divergence in some cases of implicit searches triggered by macros.

core/src/main/scala/pureconfig/DerivedReaders.scala
```diff
@@ -40,17 +40,7 @@ trait DerivedReaders extends DerivedReaders1 {
...
-  private[pureconfig] trait WrappedConfigReader[Wrapped, SubRepr] extends ConfigReader[SubRepr]
-
-  private[pureconfig] trait WrappedDefaultValue[Wrapped, SubRepr <: HList, DefaultRepr <: HList] {
-    def fromWithDefault(config: ConfigValue, default: DefaultRepr): Either[ConfigReaderFailures, SubRepr] = config match {
-      case co: ConfigObject => fromConfigObject(co, default)
-      case other => fail(WrongType(other.valueType, Set(ConfigValueType.OBJECT), ConfigValueLocation(other), ""))
-    }
-    def fromConfigObject(co: ConfigObject, default: DefaultRepr): Either[ConfigReaderFailures, SubRepr]
-  }
+trait DerivedReaders1 extends DerivedReaders2 {
```
leifwickland: Nitpick: These seem like poor names. I'll admit I hardly understand what's going on here, so it could be that it's simply my ignorance speaking.

ruippeixotog: It's a pattern commonly used in libraries like scalaz and cats to deal with several layers of implicits. That means the implicits in `DerivedReaders` get higher priority, followed by `DerivedReaders1` (since `DerivedReaders` includes `DerivedReaders1`), followed by `DerivedReaders2`. They could be `DerivedReaders`, `LowerPriorityDerivedReaders` and `LowererPriorityDerivedReaders` :)

I had to do this because both `deriveOption` and `deriveCoproductInstance` can be used to derive a reader for `Option`, and the macro would see both of them as equal choices.

melrief: @leifwickland this kind of naming for different priority traits is common in scala libraries

melrief:
@ruippeixotog just FYI, the links to `Derivation.scala` and `DerivationSuite.scala` in the first PR message are broken.

ruippeixotog:
When trying to add `Derivation` around `ConfigConvert`, I found out the reason for [this bug](https://github.com/pureconfig/pureconfig/pull/279/files#diff-e9aaf3890bef489a3ba03f77714eef87R92) and unfortunately it is more critical than I expected, since it may make good code fail when wrapped in a `Derivation`.

After some thought, I realized that a macro like this is bound to fail when deriving some stranger kinds of implicit macros. Since a macro may do close to everything, from rendering any kind of Scala AST to triggering any number of implicit searches (including `Derivation` ones), there is no way to make a macro like this fool-proof when used in tandem with others. The current implementation is actually my second major one - the first one having been shot down by `Lazy`, which is a macro as "strange" as it can be for `Derivation` purposes.

I have an idea to fix that issue, which I'll try to implement next week. The general idea is to have special support for `Lazy` (since we really need it for derivations) and providing no guarantees for any other kind of macros (the most common ones, which just rewrite expressions for reducing boilerplate, most probably will still work).

I think it is not worth it to review this for now. While many parts will remain unchanged after I try the new approach, you would be reviewing some parts which would have to be changed.

melrief:
ack @ruippeixotog, thanks for doing this.

ruippeixotog:
Well, I think this is finally ready for some serious reviewing. Since my last comment I changed the algorithm for extracting failed paths and now everything seems to be working as expected. I actually have more confidence in the current implementation now and I feel this is robust enough to be merged as an experimental feature.

The only downside from the new approach is that we now treat `Lazy` in a special way, doing some implementation-specific pattern matching to extract the correct hierarchies from there. I don't think the `Lazy` internals will change that much in the future, so I'm not too worried. However, if it ever happens it should be easy enough to update pureconfig.

From the point of view of a pureconfig user, a summary of the relevant aspects of `Derivation`:

* `Derivation` is a feature hidden behind the compiler flag `-Xmacro-settings:materialize-derivations` and available from Scala 2.11 upwards. If that flag is not passed or Scala 2.10 is being used, the `Derivation` class will still be available (so the users' code is always compatible) but the macro falls back to a regular implicit search;
* It may not work properly if some of the derivation steps are generated by implicit macros. It has special support for `Lazy` and most simple macros should work without a problem. However, `Derivation` may have trouble with implicit macros that do implicit searches for `Derivation` or severely mess up the syntax tree hierarchy;
* It may be used with custom code for loading configs. In order to require an implicit `ConfigReader[A]` with better derivation failures, it is enough to ask for a `Derivation[ConfigReader[A]]` instead. The single `value` method of `Derivation` returns the requested `ConfigReader[A]`;
* It may be used in custom readers and writers in the same way. If a custom reader requires another class as an implicit and it should be shown as a derivation step, it is enough to require a `Derivation` of that class instead. The macro code prints out all the failed derivations it can find in the generated implicit tree.

I'll take the time later to write this as documentation for the README. Right now I can't look at this macro anymore 😄

core/src/main/scala/pureconfig/DerivedReaders.scala
ruippeixotog: @leifwickland After the latest algorithm revamp, I think this separation isn't needed anymore (the macro seeing them as equal was a smell a of a bigger problem which I had to fix). I think I'm going to revert this in order to reduce complexity here.

melrief: @ruippeixotog the separation is still here, do you want to fix it or we keep it like this?

core/src/main/scala/pureconfig/package.scala
```diff
@@ -19,6 +19,9 @@ import pureconfig.error._
 
 package object pureconfig {
 
+  // an alias for a `Derivation` of a `ConfigReader`
+  private type DerivedConfigReader[A] = Derivation[ConfigReader[A]]
```
melrief: An optional comment that you can ignore if you want: this alias doesn't buy you anything but makes the code more obscure to read. I would just delete it and use `Derivation[ConfigReader[A]]` everywhere.

macros/src/main/scala-2.11+/pureconfig/Derivation.scala
```diff
@@ -0,0 +1,196 @@
...
+
+import scala.collection.mutable
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+
+import pureconfig.derivation._
+
+sealed trait Derivation[A] {
+  def value: A
+}
```
melrief, line 11: This comment is really not important but I'm curious: why `Derivation` is not in a shared package that is then used by all the scala versions?

melrief (approved):
It took a hour to review but I think I understand the code and LGTM. Nice touch the scalac option to enable/disable this feature. I'm looking forward to see the feedback from the users.

core/src/main/scala/pureconfig/package.scala
ruippeixotog: I guess I was lazy there :) without the alias I'd have to replace the syntatic sugar of writing `def loadConfig[Config: DerivedConfigReader]` with `def loadConfig[Config](implicit reader: Derivation[ConfigReader[Config]])`. While I prefer the shorter syntax I agree that it makes the code more obscure to read for users. I'll change it.

macros/src/main/scala-2.11+/pureconfig/Derivation.scala
ruippeixotog: I didn't do it because the companion object and the trait have to be in the same file, and the companion objects of the two versions are different. Moreover, I wouldn't be able to make the trait `sealed` (unless I also defined `Failed` in 2.10).

core/src/main/scala/pureconfig/package.scala
ruippeixotog: Fixed in 08fa51e.

core/src/main/scala/pureconfig/DerivedReaders.scala
ruippeixotog: Fixed in 12d0082 and fd79b6c.

macros/src/main/scala-2.11+/pureconfig/derivation/MacroCompat.scala
```diff
@@ -0,0 +1,37 @@
...
+  // to make the compiler print our custom message. That's done by setting a @implicitNotFound annotation on our
+  // `Derivation` class (idea taken from shapeless `Lazy`).
+  def setImplicitNotFound(msg: String): Unit = {
+    import c.internal.decorators._
+    val infTree = c.typecheck(q"""new _root_.scala.annotation.implicitNotFound($msg)""", silent = false)
+    typeOf[Derivation[_]].typeSymbol.setAnnotations(Annotation(infTree))
+  }
+
+  // This should be simply defined as `c.inferImplicitValue(c.weakTypeOf[A])`, but divergent implicits are wrongly
+  // reported up to Scala 2.12.2. See https://github.com/scala/scala-dev/issues/398 for more information.
```
jcazevedo: Can you point here to https://github.com/scala/bug/issues/10398 instead? It seems like a better place to follow the history of the issue and the implementation that solves it.

jcazevedo (approved):
This is a great feature! Thanks!

macros/src/main/scala-2.11+/pureconfig/derivation/MacroCompat.scala
ruippeixotog: Good catch, the link you mentioned is the best one to show the issue. Changed in e3821af.

PR #160, opened 2017-03-09
Join in a Contributors :)

sh0hei:
By https://github.com/melrief/pureconfig/pull/158

PR #152, opened 2017-03-02
Add stsatlantis to CONTRIBUTORS

stsatlantis:

PR #118, opened 2017-02-09
add derekmorr to contributors

derekmorr:
Add me to the contributors list.

PR #280, opened 2017-06-28
Include msg of CannotParse failure into its description.

jbgi:


jcazevedo (changes requested):
Thanks for your PR! I agree it makes sense to include the message of a `CannotParse` failure into its description.

Unfortunately, this way we end up including the location of the failure twice, as the `ConfigException.Parse` exception message includes it and we also include it when we list the failures in the `ConfigReaderException` message. Maybe we can strip the prefix of the `ConfigOrigin`'s description from the `ConfigException.Parse` message (see [ConfigException.java#L23](https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/ConfigException.java#L23)).

Please also update the tests in the `ConfigReaderExceptionSuite` accordingly.

melrief:
@jbgi thanks for the PR! I agree with @jcazevedo about the changes that should be done to get this merged in master. Both the breaking test and the double inclusion of the location should be addressed before we continue.

jbgi:
@jcazevedo, @melrief PR updated with requested changes.

core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
```diff
@@ -34,7 +34,13 @@ object ConfigFactoryWrapper {
   private def unsafeToEither[A](f: => A, path: Option[Path] = None): Either[ConfigReaderFailures, A] = {
     try Right(f) catch {
       case e: ConfigException.IO if path.nonEmpty => fail(CannotReadFile(path.get, Option(e.getCause)))
-      case e: ConfigException.Parse => fail(CannotParse(e.getLocalizedMessage, ConfigValueLocation(e.origin())))
+      case e: ConfigException.Parse =>
+        val msg = (if (e.origin != null)
+          // Removing the error origin from the exception message since origin is stored and used separately:
+          e.getMessage.stripPrefix(s"${e.origin.description}: ")
+        else
+          e.getMessage).stripSuffix(".")
```
melrief, line 42: I'll link the source code of [`ConfigException`](https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/ConfigException.java#L21) in case somebody else wants to check why we need do strip the prefix in this way. `ConfigException` takes in the constructor origin and [appends the description to the exception message](https://github.com/typesafehub/config/blob/master/config/src/main/java/com/typesafe/config/ConfigException.java#L23).

melrief (approved):
LGTM @jbgi, thanks for doing this. I'll wait for another review and then merge.

jcazevedo (approved):

jcazevedo:
Thanks for the PR @jbgi! We would like to add you to the [CONTRIBUTORS](https://github.com/pureconfig/pureconfig/blob/master/CONTRIBUTORS) file. If you agree, please submit another PR in which you add a line with your name, your email (optional) and your username, in line with PRs such as #118, #152 or #160.

jbgi:
Thanks @jcazevedo, however I am fine with only the git log attribution.

PR #281, opened 2017-06-29
Use -Xfatal-warnings just on Travis

ruippeixotog:
This PR removes the `-Xfatal-warnings` flag from the SBT build and adds it only on the Travis build. It's very cumbersome for me to have this flag active during development - when I'm testing stuff I sometimes comment things or move code around and I always end up having to comment or remove the imports too. I should be able to easily compile the code with warnings, as long as I submit it without them.

Let me know what you think of this :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,18 @@
 language: scala
 scala:
-    - 2.10.6
-    - 2.11.11
-    - 2.12.2
+  - 2.10.6
+  - 2.11.11
+  - 2.12.2
 jdk:
-    - oraclejdk8
+  - oraclejdk8
 script:
-  - sbt coverage "+++$TRAVIS_SCALA_VERSION test" "+++$TRAVIS_SCALA_VERSION tut" "+++$TRAVIS_SCALA_VERSION doc"
-  - (cd example ; sbt ++$TRAVIS_SCALA_VERSION test)
+  - sbt -S-Xfatal-warnings
+      coverage
+      "+++$TRAVIS_SCALA_VERSION test"
+      "+++$TRAVIS_SCALA_VERSION tut"
+      "+++$TRAVIS_SCALA_VERSION doc"
+
+  - cd example && sbt -S-Xfatal-warnings ++$TRAVIS_SCALA_VERSION test
 
   # check if there are no changes after `tut` runs
   - if [[ $TRAVIS_SCALA_VERSION =~ ^2\.12.* ]]; then
--- a/build.sbt
+++ b/build.sbt
@@ -80,7 +80,6 @@ lazy val allVersionLintFlags = Seq(
   "-encoding", "UTF-8", // yes, this is 2 args
   "-feature",
   "-unchecked",
-  "-Xfatal-warnings",
   "-Yno-adapted-args",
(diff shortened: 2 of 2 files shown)


leifwickland:
@ruippeixotog <https://github.com/ruippeixotog>, what do you think about
instead disabling the lint flags outside of Jenkins? I think those are what
are causing you the problems more so than fatal warnings. In any case I
will defer to your preferences as you do far more work in the code than I.

On Jun 29, 2017 3:05 AM, "Rui Gonçalves" <notifications@github.com> wrote:

@ruippeixotog <https://github.com/ruippeixotog> requested your review on:
pureconfig/pureconfig#281
<https://github.com/pureconfig/pureconfig/pull/281> Use -Xfatal-warnings
just on Travis.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/pureconfig/pureconfig/pull/281#event-1143817142>, or mute
the thread

jcazevedo (approved):
I've also commented out the `-Xfatal-warnings` flag in the build on some past occasions, so I approve this change. However, I don't mind leaving it in if someone really wants it.

I guess I'm OK either way as long as this is enabled on Travis.

ruippeixotog:
@leifwickland in my own experience, the unused import warning is the worst warning to have as a failure during development. As such, I guess I'd be ok with having other warnings as fatal during development.

But can you think of a warning that you'd like to keep as fatal even during development? Moreover, if we disable linting in development and have it in Travis, we could easily get into a situation where we are to submit a PR without any warning or error locally, only to have Travis failing it unexpectedly later on.

derekmorr:
I find the unused import warning annoying, so I'd be in favor of disabling that. I'd prefer to keep fatal warnings enabled. I've found it useful on other projects, for detecting deprecation warnings or unused code.

ruippeixotog:
@derekmorr are you saying to leave the fatal warnings enabled (without unused imports), even in development? Or are you just talking about the Travis build?

I'm all for leaving all linting options in Travis as errors - even the unused imports one. It's good to require the code clean before sending it to master. I'm only in favor of showing warnings during development as what they are - I don't think there's any value in preventing the compilation of transient code if the compiler can compile that code. Warnings are still shown, so there's not even loss of information or loss of useful hints for the developer...

derekmorr:
Yes, I'm suggesting leaving fatal warnings enabled, even outside of CI. I find it useful, and it helps prevent bad practices from sneaking in.

leifwickland:
That's what I was trying to suggest as well. My goal was avoid the most
irritating of the warnings.

melrief:
I also agree we should leave this option. I prefer to have it anyway. If I'm experimenting, I prefer to disable it locally until I'm ready to re-enable it. @ruippeixotog @jcazevedo what do you think about this?

ruippeixotog:
No problem, I'll close this since there seems to be no way to disable `-Xfatal-warnings` just for one type of warning and disabling the warning locally only will lead to many failed CI builds. While I have some trouble finding arguments other than personal preference in favor of enabling fatal warnings locally, it's no trouble for me to keep my build modified to disable the flag :)

PR #282, opened 2017-07-07
Wrap exceptions in ConfigReaderFailure when using map and xmap

jcazevedo:
This PR modifies the `xmap` method of `ConfigConvert` and the `map` method of `ConfigReader` to wrap exceptions that the functions used to map might throw in a `ConfigReaderFailure`.

### DIFF
--- a/core/src/main/scala/pureconfig/ConvertHelpers.scala
+++ b/core/src/main/scala/pureconfig/ConvertHelpers.scala
@@ -32,6 +32,9 @@ trait ConvertHelpers {
         ConfigReaderFailures(headImproved, tailImproved)
     }
 
+  private[pureconfig] def toResult[A, B](f: A => B): A => Option[ConfigValueLocation] => Either[ConfigReaderFailures, B] =
+    v => l => eitherToResult(tryToEither(Try(f(v)))(l))
+
   private[pureconfig] def eitherToResult[T](either: Either[ConfigReaderFailure, T]): Either[ConfigReaderFailures, T] =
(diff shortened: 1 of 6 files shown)


core/src/main/scala/pureconfig/ConvertHelpers.scala
```diff
@@ -32,6 +32,9 @@ trait ConvertHelpers {
         ConfigReaderFailures(headImproved, tailImproved)
     }
 
+  private[pureconfig] def toResult[A, B](f: A => B): A => Option[ConfigValueLocation] => Either[ConfigReaderFailures, B] =
```
ruippeixotog, line 35: We should probably do a cleanup of these private methods in pureconfig. I did not look at them throughly, but I believe that we can simplify these utility methods to a few core ones. For this PR this is fine though, let's leave it for another time :)

ruippeixotog (approved):
Nice improvement, thanks!

jcazevedo:
> I did not look at them throughly, but I believe that we can simplify these utility methods to a few core ones.

Yeah. Now that we have combinators on `ConfigConvert`, I think a lot of the helpers that deal with `String`s can be simplified to map over a `ConfigConvert[String]`.

core/src/main/scala/pureconfig/ConvertHelpers.scala
melrief: +1

melrief (approved):

PR #283, opened 2017-07-10
Allow periods in Map keys.

LeifW:
The current conversion just uses `ConfigObject` from the `ConfigValue`.

Given e.g. `{ foo.bar = 10 }`, you get `Map(foo -> Map(bar -> 10))`, whereas I think what would be wanted for a `Map[String, Int]` would be  `Map(foo.bar -> 10)`. Kafka takes its config as a `Map[String, Object]` (see https://kafka.apache.org/0100/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html) where the keys are things like `client.id` and `enable.auto.commit`. We have those stored in sections of reference.conf and application.conf, but the current PureConfig `Map` reader doesn't work for this. The `.entrySet` method on `Config` returns the contents flattened out with the keys being the full (period-delimited) path to each leaf node value.

This change seems to pass the tests, except for the case where you're turning a `Map` with an empty string into Config - this `.entrySet` method seems to turn that empty string into `""` (a string that's just a pair of quotes) on the way back. I'm not sure empty string makes sense as a config value - how could an empty string be written as a key in a .conf file? Could restrict those test cases to "non-empty string". Or modify the ConfigWriter behaviour to add the empty quotes in that case, to match the ConfigReader?

Also, PureConfig doesn't support converting to `Object`, but it seems easy enough to just an a `ConfigReader` that calls `.unwrapped` on any `ConfigValue`.

### DIFF
--- a/core/src/main/scala/pureconfig/DerivedReaders.scala
+++ b/core/src/main/scala/pureconfig/DerivedReaders.scala
@@ -183,12 +183,12 @@ trait DerivedReaders1 {
         case co: ConfigObject =>
           val z: Either[ConfigReaderFailures, Map[String, T]] = Right(Map.empty[String, T])
 
-          co.asScala.foldLeft(z) {
-            case (acc, (key, value)) =>
+          co.toConfig.entrySet.asScala.foldLeft(z) {
+            (acc, entry) =>
               combineResults(
                 acc,
-                improveFailures(configConvert.value.from(value), key, ConfigValueLocation(value))) {
-                  (map, valueConverted) => map + (key -> valueConverted)
+                improveFailures(configConvert.value.from(entry.getValue), entry.getKey, ConfigValueLocation(entry.getValue))) {
+                  (map, valueConverted) => map + (entry.getKey -> valueConverted)
                 }
           }
 


ruippeixotog:
Hi @LeifW! Thanks for the PR.

In HOCON (the format used by Typesafe config), a `{ foo.bar = 10 }` is not a map with a "foo.bar" key; the [specification](https://github.com/typesafehub/config/blob/master/HOCON.md#paths-as-keys) explains that path expression keys are semantically equivalent to nested objects (i.e. `{ foo { bar = 10 } }`), which is in turn reflected in the [object merging](https://github.com/typesafehub/config/blob/master/HOCON.md#duplicate-keys-and-object-merging) rules.

If you can afford to edit the config file where those keys are defined, either in HOCON or in property files, you can define them this way:

```scala
scala> val conf = ConfigFactory.parseString("""
     | "foo.bar" = 10
     | """)
conf: com.typesafe.config.Config = Config(SimpleConfigObject({"foo.bar":10}))
```

If you can't afford to do that, having a flattened `ConfigReader` for a `Map[String, T]` is indeed an option and your case seems to be a valid one for it. However, I think this can't be the default behavior for reading `Map` instances. It can be a good idea to have somewhere in pureconfig a set of non-standard readers that are only used if the users import them explicitly.

I'm not fond of having a reader for `Object` in pureconfig just because it is against the idea of having a pure, type-safe configuration. I'd consider having it as a non-standard reader like above, but even that way I wouldn't be happy to have it...

If you really need to read a given key with a value that can be of _any_ type, can you consider reading it as a `ConfigValue` (which is already supported)? At least that type specifies clearly the possible types of values inside it, and you can still do `.mapValues(_.unwrapped)` afterwards.

LeifW:
I didn't know you could quote the keys like that. I'll go ahead and make those edits to our config files. Thanks!

Issue #292, opened 2017-08-01
could not find implicit value


sidhant92:
Hey guys im getting the following error:

could not find implicit value for parameter reader: pureconfig.ConfigReader

And the funny thing is this error is very random. Sometimes its coming sometime its not. What should i do to fix it.

sidhant92:
Okay i noticed that im getting this error whenever my case class includes a variable whose type is a sealted trait having many sub case classes.

leifwickland:
Hi, @sidhant92.

Sorry you're having trouble. 

To encode an ADT like your sealed hierarchy, PureConfig needs a little help because there's no native representation of it. Either you'll need to create an enum for it (see https://github.com/pureconfig/pureconfig/tree/master/modules/enumeratum or  https://github.com/pureconfig/pureconfig/tree/master/modules/enum) or provide a `CoproductHint` that tell PureConfig how you want to encode the members of your ADT. See the test for `EnumCoproductHint` for an example of how to do the latter. https://github.com/pureconfig/pureconfig/blob/67af0decdbef67d64e25e33c8f200a1f626faac3/core/src/test/scala/pureconfig/CoproductHintSuite.scala#L42

Please let me know if that helps.

Leif

sidhant92:
@leifwickland I used this https://github.com/pureconfig/pureconfig/tree/master/modules/enumeratum

But here instead of case object i had case classes.
The implicit issue was solved after this.
But then i started getting this error:
Left(ConfigReaderFailures(CannotConvert.
It cant map it to the correct case class.

sidhant92:
@leifwickland And with no changes done the config is loading absolutely fine. Its just that its not compiling it. 
And once i load the config in console after that it compiles also successfully until i do a clean compile.

leifwickland:
@sidhant92 Apologies. I gave you bad advice. I missed the fact that you had `case class`es, not `object`s. I'm sorry.

This page documents the behavior that should be happening in your case:
https://github.com/pureconfig/pureconfig/blob/master/docs/override-behaviour-for-sealed-families.md

Theoretically, it should compile without anything additional. To successfully read the config, the config element which represents the ADT member would need a `type` member which contained a value equal to the name of the class.

If that's not the behavior you're seeing, please give us a reproducible case we can investigate.

Again, I'm sorry I led you astray originally.

Leif

sidhant92:
[...]

PR #294, opened 2017-08-03
Warn about using old minor versions of Scala.

leifwickland:
Try to help others avoid the pain we're seeing in #292.

### DIFF
--- a/core/src/main/tut/README.md
+++ b/core/src/main/tut/README.md
@@ -88,7 +88,8 @@ libraryDependencies ++= Seq(
 For a full example of `build.sbt` you can have a look at this [build.sbt](https://github.com/melrief/pureconfig/blob/master/example/build.sbt)
 used for the example.
 
-Earlier versions of Scala had bugs which can cause subtle compile-time problems in PureConfig. As a result we recommend only using 2.11.9 or 2.12.1 or newer within the minor series.
+Earlier versions of Scala had bugs which can cause subtle compile-time problems in PureConfig.
+As a result we recommend only using 2.11.11 or 2.12.1 or newer within the minor series.
 
 ## Use PureConfig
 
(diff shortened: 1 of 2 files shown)


ruippeixotog:
If #292 is found out to be a real problem with some patch versions this LGTM, but as I mentioned [here](https://github.com/pureconfig/pureconfig/issues/292#issuecomment-320111390) I think we should investigate the issue a bit further.

leifwickland:
@ruippeixotog Yup. I'm totally OK with that. I'm already beyond my understanding of PureConfig in #292, so I'll leave it to more capable hands.

core/src/main/tut/README.md
```diff
@@ -88,6 +88,7 @@ libraryDependencies ++= Seq(
 For a full example of `build.sbt` you can have a look at this [build.sbt](https://github.com/melrief/pureconfig/blob/master/example/build.sbt)
 used for the example.
 
+Earlier versions of Scala had bugs which can cause subtle compile-time problems in PureConfig. As a result we recommend only using 2.11.9 or 2.12.1 or newer within the minor series.
```
ruippeixotog: 2.11.9 is a "forbidden" version :) Both 2.11.9 and 2.11.10 were intended to be the last version in the 2.11.x series, but they both introduced binary compatibility issues and were quickly superseded by 2.11.11. for that reason, I think we should recommend using 2.11.11.

See https://github.com/scala/scala/releases/tag/v2.11.11 for more details.

melrief:
I think having this in the README can help developers. I agree with the content of the message regardless of #292, last versions of Scala should be much better for libraries using shapeless, like our.

ruippeixotog (approved):
LGTM!

core/src/main/tut/README.md
leifwickland: Good point, @ruippeixotog. Updated.

PR #295, opened 2017-08-06
Use ConfigList when reading and writing Scala tuples

### LINKED ISSUE #274: Use ConfigList by default when reading and writing Scala tuples
Scala tuples currently get serialized as `ConfigObject`s with keys `_1`, `_2`, etc. Even though I think we should still support reading from this format, I think it's more natural for tuples to be serialized as lists (and consequently `ConfigReader`s for tuples should be able to read from lists).

jcazevedo:
This PR modifies the way tuples are read and written in order to use `ConfigList`s. This implementation still allows reading tuples in the previous format (`ConfigObject`s with keys `_1`, `_2` and so on) but never writes them that way.

Incidentally, this also implements `ConfigReader` and `ConfigWriter` instances for shapeless' `HList`s. We usually have this kind of readers and writers in a separate module, but since `core` already depends on shapeless I think it's OK to leave them there. I can move them to a separate `pureconfig-shapeless` module if you prefer.

This PR fixes #274.


core/src/main/scala/pureconfig/DerivedReaders.scala
```diff
@@ -35,6 +36,20 @@ trait DerivedReaders extends DerivedReaders1 {
       def from(value: ConfigValue): Either[ConfigReaderFailures, T] =
         reader.from(value).right.map(unwrapped.wrap)
     }
+
+  // used for tuples
+  implicit def deriveTupleInstance[F: IsTuple, Repr](
+    implicit
+    g: Generic.Aux[F, Repr],
+    gcr: ConfigReader[Repr],
+    pr: ConfigReader[F]): ConfigReader[F] = new ConfigReader[F] {
```
ruippeixotog: This implicit seems a bit strange to me... it seems that you are recursively asking for the very same implicit and this only works because of the compiler's divergent implicit checker (which isn't a good thing to rely on and has different behavior in different environments, such as macros). What did you mean with this?

If I was right in the assessment above, maybe I would suggest writing a non-implicit `def deriveTupleInstanceAsObject` and a `def deriveTupleInstanceAsList` and making this method use the other two. That way the code would be more explicit in its intent and users could even choose to use only one of the strategies.

core/src/main/scala/pureconfig/DerivedReaders.scala
```diff
@@ -35,6 +36,20 @@ trait DerivedReaders extends DerivedReaders1 {
...
+    override def from(value: ConfigValue) = {
+      // Try to read first as the product representation (i.e.
+      // ConfigObject with '_1', '_2', etc. keys) and afterwards as the Generic
+      // representation (i.e. ConfigList).
+      pr.from(value).left.flatMap(_ => gcr.from(value).right.map(g.from))
```
ruippeixotog: We should combine the errors of both strategies properly (the `WrongType` failure should specify that we were expecting either a list or an object). Can you write tests for this?

core/src/main/scala/pureconfig/DerivedReaders.scala
```diff
@@ -198,6 +213,61 @@ trait DerivedReaders1 {
     }
   }
 
+  // We introduce this special type here to keep accumulating the numbers for
+  // building the `WrongSizeList` ConfigReaderFailure. We can't rely on the
+  // `WrongSizeList` type alone since we can have HLists inside HLists and we
+  // don't wan't to keep accumulating for inner HLists.
+  protected[pureconfig] trait HListConfigReader[T <: HList] {
```
ruippeixotog: In order to avoid this extra type of reader, note that you shouldn't have to build the number of elements the "recursive" way (i.e. by starting with 0 and summing 1 one in each recursive step) in case of a failure. Instead, it suffices to check the sizes first in `deriveHConsConfigReader` and only use the tail reader if the sizes are correct. Something like:

```scala
case cl: ConfigList if cl.size == tl().length + 1 =>
  fail(WrongSizeList(tl.length() + 1, cl.size, ...))

case cl: ConfigList =>
  // read using the head and tail reader as usual; since you verified the
  // size before, every inner reader will also have matching sizes
```

I think it should make the code much simpler and also more flexible, since it uses plain `ConfigReader` instances. Am I missing something?

core/src/main/scala/pureconfig/DerivedReaders.scala
```diff
@@ -198,6 +213,61 @@ trait DerivedReaders1 {
...
+          case cl: ConfigList =>
+            val sl = cl.asScala
+            val hv = hr.from(sl.head).left.map(_.toList.map(Right(_)))
+            val tv = tr.from(ConfigValueFactory.fromAnyRef(sl.tail.asJava)).left.map {
+              _.map {
+                case Left((ex, fo)) => Left((ex + 1, fo + 1))
+                case other => other
+              }
+            }
+            (hv, tv) match {
```
ruippeixotog: `ConvertHelpers.combineResults` is a very handy helper for doing this.

core/src/main/scala/pureconfig/DerivedWriters.scala
```diff
@@ -108,8 +119,27 @@ trait DerivedWriters1 {
     }
   }
 
+  // Auxiliary type to make sure we always return `ConfigList` when writing HLists.
+  private[pureconfig] trait HListConfigWriter[T <: HList] {
```
ruippeixotog: In this case, why didn't you define just `deriveHNilConfigWriter` and `deriveHConsConfigWriter` publicly? Since those are the "canonical" ways of writing HLists to configs and `deriveLabelledGenericInstance` asks specifically for a `WrappedConfigWriter`, we could simplify this. I think the same thing applies to readers, if you can get rid of `HListConfigReader`.

core/src/main/scala/pureconfig/DerivedWriters.scala
```diff
@@ -18,14 +18,25 @@ trait DerivedWriters extends DerivedWriters1 {
...
  * Trait containing `ConfigWriter` instances for collection, product and coproduct types.
  */
 trait DerivedWriters1 {
 
-  private[pureconfig] trait WrappedConfigWriter[Wrapped, SubRepr] extends ConfigWriter[SubRepr]
+  private[pureconfig] trait WrappedConfigWriter[Wrapped, SubRepr] {
+    def to(cv: SubRepr): ConfigValue
+  }
 
   implicit final def hNilConfigWriter[Wrapped]: WrappedConfigWriter[Wrapped, HNil] = new WrappedConfigWriter[Wrapped, HNil] {
```
ruippeixotog: We should rename this and `hConsConfigWriter` to `labelledHNilConfigWriter` and `labelledHConsConfigWriter` in order to clearly distinguish them from the new ones. The same thing applies for readers.

ruippeixotog (changes requested):
This is a great feature, thank you @jcazevedo :) Reading tuples and hlists will surely come in handy. Take a look at the comments below, mostly related to code organization.

ruippeixotog:
One remark: I believe there are very few people using tuples for reading/writing configs, since the current syntax is so strange. Nonetheless, do we want to deprecate (even though we don't disable for now) reading tuples as objects? I think it would be a good measure.

core/src/main/scala/pureconfig/DerivedReaders.scala
jcazevedo: I was being a bit lazy and relying on the diverging implicit checker to have access to the `ConfigReader` that reads tuples from `ConfigObject`s. I modified the code as you suggested in f806a5cf7ffd0624a24ece65c462067cb1d8927b.

core/src/main/scala/pureconfig/DerivedReaders.scala
jcazevedo: Fixed this and added tests for it in f806a5c and 27b0dc7.

core/src/main/scala/pureconfig/DerivedReaders.scala
jcazevedo: Visiting the whole `ConfigList` allowed us to potentially fetch more errors. However, I agree that it does complicate the code and the usefulness of it is debatable. I've modified it in fa049e5c0c6b079a2545d498cf1ac59c9449301f.

core/src/main/scala/pureconfig/DerivedReaders.scala
jcazevedo: I've modified this code to use it in fa049e5c0c6b079a2545d498cf1ac59c9449301f.

core/src/main/scala/pureconfig/DerivedWriters.scala
jcazevedo: I've named the methods as you suggested in 3a758ff826f72c37c2cf2a10eba18027cdfea51c.

core/src/main/scala/pureconfig/DerivedWriters.scala
jcazevedo: When I originally wrote this we had `WrappedConfigWriter` extending `ConfigWriter` which provided a conflicting way of writing `HLists` as empty `ConfigObject`s. I ended up working around it by introducing this `HListConfigWriter`. 

In the end I also had to make `WrappedConfigWriter` not extending `ConfigWriter` as otherwise we couldn't derive a proper `ConfigWriter` for `HNil`. In that case, I can make this extend `ConfigWriter` and make those methods public. I've done that in 3a758ff826f72c37c2cf2a10eba18027cdfea51c.

core/src/main/scala/pureconfig/DerivedWriters.scala
```diff
@@ -108,8 +119,23 @@ trait DerivedWriters1 {
     }
   }
 
+  // Auxiliary type to make sure we always return `ConfigList` when writing HLists.
+  private[pureconfig] trait HListConfigWriter[T <: HList] extends ConfigWriter[T] {
```
ruippeixotog: At this point, I think that we should drop this hlist-specific trait and turn the two readers below into pure `ConfigReader` instances, even if we have to return an additional failure if an unexpected config type is returned from a tail writer. I think it would be strange as an user if I wanted to provide my own writer for a concrete HList (e.g. `String :: String :: HNil`) and realize that it can't be used as part of a larger HList even if it generates a `ConfigList`.

The concept of restricting the types of configs read/written by some readers/writers in the type system is very interesting, though. We could think of a generic way (via shapeless tags, maybe) to build writers with the guarantee that all generated config values are of a given type, increasing the type-safety of combining them and maybe even serving as a base for generating config schemas.

I'm not too strong in removing this right now, so feel free to leave it as it is :)

ruippeixotog (approved):

core/src/main/scala/pureconfig/DerivedReaders.scala
```diff
@@ -198,6 +232,32 @@ trait DerivedReaders1 {
...
+      def from(cv: ConfigValue): Either[ConfigReaderFailures, HNil] = {
+        cv match {
+          case cl: ConfigList if cl.size == 0 => Right(HNil)
+          case cl: ConfigList => fail(WrongSizeList(0, cl.size, ConfigValueLocation(cv), ""))
+          case other => fail(WrongType(other.valueType, Set(ConfigValueType.LIST), ConfigValueLocation(other), ""))
+        }
+      }
+    }
+
+  implicit final def hConsConfigReader[H, T <: HList](implicit hr: Derivation[Lazy[ConfigReader[H]]], tr: Derivation[Lazy[ConfigReader[T]]], tl: HKernelAux[T]): ConfigReader[H :: T] =
```
ruippeixotog: Note that the recursive part of the derivation should not be wrapped in `Derivation`. If you do so, when you try to derive a reader for `String :: Int :: Custom :: HNil`, for example, you'll get a message like this:

```
Could not derive a ConfigReader instance for String :: Int :: Custom :: HNil, because:
  - could not derive a ConfigReader instance for Int :: Custom :: HNil, because:
    - could not derive a ConfigReader instance for Custom :: HNil, because:
      - could not derive a ConfigReader instance for Custom
```

When you probably want just:

```
Could not derive a ConfigReader instance for String :: Int :: Custom :: HNil, because:
  - could not derive a ConfigReader instance for Custom
```

core/src/main/scala/pureconfig/DerivedWriters.scala
jcazevedo: I've removed `HListConfigWriter` in 6bb77972746e7f4b39cd771a4e795eaee053741e and now throw an exception whenever the `to` method doesn't return the appropriate type.

core/src/main/scala/pureconfig/DerivedReaders.scala
jcazevedo: Yeah. I've removed them in a4461937cdf694fb6d3a75fd70f5f4cdcb8131cc.

jcazevedo:
@ruippeixotog: I think I've added `Derivation`s to all the appropriate places, but please do check.

melrief:
I think @ruippeixotog proposal to disable reading tuples from objects make sense, a tuple doesn't have names for fields and the way to specify a tuple without this PR is awkward. I propose to deprecate that format too. @ruippeixotog do you propose a deprecation message at compile time?

ruippeixotog:
I don't think we should add a compile-time message, since we can't know then if the user is using the object format for tuples. It would quickly become too annoying for other users.

Assuming there are not many users relying on this format, I think that a mention about that in the CHANGELOG of the version in which we remove it is good enough.

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
```diff
@@ -21,4 +21,9 @@ object ConfigReaderFailures {
 
   def apply(configReaderFailure: ConfigReaderFailure): ConfigReaderFailures =
     new ConfigReaderFailures(configReaderFailure, List.empty[ConfigReaderFailure])
+
+  def apply(configReaderFailures: List[ConfigReaderFailure]): ConfigReaderFailures = {
+    require(configReaderFailures.length > 0)
```
melrief: I don't like this, `ConfigReaderFailures` should be a `NonEmptyList` and should always be built from either just the head or the head and a tail. Why do we want to add this method that can throw exception?

core/src/test/scala/pureconfig/ConfigConvertChecks.scala
```diff
@@ -23,8 +26,8 @@ trait ConfigConvertChecks { this: FlatSpec with Matchers with GeneratorDrivenPro
    * because, for instance, multiple representation of `t: T` are possible. Use [[checkRead()]] for those
    * representations.
    */
-  def checkArbitrary[T](implicit cc: ConfigConvert[T], arb: Arbitrary[T], tag: ClassTag[T], equality: Equality[T]): Unit =
-    it should s"read an arbitrary ${tag.runtimeClass.getCanonicalName}" in forAll {
+  def checkArbitrary[T](implicit cc: ConfigConvert[T], arb: Arbitrary[T], tpe: TypeTag[T], equality: Equality[T]): Unit =
```
melrief, line 29: Good idea to use `TypeTag`. We should switch to `TypeTag` when we just need the type name.

melrief (changes requested):
LGTM except for the unsafe constructor of `ConfigReaderFailures`.

melrief:
@jcazevedo can you add to the supported types in the doc that we support `shapeless.HList`?

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
jcazevedo: I've removed it in ba8e6e1.

jcazevedo:
I've added `shapeless.HList` to the list of supported types in the README in 42a2968.

melrief (approved):
LGTM. Nice new features we have with this PR, thanks @jcazevedo!

PR #297, opened 2017-08-08
Update dependency versions

ruippeixotog:
Updates dependency versions in preparation for the next release.


leifwickland (approved):
LGTM. 👍

build.sbt
```diff
@@ -13,7 +13,7 @@ lazy val docs = (project in file("docs")).
   dependsOn(core)
 
 def module(proj: Project) = proj.
-  enablePlugins(TutPlugin).
+  enablePlugins(TutPlugin, SbtOsgi).
```
jcazevedo: I think `SbtOsgi` should also be enabled for `core`.

core/build.sbt
```diff
@@ -46,8 +46,6 @@ pomExtra := {
   </developers>
 }
 
-osgiSettings
```
jcazevedo: I'm not familiar with OSGi, but removing this changes the way the artifacts are published. Instead of a `bundles` folder we get a `jar` folder inside the archive and the OSGi headers seem to be absent from the JAR's manifest.

jcazevedo (changes requested):

core/build.sbt
ruippeixotog: Hmm, I'm not familiar either... I though this was a straightforward change to auto-plugins. I'll take a deeper look later. Thanks :)

build.sbt
ruippeixotog: Yeah, fixed in db858f4.

core/build.sbt
ruippeixotog: I misread the documentation for sbt-osgi, it seems that `osgiSettings` is still needed if we want to override the `publish` behavior. I reverted those changes at 131df70.

jcazevedo (approved):

PR #299, opened 2017-08-15
Add PureConfig logo and extended summary to README header.

leifwickland:


README.md
```diff
@@ -1,12 +1,17 @@
...
 [![Coverage Status](https://coveralls.io/repos/github/pureconfig/pureconfig/badge.svg?branch=master)](https://coveralls.io/github/pureconfig/pureconfig?branch=master)
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.github.pureconfig/pureconfig_2.11/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.github.pureconfig/pureconfig_2.11)
 [![Join the chat at https://gitter.im/melrief/pureconfig](https://badges.gitter.im/melrief/pureconfig.svg)](https://gitter.im/melrief/pureconfig?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
+PureConfig is a Scala library for loading configuration files. It reads [Typesafe Config](https://github.com/typesafehub/config) configurations written in [HOCON](https://github.com/typesafehub/config/blob/master/HOCON.md#hocon-human-optimized-config-object-notation), properties, or JSON to native Scala classes in a boilerplate-free way. Sealed traits, case classes, collections, optional values, and many other [types are all supported out-of-the-box](#supported-types). Users also have many ways to add support for custom types or customize existing ones.
```
ruippeixotog: I wonder if saying that PureConfig reads HOCON, properties and JSON is a bit redundant and misleading. PureConfig is currently focused on being a library for Typesafe Config / HOCON, which is indeed a superset of the two others. Enumerating them may lead users to believe that PureConfig is extensible to other backends (which it could be in the future, but not right now).

ruippeixotog (commented):
Looks good to me, thanks for the effort of designing the logo and trying all the layouts!

README.md
leifwickland: Yeah, I was quoting what Typesafe Config says about itself:
> supports files in three formats: Java properties, JSON, and a human-friendly JSON superset

That's also what PureConfig's docs say elsewhere.

I don't want someone to be turned off from PureConfig initially by making them think they'd have to migrate their `properties` or JSON file format to HOCON.

If you feel strongly about it, I'm happy to simply to just HOCON.

ruippeixotog: I'm not too strong about it, it's fine as it is right now :) Only one more question, do you think "properties" is clear enough for users to know we're talking about Java `.properties` files? (honest question, I do not see the format mentioned in many places)

ruippeixotog (approved):

README.md
leifwickland: Fair point. I expanded it to "Java `.properties`".

melrief (approved):

PR #300, opened 2017-08-15
Add user documentation for Derivation

ruippeixotog:
Adds a new section containing end-user documentation for debugging implicits not found by activating the new `Derivation` macro.

We may add some developer-facing documentation later for users that have their own derived readers, but for now I think this is enough.


docs/debugging-implicits-not-found.md
```diff
@@ -0,0 +1,56 @@
...
+// <console>:18: error: could not find implicit value for parameter reader: pureconfig.Derivation[pureconfig.ConfigReader[Conf]]
+//        ConfigFactory.load.to[Conf]
+//                             ^
+```
+
+In PureConfig, the derivation of config readers and writers is done by chaining implicits - the converters of larger
+structures (like `Conf`) depend on the implicit converters of smaller ones (like `Boolean` or `Custom`). However, Scala
+is not helpful in case one of those upstream dependencies is missing, limiting itself to showing the message above.
+
+Since version 0.7.3, PureConfig provides a way to obtain better error messages for those cases. To enable it, we need to
```
melrief: `0.8.0`?

ruippeixotog: Yes, my mistake, I used the current SNAPSHOT version without thinking. Updated it in 08efa7c.

melrief (approved):

jcazevedo (approved):

Issue #87, opened 2017-01-16
Abstract from the typesafe config backend


melrief:
pureconfig started as layer on top of the typesafe config library. The idea was to keep intact the parsing of files and the resolution rules while adding a generic derivation system that converts a `ConfigValue` to a value of an arbitrary type. The initial idea was to not tie the library to typesafe config and in future releases to allow to plug in other backends. I would like to start a discussion about this and eventually implement this abstraction in our codebase. My thoughts about this:

1. **abstraction**: right now the structure that represents a raw configuration is [`ConfigValue`](https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigValue.html). `ConfigValue` represents JSON-like values with support for [boolean, list, null, number, object and string](https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigValueType.html). If we decide to abstract from the backend, using typesafe structures makes no sense because we don't want to have typesafe config as dependency when it's not required. What I propose it to create our own representation of a JSON-like value and keep it as small and minimalistic as possible. For each backend, a new adapter from the backend specific structure to our representation should be coded and that's the only required piece of code to make the new backend work with pureconfig. I am open to use an external library to represent a configuration if there is a good reason for it.

2. **loading**: there are multiple ways to load a backend implementation at runtime. In Java it is common to set the implementor class somehow, e.g. via configuration files, and then use the class loader to load the specified class. This style allows to create a new backend without recompiling pureconfig but has the side-effect that there are no checks at compile time. I personally don't like this style even if I understand the great flexibility it provides. I like to have my compiler to link all the pieces together before I use the library. Consequently I propose to make the core of pureconfig abstract from the implementation and then have one subproject for each backend. If a user wants to implement a new backend, he/she just have to add a new subproject with minimum effort. Each backend would have its own artifact on maven.

Let me know what you think about these two points and if you want to add other points to the discussion. I would prefer to have a clear idea on which everybody agree before we implement it.

leifwickland:
I believe the current guts of `ConfigConvert` don't really use the `boolean`, `number`, etc parts of the `ConfigValue` interface. `ConfigConvert` type-tests and then extracts a list, object, or string representation of the value. To some degree that couples the type-specific `ConfigConvert` instances to the string representation of data that are produced would be produced by the various backends. I'm not sure what an alternative would look like. 

All of that to say, I'm not sure if the solution to abstraction needs to include any more than:
- get type
- get list
- get map/object
- get string representation

[...]

PR #314, opened 2017-10-05
Use a ConfigCursor for reading configs in a safe way

### LINKED ISSUE #287: Introduce cursors in ConfigReaders and possibly ConfigWriters
Currently, `ConfigReader`s lack context on where in the configuration we are, since we deal with `ConfigValue`s directly. Amongst other things, this forces us to rely on calling the `improveFailures` method explicitly to add better context to reading failures. Custom implementations of `ConfigReader`s might not call it and therefore the error locations become unreliable. Introducing cursors would allow us to keep track of context in a consistent way.

Besides improving the context of errors, we can also use cursors to pass information relevant for reading `ConfigValue`s to readers down in the tree.

ruippeixotog:
This PR introduces a `ConfigCursor` as a wrapper for `ConfigValue` on readers. Most notably, it provides methods to safely cast the cursor to a specific type and to navigate through the config. The effort of creating most failures (most notably `WrongType` and `KeyNotFound`) was moved from `ConfigReader` implementations to the `ConfigCursor` internals, making `ConfigReader`s shorter and providing them some automatic error handling.

`ConfigCursor` also keeps the path in the config where the value is located, which means that cursors always provide a full path ready to be used in all failures - both those created by `ConfigCursor` and those created by users.

I also used this PR to remove the previously deprecated `ConfigConvert` constructors. They are deprecated since version 0.6 and 0.7. 

I think that this ~~closes~~ is a first step for #287, even though there is still some effort to be spent on improving failures. Having to provide the config path and location in the `ConfigReaderFailure` constructor, there's always the risk of a user not filling those parameters with the values provided by `ConfigCursor`. I have some ideas to improve this in a future PR.

Reviewing this commit by commit may be easier.


core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -0,0 +1,273 @@
...
+ */
+sealed trait ConfigCursor {
+
+  /**
+   * The `ConfigValue` to which this cursor points to.
+   */
+  def value: ConfigValue
+
+  /**
+   * The path in the config to which this cursor points as a list of keys.
```
melrief: In reverse order?

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -0,0 +1,273 @@
...
+    asObjectCursor.right.map(_.map)
+
+  /**
+   * Casts this cursor as either a `ConfigListCursor` or a `ConfigObjectCursor`.
+   *
+   * @return a `Right` with this cursor as a list or object cursor if the cast can be done, `Left` with a list of
+   *         failures otherwise.
+   */
+  def asCollectionCursor: Either[ConfigReaderFailures, Either[ConfigListCursor, ConfigObjectCursor]] = {
+    if (value == null) {
```
melrief: I would use `isUndefined` here instead of `value == null`.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -0,0 +1,273 @@
...
+      val listAtLeft = asListCursor.right.map(Left.apply)
+      lazy val mapAtRight = asObjectCursor.right.map(Right.apply)
+      listAtLeft
+        .left.flatMap(_ => mapAtRight)
+        .left.flatMap(_ => fail(WrongType(value.valueType, Set(ConfigValueType.LIST, ConfigValueType.OBJECT), location, path)))
+    }
+  }
+
+  private[this] def castOrFail[A](expectedType: ConfigValueType, cast: => A): Either[ConfigReaderFailures, A] = {
+    if (value == null) {
```
melrief: I would use `isUndefined` instead

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -0,0 +1,273 @@
...
+      fail(KeyNotFound(path, location, Set()))
+    } else if (value.valueType != expectedType) {
+      fail(WrongType(value.valueType, Set(expectedType), location, path))
+    } else {
+      Try(cast) match {
```
melrief: I have to admit I find this unconfortable: `cast` is just a parameter of the function not necessarily related to `value`. What if instead of `cast: => A` you pass `cast: ConfigValue => A`? The you can apply `cast` to `this.value`.

melrief (approved):
Sorry to make you wait @ruippeixotog but I've been busy and this PR is not small. But it's very good and I like the new cursor API. I left some comments but overall I think it's a great improvement over the current API. It's also a good step forward in having different backends (I may change the sealed trait `ConfigCursor` in future to do some tests with YAML).

I have only one comment that is more a question than a real request for changes: why keeping `value` public? I see two reasons to not keep it public

1. We know that `ConfigValue` can be `null`, that's why `isNull` and `isUndefined` exist, and we may want to avoid to expose it directly to the developer. Instead, one of the `as` functions should be used. What do you think about restricting the API such that it's safe? Note that it's still possible to check if the cursor points to `null` with `isNull`, like you do in `OptionConfigReader`, and this method is safe.
2. We can avoid exposing the typesafe config types and this would allow us to change the backend in future. I know it's not the main focus of the PR or pureconfig but looking at your API I really don't see a good reason to expose typesafe config types at all. Sorry for pushing for this but, as I said, I see this PR as a step forward in trying other backends. We just need other implementations of the cursor.

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: Right, improved in e7147e0.

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: Sure, changed it in c0372c7.

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: Changed it in c0372c7.

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: Yeah, that makes all sense. I was previously using that function for more general expressions, but right now I'm really only doing casts acting on `value`. I did what you suggested in 3c3ff3e.

ruippeixotog:
@melrief I also pondered whether we should keep `value` public or not. I ended up exposing it because it may still be needed by users:
- it makes migrating old code way easier, as it suffices to use `cursor.value` instead of `configValue` in custom readers or functions;
- our readers using `Config`, `ConfigValue` and other Typesafe structures would be impossible to implement without access to the underlying config. There may be user code requiring access to it (for example, a user function manipulating `ConfigValue` entities) which we shouldn't invalidate. 

We also should be aware that there are _many_ things that the cursor still can't do, such as converting to int/bool, obtaining the type of a value, something like `withFallback`...

Abstracting over `value` (and `location`) are things definitely on the realm of #87, but I don't think they would be difficult to add later. Regarding `value` it suffices to turn it later into a deprecated method, to be removed in a subsequent version - and if users already started to migrate to `as` methods that wouldn't cause much impact.

While I understand that this is a very useful step if we want to have multiple backends, I don't think this is the right time (or even version) to abstract completely from Typesafe config.

core/src/main/scala/pureconfig/ConfigReader.scala
```diff
@@ -31,7 +40,7 @@ trait ConfigReader[A] {
    * @return a `ConfigReader` returning the results of this reader mapped by `f`.
    */
   def map[B](f: A => B): ConfigReader[B] =
-    fromFunction[B](v => from(v).right.flatMap(toResult(f)(_)(ConfigValueLocation(v))))
+    fromCursor[B](v => from(v).right.flatMap(toResult(f)(_)(ConfigValueLocation(v.value))))
```
jcazevedo: We can avoid accessing the `ConfigCursor`'s `value` by using `v.location` instead of `ConfigValueLocation(v.value)` here.

core/src/main/scala/pureconfig/DerivedReaders.scala
```diff
@@ -175,85 +163,70 @@ trait DerivedReaders1 {
...
-            val keyResult = catchReadError(_.toInt)(implicitly)(key)(ConfigValueLocation(value)).left.flatMap(t => fail(CannotConvert(key, "Int",
-              s"To convert an object to a collection, its keys must be read as Int but key $key has value" +
-                s"$value which cannot converted. Error: ${t.because}", ConfigValueLocation(value), key)))
-            val valueResult = configConvert.value.value.from(value)
+
+        case Right(objCur) =>
+          def keyValueReader(key: String, valueCur: ConfigCursor): Either[ConfigReaderFailures, (Int, T)] = {
+            val keyResult = catchReadError(_.toInt)(implicitly)(key)(valueCur.location).left
+              .flatMap(t => fail(CannotConvert(key, "Int",
+                s"To convert an object to a collection, its keys must be read as Int but key $key has value" +
```
jcazevedo: There's a space missing between `value` and `${valueCur.value}`. Also, I think this error message is wrong. We can't convert an object to a traversable when its keys can't be converted to integers, but we're mistakenly pointing to the values, which are not the issue here.

I understand this was already here before, but perhaps we can use this PR to fix it, since it is an easy fix.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -0,0 +1,273 @@
...
+  /**
+   * Returns a cursor to the config at a given key.
+   *
+   * @param key the key of the config for which a cursor should be returned
+   * @return a `Right` with a cursor to the config at `key` if such a config exists, a `Left` with a list of failures
+   *         otherwise.
+   */
+  def atKey(key: String): Either[ConfigReaderFailures, ConfigCursor] = {
+    atKeyOrUndefined(key) match {
+      case keyCur if keyCur.isUndefined => fail(KeyNotFound(keyCur.path, location, key, map.values.map(_.path)))
```
jcazevedo: I don't think building `KeyNotFound` like this would produce a valid list of candidate keys for the missing one, because we'll be comparing complete paths with a key at this level.

Also, I'm not sure if we should provide candidate keys in the error message for custom `ConfigReader`s that use the `atKey` method. What do you think?

jcazevedo (changes requested):
This is very good! Thanks!

I think we should use this PR to remove the `improveFailures` method of `ConvertHelpers` as well as the `withImprovedContext` method of `ConfigReaderFailure` and `ConfigReaderFailures`.

core/src/main/scala/pureconfig/ConfigReader.scala
ruippeixotog: Yep, improved in 9ffb2ce30e0547102c031186743b8f7b42f2ae9e!

core/src/main/scala/pureconfig/DerivedReaders.scala
ruippeixotog: Sure. Fixed in 6a7edfd1f337c9024c558af4468c8cbce5feb9e5.

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: Sounds good. I'll pass only the keys themselves to `KeyNotFound`, since the full path of each candidate does not convey additional information when printing the error message.

I think that the error message is still useful even for custom `ConfigReader`s. While it points to a misconfigured product hint when it's not the case, I think that it still provides an important clue to the developer.

Improved in 228518d1c28ad9cd9a47684c984773998f50e823.

ruippeixotog:
@jcazevedo thanks for reminding me about `improveFailures`, I forgot to remove it. I did it now in 7a1ecde8b86a8cecaa8290fe86aa6c2f68f87a0f.

Unfortunately, `withImprovedContext` cannot be removed yet because `stringToEitherConvert` depends on it. In general, the current failure model is not adequate when you pass control from `ConfigCursor` to a Scala value such as in `ConfigReader.fromString`, since you have to have a location and a path on the places you instantiate failures.

In my view, it would be better to have base failures (such as `KeyNotFound` and `WrongType`) agnostic of the context and then add context to them via decorators. I'll try to implement a version of this in a future PR so that we can evaluate that design :)

jcazevedo:
> In my view, it would be better to have base failures (such as KeyNotFound and WrongType) agnostic of the context and then add context to them via decorators. I'll try to implement a version of this in a future PR so that we can evaluate that design :)

I agree. I'd rather only close #287 when that's done, though.

jcazevedo (approved):

PR #315, opened 2017-10-10
Upgrade build to SBT 1.0

ruippeixotog:
Upgrades SBT to version 1.0.

Besides the usual plugin updates, the only non-trival change is the removal of `sbt-doge`, since its behavior is now [integrated in SBT 1.0](https://github.com/sbt/sbt-doge/issues/18).


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.16
+sbt.version=1.0.1
```
melrief: [`1.0.2`](http://developer.lightbend.com/blog/2017-09-19-sbt-1-migration-status-and-1-0-2-hotfix/#sbt-hotfix-1-0-2) is available with other bugfixes

ruippeixotog: Thanks, I'll update this tomorrow! Unfortunately, not all sbt-doge features were moved to SBT (https://github.com/sbt/sbt-doge/issues/18)... We only need the missing operator for Travis, though.

jcazevedo (changes requested):
We should also upgrade the sample project build to use SBT 1.0.

ruippeixotog:
@jcazevedo I updated SBT in the example project in 1cefebc.

At this point our Travis build still doesn't work due to its reliance on the strict aggregation operator from sbt-doge, which wasn't ported to SBT 1.0. https://github.com/sbt/sbt/issues/3698 is tracking this problem. We can wait for SBT to fix this behavior or change our Travis config not to rely on `+++`. If we choose the latter we can either:

- Not use the `scala` key in .travis.yml, merging everything in a single build, and use `+test` for the test step;
- Create our custom SBT commands for building against a given Scala version.

jcazevedo:
I'd rather wait for the problem to be fixed.

If we opt to change our Travis config, I prefer having a different build for each Scala version.

melrief:
sbt `1.0.4` is available!

ruippeixotog:
Yes, but it doesn't solve our problem with the lack of sbt-doge :( I'll consider changing our code for sbt-cross in order for us not to be blocked by this.

ruippeixotog:
I think that we'll finally be able to update SBT after 1.2.0 is released (https://github.com/sbt/sbt/pull/3995) 🎉

ruippeixotog:
SBT 1.2.0 was finally released today 🎉 I updated this PR to clean up the history and update everything in a clean way.

@jcazevedo, can you review this again?

example/project/build.properties
```diff
@@ -1 +1 @@
-sbt.version = 0.13.15
+sbt.version=1.2.0
```
jcazevedo: We should update to sbt version 1.2.1, which fixes a [forward binary compatibility breakage](https://developer.lightbend.com/blog/2018-08-07-sbt-1-2-1/) introduced in 1.2.0.

jcazevedo (changes requested):

example/project/build.properties
ruippeixotog: The first part seems more important for plugins than for end users like us, but we were affected by the second point, regarding unknown configurations. Updated to SBT 1.2.1 in d53bc33.

jcazevedo (approved):

leifwickland (approved):

PR #317, opened 2017-10-19
Update to Scala 2.12.4

### LINKED ISSUE #205: [error] knownDirectSubclasses of ... observed before subclass ... registered

ruippeixotog:
Updates Scala to version 2.12.4. This is important since this finally closes #205 🎉

### DIFF
--- a/core/src/test/scala/pureconfig/CoproductConvertersSuite.scala
+++ b/core/src/test/scala/pureconfig/CoproductConvertersSuite.scala
@@ -21,12 +21,6 @@ class CoproductConvertersSuite extends BaseSuite {
     ConfigConvert[AnimalConfig].from(conf.root()) shouldEqual Right(DogConfig(2))
   }
 
-  it should "read disambiguation information on sealed families with the cases nested in the companion" in {
-    import CarMaker._
-    val conf = ConfigFactory.parseString("{ type = bmw }")
-    ConfigConvert[CarMaker].from(conf.root()) shouldEqual Right(BMW)
-  }
-
   it should "write disambiguation information on sealed families by default" in {
     val conf = ConfigConvert[AnimalConfig].to(DogConfig(2))
     conf shouldBe a[ConfigObject]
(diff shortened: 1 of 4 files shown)


core/src/test/scala/pureconfig/CoproductConvertersSuite.scala
```diff
@@ -21,6 +21,12 @@ class CoproductConvertersSuite extends BaseSuite {
     ConfigConvert[AnimalConfig].from(conf.root()) shouldEqual Right(DogConfig(2))
   }
 
+  it should "read disambiguation information on sealed families with the cases nested in the companion" in {
+    import CarMaker._
+    val conf = ConfigFactory.parseString("{ type = bmw }")
+    ConfigConvert[CarMaker].from(conf.root()) shouldEqual Right(BMW)
```
melrief: Strange bug, I was expecting this to work because `type` is set and correct and no other arguments are needed to build `BMW`. @ruippeixotog do you know what's this about?

ruippeixotog:
I forgot that the test obviously doesn't work in Scala 2.10 and 2.11...

I'll wait until macro-paradise is released for the new Scala 2.10.7 and 2.11.12 (https://github.com/scalamacros/paradise/issues/108) to see if that particular bug fix was backported. If not, I'll move the new test to be 2.12-only.

ruippeixotog:
After some analysis, I believe that Scala 2.10.7 and 2.11.12 won't fix this issue. Therefore, I moved the test so that it only runs on Scala 2.12+. This is now ready to be reviewed and merged.

melrief (approved):
LGTM. I don't think we need to update the CHANGELOG right?

ruippeixotog:
I don't think so, since the change is invisible for pureconfig users.

jcazevedo (approved):

PR #319, opened 2017-10-27
sbt-microsite

melrief:
I finally found some time to work on the sbt-microsite and I've pushed a first attempt into the `microsites` branch. Before pushing for a final solution I'd like to know your opinion about it. I attached some screenshots to the PR. Follow [this](https://47deg.github.io/sbt-microsites/docs/build-the-microsite.html) to test it locally.

Note that I stripped the original `README.md` of most of the content and moved it to the new website because maintaining both would require a lot of effort. The `README.md` now has a link to the documentation that will work once we upload it to `pureconfig.github.io`. `README.md` will still be produced manually by tut.

I also renamed some of the section of the documentation to better fit the sidebar, e.g. `Add PureConfig to your project` is not `Setup PureConfig` and also moved some of the content to their own files.

Once we agree on the PR, I will first publish the website and then we can proceed with merging the PR. Because we are all new to sbt-microsites, I propose to publish manually the website. We can move to automatically deployment in future if we prefer.

Closes #289 .

screen shot 2017-10-27 at 16 52 06
screen shot 2017-10-27 at 16 52 26

PR #320, opened 2017-10-27
Documentation Clarity

skeet70:
I've had a few people at work get confused by the structure of the pureconfig documentation. Issues addressed here:

* Assumption that "Table of Contents" applied only to the README after clicking the first few links.
* Confused by "Whence the config files?"
* The `use` section showed a lot of magic without links to explanation. 
* The `example` section was an example of `parseString`, which is not the recommended way of using the library. The README usage/example should be the preferred way, so someone could could conceivably get to using it immediately.
* Fixed a broken link in the doc discussing how to use `loadConfigFromFiles`.

### DIFF
--- a/core/src/main/tut/README.md
+++ b/core/src/main/tut/README.md
@@ -36,7 +36,7 @@ Click on the demo gif below to see how PureConfig effortlessly translates your c
 - [Error handling](docs/error-handling.md)
 - [Handling missing keys](docs/handling-missing-keys.md)
 - [Example](docs/example.md)
-- [Whence the config files](docs/whence-the-config-files.md)
+- [Config files](docs/config-files.md)
 - [Support for Duration](docs/support-for-duration.md)
 - [Integrating with other libraries](#integrating-with-other-libraries)
 - [Contribute](#contribute)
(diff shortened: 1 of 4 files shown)


ruippeixotog:
Hi @skeet70, thank you for the contribution! I like the changes you propose; I agree that our documentation needs some work and every contribution helps.

We use [tut](https://github.com/tpolecat/tut) for generating our README and the rest of the documentation and our build checks that the "source code" of the documentation is in sync with the generated docs. Because of that, changes to the README must be done in [core/src/main/tut](https://github.com/pureconfig/pureconfig/tree/master/core/src/main/tut) and changes to other docs should be done in [docs/src/main/tut](docs/src/main/tut). Afterwards, just run `sbt tut` and SBT will generate the "real" documentation. Be aware that code inside tut code blocks is actually compiled and run, so any config loading must point to real files.

The reorganization of the Table of Contents will most probably conflict with https://github.com/pureconfig/pureconfig/pull/319 which also simplifies that, so I guess it's better to leave it for later. Other than that, this looks good :)

skeet70:
Ah, was wondering how the build had failed on doc changes. Thanks, I definitely didn't notice `tut` was involved, but it is a great tool. I'm pushing up a new commit generated using that.

melrief (approved):

melrief:
Hi @skeet70, thank you for the PR! LGTM for me, I found the changes well done and the resulting README more readable. I'll wait for another review before merge.

core/src/main/tut/README.md
```diff
@@ -101,15 +101,46 @@ As a result we recommend only using the latest Scala versions within the minor s
...
+  }, 
+  "list": ["1", "20%"], 
+  "map": { "key": "value" } 
+}
+```
+
+Then, load the configuration ([in this case from the classpath](docs/config-files.md)):
 
-val config: Either[pureconfig.error.ConfigReaderFailures,YourConfClass] = loadConfig[YourConfClass]
+```scala
```
ruippeixotog, line 141: My only concern with this particular change is that it's not being run by tut anymore. While I understand that loading a config from a file is much more useful than from a string for showcase purposes, this will get outdated if we ever change the `loadConfig` signature or the behavior of one of the readers.

I'm willing to let this without tut as it's a very short code section and we shouldn't change such an entrypoint to our library anyway. @melrief, do you agree?

melrief: Most of the people use pureconfig to load from `application.conf` and I think @skeet70 changes are good. I'd like to have `tut` to check also this example but we can look into this later. Maybe there is way to still use `tut` somehow. For now, let's keep it unchecked @ruippeixotog.

ruippeixotog (approved):

PR #321, opened 2017-11-02
Improve ConfigReaderFailure for better scoping from ConfigCursor

### LINKED ISSUE #287: Introduce cursors in ConfigReaders and possibly ConfigWriters

ruippeixotog:
This PR focuses on refactoring our failure models to better support scoping by `ConfigCursor`.

The main problem with our current failures is that they all require a `ConfigValueLocation` and a config path to be provided the moment they are instantiated. This is not a very flexible design because:

- That information is not always available the moment the failure is being created. In particular, when we delegate the reading part to the user without providing a `ConfigCursor` (for example, when they use `ConfigReader.fromString` or use `emap` on an existing reader) they have no access to that information and have to fill those with `None` and `""`, respectively. Because of that, many functions in `ConvertHelpers` are unecessarily complicated by having an extra `Option[ConfigValueLocation]` argument;
- Users always have to remember to fill those fields, even when they have access to the necessary context;
- Functions like `improveFailures` and `withImprovedContext` had to be created to correctly scope the pre-created failures.

The new model improves that by splitting our failures into two distinct types:

- A `FailureReason` represents the reason why a conversion failed, agnostic of locations and paths in the config. Most of our `ConfigReaderFailure` subclasses were converted to `FailureReason` subclasses;
- A `ConfigReaderFailure` represents a higher-level failure. This can be a `ConvertFailure(reason: FailureReason, location: Option[ConfigValueLocation], path: String)` or a more fundamental failure such as `CannotReadFile` or `CannotParse`.

`ConfigCursor` was updated to provide three new methods for users to easily create failures with the correct context. Since now classes such as `KeyNotFound` and `WrongType` are `FailureReason`s and cannot be returned directly as `ConfigReaderFailures`, users are encouraged to use the `ConfigCursor` methods that provide that bridge.

I'm not very convinced of the names I chose for the failure traits and for the new `ConfigCursor` methods. Suggestions are welcome :)

I think that after the `ConfigCursor` introduction and this PR, we're in dire need of a thorough documentation review in order to update best practises and outdated text. I didn't do anything about it on this PR because I'll dedicate a PR to it later.

Closes #287.

### DIFF
--- a/core/src/main/scala/pureconfig/ConfigCursor.scala
+++ b/core/src/main/scala/pureconfig/ConfigCursor.scala
@@ -4,7 +4,6 @@ import scala.collection.JavaConverters._
 import scala.util.{ Failure, Success, Try }
 
 import com.typesafe.config._
-import pureconfig.ConvertHelpers._
 import pureconfig.backend.PathUtil
 import pureconfig.error._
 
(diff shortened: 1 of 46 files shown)


core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -6,271 +6,94 @@ package pureconfig.error
...
+  def description = "The config files to load must not be empty."
+  def location = None
 }
 
 /**
- * A failure representing the presence of an unknown key in a ConfigObject. This
- * error is raised when a key of a ConfigObject is not mapped into a field of a
- * given type, and the allowUnknownKeys property of the ProductHint for the type
- * in question is false.
+ * A failure occurred because an exception was thrown duing the reading process.
```
jcazevedo: There's a typo on `duing`.

jcazevedo (changes requested):
This looks good to me as is. I just identified a small typo.

jcazevedo:
> I'm not very convinced of the names I chose for the failure traits and for the new ConfigCursor methods. Suggestions are welcome :)

I think the names are accurate. My only issue is with using `ConfigReaderFailure`s on the write side, but I think that's a problem for later.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
ruippeixotog: Fixed in aeaafcd.

jcazevedo (approved):

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -103,27 +102,65 @@ sealed trait ConfigCursor {
...
+   * @tparam A the returning type of the `ConfigReader`
+   * @return a failed `ConfigReader` result built by scoping `reason` into the context of this cursor.
+   */
+  def failed[A](reason: FailureReason): Either[ConfigReaderFailures, A] =
+    Left(ConfigReaderFailures(failureFor(reason)))
+
+  /**
+   * Returns a `ConfigReaderFailure` resulting from scoping a `FailureReason` into the context of this cursor.
+   *
+   * This operation is useful for constructing `ConfigReaderFailures` when there are multiple `FailureReason`s.
```
melrief, line 130: What's the reason? Because it is easier to scope multiple `FailureReason`s?

docs/add-support-for-complex-types.md
```diff
@@ -88,54 +88,35 @@ Similarly to adding support for simple types, it is possible to manually create
...
-  Left(ConfigReaderFailures(WrongType(wrong.valueType(), Set(ConfigValueType.OBJECT), ConfigValueLocation(wrong), "")))
-})
+implicit val identifiableConfigReader = ConfigReader.fromCursor { cur =>
+  for {
+    objCur <- cur.asObjectCursor.right
+    typeCur <- objCur.atKey("type").right
+    typeStr <- typeCur.asString.right
+    ident <- extractByType(typeStr, objCur).right
+  } yield ident
+}
```
melrief, line 119: This is much better, well done!

melrief (approved):

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: Because you can easily do `listOfFailureReasons.map { fr => cur.failureFor(fr) }` and construct `ConfigReaderFailures` from a `List[ConfigReaderFailure]`. We could as well provide a `ConfigCursor#failed(FailureReason, FailureReason*)` variant, but I thought it would be more useful to provide the low-level way to "lift" a `FailureReason`. Do you think the API or this documentation should be improved?

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -103,27 +102,65 @@ sealed trait ConfigCursor {
...
+   *
+   * This operation is needed when control of the reading process is passed to a place without a `ConfigCursor`
+   * instance providing the nexessary context (for example, when `ConfigReader.fromString` is used. In those scenarios,
+   * the call should be wrapped in this method in order to turn `FailureReason` instances into `ConfigReaderFailures`.
+   *
+   * @param result the result of a config reading operation
+   * @tparam A the returning type of the `ConfigReader`
+   * @return a `ConfigReader` result built by scoping `reason` into the context of this cursor.
+   */
+  def scopeFailures[A](result: Either[FailureReason, A]): Either[ConfigReaderFailures, A] =
```
melrief: I'm not sure about naming
- `scopeFailures` should be called `scopeFailure` imho because it's still one single `FailureReason`
- `failed` and `failureFor` names are strange when you compare them to `scopeFailures`: `scopeFailures` clearly tells the developer that its main task is "scoping" the given failure into the context of the cursor while `failed` and `failureFor` names don't contain scope at all. This is nitpicking :) but I find the names difficult to understand. I don't have a counterproposal though.

core/src/main/scala/pureconfig/ConfigCursor.scala
melrief: No I think it's clear, thank you.

melrief (approved):

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: I renamed `scopeFailures` to `scopeFailure` in ace155f, it makes more sense indeed. Regarding the others I am also not sure about the naming.

I think my rationale behind the names was that `failureFor` was a simple constructor for a `ConfigReaderFailure` from a `FailureReason`. Therefore, the focus of the method was not on scoping, but on the conversion.

`failed`, on the other hand, is a method targeted clearly at failure cases in `ConfigReader` implementations due to its signature, so I wanted to give a more fluent API (as in, calling `cur.failed` in a reader to signal a failed conversion).

`scopeFailures` is the only method of the three that doesn't necessarily result in a failed result. While `failed` always returns a `Left` and `failureFor` returns a `ConfigReaderFailure`, `scopeFailures` may return a `Right` result if its argument is not a failure. That's why I focused on the "scoping" part and not on the "failure" part.

I agree that the naming may be confusing, I'm open to ideas :)

PR #331, opened 2017-12-02
Fix Derivation usage outside the pureconfig package

### LINKED ISSUE #330: -Xmacro-settings:materialize-derivations failing to provide improved error message
@ruippeixotog When the better error messages work, they're awesome. In the following case, they don't seem to work. 

I set up a [project](https://github.com/leifwickland/pureconfig-debugging-implicits) to demonstrate.

Am I doing it wrong?

ruippeixotog:
Even though the `Failed` derivation constructor is only used internally by the macro, apparently it still needs to be accessible on the macro application points. As such, (some?) `Derivation`s outside the `pureconfig` package were failing without the full error description.

Fixes #330.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
     of `ConfigReaderFailures`;
   - The `CannotConvertNull` failure was removed, being superseeded by `KeyNotFound`;
   - Methods deprecated in previous versions were removed.
+  
+- Bug fixes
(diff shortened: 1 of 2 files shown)


leifwickland (approved):
@ruippeixotog This fixes the problems I was having. Thanks!

melrief:
Thanks guys for taking care of this. We should consider the release of `0.9.0` soon.

Issue #323, opened 2017-11-05
Reorganize the documentation


ruippeixotog:
Our documentation has grown organically as we added more and more information there. As it is now I think that we can improve it in order to better reflect the hierarchy of concepts involved. Right now we have too many sections at the top-level, some of them very disparate in terms of depth.

The forthcoming website (https://github.com/pureconfig/pureconfig/pull/319) will also provide us the flexibility to include different content in the repository's README (the landing page for developers who want to get started quickly or contribute) and in the new documentation page of the website (a place for developers looking to do a specific thing with pureconfig or wanting deeper information).

This issue is only concerned with the high-level organization and not with the contents themselves, which also need some rewriting to cope with the newest pureconfig features.

I suggest we reorganize it as follows:

README contents:

- Why/Motivation (from [Why](https://github.com/pureconfig/pureconfig#why))
- Quick Start (a compact version of [Add PureConfig to your project](https://github.com/pureconfig/pureconfig#add-pureconfig-to-your-project) + [Use PureConfig](https://github.com/pureconfig/pureconfig#use-pureconfig) + [Example](https://github.com/pureconfig/pureconfig#example) + [Whence the config files](https://github.com/pureconfig/pureconfig/blob/master/docs/whence-the-config-files.md))
- Documentation (short section linking to the documentation page)
- Contribute (from [Contribute](https://github.com/pureconfig/pureconfig#contribute))
- License (from [License](https://github.com/pureconfig/pureconfig#license))
- Special Thanks (from [Special Thanks](https://github.com/pureconfig/pureconfig#special-thanks))

Documentation contents:

- Why/Motivation (copy of the README section)
- Quick Start (copy of the README section)
- Built-in Supported Types (from [Supported Types](https://github.com/pureconfig/pureconfig#supported-types))
- Supporting New Types (from [Support new Types](https://github.com/pureconfig/pureconfig/blob/master/docs/support-new-types.md))
  - Simple Types (from [Add support for simple types](https://github.com/pureconfig/pureconfig/blob/master/docs/add-support-for-simple-types.md))
  - Complex Types (from [Add support for complex types](https://github.com/pureconfig/pureconfig/blob/master/docs/add-support-for-complex-types.md))
- Overriding Behavior for Types (a short section indicating that all techniques exposed in "Supporting New Types" can be used to override the default behavior for a type)
[...]

PR #332, opened 2017-12-03
Reorganize docs webpage sections

ruippeixotog:
Reorganizes the documentation webpage sections as I described in #323.

I touched the content of the sections as little as possible in order to make this easy to review. I'll follow this up with a review of the content in order to update outdated information and to fix all errors I can find.


ruippeixotog:
It is insanely easy to leave broken links on our website while updating stuff, since most links are internal and file renames lead to different page locations. In 68d0562 I updated our Travis build to check for broken links in the generated page using their [build stages and jobs](https://docs.travis-ci.com/user/build-stages).

README.md
```diff
@@ -7,20 +7,53 @@
...
 
-PureConfig is a Scala library for loading configuration files. It reads [Typesafe Config](https://github.com/typesafehub/config) configurations written in [HOCON](https://github.com/typesafehub/config/blob/master/HOCON.md#hocon-human-optimized-config-object-notation), Java `.properties`, or JSON to native Scala classes in a boilerplate-free way. Sealed traits, case classes, collections, optional values, and many other [types are all supported out-of-the-box](docs/supported-types.html). Users also have many ways to add support for custom types or customize existing ones.
+PureConfig is a Scala library for loading configuration files. It reads [Typesafe Config](https://github.com/typesafehub/config) configurations written in [HOCON](https://github.com/typesafehub/config/blob/master/HOCON.md#hocon-human-optimized-config-object-notation), Java `.properties`, or JSON to native Scala classes in a boilerplate-free way. Sealed traits, case classes, collections, optional values, and many other [types are all supported out-of-the-box](https://pureconfig.github.io/docs/built-in-supported-types.html). Users also have many ways to add support for custom types or customize existing ones.
```
melrief, line 10: If you hardcode the URL then you cannot test the output anymore locally. Relative URLs can be tested locally.

README.md
```diff
@@ -7,20 +7,53 @@
...
+
+## Quick Start
 
-## Use PureConfig
+To use PureConfig in an existing SBT project with Scala 2.10 or a later version, add the following dependency to your
+`build.sbt`:
 
-First, import the library, define data types, and a case class to hold the configuration:
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig" % "0.7.2"
```
melrief: `0.7.2`?

docs/src/main/tut/docs/faq.md
```diff
@@ -96,3 +97,57 @@ But, of course, the values we need are still there:
 secret.password.value 
 secret.apiKey.value
 ``` 
+
+### How do I debug "implicit not found" errors?
```
melrief, line 101: I think this should be in its own section because many developers could be interested

melrief (changes requested):

README.md
ruippeixotog: This is the README of the repository, so there isn't any relative link to put here. Note that `docs/src/main/tut/index.md` has a relative link at this part.

melrief: Right, never mind this comment :)

README.md
ruippeixotog: Good catch, updated in 07f4c3cb7e8fb90b45a965106f5eb045def3c813 everywhere I saw the old version.

docs/src/main/tut/docs/faq.md
ruippeixotog: I guess so. But even though it's technically important, I can't seem to like it as a top-level section on our documentation, on par with things like "Supporting New Types" and "Error Handling". The problem is not even specific to PureConfig...

I can see two alternatives here:

- A "Troubleshooting" top-level section that has not only this, but also common problems our users have (like mixing up naming conventions). If the section is called "Troubleshooting" users having problems with implicits not found will go there to check;
- Add all FAQ questions to the side bar, giving users direct links to the common questions they are having. Similarly, by seeing a related question at the side bar users will find it easily. 

What do you think?

README.md
```diff
@@ -7,20 +7,53 @@
...
 
 Click on the demo gif below to see how PureConfig effortlessly translates your configuration files to well-typed objects without error-prone boilerplate.
 <br clear="right"> <!-- Turn off the wrapping for the logo image. -->
 
 ![](http://i.imgur.com/P6sda06.gif)
 
-## Documentation
 
-Read the [documentation](https://pureconfig.github.io/docs).
+## Why
```
leifwickland, line 18: Thanks for restoring this. I'm glad to see a raison d'etre right up front.

core/src/main/tut/README.md
```diff
@@ -7,20 +7,53 @@
...
+
+For a full example of `build.sbt` you can have a look at this [build.sbt](https://github.com/pureconfig/pureconfig/blob/master/example/build.sbt).
+
+Earlier versions of Scala had bugs which can cause subtle compile-time problems in PureConfig.
+As a result we recommend only using the latest Scala versions within the minor series.
+
+In your code, first import the library and define data types and a case class to hold the configuration:
 
 ```scala
 import pureconfig.loadConfig
```
leifwickland, line 59: Nitpick: I suggest removing this line and making the call to `loadConfig` below fully qualified or placing the import closer to the usage.
Feel free to ignore.

core/src/main/tut/README.md
```diff
@@ -55,21 +87,41 @@ the [config files documentation](docs/config-files.html):
 }
 ```
 
-Then, load the configuration ([in this case from the classpath](docs/config-files.html)):
+Finally, load the configuration:
 
 ```scala
 loadConfig[MyClass](conf)
```
leifwickland, line 93: Stupid question: Why does this work? I don't see a `conf` identifier defined anywhere in this file.

docs/src/main/tut/docs/index.md
```diff
@@ -27,9 +21,62 @@ libraryDependencies ++= Seq(
...
+  }, 
+  "list": ["1", "20%"], 
+  "map": { "key": "value" } 
+}
+```
+
+Finally, load the configuration:
+
+```scala
+loadConfig[MyClass](conf)
```
leifwickland, line 66: Ditto the question about why does this work? I swear this code shouldn't work.

leifwickland (approved):
Thanks for your work on this, @ruippeixotog!

I'm definitely curious about that code snippet I mentioned. Please take a look at it if you have time in the context of this PR.

core/src/main/tut/README.md
ruippeixotog: Yep, I agree. But if you don't mind, I'll do it in my next PR. As I mentioned on the PR description, I think it's better for reviewing if we set out the structure first and update/fix the content afterwards separately. Moving files and sections around like this always leads to very messy diffs (as you can see in the "Files changed" tab) and I usually avoid doing changes (even small ones) since they get easily overlooked by reviewers.

core/src/main/tut/README.md
ruippeixotog: The response is also stupid; that works because, unfortunately, we aren't running that by tut (note the `scala` block in the markdown) :(

#320 updated some aspects of our documentation and one of the changes introduced was to show an example of an `application.conf` instead of showing a `val conf = ConfigFactory.parseString(...)` call, since the latter is not at all a common use case for pureconfig. Apparently [my concerns](https://github.com/pureconfig/pureconfig/pull/320#discussion_r151871712) with not running this with tut were shown to be justified much sooner than I would like to... 

On the next PR I'll make this run with tut, even if I have to add some sample `application.conf` to the resources in order for the code to compile. Right now, I'll leave this as it is since the "bug" wasn't introduced with the changes on this PR.

leifwickland: Totally OK. Thanks for taking a look. Thanks again for your efforts.

melrief (approved):

PR #339, opened 2017-12-11
Make emap use FailureReason instead of ConfigReaderFailures

ruippeixotog:
`ConfigReader#emap` should require an `Either` with `FailureReason` as a failure type and not a `ConfigReaderFailures`, since users do not have the necessary context to create a `ConfigReaderFailure`. I must have forgotten about this method when I wrote #321.


leifwickland:
@ruippeixotog Are these changes to the API relative to an unreleased version? I.e this would normally be a breaking change, but is this *not* a breaking change because it's a change something new since 0.8, right?

ruippeixotog:
That's more or less correct :) The `emap` method already existed; however, #321 introduced the new failures model some existing methods that deal with failures were changed. The CHANGELOG was modified to account for those changes at that time.

leifwickland (approved):

jcazevedo (approved):

PR #340, opened 2017-12-11
Make `asString` read any scalar config type

### LINKED ISSUE #338: ConfigCursor#asString should work with all scalar config values
Due to the nature of the HOCON format, any config value can be a string, since strings are allowed to be defined without quotes. The Typesafe `Config` inferface supports that flexible typing by transforming the original values according to the requested type, using their internal [DefaultTransformer](https://github.com/lightbend/config/blob/master/config/src/main/java/com/typesafe/config/impl/DefaultTransformer.java). Our `ConfigCursor` is not so flexible, though, and calling `asString` only gives a `Right` result if the value cannot be any other type.

We should improve `asString` to allow reading any scalar config value (i.e. non-collection) as a string. Ideally, we would use the `DefaultTransformer` to ensure full compatibility, but the class is not accessible.

ruippeixotog:
Makes `asString` read any scalar config type, just like the Typesafe Config API does. Since the underlying config format allows strings to be defined without quotes and primitives like `true` and `null` can arguably be seen as strings, this "cast" makes sense to me in PureConfig.

Fixes #338.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,4 +47,3 @@ cache:
   directories:
     - $HOME/.ivy2/cache
     - $HOME/.sbt
-    - $HOME/.rvm
--- a/core/src/main/scala/pureconfig/ConfigCursor.scala
+++ b/core/src/main/scala/pureconfig/ConfigCursor.scala
@@ -57,7 +57,7 @@ sealed trait ConfigCursor {
    *         failures otherwise.
    */
   def asString: Either[ConfigReaderFailures, String] =
-    castOrFail(STRING, Set(NUMBER, BOOLEAN, NULL, STRING), { cv => String.valueOf(cv.unwrapped) })
+    castOrFail(STRING, Set(NUMBER, BOOLEAN, STRING), { cv => String.valueOf(cv.unwrapped) })
 
   /**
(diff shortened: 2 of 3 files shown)


leifwickland (approved):
Seems plausible.

jcazevedo (changes requested):
I think we want `ConfigCursor` behaviour regarding casts to be similar to what Typesafe Config does. As such, I'm not sure we should be supporting `null` as `String`, since Typesafe Config [doesn't](https://github.com/lightbend/config/blob/master/config/src/main/java/com/typesafe/config/impl/DefaultTransformer.java#L60).

ruippeixotog:
@jcazevedo that's right, I read the code wrong, I added support for `null` thinking Typesafe Config would convert it.

The comment explaining their decision seems to be not much about semantics, but about the inability for them to distinguish a null from a missing value on their API:

```
// if we converted null to string here, then you wouldn't properly
// get a missing-value error if you tried to get a null value
// as a string.
```

However, I think I prefer to keep the behavior close to Typesafe Config. Readers can always test if the cursor `isNull` before casting `asString` and handle the special case manually.

jcazevedo (approved):

PR #344, opened 2017-12-17
Add cats-effect module

### LINKED ISSUE #343: Add support for cats-effect
I'm trying to use pureconfig in combination with cats-effect's IO however as it stands the code to do this is a little faffsome, here is what it looks like in my codebase: 

```scala
  def loadSettings(path: Path): IO[MyConfigCaseClass] = {
    val load: IO[Either[Throwable, MyConfigCaseClass]] = IO {
      val config = loadConfig[MyConfigCaseClass](path)
      config.left.map(x => new RuntimeException("Should actually convert ConfigReaderFailures into exception here"))
    }
    load.flatMap(IO.fromEither)
  }
```

Would it be possible to add support for cats-effect, such that we could have a loadconfig that returns an IO of the config class? This would simplify usage with cats-effect, which it seems to me might be a fairly common use case given pureconfig seems to be the preferred config library within the cats ecosystem.

keirlawson:
Resolves #343

### DIFF
--- a/modules/cats-effect/README.md
+++ b/modules/cats-effect/README.md
@@ -1,0 +1,54 @@
+# Cats-effect module for PureConfig
+
+Adds support for loading configuration using [cats-effect](https://github.com/typelevel/cats-effect) to control side effects.
+
+## Add pureconfig-cats-effect to your project
+
+In addition to [core pureconfig](https://github.com/pureconfig/pureconfig), you'll need:
+
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-cats-effect" % "0.8.0"
+```
+
+## Example
+
+### Reading configuration
+
+To load a configuration file from a path using cats-effect's `IO`:
+
+
+
+
+```scala
+import pureconfig._
+import pureconfig.module.catseffect._
+import cats.effect.IO
+
+case class MyConfig(somefield: Int, anotherfield: String)
+
+val load: IO[MyConfig] = loadConfigF[IO, MyConfig](somePath)
+```
+
+To test that this `IO` does indeed return a `MyConfig` instance:
+```scala
+//Show the contents of the file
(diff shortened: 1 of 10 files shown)


keirlawson:
Build failing due to rvm issue seemingly unrelated to PR

keirlawson:
Will rebase on master when build fix in https://github.com/pureconfig/pureconfig/pull/340/commits/2f60776a4deec9479f2a8043d5eb1a424b305e89 lands

ruippeixotog:
@keirlawson the fix just got merged 👍

keirlawson:
Rebased, awaiting review :)

modules/cats-effect/README.md
```diff
@@ -0,0 +1,51 @@
+# Cats-effect module for PureConfig
```
melrief, line 1: Not mandatory but we highly recommended to use tut to generate the markdown files. See what we do for the [cats module](https://github.com/pureconfig/pureconfig/tree/master/modules/cats/src/main/tut).

melrief (changes requested):
Very nice! Would be LGTM but I have one comment about how to generate documentation. We usually use `tut`.

modules/cats-effect/README.md
keirlawson: Ok, will do so

keirlawson:
Switched to tut for documentation as requested

melrief (approved):
LGTM! I like this new module, thank you @keirlawson ! I'd like to merge it if it is ready.

keirlawson:
Yep, go ahead :)

melrief:
Done! I will wait for the release of 0.9.0 to release this module (it should be soon). In the meanwhile, I'd like to add you to the [CONTRIBUTORS](https://github.com/pureconfig/pureconfig/blob/master/CONTRIBUTORS) file. If you agree, please make a PR with a single line with your name, username and your email and I'll merge it.

Thank you for the contribution!

PR #347, opened 2017-12-22
Fix supported types documentation link.

mikla:


leifwickland (approved):
Thanks, @mikla!

ruippeixotog:
The link is actually correct with regard to the [version of the website currently on master](https://github.com/pureconfig/pureconfig/tree/master/docs/src/main/tut/docs). In particular, #332 changed the name of the webpage (including this link) and we didn't update the website since then.

We should decide how to manage the website regarding this. Instead of reverting the changes to the link, we should probably make sure that the website is always up-to-date with the master branch. However, that would pose a problem when we updated the content of the website with documentation for unreleased features.

leifwickland:
@ruippeixotog Good catch. Sorry I missed that.

melrief:
@ruippeixotog we can automatize publication of the website from master if we want. It's complicated but doable. The reason why it is manual right now is because that gave me more control but I can work on automatize. What do you think @leifwickland @ruippeixotog ?

The other solution is that everytime somebody merges a PR, he also pushes the new website if needed. We are all part of the `pureconfig` organization and we should all be able to push the website. I can publish guidelines for the authors to be used to publish the next version of the website.

The issue with the documentation referring to the last version of PureConfig and not the last published version of PureConfig is difficult to solve. A solution is to use a long-living branch `website` for the website of the last published version of PureConfig and master for the documentation of the next version of PureConfig. When we do a new release, we replace the `website` branch with a copy of master and continue from there. The main issue with this approach is that fixes to the documentation must be maintained for both the long-living branch and master.

ruippeixotog:
Yeah, I'm ok with the website having either the latest (master) version or the latest stable version. We should definitely automatize this - I can try to put a step on Travis for this every time a new commit gets pushed to `master` if you want.

One thing that we lost when we changed from `tut`-generated markdowns to a website is that we can't link to the documentation of a specific PureConfig version. For example, while I can link to the [v0.8.0](https://github.com/pureconfig/pureconfig/tree/v0.8.0) or [v0.7.0](https://github.com/pureconfig/pureconfig/tree/v0.7.0) pages to see the documentation for each version, when we release 0.9.0 we won't have a permalink for the docs, since they aren't committed to the repository anymore (at least not after they are compiled by tut). Incidentally, we also lost a way to check that the code in the docs didn't change when we review a PR; for example, we can now easily introduce a bug where the result of a `loadConfig` wrongly changes from a `Right` to a `Left`.

What do you think of continuing to generate and commit Markdown files generated by `tut` for all the documentation just for the "time machine" feature, while continuing to have the website up-to-date as the main source for documentation?

jcazevedo:
Since we don't publish snapshots, I prefer having the latest stable version as the version on the website, particularly because that's the version we indicate to add as a dependency in the [first page](https://pureconfig.github.io/docs/) of the documentation. If we had snapshots, then I'd be OK with having the latest published version as the version of the documentation.

> What do you think of continuing to generate and commit Markdown files generated by tut for all the documentation just for the "time machine" feature, while continuing to have the website up-to-date as the main source for documentation?

I think it'd also be nice to have permalinks on the site for the older versions of the library. Since we control the way the website is published, I think we can include the documentation for older versions under a different directory, so that we have the latest stable version available at https://pureconfig.github.io/ and the older versions at https://pureconfig.github.io/vX.Y.Z/.

> Incidentally, we also lost a way to check that the code in the docs didn't change when we review a PR; for example, we can now easily introduce a bug where the result of a `loadConfig` wrongly changes from a `Right` to a `Left`.

We're still running `tut` on our CI build, and `makeMicrosite` also runs it. I think we can keep running `tut` (probably running `makeMicrosite` would be preferable) to have those checks at the CI build, no?

ruippeixotog:
@jcazevedo I like the idea of having permalinks to the website at the time of our stable versions! I think it wouldn't be too difficult to automatize the process, I'll investigate it.

> We're still running `tut` on our CI build, and `makeMicrosite` also runs it. I think we can keep running `tut` (probably running `makeMicrosite` would be preferable) to have those checks at the CI build, no?

We run `makeMicrosite` (and even check for broken links in the generated pages), but while this detects code that stops compiling, it doesn't detect changes in the `tut` output, which can lead to overlooking potential bugs (for example, a `Right` result can change to a `Left` unintentionally). This behavior can only be kept if we commit the tut results to the repository, as we were doing before.

ruippeixotog:
I'll close this issue since it will be fixed on the next website update. Thanks anyway for noticing this, @mikla, the issue you noticed led to a good discussion about this!

jcazevedo:
> We run `makeMicrosite` (and even check for broken links in the generated pages), but while this detects code that stops compiling, it doesn't detect changes in the `tut` output, which can lead to overlooking potential bugs (for example, a `Right` result can change to a `Left` unintentionally).

Ah, right! Then I think it's reasonable to generate and commit files generated by `tut` for all the documentation in order to detect changes in the output.

PR #361, opened 2018-02-19
Add http4s module

jcranky:
Very simple for now, but adds the whole skeleton in case someone needs something more.

### DIFF
--- a/modules/http4s/README.md
+++ b/modules/http4s/README.md
@@ -1,7 +1,8 @@
 
 # Http4s module for PureConfig
 
-Adds support for selected [Http4s](http://http4s.org/) classes to PureConfig.
+Adds support for [Http4s](http://http4s.org/)'s Uri class to PureConfig. PRs adding support
+for other classes are welcome :)
 
 ## Add pureconfig-http4s to your project
 
--- a/modules/http4s/README.md
+++ b/modules/http4s/README.md
@@ -1,0 +1,35 @@
+
+# Http4s module for PureConfig
+
+Adds support for selected [Http4s](http://http4s.org/) classes to PureConfig.
+
+## Add pureconfig-http4s to your project
+
+In addition to [core PureConfig](https://github.com/pureconfig/pureconfig), you'll need:
+
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-http4s" % "0.9.0"
+```
+
+## Example
+
+To load an `Uri` into a configuration, create a class to hold it:
+
+```scala
+import org.http4s.Uri
+import com.typesafe.config.ConfigFactory.parseString
+import pureconfig._
+import pureconfig.modules.http4s._
+
+case class MyConfig(uri: Uri)
+```
+
+We can read a `MyConfig` with the following code:
+
+```scala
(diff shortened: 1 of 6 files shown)


leifwickland:
@jcranky, Thanks for the PR! Aside from the broken 2.10 build, it looks good to me.

The 2.10 build is failing because http4s isn't published for that version of scala. I don't recall how to exclude your new sbt subproject from being built for that part of the cross build. Do you?

jcranky:
I'm not sure, but I'll push a commit trying something in a minute. Lets see what happens :)

leifwickland:
@jcranky I'm really baffled how that worked. Don't get me wrong: I'm glad it did, but the 2.10 build seems to have run the same course as the other builds, but succeeded this time.

jcranky:
What do you mean? The http4s module is not there in the 2.10 build anymore, which makes sense to me :) - I was just not sure if that line I added would have the overriding behaviour I wanted, but it did :D

leifwickland:
@jcranky Oh, gosh. My bad. Please pardon my pre-coffee lack of awareness.

Thanks again.

modules/http4s/build.sbt
```diff
@@ -0,0 +1,25 @@
+name := "pureconfig-http4s"
+
+libraryDependencies ++= Seq(
+  "org.http4s" %% "http4s-core" % "0.18.0",
+  Dependencies.scalaMacrosParadise,
+  Dependencies.scalaTest
+)
+
+crossScalaVersions := Seq("2.11.11", "2.12.4")
```
ruippeixotog: Even though this works perfectly, it is not so convenient when we have to update Scala versions (with this we'd have to update in multiple places). Can you use something like we did with [the Akka module](https://github.com/pureconfig/pureconfig/blob/master/modules/akka/build.sbt#L4) instead?

modules/http4s/README.md
```diff
@@ -0,0 +1,35 @@
+
+# Http4s module for PureConfig
```
ruippeixotog, line 2: I wonder if we should add something here saying that only `Uri` is supported for now. People may come here expecting readers for many HTTP model classes and find out that it's not the case. I'm ok with leaving this as it is, though :)

ruippeixotog (changes requested):
Thanks for the new module @jcranky, this looks good :) I hope that readers for more HTTP model classes get added with time.

modules/http4s/build.sbt
jcranky: Ah, of course, that is better, will do.

modules/http4s/README.md
jcranky: Well it wouldn't hurt to do that, so I'll do it :)

jcranky:
@ruippeixotog done

ruippeixotog (approved):
LGTM now. Thanks again!

leifwickland:
@jcranky, I manually merged this branch into master in 9e21c48 to work around the failing Travis check.

I'd like to add you to the CONTRIBUTORS file. If you agree, please make a PR with a single line with your name, username and your email and I'll merge it.

Thank you for the contribution!

jcranky:
thanks :)

PR #364, opened 2018-02-26
Fix reading http4s Uris

jcranky:
Make sure that errors are not ignored and that the values are put in the
correct fields.

### DIFF
--- a/modules/http4s/src/test/scala/pureconfig/modules/http4s/Http4sTest.scala
+++ b/modules/http4s/src/test/scala/pureconfig/modules/http4s/Http4sTest.scala
@@ -1,9 +1,10 @@
 package pureconfig.modules.http4s
 
-import pureconfig.syntax._
 import com.typesafe.config.ConfigFactory
 import org.http4s.Uri
 import pureconfig.BaseSuite
+import pureconfig.error.{ CannotConvert, ConfigReaderFailures, ConvertFailure }
+import pureconfig.syntax._
 
 class Http4sTest extends BaseSuite {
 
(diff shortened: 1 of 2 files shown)


ruippeixotog:
The code isn't compiling on Scala 2.11 because `toTry` is 2.12-only. Can you fix that issue?

jcranky:
@ruippeixotog I didn't notice the build failure, I'll look into that

jcranky:
PR updated :)

ruippeixotog (approved):
LGTM!

modules/http4s/src/test/scala/pureconfig/modules/http4s/Http4sTest.scala
```diff
@@ -12,6 +12,6 @@ class Http4sTest extends BaseSuite {
   "reading the uri config" should "parse the uri" in {
```
leifwickland, line 13: Please add a case which tests the unhappy path of a bad URI.

leifwickland (changes requested):

modules/http4s/src/test/scala/pureconfig/modules/http4s/Http4sTest.scala
jcranky: will do

jcranky:
@leifwickland done

modules/http4s/src/test/scala/pureconfig/modules/http4s/Http4sTest.scala
leifwickland: @jcranky, Thanks!

leifwickland (approved):
LGTM! 👍 Thanks, @jcranky

PR #365, opened 2018-02-26
Add ConfigReader and ConfigWriter for Char

stefanobaghino:
There are no tests right now, I looked around but found no tests for primitive types. I guess I could use some guidance in this regard. 🙂

### DIFF
--- a/core/src/main/scala/pureconfig/BasicReaders.scala
+++ b/core/src/main/scala/pureconfig/BasicReaders.scala
@@ -24,6 +24,12 @@ import pureconfig.error._
 trait PrimitiveReaders {
 
   implicit val stringConfigReader = ConfigReader.fromString[String](s => Right(s))
+  implicit val charConfigReader = ConfigReader.fromNonEmptyString[Char](s =>
+    s.size match {
+      case 1 => Right(s.charAt(0))
+      case len => Left(WrongSizeString(1, len))
+    }
+  )
   implicit val booleanConfigReader = ConfigReader.fromNonEmptyString[Boolean](catchReadError({
     case "yes" | "on" => true
     case "no" | "off" => false
--- a/core/src/main/scala/pureconfig/BasicReaders.scala
+++ b/core/src/main/scala/pureconfig/BasicReaders.scala
@@ -28,8 +28,7 @@ trait PrimitiveReaders {
     s.size match {
       case 1 => Right(s.charAt(0))
       case len => Left(WrongSizeString(1, len))
-    }
-  )
+    })
   implicit val booleanConfigReader = ConfigReader.fromNonEmptyString[Boolean](catchReadError({
     case "yes" | "on" => true
     case "no" | "off" => false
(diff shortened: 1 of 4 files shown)


ruippeixotog:
This looks good!

Tests for basic converters are in [BasicConvertersSuite.scala](https://github.com/pureconfig/pureconfig/blob/master/core/src/test/scala/pureconfig/BasicConvertersSuite.scala). Adding `checkArbitrary[Char]` there will use scalacheck to verify that all `Char` values convert to and from a config in a consistent way. For completeness, you should also add `checkFailure` tests for an empty string and a string with more than one character.

stefanobaghino:
Ah-ah, cool, that's why IntelliJ couldn't find tests for me. Thanks, I've added the test. There seems to be something not working with Scala 2.12 on Travis CI but I can't tell what it its.

stefanobaghino:
Wow, I'm finding out stuff about `String`s in Java that I didn't know, fix is coming.

stefanobaghino:
Now it's ok. For some reason `ConfigWriter.forPrimitive[Char]` doesn't work.

stefanobaghino:
I have no idea why it's failing in Scala 2.12, the jobs exits with an error code but all tests pass. 😐

ruippeixotog:
> Now it's ok. For some reason ConfigWriter.forPrimitive[Char] doesn't work.

`forPrimitive` relies on `ConfigValueFactory.forAnyRef`, which [doesn't support Char](https://github.com/lightbend/config/blob/master/config/src/main/java/com/typesafe/config/impl/ConfigImpl.java#L204). `toDefaultString` seems fine.

> I have no idea why it's failing in Scala 2.12, the jobs exits with an error code but all tests pass. 😐

According to Travis, `The command "if [[ $TRAVIS_SCALA_VERSION =~ ^2\.12.* ]]; then git diff --exit-code; fi" exited with 1.`. This happens because the code isn't formatted according to the rules defined for the project. If you run `sbt compile` on your computer you'll notice that the code gets reformatted automatically. You should commit those changes in order for the build to pass :)

stefanobaghino:
I see, thanks.

stefanobaghino:
Yay, thanks for the guidance!

ruippeixotog (approved):
LGTM. Thanks, @stefanobaghino!

jcazevedo:
LGTM. Thanks for this, @stefanobaghino!

jcazevedo:
Thanks for the PR @stefanobaghino! We would like to add you to the [CONTRIBUTORS](https://github.com/pureconfig/pureconfig/blob/master/CONTRIBUTORS) file. If you agree, please submit another PR in which you add a line with your name, your email (optional) and your username, in line with PRs such as #118, #152 or #160.

stefanobaghino:
Thanks for helping me out with this modest contribution, it'll be my pleasure to add myself to the `CONTRIBUTORS` file. Cheers!

PR #367, opened 2018-03-03
Add fs2 module

keirlawson:


modules/fs2/src/test/scala/pureconfig/module/fs2Suite.scala
```diff
@@ -0,0 +1,63 @@
+package pureconfig.module.fs2
```
ruippeixotog: Nitpick: this file should be in `modules/fs2/src/test/scala/pureconfig/module/fs2/fs2Suite.scala` (with a fs2 subfolder) in order to match the package name.

modules/fs2/build.sbt
```diff
@@ -0,0 +1,32 @@
...
+  "co.fs2" %% "fs2-io" % "0.10.2",
+  Dependencies.scalaMacrosParadise,
+  Dependencies.scalaTest)
+
+// fs2 0.10 isn't published for Scala 2.10
+crossScalaVersions ~= { oldVersions => oldVersions.filterNot(_.startsWith("2.10")) }
+
+pomExtra := {
+  <scm>
+    <url>git@github.com:pureconfig/pureconfig.git</url>
```
ruippeixotog: You shouldn't specify the SCM information here. Doing that clashes with `sbt-git`, which also injects SCM information in the POM file, leading to invalid (duplicate) SCM entries.

ruippeixotog (changes requested):
LGTM overall, thanks for the new module!

modules/fs2/src/test/scala/pureconfig/module/fs2Suite.scala
keirlawson: I feel that is is more important to mirror the src tree, which is `modules/fs2/src/main/scala/pureconfig/module/fs2.scala` as it is a package object, however it is pretty minor so if you insist on it I'm happy to change :)

modules/fs2/build.sbt
keirlawson: Fixed, not sure how that slipped in!

modules/fs2/src/test/scala/pureconfig/module/fs2Suite.scala
ruippeixotog: Sure, it makes all sense to mirror the source tree 👍 But then this package should be `pureconfig.module` just like the `fs2` package object has, isn't it?

keirlawson:
Changes made to address feedback

ruippeixotog (approved):
LGTM!

leifwickland (approved):
Thanks for the addition, @keirlawson!

PR #369, opened 2018-03-04
Check PureConfig's GitHub links against the local tree

ruippeixotog:
This should better handle links from our documentation to our own project page on GitHub (not only to added pages, but also to deleted ones). Fixes the build issues on #361 and #367.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,11 @@ jobs:
       script:
         - sbt makeMicrosite
           # we're ignoring www.47deg.com due to SSL errors; we should check this again later
-        - htmlproofer --allow_hash_href --url-ignore '/www.47deg.com/' docs/target/site
+        - htmlproofer \
+            --allow_hash_href \
+            --url-ignore '/www.47deg.com/' \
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,7 +41,11 @@ jobs:
       script:
         - sbt makeMicrosite
           # we're ignoring www.47deg.com due to SSL errors; we should check this again later
```
leifwickland, line 43: Nitpick: I would prefer to retain the "why" part of the comment about 47deg.

leifwickland (approved):
Thanks for figuring this out, @ruippeixotog!

.travis.yml
ruippeixotog: I'm not sure what you mean. I didn't change the message on this PR; you you want me to add more information to the message? I can add a link to the PR where I introduced this, if you want.

leifwickland: My bad. I've got to stop reviewing PRs before coffee. I read that as removing the comment. Apologies.

PR #370, opened 2018-03-05
Added ConfigConvert.viaReaderAndWriter method

LMnet:
It could be useful in cases when you have explicit `ConfigReader` and `ConfigWriter` and want to expose them as the single `ConfigConvert` instance.

### DIFF
--- a/core/src/main/scala/pureconfig/ConfigConvert.scala
+++ b/core/src/main/scala/pureconfig/ConfigConvert.scala
@@ -49,6 +49,11 @@ object ConfigConvert extends ConvertHelpers {
     def to(t: T) = writer.value.to(t)
   }
 
+  def viaReaderAndWriter[T](reader: ConfigReader[T], writer: ConfigWriter[T]): ConfigConvert[T] = new ConfigConvert[T] {
+    def from(cur: ConfigCursor): Either[ConfigReaderFailures, T] = reader.from(cur)
+    def to(t: T): ConfigValue = writer.to(t)
+  }
+
   def viaString[T](fromF: String => Either[FailureReason, T], toF: T => String): ConfigConvert[T] = new ConfigConvert[T] {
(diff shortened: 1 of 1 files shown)


ruippeixotog:
Hi @LMnet! Thank you for the contribution. However, I don't think this should be the way to go. A bit of context: the `Derivation` type class was introduced just to help with "implicit not found" messages as described [here](https://pureconfig.github.io/docs/faq.html#how-do-i-debug-implicit-not-found-errors). It aims to be completely invisible to the end user, acting just as a compile-time only wrapper for `ConfigReader`s and `ConfigWriter`s.

Just as you were unable to use `fromReaderAndWriter` (the proper method to join the two parts of `ConfigConvert`), other people may need to call one of the several implicit `ConfigReader` and `ConfigWriter` methods and find it hard because of `Derivation`. A more general approach may be to add an implicit conversion in the `Derivation` companion object like this:

```scala
implicit def liftDerivation[A](a: A): Derivation[A] = Successful(a)
```

And write your example as you would without `Derivation`:

```scala
case class MyConfig(a: Int)
implicit val conv: ConfigConvert[MyConfig] = {
  implicit val reader: ConfigReader[MyConfig] = ???
  implicit val writer: ConfigWriter[MyConfig] = ???
  ConfigConvert.fromReaderAndWriter(reader, writer)
}
```

I need to think a bit about the implications of adding this method, but at first sight I don't see any harm with it.

LMnet:
Actually, in my current code I have something like this: 
```scala
ConfigConvert.fromReaderAndWriter(Derivation.Successful(reader), Derivation.Successful(writer))
```
And I'm not happy with this. `Derivation` should be hidden from me. That's why I created this pull request. I talked about this issue a little bit on gitter and didn't found any better solution. 

I like your suggestion about `liftDerivation`. Less amount of methods is always better.

ruippeixotog:
`Derivation` is gone from PureConfig, so this is now obsolete. Thank you for the contribution anyway!

PR #371, opened 2018-03-06
Hadoop module

LMnet:
Added Hadoop module. Currently, it supports only `Path` class, but it could be a good start.

### DIFF
--- a/docs/src/main/tut/docs/library-integrations.md
+++ b/docs/src/main/tut/docs/library-integrations.md
@@ -18,7 +18,8 @@ The core of PureConfig eschews unnecessary dependencies. Separate modules exist 
 - [`pureconfig-joda`](https://github.com/pureconfig/pureconfig/tree/master/modules/joda) provides configurable converters for [Joda-Time](http://www.joda.org/joda-time/) types;
 - [`pureconfig-scala-xml`](https://github.com/pureconfig/pureconfig/tree/master/modules/scala-xml) provides support for XML via [scala-xml](https://github.com/scala/scala-xml);
 - [`pureconfig-squants`](https://github.com/pureconfig/pureconfig/tree/master/modules/squants) provides converters for [Squants](http://www.squants.com/)'s beautiful types representing units of measure;
-- [`pureconfig-http4s`](https://github.com/pureconfig/pureconfig/tree/master/modules/http4s) provides converters for [Http4s](http://http4s.org/) types.
+- [`pureconfig-http4s`](https://github.com/pureconfig/pureconfig/tree/master/modules/http4s) provides converters for [Http4s](http://http4s.org/) types;
+- [`pureconfig-hadoop`](https://github.com/pureconfig/pureconfig/tree/master/modules/hadoop) provides converters for [Hadoop](http://hadoop.apache.org//) types.
 
 ### External Integrations
 
--- a/modules/hadoop/src/main/tut/README.md
+++ b/modules/hadoop/src/main/tut/README.md
@@ -1,0 +1,48 @@
+# Hadoop module for PureConfig
+
+Adds support for selected [Hadoop](http://hadoop.apache.org/) classes to PureConfig.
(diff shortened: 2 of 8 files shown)


docs/src/main/tut/docs/library-integrations.md
```diff
@@ -19,6 +19,7 @@ The core of PureConfig eschews unnecessary dependencies. Separate modules exist
 - [`pureconfig-scala-xml`](https://github.com/pureconfig/pureconfig/tree/master/modules/scala-xml) provides support for XML via [scala-xml](https://github.com/scala/scala-xml);
 - [`pureconfig-squants`](https://github.com/pureconfig/pureconfig/tree/master/modules/squants) provides converters for [Squants](http://www.squants.com/)'s beautiful types representing units of measure;
 - [`pureconfig-http4s`](https://github.com/pureconfig/pureconfig/tree/master/modules/http4s) provides converters for [Http4s](http://http4s.org/) types.
+- [`pureconfig-hadoop`](https://github.com/pureconfig/pureconfig/tree/master/modules/hadoop) provides converters for [Hadoop](http://hadoop.apache.org//) types.
```
leifwickland: Nitpick: Please follow, the slightly odd punctuation style of the bullets above, i.e. `;` for all lines except the final one.

modules/hadoop/src/main/tut/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Add pureconfig-hadoop to your project
+
+In addition to [core PureConfig](https://github.com/pureconfig/pureconfig), you'll need:
+
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-hadoop" % "0.9.0"
+```
+
+Also, `pureconfig-hadoop` depends on `hadoop-common` with `provided` scope. It's mean that you should explicitly add a dependency on `hadoop-common` or any other Hadoop library who depends on `hadoop-common`. Usually it would be something like this:
```
leifwickland: The wording here is a little odd. I would suggest something like "This means that ..." and "... library which depends ...".

modules/hadoop/src/main/tut/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+```scala
+libraryDependencies += "org.apache.hadoop" % "hadoop-client" % "3.0.0"
```
leifwickland: Do you know if the `Path` class is binary compatible across Hadoop v2 and v3? 

I'm curious since I'm stuck on v2. If they're not compatible, it might nice to mention that here.

modules/hadoop/src/main/tut/README.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+To use Hadoop module you need to import:
```
leifwickland: Typically this would be worded as "... use _the_ Hadoop ...".

leifwickland (changes requested):
I have a few wording nitpicks, but in general this is a nice addition that I could've used a long time ago.

modules/hadoop/src/main/tut/README.md
LMnet: I checked - `Path` is binary compatible. But overall there is no strong guarantee (SHOULD and SHELL instead of MUST) for binary 2 and 3 major releases. It may be a problem for some cases, but at the current moment we could ignore this.

LMnet:
@leifwickland I fixed all nitpicks.

modules/hadoop/src/main/tut/README.md
leifwickland: That seems entirely reasonable. Thanks for checking into it.

leifwickland (approved):
LGTM. 👍 Thanks, @LMnet.

ruippeixotog (approved):
LGTM! Thank you for the contribution and sorry for the delay reviewing this.

PR #372, opened 2018-03-18
Use SBT Developer model for writing the POM

ruippeixotog:
Defining `pomExtra` in SBT is erroneous because of its unstructured nature. We had problems in the past with the automated `release` process because we were defining a block twice there and there's always the possibility that we're defining the XML wrong. This PR converts the `pomExtra` definition into a `developers` setting using SBT models.

I took the emails of the contributors for the CONTRIBUTORS file and, in a few cases, from the commit email. @leifwickland, since you don't have your email in the CONTRIBUTORS file, if you don't want your email to appear in the POM files please let me know and I'll leave the field blank.


leifwickland (approved):
Thanks for cleaning this up, @ruippeixotog!

Yes, feel free to include my email address.

PR #373, opened 2018-03-18
Organize package object files in the standard Scaladoc way

ruippeixotog:
The [Scaladoc guide](https://docs.scala-lang.org/style/scaladoc.html#packages) recommends all package objects to be in a `package.scala` file inside the usual package folder (that's how IDEs generate package objects, too). Not all of our modules followed that standard, leading to potential problems with Scaladoc generation.

In logical terms, all the packages remain exactly the same, so this won't lead to breaking changes. The only thing changed was the directory structure.


leifwickland (approved):
Thanks for cleaning this up, @ruippeixotog!

Just out of curiosity, what suggested the import changes?

ruippeixotog:
@leifwickland the "Move" action in IntelliJ reorganizes the imports automatically and I opted to leave the changes.

PR #377, opened 2018-03-27
Define a method `ConfigCursor#atPath`

missingfaktor:
Similar to `ConfigCursor#atKey` but for deeper nested paths. We have this in our project's utils, but were wondering if this could be put in PureConfig itself.

### DIFF
--- a/core/src/main/scala/pureconfig/ConfigCursor.scala
+++ b/core/src/main/scala/pureconfig/ConfigCursor.scala
@@ -95,6 +95,26 @@ sealed trait ConfigCursor {
   def asMap: Either[ConfigReaderFailures, Map[String, ConfigCursor]] =
     asObjectCursor.right.map(_.map)
 
+  @inline private final def atPathSegment(pathSegment: PathSegment): Either[ConfigReaderFailures, ConfigCursor] = {
+    pathSegment match {
+      case PathSegment.Key(k) => this.asObjectCursor.right.flatMap(_.atKey(k))
+      case PathSegment.Index(i) => this.asListCursor.right.flatMap(_.atIndex(i))
+    }
+  }
+
+  /**
+   * Returns a cursor to the config at the path composed of given path segments.
+   *
+   * @param pathSegments the path of the config for which a cursor should be returned
+   * @return a `Right` with a cursor to the config at `pathSegments` if such a config exists, a `Left` with a list of
+   *         failures otherwise.
+   */
+  final def atPath(pathSegments: PathSegment*): Either[ConfigReaderFailures, ConfigCursor] = {
+    pathSegments.foldLeft(Right(this): Either[ConfigReaderFailures, ConfigCursor]) {
+      case (soFar, segment) => soFar.right.flatMap(_.atPathSegment(segment))
+    }
+  }
+
   /**
    * Casts this cursor as either a `ConfigListCursor` or a `ConfigObjectCursor`.
    *
--- a/core/src/test/scala/pureconfig/ConfigCursorSuite.scala
+++ b/core/src/test/scala/pureconfig/ConfigCursorSuite.scala
@@ -92,6 +92,14 @@ class ConfigCursorSuite extends BaseSuite {
       Right(Right(ConfigObjectCursor(conf("{ a: 1, b: 2 }").asInstanceOf[ConfigObject], defaultPath)))
   }
 
+  it should "allow access to a given path" in {
+    cursor("{ a: 2 }").atPath() shouldBe Right(cursor("{ a: 2 }", defaultPath))
+    cursor("{ a: 2 }").atPath("a") shouldBe Right(cursor("2", "a" :: defaultPath))
+    cursor("{ a: { b: 2 } }").atPath("a", "b") shouldBe Right(cursor("2", "b" :: "a" :: defaultPath))
+    cursor("{ a: { b: [{ c: 2 }] } }").atPath("a", "b", 0, "c") shouldBe Right(cursor("2", "c" :: "0" :: "b" :: "a" :: defaultPath))
+    cursor("{ a: { b: { c: 2 } } }").atPath("a", "c") should failWith(KeyNotFound("c", Set()), s"$defaultPathStr.a")
+  }
+
   it should "handle in a safe way cursors to undefined values" in {
     val cur = ConfigCursor(null, defaultPath)
     cur.path shouldBe defaultPathStr
--- a/core/src/main/scala/pureconfig/PathSegment.scala
+++ b/core/src/main/scala/pureconfig/PathSegment.scala
@@ -1,0 +1,13 @@
+package pureconfig
+
+import language.implicitConversions
+
+sealed trait PathSegment extends Product with Serializable
+
+object PathSegment {
+  final case class Key(k: String) extends PathSegment
+  final case class Index(i: Int) extends PathSegment
+
+  implicit def intToPathSegment(i: Int): PathSegment = Index(i)
+  implicit def stringToPathSegment(k: String): PathSegment = Key(k)
+}


ruippeixotog:
That's a great idea, thanks for the contribution!

I have only one comment: I think this can be part of `ConfigCursor`, and not of `ConfigObjectCursor`. `atKey` is kind of a primitive operation for cursors - since we forced the user to cast the cursor to an object cursor first, that operation does a very specific thing, and without this operation we can't fully navigate through the config. On the other hand, `atPath` already has to do "casts" internally; since the operation can already fail with `WrongType`, I'd say it makes sense to make it even easier to use and allow it on generic cursors.

Incidentally, that would also greatly simplify your code, since you wouldn't have to handle the last key differently anymore.

missingfaktor:
Okay, let me submit a change in a separate commit. We can squash later if you approve of this change.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -95,6 +95,17 @@ sealed trait ConfigCursor {
   def asMap: Either[ConfigReaderFailures, Map[String, ConfigCursor]] =
     asObjectCursor.right.map(_.map)
 
+  final def atPath(key: String, moreKeys: String*): Either[ConfigReaderFailures, ConfigCursor] = {
+    val allKeysLeadingUpToLast :+ lastKey = key +: moreKeys
+    val finalObject = allKeysLeadingUpToLast
+      .foldLeft(this.asObjectCursor) {
```
ruippeixotog: Now that the method is in `ConfigCursor`, you don't need to extract and treat `lastKey` in a separate way. You can simplify this expression by folding over `key +: moreKeys` with an initial value of `this`.

ruippeixotog (changes requested):
Since `Either` in Scala 2.11 and below isn't right-biased, you have to call `.right` before each `.flatMap` on an `Either`.

missingfaktor:
Thanks, @ruippeixotog. I have added an improved version. A few notes about it:
- It can navigate not only through objects but arrays too; something I missed earlier.
- The new definition is recursive. A simple reasoning shows it will terminate, but I am unable to convince the Scala compiler of it. If you could suggest changes that would make it tail-recursive, that would be great.
- I am using `implicitConversions` here. Sadly Scala's `implicit class` mechanism does not cater to sum types, even when newly defined, so I had to use this special language feature here. Hope that's alright.

core/src/main/scala/pureconfig/Path.scala
```diff
@@ -0,0 +1,9 @@
+package pureconfig
+
+final case class Path(segments: List[Path.Segment]) extends AnyVal
```
ruippeixotog: Is there any advantage of having this class instead of using always a `Seq[Path.Segment]`? You could have a single `atPath` method if you were to remove this.

core/src/main/scala/pureconfig/syntax/package.scala
```diff
@@ -31,4 +32,7 @@ package object syntax {
     def to[T](implicit reader: Derivation[ConfigReader[T]]): Either[ConfigReaderFailures, T] = conf.root().to[T]
     def toOrThrow[T](implicit reader: Derivation[ConfigReader[T]], cl: ClassTag[T]): T = getResultOrThrow(conf.root().to[T])(cl)
   }
+
+  implicit def intToPathSegment(i: Int): Path.Segment = Path.Index(i)
```
ruippeixotog: These conversions should be better inside the companion object of `Path` or `Path.Segment`. They do not need to be imported explicitly if they're there.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -95,6 +95,18 @@ sealed trait ConfigCursor {
   def asMap: Either[ConfigReaderFailures, Map[String, ConfigCursor]] =
     asObjectCursor.right.map(_.map)
 
+  final def atPath(path: Path): Either[ConfigReaderFailures, ConfigCursor] = {
+    path.segments match {
```
ruippeixotog: In the general case, monads such as `Either` are not tail-recursive, so you can't easily make this stack safe. Can you revert this to the old `foldLeft` formulation?

ruippeixotog (changes requested):

core/src/main/scala/pureconfig/Path.scala
missingfaktor: The only advantage is a separation between a more fundamental method (one using `Path`) and one defined in a way to make it syntactically more convenient for direct use (one using `Path.Segment*`). The former is more suitable for composition. But we can always reintroduce it when needed. For now I will remove it.

core/src/main/scala/pureconfig/syntax/package.scala
missingfaktor: Okay, will do.

core/src/main/scala/pureconfig/ConfigCursor.scala
missingfaktor: Okay, will do.

missingfaktor:
@ruippeixotog I have made the requested changes. Please re-review!

ruippeixotog (approved):
LGTM, thanks again for the contribution!

missingfaktor:
Thanks, @ruippeixotog! Do you need me to add documentation, CHANGELOG entries etc, or is this ready for merge as it is?

ruippeixotog:
I don't think so, don't worry about it :) We typically wait for the approvals of two authors before merging non-trivial changes, so let's wait until someone else reviews this.

missingfaktor:
Alright, no problem. Thanks again. :)

core/src/test/scala/pureconfig/ConfigCursorSuite.scala
```diff
@@ -177,4 +177,12 @@ class ConfigCursorSuite extends BaseSuite {
     objCursor("{ a: 1, b: 2 }").map shouldBe Map("a" -> cursor("1", "a" :: defaultPath), "b" -> cursor("2", "b" :: defaultPath))
     objCursor("{}").map shouldBe Map.empty
   }
+
+  it should "allow access to a given path" in {
```
jcazevedo: Can you please move this to the behavior of `ConfigCursor` (and also use `cursor` instead of `objCursor` to create the test configurations)? The `at` method isn't specific of `ConfigObjectCursor`.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -95,6 +95,19 @@ sealed trait ConfigCursor {
...
     asObjectCursor.right.map(_.map)
 
+  final def at(pathSegment: PathSegment): Either[ConfigReaderFailures, ConfigCursor] = {
+    pathSegment match {
+      case PathSegment.Key(k) => this.asObjectCursor.right.flatMap(_.atKey(k))
+      case PathSegment.Index(i) => this.asListCursor.right.flatMap(_.atIndex(i))
+    }
+  }
+
+  final def at(pathSegments: PathSegment*): Either[ConfigReaderFailures, ConfigCursor] = {
```
jcazevedo: Can you please also add Scaladoc for these methods?

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -95,6 +95,19 @@ sealed trait ConfigCursor {
   def asMap: Either[ConfigReaderFailures, Map[String, ConfigCursor]] =
     asObjectCursor.right.map(_.map)
 
+  final def at(pathSegment: PathSegment): Either[ConfigReaderFailures, ConfigCursor] = {
```
jcazevedo: Even though the compiler always chooses the most specific method to call, having this overloaded method in the public API of `ConfigCursor` might be a bit confusing to read. What do you think of making this definition private (or just inlining it in the other `at` definition)?

jcazevedo (changes requested):
Thanks for your contribution! Any reason why you prefer `at` instead of `atPath`? I think `atPath` describes the functionality more clear and is more consistent with the `atKey` and `atIndex` methods of `ConfigObjectCursor` and `ConfigListCursor`.

core/src/test/scala/pureconfig/ConfigCursorSuite.scala
missingfaktor: Good point.

core/src/main/scala/pureconfig/ConfigCursor.scala
missingfaktor: Sure thing.

core/src/main/scala/pureconfig/ConfigCursor.scala
missingfaktor: * I named it `at` because it is a more general version of `atKey` and `atIndex`. `*Path` suffix didn't seem to add anything of value. But if you think it makes it more readable, happy to change the name back.
* I would keep the two methods separate because each does one specific thing, and I think that makes them easier to understand. I can make the non-varargs one private as the varargs one subsumes it for the users.

missingfaktor:
@jcazevedo All review points addressed.

missingfaktor:
Hello, @jcazevedo and @ruippeixotog. Do you need me to make any more changes?

jcazevedo (approved):

jcazevedo:
> Hello, @jcazevedo and @ruippeixotog. Do you need me to make any more changes?

No. It's good. Thanks again for your contribution, and sorry for the delay in reviewing this!

missingfaktor:
Thanks, @jcazevedo! :-)

PR #383, opened 2018-04-26
Use the fully qualified name of `implicitly` in Derivation

### LINKED ISSUE #382: Fail deriving ConfigReader in projects with custom Predef
I'm attaching a minimal sbt project using custom predef called `std`, if you run `sbt compile` build will fail with implicit resolution error. If you remove `import std._` from `config.scala` and remove `"-Yno-imports" and "-Yno-predef"` from `build.sbt` the project compiles

[pureconfig-custom-predef.zip](https://github.com/pureconfig/pureconfig/files/1937464/pureconfig-custom-predef.zip)

ruippeixotog:
Fixes #382.

### DIFF
--- a/macros/src/main/scala-2.10/pureconfig/Derivation.scala
+++ b/macros/src/main/scala-2.10/pureconfig/Derivation.scala
@@ -21,6 +21,6 @@ class DerivationMacros(val c: whitebox.Context) {
 
   // Derivations are not supported on Scala 2.10. This is a shim executing a regular implicit search.
   def materializeDerivation[A: WeakTypeTag]: Tree = {
-    q"_root_.pureconfig.Derivation.Successful(implicitly[${weakTypeOf[A]}])"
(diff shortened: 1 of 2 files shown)


leifwickland (approved):
Thanks, @ruippeixotog!

Do you have any ideas how we can test for this to ensure it stays fixed? My initial idea is to have another sub-project in the sbt build that is configured with no-predef.

I think projects using no-predef is a rare enough case that it might not be worth the additional effort.

ruippeixotog:
I also thought about that... While using custom predefs is indeed a rare case, this could also happen with local definitions (i.e. having a `def implicitly[A](...)` in local scope). Unfortunately, Scala macros are not hygienic 😞 I believe this is difficult to test thoroughly and any test cases we added now would probably not generalize well, as we'd be creating them based on the functions and types we use now.

PR #386, opened 2018-05-06
Update Scala, dependency and SBT plugin versions

ruippeixotog:
This PR also removes Joda as a test dependency of the core module, since it wasn't really needed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -66,11 +66,6 @@ lazy val commonSettings = Seq(
   scalacOptions in (Test, console) := (scalacOptions in (Compile, console)).value,
   scalacOptions in Tut --= Seq("-Ywarn-unused-import", "-Xmacro-settings:materialize-derivations"),
 
-  // use sbt <module_name>/test:console to run an ammonite console
-  // not published for Scala 2.12.6 (maybe now there's another way of integrating this?)
-  // libraryDependencies += "com.lihaoyi" % "ammonite" % "1.1.0" % "test" cross CrossVersion.full,
-  // initialCommands in (Test, console) := """ammonite.Main().run()""",
-
   scalariformPreferences := scalariformPreferences.value
     .setPreference(DanglingCloseParenthesis, Prevent)
     .setPreference(DoubleIndentConstructorArguments, true)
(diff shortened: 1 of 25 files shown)


docs/src/main/tut/docs/built-in-supported-types.md
```diff
@@ -36,6 +36,7 @@ Additionally, PureConfig also handles the following collections and composite Sc
   [Configurable Converters](configurable-converters.html));
 - `shapeless.HList`s of elements whose type is on this list;
 - case classes;
+- classes with only public `val` and `var` parameters in their constructor (Scala 2.11+);
```
jcazevedo, line 39: We get this from free with the new shapeless version, but perhaps we should also mention it in the change log.

ruippeixotog: Yes, I agree. Lately we've only been updating the CHANGELOG just before each version release, though.

build.sbt
```diff
@@ -67,8 +67,9 @@ lazy val commonSettings = Seq(
   scalacOptions in Tut --= Seq("-Ywarn-unused-import", "-Xmacro-settings:materialize-derivations"),
 
   // use sbt <module_name>/test:console to run an ammonite console
-  libraryDependencies += "com.lihaoyi" % "ammonite" % "1.0.3" % "test" cross CrossVersion.patch,
-  initialCommands in (Test, console) := """ammonite.Main().run()""",
+  // not published for Scala 2.12.6 (maybe now there's another way of integrating this?)
```
leifwickland: Should we just dial back to using ammonite with 2.11.12 or pull this code out entirely?

leifwickland (approved):
Thanks for doing this, @ruippeixotog.

Because of all the dependency changes, I think our next version bump should increment minor and the changelog add a passing reference about all those updates.

build.sbt
ruippeixotog: I don't use this and it's a bit cumbersome to maintain, since this seems a pretty custom or old integration that breaks frequently. I'd remove it, but if @jcazevedo, @melrief or @derekmorr use it we can leave this commented and reenable later :)

ruippeixotog:
@leifwickland I agree with the bump in the minor version and references in the CHANGELOG where applicable 👍

build.sbt
leifwickland: I'm OK killing it and looking back in the git history if we need to figure out how to restore it.

jcazevedo: I'm also OK with removing it.

ruippeixotog: Ok, I removed it in 8368801ad94cda35cf0602840f3d05e8f9507201 then! We can always turn it on again later by looking at the Git history, as @leifwickland said.

jcazevedo (approved):

PR #387, opened 2018-05-12
Support the HOCON unit syntax in Period

### LINKED ISSUE #359: Support the HOCON 1.3.2 Period Format for java.time.Period
I hoped this would Just Work™, but it appears pureconfig's `Period` support predates that of Typesafe Config 1.3.2 (lightbend/config#478) and doesn't know about the new HOCON unit syntax yet:

```scala
case class Conf(period: java.time.Period)

val config = ConfigFactory.parseString("""{ period = 5 years }""")

config.getPeriod("period") // => java.time.Period = P5Y

pureconfig.loadConfigOrThrow[Conf](config) // BOOM
```

pureconfig does handle `ConfigFactory.parseString("""{ period: P5Y }""")`, but Typesafe `getPeriod` does not, and it's not quite as nice for humans.

It's unfortunate that Typesafe Config doesn't seem to expose its parsers, so users would need to copy-and-paste a big hunk of its package-private code to define an overriding `ConfigReader[Period]` instance supporting the syntax.

I thought I wouldn't be the last to expect this to work, thus opening this issue. If I have the chance, I'll look more closely at the existing `Duration` support and try to come up with a suitable PR. Should pureconfig try HOCON format parsing and fall back to ISO-8601 `Period.parse` transparently? This doesn't quite fit the mold of existing configurable converters.

Thank you for the library, I'm enjoying the type richness and boilerplate reduction!

ruippeixotog:
I also did a little refactoring on our `checkRead` test utility method, which had a misleading signature, and made use of `checkReadString` in more places.

3faed69 adds `checkWriteString`, `checkReadWrite` and `checkReadWriteString`, which I use to increase our test coverage.

Closes #359.

### DIFF
--- a/core/src/main/scala/pureconfig/PeriodUtils.scala
+++ b/core/src/main/scala/pureconfig/PeriodUtils.scala
@@ -1,0 +1,64 @@
+package pureconfig
+
+import java.time.Period
+import java.time.temporal.ChronoUnit
+import java.time.temporal.ChronoUnit._
+
+import scala.util.Try
+
+import com.typesafe.config.impl.ConfigImplUtil
+import pureconfig.error.{ CannotConvert, FailureReason }
+
+/**
+ * Utility functions for converting a `String` to a `Period`. The parser accepts the HOCON unit syntax.
+ */
+private[pureconfig] object PeriodUtils {
+
+  /**
+   * Convert a string to a Period while trying to maintain compatibility with Typesafe's abbreviations.
+   */
+  val fromString: String => Either[FailureReason, Period] = { str =>
+    Try(Right(Period.parse(str))).getOrElse(typesafeConfigParsePeriod(str))
+  }
+
+  // This is a copy of the period parser in `com.typesafe.config.impl.SimpleConfig` adapted to be safer (dealing with
+  // `Either` values instead of throwing exceptions). Try not to refactor this too much so as to be easy for anyone
+  // to compare this code with the original.
+  def typesafeConfigParsePeriod(input: String): Either[FailureReason, Period] = {
+    val s = ConfigImplUtil.unicodeTrim(input)
+    val originalUnitString = getUnits(s)
+    var unitString = originalUnitString
+    val numberString = ConfigImplUtil.unicodeTrim(s.substring(0, s.length - unitString.length))
+    var units: ChronoUnit = null
+    // this would be caught later anyway, but the error message
+    // is more helpful if we check it here.
+    if (numberString.length == 0) return Left(CannotConvert(input, "Period", "No number in period value"))
+    if (unitString.length > 2 && !unitString.endsWith("s")) unitString = unitString + "s"
+    // note that this is deliberately case-sensitive
+    if (unitString == "" || unitString == "d" || unitString == "days") units = ChronoUnit.DAYS
+    else if (unitString == "w" || unitString == "weeks") units = ChronoUnit.WEEKS
+    else if (unitString == "m" || unitString == "mo" || unitString == "months") units = ChronoUnit.MONTHS
+    else if (unitString == "y" || unitString == "years") units = ChronoUnit.YEARS
+    else return Left(CannotConvert(input, "Period", s"Could not parse time unit '$originalUnitString' (try d, w, mo, y)"))
+    try
+      periodOf(numberString.toInt, units).left.map(CannotConvert(input, "Period", _))
+    catch {
+      case _: NumberFormatException =>
+        Left(CannotConvert(input, "Period", s"Could not parse duration number '$numberString'"))
+    }
+  }
+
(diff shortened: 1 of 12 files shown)


core/src/main/scala/pureconfig/PeriodUtils.scala
```diff
@@ -0,0 +1,61 @@
...
+  // to compare this code with the original.
+  def typesafeConfigParsePeriod(input: String): Either[FailureReason, Period] = {
+    val s = ConfigImplUtil.unicodeTrim(input)
+    val originalUnitString = getUnits(s)
+    var unitString = originalUnitString
+    val numberString = ConfigImplUtil.unicodeTrim(s.substring(0, s.length - unitString.length))
+    var units: ChronoUnit = null
+    // this would be caught later anyway, but the error message
+    // is more helpful if we check it here.
+    if (numberString.length == 0) return Left(CannotConvert(input, "Period", "No number in period value"))
```
leifwickland: Prefer `.isEmpty` to `length == 0`.

core/src/main/scala/pureconfig/PeriodUtils.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  // This is a copy of the period parser in `com.typesafe.config.impl.SimpleConfig` adapted to be safer (dealing with
+  // `Either` values instead of throwing exceptions). Try not to refactor this too much so as to be easy for anyone
+  // to compare this code with the original.
+  def typesafeConfigParsePeriod(input: String): Either[FailureReason, Period] = {
+    val s = ConfigImplUtil.unicodeTrim(input)
+    val originalUnitString = getUnits(s)
+    var unitString = originalUnitString
+    val numberString = ConfigImplUtil.unicodeTrim(s.substring(0, s.length - unitString.length))
+    var units: ChronoUnit = null
```
leifwickland: Yikes. I feel bad about allowing `var` and `null` into here just to be like the typesafe code.

core/src/main/scala/pureconfig/PeriodUtils.scala
```diff
@@ -0,0 +1,61 @@
...
+    // this would be caught later anyway, but the error message
+    // is more helpful if we check it here.
+    if (numberString.length == 0) return Left(CannotConvert(input, "Period", "No number in period value"))
+    if (unitString.length > 2 && !unitString.endsWith("s")) unitString = unitString + "s"
+    // note that this is deliberately case-sensitive
+    if (unitString == "" || unitString == "d" || unitString == "days") units = ChronoUnit.DAYS
```
leifwickland: Translating an empty string to days seems weirdly dangerous. I know you're just copying what typesafe config did, but surprises me that they made that choice. Do you know what the rationale was for defaulting to days?

ruippeixotog: Not sure, but it's [well documented](https://github.com/lightbend/config/blob/master/HOCON.md#period-format), so I don't think we should decide against it. If I was to guess, defaulting to days results in HOCON also accepting integers in `Period` configurations, a type that's arguably nicer to handle when generating or modifying configs programatically.

core/src/main/scala/pureconfig/PeriodUtils.scala
ruippeixotog: Yeah... it's originally Java code and I used IntelliJ to automatically convert it, which I think is the reason why the code looks even uglier than it's Java counterpart.

As the comment says, the reason I just copied the code was to keep the code easy to sync in case the Typesafe Config guys change anything in the parser. I think that the most important thing is to accept every format that the underlying lib accepts out-of-the-box because it eases a lot the adoption of Pureconfig by users using vanilla Typesafe Config.

If it was easy to implement, I'd add Scalacheck tests capturing this compatibility and just refactor this code to something nice. However, it's difficult to capture newly supported formats with naive generators... I think I'll just refactor this code anyway since it's not that complex.

ruippeixotog: Refactored in dcdb3ae. I also changed `DurationUtils` to use milliseconds as the default unit for durations, since it's also assumed by Typesafe Config...

ruippeixotog:
@leifwickland I think I addressed all of your comments. Let me know if you find out anything else :)

jcazevedo (approved):

ruippeixotog:
@leifwickland can you take a look at my last commit and see if the changes address all of your concerns in the comments? It would be nice to merge this and do a release soon :)

leifwickland:
@ruippeixotog I'll try to look in the next 24 hours. Sorry for the terribly long delay.

leifwickland (approved):
@ruippeixotog LGTM. 👍 
Sorry again for the terribly long delay.

ruippeixotog:
No problem, @leifwickland :) merging it now.

PR #392, opened 2018-07-08
YAML loading support

### LINKED ISSUE #61: YAML Support
It does JSON, could it also do YAML?

ruippeixotog:
Adds support to PureConfig for reading YAML files as configurations. The module uses [SnakeYAML](https://bitbucket.org/asomov/snakeyaml) to parse YAML documents and reuses the existing PureConfig structure of `ConfigReader`s and hints to read configurations to domain objects without boilerplate.

The idea of having modules for reading configurations from different formats is an alternative solution for #87 - while it still uses the Typesafe Config AST behind the scenes (which is definitely not the cleanest one around), it keeps the PureConfig core simple by not adding any extra abstraction layer.

Closes #61.


modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
```diff
@@ -0,0 +1,151 @@
...
+import pureconfig.module.yaml.error.{ NonStringKeyFound, UnsupportedYamlType }
+
+package object yaml {
+
+  // Converts an object created by SnakeYAML to a Typesafe `ConfigValue`.
+  // (https://bitbucket.org/asomov/snakeyaml/wiki/Documentation#markdown-header-loading-yaml)
+  private[this] def yamlObjToConfigValue(obj: AnyRef): Either[ConfigReaderFailures, ConfigValue] = {
+
+    def aux(obj: AnyRef): Either[ConfigReaderFailures, AnyRef] = obj match {
+      case m: java.util.Map[Object @unchecked, Object @unchecked] =>
```
leifwickland: Just curious: Why represent the types in the `Map` as `Object` rather than `AnyRef`? Also, why the `@unchecked` annotation.

When I was in a similar place with SnakeYAML I used `AnyRef`.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
```diff
@@ -0,0 +1,151 @@
...
+      case s: java.util.Set[Object @unchecked] =>
+        s.asScala.map(aux)
+          .foldLeft(Right(Set.empty): Either[ConfigReaderFailures, Set[AnyRef]])(combineResults(_, _)(_ + _))
+          .right.map(_.asJava)
+
+      case _: java.lang.Integer | _: java.lang.Long | _: java.lang.Double | _: java.lang.String | _: java.lang.Boolean =>
+        Right(obj) // these types are supported directly by `ConfigValueFactory.fromAnyRef`
+
+      case i: java.math.BigInteger =>
+        Right(i.longValue(): java.lang.Long)
```
leifwickland: Could this convert the `BigInteger` to a string and then depend on the existing `ConfigReader[BigInteger]` later?

If it worked, that'd have the advantage of avoiding overflow.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  /**
+   * Loads a configuration of type `Config` from the given YAML file.
+   *
+   * @return A `Success` with the configuration if it is possible to create an instance of type
+   *         `Config` from the YAML file, else a `Failure` with details on why it isn't possible
+   */
+  def loadYaml[Config](path: Path)(implicit reader: Derivation[ConfigReader[Config]]): Either[ConfigReaderFailures, Config] = {
```
leifwickland: I'd appreciate having a method that would read YAML from a `String`.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
```diff
@@ -0,0 +1,151 @@
...
+        Right(obj) // these types are supported directly by `ConfigValueFactory.fromAnyRef`
+
+      case i: java.math.BigInteger =>
+        Right(i.longValue(): java.lang.Long)
+
+      case _: java.util.Date | _: java.sql.Date | _: java.sql.Timestamp =>
+        Right(obj.toString)
+
+      case ba: Array[Byte] =>
+        Right(new String(ba))
```
leifwickland: This one seems sketchy, but I have no idea what else to do about it. Converting it to a String will result in some byte sequences being corrupted in the conversion to UCS-2 from the platform's default char set, I think. In most cases I believe those will be translated to a `?`.

One possible solution is to create a new `AbstractConfigValue` derived type to wrap an array of byte and then create a `ConfigReader` that is cool with that as its input.

I'd be tempted to report an error about not supporting byte strings than to support them in a way which was likely to result in corruption.

modules/yaml/src/test/scala/pureconfig/module/yaml/YamlApiSuite.scala
```diff
@@ -0,0 +1,92 @@
...
+  it should "loadYaml from a simple YAML file" in {
+    loadYaml[Conf](resourcePath("basic.yaml")) shouldBe Right(Conf(
+      "abc",
+      true,
+      Set(4, 6, 8),
+      List(10L, 10000L, 10000000L, 10000000000L),
+      Map("a" -> 1.5, "b" -> 2.5, "c" -> 3.5),
+      InnerConf(42, "def")))
+
+    loadYaml[Conf](Paths.get("nonexisting.yaml")) should failWithType[CannotReadFile]
```
leifwickland: Suggestion: I would prefer this assertion in a separate `it should ...` block.

modules/yaml/src/test/scala/pureconfig/module/yaml/YamlApiSuite.scala
```diff
@@ -0,0 +1,92 @@
...
+  it should "loadYamlOrThrow from a simple YAML file" in {
+    loadYamlOrThrow[Conf](resourcePath("basic.yaml")) shouldBe Conf(
+      "abc",
+      true,
+      Set(4, 6, 8),
+      List(10L, 10000L, 10000000L, 10000000000L),
+      Map("a" -> 1.5, "b" -> 2.5, "c" -> 3.5),
+      InnerConf(42, "def"))
+
+    a[ConfigReaderException[_]] should be thrownBy loadYamlOrThrow[Conf](Paths.get("nonexisting.yaml"))
```
leifwickland: Suggestion: I would prefer this assertion in a separate `it should ...` block.

leifwickland (commented):
Thanks for doing this! 

Please let me know what you think of my suggestions.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
ruippeixotog: The `Object` references were copy-paste, changed them to `AnyRef` in 3db81a762adb60311dedf68dc885b0743278762d.

We need the `@unchecked` since the Scala compiler cannot ensure that the type params are really we ones we asked for in pattern matches, due to type erasure. We need this annotation so that the compiler doesn't yield a warning.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
ruippeixotog: That makes total sense! Changed it in 3037283d7dcedb269e06511f525ad80e81951556.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
ruippeixotog: Agreed, should be useful. Added string loading variants in 68f76e3f3b75e97ed8bfe32d3cfe6a80eb83c1a7.

modules/yaml/src/test/scala/pureconfig/module/yaml/YamlApiSuite.scala
ruippeixotog: That's ok by me, refactored the tests in 192dec62884ed19dc8b4e0029d3911e206346834.

modules/yaml/src/test/scala/pureconfig/module/yaml/YamlApiSuite.scala
ruippeixotog: That's ok by me, refactored the tests in 192dec62884ed19dc8b4e0029d3911e206346834.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
ruippeixotog: Hmm, good point... You are saying that there are some byte sequences `byteArr` such that `new String(byteArr).getBytes != byteArr`, is that right?

SnakeYAML reads YAML fields as byte arrays if a Base64 string is specified with the official `!!binary` (http://yaml.org/type/binary.html) annotation. Since PureConfig doesn't even have a default reader for byte arrays (which seems reasonable, as most config formats support only text), I say the best way to handle this is to provide readers a Base64 string, since that's the original content of the YAML file. What do you think?

leifwickland: Yes, `new String(byteArr).getBytes != byteArr` is what I'm saying.

Converting to base 64 doesn't quite feel right to me. I can't think of why that would be more useful to the user than `Array[Byte]`. I suppose if the alternatives are no support or base 64, base 64 is preferable.

Would you like me to try to make the `AbstractConfigValue`-derived type wrapper approach I described above work?

ruippeixotog: I don't think that having a specialized `AbstractConfigValue` makes much sense, given our current abstractions. Such a solution would require a user to make a `ConfigReader` that explicitly bypassed our `ConfigCursor` abstraction and checked if the `ConfigValue` was an instance of our new `ByteArrayConfigValue` in order to take advantage of YAML binary data. Such a `ConfigReader` would be compilation-dependent on this new module and would have specialized code for a backend, which seems the type of thing we want to avoid.

Choosing to represent such fields as Base64 would be a decision based not on usefulness, but on format consistency. Given a YAML file like this:

```yaml
gif_file: !!binary |
  R0lGODlhDAAMAIQAAP//9/X17unp5WZmZgAAAOfn515eXvPz7Y6OjuDg4J+fn5
  OTk6enp56enmlpaWNjY6Ojo4SEhP/++f/++f/++f/++f/++f/++f/++f/++f/+
  +f/++f/++f/++f/++f/++SH+Dk1hZGUgd2l0aCBHSU1QACwAAAAADAAMAAAFLC
  AgjoEwnuNAFOhpEMTRiggcz4BNJHrv/zCFcLiwMWYNG84BwwEeECcgggoBADs=
```

Assuming we don't support and don't want to support byte arrays in our AST, representing the data just as it is written in the file format (`R0lG`...`ADs=`) seems the most surprise-free behavior.

ruippeixotog: (Personal opinion about this subject: even though I love the readability of the YAML format, I don't like many of their semantics; one of them is the notion of tags, that are expected to "tell" the libraries the concrete data types they should use. More often than not this leads to language-dependent config files and lack of portability, which is why I frequently advocate that parsers should concentrate on reading the data on config files just as it is represented - as very primitive data types - and let libraries such as PureConfig interpret them afterwards.)

leifwickland: If we put the `AbstractConfigValue` and an implemented `ConfigReader[ByteArray]` for in the yaml project, then I think things would basically just work out.

I'm totally find boarding the base 64 train. I just wanted to talk out what the other side would look like.

ruippeixotog: Besides my distaste for the design choice of extending the core AST to provide format-specific extensions, there's the practical aspects I can think of:
- `AbstractConfigValue`, as every class in `com.typesafe.config.impl`, is package-private. You would have to break package encapsulation and provide the new implementation in `com.typesafe.config.impl` too, which while feasible is rather hacky;
- What would happen if a user used the new `ConfigReader[Array[Byte]]` when loading a Typesafe config? What would happen if a user-provided `ConfigReader[Array[Byte]]` (for example, reading from Base64) was used with `loadYaml`?

My intention with this YAML module (and other backends) was always to support only a common subset of features, so that readers were always interchangeable. If we want to support such features, I'd prefer if we implemented a proper abstraction over the backends, with our own config AST and proper datatype extension support. But if you still think that we should investigate how an `AbstractConfigValue` implementation would look like, I can give it a try later :)

leifwickland: Oh, I didn't notice that `AbstractConfigValue` is package private. Please forget I mentioned it.

Let's go base 64.

leifwickland: @ruippeixotog, Just to make sure we're on the same page, I was going to wait to approve the PR until you'd made changes to encode the bytes as a base 64 string.

ruippeixotog: @leifwickland sure, got it :) I haven't had the chance to implement base64 encoding yet, I'll update the PR tonight or tomorrow.

ruippeixotog: Done in c0646a2. I also added some tests for null values in 6b8eaf9.

leifwickland (approved):
👍 Great work, @ruippeixotog! Thanks for talking through my quibbles.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
```diff
@@ -0,0 +1,227 @@
...
+   * Loads a configuration of type `Config` from the given YAML file.
+   *
+   * @param path the path of the YAML file to read
+   * @return A `Success` with the configuration if it is possible to create an instance of type
+   *         `Config` from the YAML file, else a `Failure` with details on why it isn't possible
+   */
+  def loadYaml[Config](path: Path)(implicit reader: Derivation[ConfigReader[Config]]): Either[ConfigReaderFailures, Config] = {
+    handleYamlErrors(Some(path)) {
+      using(Files.newBufferedReader(path)) { ioReader =>
+        val yamlObj = new Yaml(new SafeConstructor()).load(ioReader)
```
jcazevedo: Is the usage of `SafeConstructor` instead of the default `Constructor` to avoid creating custom Java instances? If so, I think we can add a comment when we create `Yaml` instances so that our intention is explicit.

jcazevedo (approved):
Sorry for the late review. Looks good!

Too bad we can't have locations for the new `ConfigReaderFailure`s. I guess we could implement our own instance of `Constructor` to keep track of these, but I don't think the extra work is justifiable.

I just left a comment on a minor issue. Feel free to ignore and merge as is.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
ruippeixotog: Yes, that's the intention. I added comments for this in 866aa34.

ruippeixotog:
Having our own `Constructor` to get error positions is a great idea, and we could use that to translate SnakeYAML `Mark` data into Typesafe `ConfigOrigin` objects. We can ship that later as an improvement, it doesn't seem very hard :)

PR #393, opened 2018-07-16
Add support of NonEmptySet into the cat's module.

NicolaeNMV:
Cats 1.1.0 added the support of `NonEmptySet`. Add the pureconfig convertors to support it.

### DIFF
--- a/modules/cats/src/main/scala/pureconfig/module/cats/package.scala
+++ b/modules/cats/src/main/scala/pureconfig/module/cats/package.scala
@@ -1,11 +1,12 @@
 package pureconfig.module
 
-import scala.language.higherKinds
-import scala.reflect.ClassTag
-
-import _root_.cats.data.{ NonEmptyList, NonEmptyVector }
+import _root_.cats.data.{ NonEmptyList, NonEmptySet, NonEmptyVector }
 import pureconfig.error.ConfigReaderFailures
 import pureconfig.{ ConfigCursor, ConfigReader, ConfigWriter }
+
+import scala.collection.immutable.SortedSet
+import scala.language.higherKinds
+import scala.reflect.ClassTag
 
 /**
  * `ConfigReader` and `ConfigWriter` instances for cats data structures.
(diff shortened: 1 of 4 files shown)


ruippeixotog (changes requested):
Thanks for your PR, @NicolaeNMV, looks good!

Can you just add tests for the `NonEmptySet` writer you wrote, so that Coveralls doesn't complain? It would also be awesome if you added writer tests for the other non-empty data structures :)

NicolaeNMV:
@ruippeixotog thanks for reviewing this so quickly ;)

I've added the missing tests plus did some refactoring to make the tests easier to read.

ruippeixotog (approved):
LGTM! Thanks again!

modules/cats/src/main/scala/pureconfig/module/cats/package.scala
```diff
@@ -31,4 +32,10 @@ package object cats {
 
   implicit def nonEmptyVectorWriter[T](implicit vectorWriter: ConfigWriter[Vector[T]]): ConfigWriter[NonEmptyVector[T]] =
     ConfigWriter.fromFunction(nonEmptyVector => vectorWriter.to(nonEmptyVector.toVector))
+
+  implicit def nonEmptySetReader[T](implicit listReader: ConfigReader[SortedSet[T]]): ConfigReader[NonEmptySet[T]] =
```
leifwickland: Nitpick: `listReader` is slightly misnamed. It's probably better called a `setReader`. Ditto on 39.

leifwickland (approved):
Thanks for this nice addition, @NicolaeNMV, and for improving the test coverage.

I sent along a trivial suggestion.

Approved either way.

NicolaeNMV:
@leifwickland I've slightly changed the naming of the vals to avoid repetition of the information that is in the type. I also made a better use of new lines to group the reader and the writer together. Can revert if that's too much. WDYT ?

leifwickland:
@NicolaeNMV That's even better than what I suggested. Thank you.

leifwickland:
@NicolaeNMV Thanks again for the contribution. Would you like to send a PR to add yourself to the [pureconfig contributors list](https://github.com/pureconfig/pureconfig/blob/6b6e0ffb118f1b80fa2254c0ddc4679cdfdc4837/CONTRIBUTORS)?

PR #401, opened 2018-07-30
Fix DateTime arbitrary in tests

### LINKED ISSUE #266: Non-deterministic failures at DateTime tests
I just saw an error occurring in Travis on d57a9c5. Since Travis does not keep a history of retried builds, I'll leave the information here:

```
[info] - should parse DateTime *** FAILED ***
[info]   IllegalInstantException was thrown during property evaluation.
[info]     Message: Illegal instant due to time zone offset transition (daylight savings time 'gap'): 1970-01-01T00:31:29.331 (America/Mazatlan)
[info]   
[info]     )
```

The full log is here: [log.txt](https://github.com/pureconfig/pureconfig/files/1065408/log.txt)

ruippeixotog:
As explained [here](http://joda-time.sourceforge.net/faq.html#illegalinstant), not every combination of `DateTimeZone` and `LocalDateTime` form a valid `DateTime`, as in spring there's a gap where a local time for that time zone doesn't exist. This PR makes us validate the pair before trying to create a `DateTime`.

Hopefully fixes #266.

### DIFF
--- a/modules/joda/src/test/scala/pureconfig/module/joda/configurable/ConfigurableSuite.scala
+++ b/modules/joda/src/test/scala/pureconfig/module/joda/configurable/ConfigurableSuite.scala
@@ -76,8 +76,8 @@ object ConfigurableSuite {
   implicit val dateTimeArbitrary: Arbitrary[DateTime] =
     Arbitrary(
       for {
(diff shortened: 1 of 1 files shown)


leifwickland (approved):
Nice find! I didn't know that existed.
👍 LGTM

PR #402, opened 2018-07-31
Switch to parallel effects in cats instances

### LINKED ISSUE #397: Cats: provide a Parallel instance for ConfigReader
`ConfigReader.zip` allows us to run two readers and get the errors from both. However the provided `MonadError[ConfigReader]` means that the `ConfigReaders` built with cats syntax (e.g. with `.mapN` combinator) will not run the second reader if the first one fails.

`cats` provides an idiomatic way to solve this. Given a newtype
```scala
final class ParConfigReader[A](val self: ConfigReader[A]) extends AnyVal
```
we can define a parallel `Applicative[ParConfigReader]` (non-implicit) and an obvious `Parallel[ConfigReader, ParConfigReader]`. This would give us combinators like `.parMapN` and `.parTraverse`, which can be used to build readers that would provide the full error list.

Otherwise, we can just make the parallel behavior the default, replacing the `MonadError` instance with an `ApplicativeError`. `cats.data.Validated` is one such type.

nigredo-tori:
Fixes #397.

The instances follow `cats.data.Validated` in that the errors are accumulated.

Tuple syntax is now supported for all three typeclasses:
```scala
val configReader1: ConfigReader[T1] = ...
...
val f: (T1, T2) => X
val g: X => (T1, T2)

// This was already present, but now it's parallel.
(configReader1,
  configReader2
).mapN(f)

// This is new
(configWriter1,
  configWriter2
).contramapN(g)

(configConvert1,
  configConvert2
).imapN(f)(g)
```

The new `Semigroup[ConfigReaderFailures]` also means we can use `.parMapN` syntax for the conversion result (`Either[ConfigReaderFailures, A]`):
```scala
(cursor.to[T1],
  cursor.to[T2]
).parMapN(f)
```

Some issues:
1. This is definitely breaking. We don't have a full `MonadError[ConfigReader]` now, and the behavior of the `ApplicativeError[ConfigReader]` has changed.
2. The `ApplicativeError[ConfigReader]` instance doesn't match the monad implied by the `ConfigReader.flatMap` method. `Validated` takes the high road here by providing the monadic bind under a different name (`.andThen`), however that would probably break a lot of existing code in our case.

### DIFF
--- a/modules/cats/src/main/scala/pureconfig/module/cats/instances/package.scala
+++ b/modules/cats/src/main/scala/pureconfig/module/cats/instances/package.scala
@@ -27,17 +27,43 @@ package object instances {
         }
     }
 
-  implicit val configWriterCatsInstance: Contravariant[ConfigWriter] = new Contravariant[ConfigWriter] {
-    def contramap[A, B](fa: ConfigWriter[A])(f: B => A): ConfigWriter[B] =
-      fa.contramap(f)
-  }
+  implicit val configWriterCatsInstance: ContravariantSemigroupal[ConfigWriter] =
+    new ContravariantSemigroupal[ConfigWriter] {
+      def contramap[A, B](fa: ConfigWriter[A])(f: B => A): ConfigWriter[B] =
+        fa.contramap(f)
 
-  implicit val configConvertCatsInstance: Invariant[ConfigConvert] = new Invariant[ConfigConvert] {
-    def imap[A, B](fa: ConfigConvert[A])(f: A => B)(g: B => A): ConfigConvert[B] =
-      fa.xmap(f, g)
-  }
+      def product[A, B](fa: ConfigWriter[A], fb: ConfigWriter[B]) =
+        ConfigWriter.fromFunction[(A, B)] {
+          case (a, b) =>
+            fb.to(b).withFallback(fa.to(a))
+        }
+    }
+
+  implicit val configConvertCatsInstance: InvariantSemigroupal[ConfigConvert] =
+    new InvariantSemigroupal[ConfigConvert] {
+      def imap[A, B](fa: ConfigConvert[A])(f: A => B)(g: B => A): ConfigConvert[B] =
+        fa.xmap(f, g)
+
+      def product[A, B](fa: ConfigConvert[A], fb: ConfigConvert[B]): ConfigConvert[(A, B)] = {
+        val reader = fa.zip(fb)
+        val writer = ConfigWriter.fromFunction[(A, B)] {
+          case (a, b) =>
+            fb.to(b).withFallback(fa.to(a))
+        }
+
+        ConfigConvert.fromReaderAndWriter(
+          Derivation.Successful(reader),
+          Derivation.Successful(writer))
+      }
+    }
 
   implicit val configValueEq: Eq[ConfigValue] = Eq.fromUniversalEquals
   implicit val configReaderFailureEq: Eq[ConfigReaderFailure] = Eq.fromUniversalEquals
   implicit val configReaderFailuresEq: Eq[ConfigReaderFailures] = Eq.fromUniversalEquals
+
+  implicit val configReaderFailuresSemigroup: Semigroup[ConfigReaderFailures] =
+    Semigroup.instance(_ ++ _)
+
+  implicit val configValueCatsSemigroup: Semigroup[ConfigValue] =
+    Semigroup.instance((a, b) => b.withFallback(a))
 }
--- a/modules/cats/src/main/scala/pureconfig/module/cats/instances/package.scala
+++ b/modules/cats/src/main/scala/pureconfig/module/cats/instances/package.scala
@@ -1,31 +1,19 @@
 package pureconfig.module.cats
 
-import scala.annotation.tailrec
-import cats.{ Contravariant, Eq, Invariant, MonadError }
+import cats._
 import com.typesafe.config.ConfigValue
 import pureconfig._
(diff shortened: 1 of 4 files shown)


nigredo-tori:
The check fails because the coverage dropped for the `Derivation` shim and for the `akka` module. Both have nothing to do with changes in this PR.

ruippeixotog:
Thanks for implementing this, @nigredo-tori! To be honest, I don't have enough experience with cats and pure FP type classes in general to know if providing an applicative instead of a monad is the right approach for `ConfigReader`.

Regarding the first issue you mentioned, I'm not too concerned about breaking changes, but rather what we lose with this change. Can you tell me what useful operations could be done previously but can't be done now? Also, wouldn't it make sense to keep providing a non-implicit `MonadError` implementation, so that users that wanted the monad behavior could use it by importing?

Regarding the second issue, I'm averse to do that change in PureConfig core right now - `flatMap` is a very common operation in the Scala library, one that non-cats users first look when they want to chain readers.

nigredo-tori:
> Can you tell me what useful operations could be done previously but can't be done now?

Basically anything that changes its parsing logic depending on a previously parsed value can get more awkward. E.g. to parse a coproduct you have to first parse the disambiguation field, and this now falls outside the default instance. It is still possible (we have the required operations), you just can't use some of the quality-of-life cats combinators (`.ifM` being the most obvious). At the moment I can't think of a good example where that would cost more than a few keystrokes... This might complicate existing code containing some convoluted iteration (e.g. parsing an array and folding it in the same time), but I can't think of a reason why such hypothetical code couldn't be rewritten using `Applicative` and `Traverse` - or using a cursor. As far as I can tell, validating or parsing from a tree generally seems to be a very `Applicative`-friendly process.

> Also, wouldn't it make sense to keep providing a non-implicit MonadError implementation, so that users that wanted the monad behavior could use it by importing?

This is called an _overlapping typeclass instance_, and is generally frowned upon in the Haskell-like FP style (which is where Cats comes from). Basically, your type should at most have one implementation of any given typeclass, so that it can be selected based solely on the type-level information. Sure, in Scala we can abuse our `implicit`-based typeclass encoding to manage this scenario, but it often adds more problems than it solves.

An idiomatic way to handle a type with two possible instances is adding a wrapper type that would use the second instance. This way we can manage our instances by explicitly switching from one type to another. We can do something like that here. The easiest way would be to provide explicit conversions between `ConfigReader` and this (with some better name):
```scala
type ConfigReaderM[A] = Kleisli[Either[ConfigReaderFailures, ?], ConfigCursor, A]
```
This looks a bit frightening, but it does have all the required instances [and then some](https://static.javadoc.io/org.typelevel/cats-core_2.12/1.2.0/cats/data/Kleisli$.html).

> Regarding the second issue, I'm averse to do that change in PureConfig core right now - flatMap is a very common operation in the Scala library, one that non-cats users first look when they want to chain readers.

I agree, this does sound extreme. I don't think that removing `flatMap` is strictly required - `pureconfig` is not FP enough to the point where such precision would outweigh convenience.

ruippeixotog (approved):
Ok, LGTM then! From my side, we can proceed with the change and reevaluate later if anyone needs operatons from other suitable instances.

Let's just wait until another review from @jcazevedo or @leifwickland before merging this.

modules/cats/src/main/scala/pureconfig/module/cats/instances/package.scala
```diff
@@ -38,19 +26,44 @@ package object instances {
...
 
   implicit val configValueEq: Eq[ConfigValue] = Eq.fromUniversalEquals
   implicit val configReaderFailureEq: Eq[ConfigReaderFailure] = Eq.fromUniversalEquals
   implicit val configReaderFailuresEq: Eq[ConfigReaderFailures] = Eq.fromUniversalEquals
+
+  implicit val configReaderFailuresSemigroup: Semigroup[ConfigReaderFailures] =
+    Semigroup.instance(_ ++ _)
+
+  implicit val configValueCatsSemigroup: Semigroup[ConfigValue] =
+    Semigroup.instance((a, b) => b.withFallback(a))
```
jcazevedo, line 68: The order of this (ando also how `ConfigWriter`s are combined in the `product` methods) got me thinking. While I think it's natural to think of a deep merge from left to right, with subsequent properties overriding the assignments of previous ones, we tend to do it the other way around in PureConfig. For example, in the `loadConfigFromFiles` method, values in later files are used to backstop missing values from previous files. I think this difference of behavior might introduce a source of confusion. What do you guys think?

nigredo-tori: Good point. I don't have a favorite here, since I only combine `ConfigObject`s with disjoint sets of keys. If we view `ConfigObject`s as maps, using right-biased union is consistent with `++` in the standard library. So no matter what we choose, we would have an inconsistency somewhere.

ruippeixotog: Good point, indeed... Even though the merge operation on config values is defined as `withFallback`, I prefer the natural, map-like semantics of `++` (and I suppose cats syntax extensions also treat it as such). `loadConfigFromFiles` doesn't seem important/idiomatic enough to set the standard in my opinion, but I understand the concern.

jcazevedo: I suppose going with map-like semantics is OK for this case.

jcazevedo (approved):

PR #404, opened 2018-08-02
Use UTF-8 charset when saving configs

nigredo-tori:
Using the default charset is generally bad practice, and HOCON must always be in UTF-8 anyway.

### DIFF
--- a/core/src/main/scala/pureconfig/package.scala
+++ b/core/src/main/scala/pureconfig/package.scala
@@ -4,7 +4,8 @@
 /**
  * @author Mario Pastorelli
  */
-import java.io.{ OutputStream, PrintStream }
+import java.io.{ OutputStream, OutputStreamWriter }
+import java.nio.charset.StandardCharsets.UTF_8
 import java.nio.file.{ Files, Path }
 
 import scala.reflect.ClassTag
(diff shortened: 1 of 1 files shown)


core/src/main/scala/pureconfig/package.scala
```diff
@@ -255,9 +256,10 @@ package object pureconfig {
    * @param outputStream The stream in which the configuration should be written
    */
   def saveConfigToStream[Config](conf: Config, outputStream: OutputStream)(implicit writer: Derivation[ConfigWriter[Config]]): Unit = {
-    val printOutputStream = new PrintStream(outputStream)
+    // HOCON requires UTF-8
```
ruippeixotog: Can you support this claim by adding a link to https://github.com/lightbend/config/blob/master/HOCON.md#unchanged-from-json on this comment?

ruippeixotog (changes requested):
LGTM! Please just take a look at the minor comment.

core/src/main/scala/pureconfig/package.scala
nigredo-tori: Done.

ruippeixotog (approved):

leifwickland:
Post facto approval from me. Sorry for my slow response time.
Thanks for the contribution, @nigredo-tori!
@nigredo-tori, please consider sending a PR to add yourself to the [pureconfig contributors file](https://github.com/pureconfig/pureconfig/blob/master/CONTRIBUTORS).

PR #405, opened 2018-08-02
Allow customizing the config rendering options

### LINKED ISSUE #403: Save functions should allow overriding ConfigRenderOptions
I'm migrating an existing config system to pureconfig, and one of the steps for this is converting the existing legacy configs to the new format. The default `ConfigRenderOptions` are not what I need here - these include origin comments, which in my case are useless. However there doesn't seem to be a way to override these in pureconfig functions (e.g. those in the `cats-effect` and `fs2` modules).

ruippeixotog:
Closes #403.

### DIFF
--- a/core/src/main/scala/pureconfig/package.scala
+++ b/core/src/main/scala/pureconfig/package.scala
@@ -252,7 +252,7 @@ package object pureconfig {
       throw new IllegalArgumentException(s"Cannot save configuration in file '$outputPath' because it already exists and is a directory")
     }
 
-    saveConfigToStream(conf, Files.newOutputStream(outputPath))
+    saveConfigToStream(conf, Files.newOutputStream(outputPath), options)
   }
 
   /**
--- a/core/src/main/scala/pureconfig/package.scala
+++ b/core/src/main/scala/pureconfig/package.scala
@@ -236,9 +236,15 @@ package object pureconfig {
    * @param conf The configuration to save
    * @param outputPath Where to write the configuration
    * @param overrideOutputPath Override the path if it already exists
+   * @param options the config rendering options
    */
   @throws[IllegalArgumentException]
-  def saveConfigAsPropertyFile[Config](conf: Config, outputPath: Path, overrideOutputPath: Boolean = false)(implicit writer: Derivation[ConfigWriter[Config]]): Unit = {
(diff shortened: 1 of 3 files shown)


leifwickland (approved):
I'm somewhat tempted to suggest that the PR ought to add a unit test to verify the efficacy of a non-default `ConfigRenderOptions`, but tht feels slightly pointless as it would only verify that TypeSafe config is doing its job.

Which is probably to say I should've said nothing and simply said good work.

nigredo-tori:
@liefwickland, if all we need is to check that the parameter is actually used, why not just enable `-Ywarn-unused:params`? This group of linting checks is [disabled due to a bug](https://github.com/pureconfig/pureconfig/blob/cd362a794c2c1d2b002b5d4bbc5f3ade2cae0a55/build.sbt#L138) that should be fixed by now.

This probably doesn't belong in this PR, though.

leifwickland:
@nigredo-tori, That's a good point. It'd be worth taking a look at updating the linting options.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -236,9 +236,15 @@ package object pureconfig {
...
    * @param overrideOutputPath Override the path if it already exists
+   * @param options the config rendering options
    */
   @throws[IllegalArgumentException]
-  def saveConfigAsPropertyFile[Config](conf: Config, outputPath: Path, overrideOutputPath: Boolean = false)(implicit writer: Derivation[ConfigWriter[Config]]): Unit = {
+  def saveConfigAsPropertyFile[Config](
+    conf: Config,
+    outputPath: Path,
+    overrideOutputPath: Boolean = false,
+    options: ConfigRenderOptions = ConfigRenderOptions.defaults())(implicit writer: Derivation[ConfigWriter[Config]]): Unit = {
```
jcazevedo, line 246: This method should supply its `options` parameter to `saveConfigToStream`.

jcazevedo (changes requested):

core/src/main/scala/pureconfig/package.scala
ruippeixotog: Well spotted, fixed in 823cccb.

jcazevedo (approved):

PR #406, opened 2018-08-02
Add a failOnReadError option to loadConfigFromFiles

### LINKED ISSUE #335: `loadConfigFromFiles` throws an exception if no files are read
The method `loadConfigFromFiles` takes a `Traversable[Path]` as an argument. It checks whether the `Traversable` is empty and handles the error by returning a `Left`.

Subsequently, if it turns out that the `Traversable[Path]` contains only missing or otherwise unreadable files, it passes this initial check but results in an exception from the call to `reduce` on an empty collection:

Here is a sample stack trace (from version `0.8.0` of pureconfig and Scala 2.12):
```scala
java.lang.UnsupportedOperationException: empty.reduceLeft
	at scala.collection.LinearSeqOptimized.reduceLeft(LinearSeqOptimized.scala:135)
	at scala.collection.LinearSeqOptimized.reduceLeft$(LinearSeqOptimized.scala:134)
	at scala.collection.immutable.List.reduceLeft(List.scala:86)
	at scala.collection.TraversableOnce.reduce(TraversableOnce.scala:208)
	at scala.collection.TraversableOnce.reduce$(TraversableOnce.scala:208)
	at scala.collection.AbstractTraversable.reduce(Traversable.scala:104)
	at pureconfig.package$.$anonfun$loadConfigFromFiles$5(package.scala:296)
	at scala.util.Either$RightProjection.map(Either.scala:714)
	at pureconfig.package$.loadConfigFromFiles(package.scala:296)
	at org.broadinstitute.gpp.aguila.Config$.load(Config.scala:43)
	at org.broadinstitute.gpp.aguila.Aguila$.main(Aguila.scala:51)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.twitter.app.App.$anonfun$nonExitingMain$3(App.scala:225)
	at scala.Option.foreach(Option.scala:257)
	at com.twitter.app.App.nonExitingMain(App.scala:224)
	at com.twitter.app.App.nonExitingMain$(App.scala:202)
	at org.broadinstitute.gpp.aguila.Aguila$.nonExitingMain(Aguila.scala:30)
	at com.twitter.app.App.main(App.scala:190)
	at com.twitter.app.App.main$(App.scala:188)
	at org.broadinstitute.gpp.aguila.Aguila$.main(Aguila.scala:30)
	at org.broadinstitute.gpp.aguila.Aguila.main(Aguila.scala)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at sbt.Run.invokeMain(Run.scala:67)
	at sbt.Run.run0(Run.scala:61)
	at sbt.Run.sbt$Run$$execute$1(Run.scala:51)
	at sbt.Run$$anonfun$run$1.apply$mcV$sp(Run.scala:55)
	at sbt.Run$$anonfun$run$1.apply(Run.scala:55)
	at sbt.Run$$anonfun$run$1.apply(Run.scala:55)
	at sbt.Logger$$anon$4.apply(Logger.scala:84)
	at sbt.TrapExit$App.run(TrapExit.scala:248)
	at java.lang.Thread.run(Thread.java:748)
```
The scaladocs include the following note:

> Note: If an element of `files` references a file which doesn't exist or can't be read, it will silently be ignored.

@jcazevedo noted in the gitter channel that the behaviour changed on #245 and reverted on #262. 

It was agreed by @jcazevedo and @melrief  that this method should not throw, but @jcazevedo pointed out that returning an error when files could not be read conflicts with the documented behavior of silently ignoring files that don't exist or can't be read.

ruippeixotog:
Closes #335.

I also chose to change the behavior of `loadConfigFromFiles` when an empty list is passed. IMO this should not be handled in a special way, since on dynamically generated lists of config files it may even make sense to call this with an empty list and get an empty config back (even if we were to actually fail it would make more sense to me to throw an `IllegalArgumentException` rather than returning a `Left`, since it was not really a PureConfig error).

### DIFF
--- a/core/src/main/scala/pureconfig/package.scala
+++ b/core/src/main/scala/pureconfig/package.scala
@@ -265,27 +265,28 @@ package object pureconfig {
   }
 
   /**
-   * Loads `files` in order, allowing values in later files to backstop missing values from prior, and converts them into a `Config`.
+   * Loads `files` in order, allowing values in later files to backstop missing values from prior, and converts them
+   * into a `Config`.
    *
    * This is a convenience method which enables having default configuration which backstops local configuration.
    *
-   * Note: If an element of `files` references a file which doesn't exist or can't be read, it will silently be ignored.
+   * The behavior of the method if an element of `files` references a file which doesn't exist or can't be read is
+   * defined by the `failOnReadError` flag. With `failOnReadError = false`, such files will silently be ignored while
+   * otherwise they would yield a failure (a `Left` value).
    *
    * @param files Files ordered in decreasing priority containing part or all of a `Config`. Must not be empty.
    */
-  def loadConfigFromFiles[Config](files: Traversable[Path])(implicit reader: Derivation[ConfigReader[Config]]): Either[ConfigReaderFailures, Config] = {
-    if (files.isEmpty) {
-      Left(ConfigReaderFailures(NoFilesToRead))
-    } else {
-      files
-        .map(parseFile)
-        .foldLeft[Either[ConfigReaderFailures, Seq[TypesafeConfig]]](Right(Seq())) {
-          case (c1, Left(failures)) if failures.toList.exists(_.isInstanceOf[CannotReadFile]) => c1
-          case (c1, c2) => ConfigConvert.combineResults(c1, c2)(_ :+ _)
-        }
-        .right.map(_.reduce(_.withFallback(_)).resolve)
-        .right.flatMap(loadConfig[Config])
-    }
+  def loadConfigFromFiles[Config](files: Traversable[Path], failOnReadError: Boolean = false)(implicit reader: Derivation[ConfigReader[Config]]): Either[ConfigReaderFailures, Config] = {
+    files.map(parseFile)
+      .map {
+        case Left(failures) if failures.toList.exists(_.isInstanceOf[CannotReadFile]) && !failOnReadError =>
+          Right(ConfigFactory.empty())
+        case conf => conf
+      }
+      .foldLeft[Either[ConfigReaderFailures, TypesafeConfig]](Right(ConfigFactory.empty())) {
+        case (c1, c2) => ConfigConvert.combineResults(c1, c2)(_.withFallback(_))
+      }
+      .right.flatMap { conf => loadConfig[Config](conf.resolve) }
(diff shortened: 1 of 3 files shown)


leifwickland (approved):
Thanks for doing this @ruippeixotog.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -265,27 +265,28 @@ package object pureconfig {
...
+   * into a `Config`.
    *
    * This is a convenience method which enables having default configuration which backstops local configuration.
    *
-   * Note: If an element of `files` references a file which doesn't exist or can't be read, it will silently be ignored.
+   * The behavior of the method if an element of `files` references a file which doesn't exist or can't be read is
+   * defined by the `failOnReadError` flag. With `failOnReadError = false`, such files will silently be ignored while
+   * otherwise they would yield a failure (a `Left` value).
    *
    * @param files Files ordered in decreasing priority containing part or all of a `Config`. Must not be empty.
```
jcazevedo: Since we now deal property with empty lists of files, I think we should remove the `Must not be empty` sentence from the parameter description.

jcazevedo (changes requested):

core/src/main/scala/pureconfig/package.scala
ruippeixotog: 👍 6b8337e

leifwickland: Good catch, @jcazevedo!

jcazevedo (approved):

PR #408, opened 2018-08-09
Add support for cats NonEmptyMap type.

NicolaeNMV:
* Rely on the existing `Map` type `ConfigReader` and `ConfigWriter`.
* The cats `NonEmptyMap` relies on `SortedMap` as the underlying implementation. The latter requires a `Ordering[A]` for the construction. That's why it is required by the implementation of NonEmptyMap's `ConfigReader`.


modules/cats/src/main/scala/pureconfig/module/cats/package.scala
```diff
@@ -34,5 +42,10 @@ package object cats {
   implicit def nonEmptySetReader[T](implicit reader: ConfigReader[SortedSet[T]]): ConfigReader[NonEmptySet[T]] =
     ConfigReader.fromCursor(fromNonEmpty[SortedSet, NonEmptySet, T](NonEmptySet.fromSet))
   implicit def nonEmptySetWriter[T](implicit writer: ConfigWriter[SortedSet[T]]): ConfigWriter[NonEmptySet[T]] =
-    ConfigWriter.fromFunction(nel => writer.to(nel.toSortedSet))
+    ConfigWriter.fromFunction(nonEmptySet => writer.to(nonEmptySet.toSortedSet))
+
+  implicit def nonEmptyMapReader[A, B](implicit reader: ConfigReader[Map[A, B]], ord: Ordering[A]): ConfigReader[NonEmptyMap[A, B]] =
```
nigredo-tori: Since we're talking `cats` here, it would make sense to use `cats.Order` instead of `Ordering`.

modules/cats/src/main/scala/pureconfig/module/cats/package.scala
```diff
@@ -20,6 +21,13 @@ package object cats {
         case Some(nonEmpty) => Right(nonEmpty)
       }
     }
+  private[pureconfig] def fromNonEmpty2[F[_, _], G[_, _], A, B](fromFT: F[A, B] => Option[G[A, B]])(cur: ConfigCursor)(implicit ct: ClassTag[F[A, B]], fReader: ConfigReader[F[A, B]]): Either[ConfigReaderFailures, G[A, B]] =
+    fReader.from(cur).right.flatMap { ft =>
+      fromFT(ft) match {
+        case None ⇒ cur.failed(EmptyTraversableFound(ct.toString))
+        case Some(nonEmpty) ⇒ Right(nonEmpty)
+      }
```
nigredo-tori: Way too complicated, IMO. Something like this should work for all `NonEmpty*` types:
```scala
private[pureconfig] def fromNonEmpty[X, Y](reader: ConfigReader[X])(fromX: X => Option[Y])(
  implicit ct: ClassTag[Y]
): ConfigReader[Y] =
  reader.emap(x => fromX(x).toRight(EmptyTraversableFound(ct.toString)))

implicit def nonEmptyListReader[T](implicit reader: ConfigReader[List[T]]): ConfigReader[NonEmptyList[T]] =
  fromNonEmpty(reader)(NonEmptyList.fromList)

implicit def nonEmptyMapReader[A, B](implicit reader: ConfigReader[Map[A, B]], ord: Order[A]): ConfigReader[NonEmptyMap[A, B]] =
  fromNonEmpty(reader)(x => NonEmptyMap.fromMap(SortedMap(x.toSeq: _*))
```
(importing `cats.syntax.option._` and `cats.instances.order._`).

**UPD**: By the way, note that we should use the `ClassTag` for the output type (`NonEmptyList`, etc) and not for the input one.

nigredo-tori (changes requested):

modules/cats/src/main/scala/pureconfig/module/cats/package.scala
NicolaeNMV: @nigredo-tori there is actually a problem with `NonEmptyMap` where `cats.Order` is not actually required. But they cannot fix it now as it would break the binary compatibility. The only thing that is required is `Ordering` to build `SortedMap`. https://github.com/typelevel/cats/pull/2318

nigredo-tori: @NicolaeNMV , I am aware that we only use `Ordering` inside. However this is probably more of a temporary (as you pointed out) implementation detail for `NonEmptyMap`, and we shouldn't leak this in our signature. See the `NonEmptyMap` implementation - it uses `Order` everywhere. We should do the same.

modules/cats/src/main/scala/pureconfig/module/cats/package.scala
NicolaeNMV: This indeed looks better.
But I've tried to compile the code you've provided but I'm getting errors.
>[error] ClassTag available for cats.data.NonEmptyMap[A,B]
[error]     fromNonEmpty(reader)(x => NonEmptyMap.fromMap(SortedMap(x.toSeq: _*)))

Any idea how to fix this ?

nigredo-tori: I see the problem now. Basically, `NonEmptyMap` is not a class/interface, but a type with some constraints, meaning that summoning a `ClassTag` for it is not possible. Reverting to `ct: ClassTag[X]` in `fromNonEmpty` should do the trick. This would be less precise, but it was good enough for the existing implementation...

Otherwise we can switch to `scala.reflect.runtime.universe.WeakTypeTag`. However, I don't remember the specifics of the `ClassTag`/`TypeTag`/`WeakTypeTag` distinction well enough to make any predictions about what this will an will not break.

modules/cats/src/main/scala/pureconfig/module/cats/package.scala
NicolaeNMV: @nigredo-tori users will have to import specific Order instances for their key type. Ex. `cats.instances.string._` to bring `Order[String]` in. Do you think that's a problem ?
There are less imports required if scala's `Ordering` is to be used, because the Ordering of the common types are in the scope by default.
A thing we could do to make the user's life easier is to have some default imports for String, we could bring Order[String] into the scope. WDYT ?
`package object cats extends StringInstances`

modules/cats/src/main/scala/pureconfig/module/cats/package.scala
NicolaeNMV: Ok, this works, thanks.

I've pushed the amendments as per PR comments.

modules/cats/src/main/scala/pureconfig/module/cats/package.scala
nigredo-tori: > Do you think that's a problem ?

@NicolaeNMV, I think that importing instances is reasonable - that's how Cats work in the first place, so this behavior would not be surprising. I just do `import cats.implicits._` most of the time, anyway.

In any case, I think it's safe to assume that if someone uses `NonEmptyMap` (not the most advertised Cats type), they know to import the corresponding `Order` instance.

> A thing we could do to make the user's life easier is to have some default imports for String, we could bring Order[String] into the scope. WDYT ?

For the love of all that is holy, no! Cats goes to great length to keep importing instances explicit and predictable. The last thing we need is some `pureconfig` package silently exposing another set of instances.

NicolaeNMV: @nigredo-tori, Ok, until now all the pureconfig cat's datatype convertors were not requiring any additional implicit. That's different for `NonEmptyMap`. I've added the information about that to the documentation so hopefully it should be easy for the users to figure it out. I usually don't have any cats specific imports inside my application where I use pureconfig. Because they are not needed. But it will be needed if the users try to use `NonEmptyMap`.

Anyway, please check my latest commit if it looks good to you.

NicolaeNMV:
@nigredo-tori I've amended the PR as discussed. Let me know if it looks good to you now.

modules/cats/README.md
```diff
@@ -16,47 +16,44 @@ libraryDependencies += "com.github.pureconfig" %% "pureconfig-cats" % "0.9.1"
 
 ### Reading cats data structures from a config
 
-To load a `NonEmptyList[Int]` into a configuration, we need a class to hold our configuration:
+The following cats data structures are supported: 
+
+* `NonEmptyList`, `NonEmptyVector`, `NonEmptySet`
+* `NonEmptyMap[K, V]` only string keys are supported by default because it relies on `Map`.
+For a custom key you'll also have to provide an implicit on Ordering[K].
```
nigredo-tori: Leftover `Ordering` reference. Also, we should mention that we also need instances for `Map[K, V]` in general case, otherwise it sounds like we only need an `Order`. Something like this:

> * `NonEmptyMap[K, V]` (from `ConfigReader[Map[K, V]]` and `Order[K]`).

Not sure it makes sense to mention that we only support strings here. That is a limitation on the instance for `Map` - let's not drag it around.

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
```diff
@@ -32,4 +36,9 @@ class CatsSuite extends BaseSuite with ConfigConvertChecks {
     val config = parseString("{ numbers: [] }")
     config.to[NumSet] should failWith(EmptyTraversableFound("scala.collection.immutable.SortedSet"), "numbers")
   }
+
+  it should "return an EmptyTraversableFound when reading empty set into NonEmptyMap" in {
```
nigredo-tori: ..,reading empty **object** ...

modules/cats/src/main/scala/pureconfig/module/cats/package.scala
```diff
@@ -1,38 +1,37 @@
...
+    fromNonEmpty(reader)(NonEmptyVector.fromVector)
   implicit def nonEmptyVectorWriter[T](implicit writer: ConfigWriter[Vector[T]]): ConfigWriter[NonEmptyVector[T]] =
     ConfigWriter.fromFunction(nonEmptyVector => writer.to(nonEmptyVector.toVector))
 
   implicit def nonEmptySetReader[T](implicit reader: ConfigReader[SortedSet[T]]): ConfigReader[NonEmptySet[T]] =
-    ConfigReader.fromCursor(fromNonEmpty[SortedSet, NonEmptySet, T](NonEmptySet.fromSet))
+    fromNonEmpty(reader)(NonEmptySet.fromSet)
   implicit def nonEmptySetWriter[T](implicit writer: ConfigWriter[SortedSet[T]]): ConfigWriter[NonEmptySet[T]] =
-    ConfigWriter.fromFunction(nel => writer.to(nel.toSortedSet))
+    ConfigWriter.fromFunction(nonEmptySet => writer.to(nonEmptySet.toSortedSet))
```
nigredo-tori: We can make this shorter:
```scala
writer.contramap(_.toSortedSet)
```
Same for other writers.

nigredo-tori (changes requested):
Other than these nitpicks, LGTM.

modules/cats/README.md
ruippeixotog: Instead of specifying the constraints of reading to `Map`s, we can explain more generally that all of these data structures rely on the instances of their unrestricted (i.e. possibly empty) variants. It would be good for users that their custom collection readers, if any, affect the behavior of these too (it is a concern of special importance for maps because there are [configurable map instances](https://github.com/pureconfig/pureconfig/blob/master/core/src/main/scala/pureconfig/configurable/package.scala#L62)).

modules/cats/src/main/scala/pureconfig/module/cats/package.scala
```diff
@@ -1,38 +1,37 @@
...
 package object cats {
 
-  private[pureconfig] def fromNonEmpty[F[_], G[_], T](fromFT: F[T] => Option[G[T]])(cur: ConfigCursor)(implicit ct: ClassTag[F[T]], fReader: ConfigReader[F[T]]): Either[ConfigReaderFailures, G[T]] =
-    fReader.from(cur).right.flatMap { ft =>
-      fromFT(ft) match {
-        case None => cur.failed(EmptyTraversableFound(ct.toString))
-        case Some(nonEmpty) => Right(nonEmpty)
-      }
-    }
+  private[pureconfig] def fromNonEmpty[X, Y](reader: ConfigReader[X])(fromX: X => Option[Y])(implicit ct: ClassTag[X]): ConfigReader[Y] =
```
ruippeixotog, line 15: Nice improvement!

ruippeixotog (changes requested):
Thank you for the contribution, @NicolaeNMV! Please just take a look at the documentation issues pointed out by @nigredo-tori and me, and afterwards we can merge this :)

leifwickland:
Note: If this is merged before 0.9.2 is released, we need to remember to update the changelog in #408.

NicolaeNMV:
Ok, thanks for the suggestions. I've made the amendments as discussed, let me know if it looks good to you.

nigredo-tori:
Looks good to me! :thumbsup:

ruippeixotog (approved):

PR #410, opened 2018-08-22
Update CHANGELOG

ruippeixotog:


leifwickland (approved):
LGTM. 👍 Thank you for doing this.
I assume you'll update PR if #408 is merged.
(That is not to say I want to hold up 0.9.2 for #408.)

ruippeixotog:
We aren't currently tracking module-only changes in CHANGELOG, so #408 isn't covered (I referenced pureconfig-yaml here just because it was the addition of a new module, but we didn't even do that for the others)... We should start doing that at some point in the future, though 👍

PR #411, opened 2018-08-24
Simplify release process

ruippeixotog:
Just a small improvement and a documentation fix I found during the last release.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -138,15 +138,17 @@ lazy val scala212LintFlags = allVersionLintFlags ++ Seq(
 // do not publish the root project
 skip in publish := true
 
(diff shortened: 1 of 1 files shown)


leifwickland (approved):
I'll approve, but I don't know enough to offer confidence that it's correct.

ruippeixotog:
Don't worry, I ended up testing this during the release process :) `releaseCrossBuild`, as the name says, makes `sbt-release` use the cross-version form of `test` by default. It also does that for `publishSigned`, but due to https://github.com/sbt/sbt-release/issues/184 we can't replace the `releaseStepCommandAndRemaining("+publishSigned")` line, too.

PR #412, opened 2018-08-25
 Move shapeless features to a separate generic module

### LINKED ISSUE #396: Disable automatic derivation by default
#304 was closed for an unrelated reason, however the issue remains.

(copied from https://github.com/pureconfig/pureconfig/issues/304#issuecomment-407960683).

Automatic derivation makes the field names etc. part of the API, which is not ideal. My team uses the explicit approach in similar cases (JSON codecs, etc.) - any encoding should be performed by an explicitly specified codec, and missing codec should cause a compilation error. Thus we need an option to disable automatic derivation entirely.

One way to achieve that (which is what `circe` project does) would be to remove `DerivedReaders`/`DerivedWriters` from `ConfigReader`/`ConfigWriter` objects, and make them accessible through an explicit import of some package object. We can even move those to a separate module, getting rid of the `shapeless` dependency in `core`. This would make this library a lot more flexible, at the cost of adding an import to the existing code.

**UPD**: to clarify, this issue is about shapeless-powered derivation for arbitrary ADTs. `Map`/`Option`/`TraversableOnce` typeclass instances should remain available by default.

ruippeixotog:
This PR does a broad refactoring of the pureconfig code base in order for users to have finer-grained control over what generic derivation features they want on by default. Auto-derivation is still used throughout all of the documentation pages and should still be the recommended way to use PureConfig. However, I added a new page to the documentation explaining the various ways to use derivation.

An overview of the major changes is shown below:

### SBT organization

- A new `generic` module is created;
- `core` does not depend on shapeless anymore and is now published as `pureconfig-core`;
- A new `bundle` SBT project that aggregates `core` and `generic` and publishes as `pureconfig` is created;
- The core tests were moved from `core` anymore to a new `tests` subproject. This prevents circular dependencies between projects, since the tests are still way too tied to the generic features.

### pureconfig-core

- Case class and sealed trait readers/writers, as well as their hints, were moved from `core` to `generic`;
- A new `Exported` wrapper was added in order to encode low-priority pluggable instances of a case class (see scaladoc);
- The `DerivedReaders` trait was split into `CollectionReaders` and `ExportedReaders` (same for writers);

### pureconfig-generic

- Simple `auto` and `semiauto` objects can be imported to access generic derivations;
- The previous `WrappedConfigReader` instances were very misleading; while their objective was to hold the original `Wrapped` instances for fetching hints, many of our derivations only worked properly because they were not defined as regular `ConfigReader`s 😞 In order to get this refactor to work I had to make this fact much more explicit, creating new type classes `SeqShapedReader` and `MapShapedReader`. These internal type classes are not exposed to end users, unless they are writing advanced derivation code. (idem for writers);
- `DerivedConfigReader` and `ExportMacros`, together with core's `Exported`, are used for us to have finer-grained control over implicit precedence while leaving user code working as they expect by following Scala implicit precedence rules. I based this structure on Circe, which does this successfully.

Closes #396.


docs/src/main/tut/docs/non-automatic-derivation.md
```diff
@@ -0,0 +1,108 @@
...
+programming Scala library.
+
+Custom readers can still be placed in the companion object of the respective classes; PureConfig will make sure that
+they are used instead of the automatically-derived ones.
+
+### Semi-Automatic
+
+With semi-automatic derivation, readers can still be derived using all the machinery presented on these documentation
+pages, but the reader instances are not provided as implicits. Instead, PureConfig provides two one-liner methods which
+that can be used to create derived instances, which you must put somewhere on the implicit scope.
```
jcazevedo: This last sentence is a bit confusing. What do you think of `PureConfig provides two one-liner methods to create derived instances, which you must put somewhere in the implicit scope`?

tests/src/test/scala/pureconfig/NonAutoDerivationSuite.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  sealed trait Entity
+  case class Person(name: String, surname: String) extends Entity
+  case class Place(name: String, lat: Double, lon: Double) extends Entity
+
+  val person = Person("John", "Doe")
+  val conf = ConfigFactory.parseString("{ type: person, name: John, surname: Doe }")
+}
+
+class NonAutoDerivationSuite extends BaseSuite {
```
jcazevedo: This tests the behavior of both `auto` and `semiauto`. I think the suite should be named differently, something along the lines of `DerivationStrategiesSuite` or similar.

jcazevedo (changes requested):
Looks good! Thanks for doing this. I just pointed out some minor nitpicks.

ruippeixotog:
@jcazevedo thanks for reviewing this! I followed both of your suggestions (see 4c0ce0b).

build.sbt
```diff
@@ -4,24 +4,42 @@ import microsites._
 
 lazy val core = (project in file("core")).
   enablePlugins(TutPlugin, SbtOsgi).
-  settings(commonSettings, tutTargetDirectory := file(".")).
-  dependsOn(macros).
+  settings(commonSettings).
+  dependsOn(macros)
+
+// A special module for now, since tests depend on it. We should improve this organization later.
```
leifwickland: What specifically would you do to improve the situation? Would you prefer to avoid the need for generic derivation in the tests and with tut? It seems like that either ought to be described in a comment or an issue.

docs/src/main/tut/docs/non-automatic-derivation.md
```diff
@@ -0,0 +1,108 @@
...
+  } yield Person(name, surname)
+}
+```
+
+```tut:book
+pureconfig.loadConfig[Person](conf)
+```
+
+If you don't need reader or writer derivation anywhere in your project, you can replace the `pureconfig` Maven
+dependency by `pureconfig-core`. `pureconfig-core` contains only the core classes needed by PureConfig, as well as
```
leifwickland: Nitpick: "dependency with" (not "by") sounds more natural.

build.sbt
```diff
@@ -4,24 +4,42 @@ import microsites._
...
   dependsOn(macros % "test->test") // provides helpers to test pureconfig macros
 
+// aggregates pureconfig-core and pureconfig-generic with the original "pureconfig" name
+lazy val bundle = (project in file("bundle")).
+  enablePlugins(TutPlugin, SbtOsgi).
+  settings(commonSettings).
+  settings(name := "pureconfig", tutTargetDirectory := file(".")).
+  dependsOn(core, generic)
+
 lazy val docs = (project in file("docs")).
```
leifwickland, line 33: Have you tried to `publish` or `publishLocal`? When tried local it failed on the docs project with:
```
[error] java.lang.IllegalStateException: Ivy file not found in cache for com.github.pureconfig#docs_2.11;0.9.3-RC1!
```

I'll admit I didn't try to determine the cause.

leifwickland (changes requested):
Sorry this review is so late in coming.

Before you merge this PR, I'd like to try to update a work project to use this branch to shake out any more surprises. Unfortunately, my first attempt at it met a sad fate when I realized all the refined types in the project weren't supported. I'm hoping I'll have time soon.

build.sbt
ruippeixotog: The issue here is that we make each module declared with the `module` function depend on `tests` in order for them to have access to test helpers like `BaseSuite`, creating a circular dependency in the case of the `generic` module. There are a lot of things module tests have access to that they don't need (like specific test classes). 

I think that the best way to organize this is to have testkit-like code (or it can simply be code in `tests/src/main/scala`, if we don't want yet another subproject) that does not depend on `generic` so that we can have the dependencies like:

```
            +---------+
   +--------> testkit +------------------------+-------+
   |        +---------+                        |       |
   |                                           |       |
+--+---+    +---------+                    +---v---+   |
| core +----> generic +--------------------> tests |   |
+--+---+    +---------+                    +-------+   |
   |                                                   |
   |        +-------------------------+    +-----------v----------+
   +--------> other modules (compile) +----> other modules (test) |
            +-------------------------+    +----------------------+
```

I didn't comment this better because I was planning to do that (as well as other minor improvements) right after this PR was accepted, but I'll add a comment here explaining this better :)

ruippeixotog: Regarding tut, I think that all our documentation should continue to rely on having `generic` as dependency, since that's by far the most common PureConfig use case.

docs/src/main/tut/docs/non-automatic-derivation.md
ruippeixotog: It does, indeed. Thanks for the correction!

build.sbt
ruippeixotog: I'm not having any problem running `publishLocal` here 😞 Maybe a `clean` on your side will help?

ruippeixotog:
> Before you merge this PR, I'd like to try to update a work project to use this branch to shake out any more surprises. Unfortunately, my first attempt at it met a sad fate when I realized all the refined types in the project weren't supported. I'm hoping I'll have time soon.

Not only that seems reasonable, but also a great way to put this refactoing to test in a real-world project. Let me know if I can be of any help, @leifwickland 👍

build.sbt
jcazevedo: I'm also being able to run `publishLocal` successfully here.

jcazevedo (approved):

ruippeixotog:
@leifwickland do you have any idea of when you might be able to test this? Since this PR is open for more than one month and there was demand for this feature (including mine, admittedly), I'd like to release a new version soon.

You mentioned that you noticed a bug with refined types; can you post a minimal example where a refined type can't be derived? I can help you debug or tackle the issue :)

leifwickland:
@ruippeixotog I think I'm probably a few weeks from having the time to look at it harder. I'll leave my approval so you can move forward.

I _think_ the problem was refined types, but that was simply guessing based on the fact that not having an updated version of the refined support package seemed like the most salient difference what works in pureconfig and what broke in my production project. My hope is that the problem will be eliminated when a new version of the refined package is built.

leifwickland (approved):

ruippeixotog:
@leifwickland ok, I'll merge this now since I still want to do some minor improvements before the release. I'll try to get a large-scale project migrated to PureConfig 0.10 before releasing it, though. Hope that's ok with you :)

leifwickland:
@ruippeixotog That's totally OK with me. Thanks for all your hard work. Sorry I'm not helpful.

jcazevedo:
Just an FYI: I've updated a work project to start using the changes introduced by this PR. With the fix of #425 everything works correctly.

I experimented with fully automatic derivation at first and then migrated to semi-automatic derivation. I was able to get a 33% speedup in compilation time by using semi-automatic derivation.

PR #415, opened 2018-09-17
Apply the same automatic type conversions HOCON allows at the cursor level

jcazevedo:
This PR adapts `ConfigCursor`s to allow traversing `ConfigObject`s with only integer keys as lists. We currently have this logic in our `ConfigReader` for traversables. However, since `com.typesafe.Config` supports conversion from objects with integer keys to `ConfigList`s (via its `get*List` methods), I think it's fitting that we also support it in `ConfigCursor`.


core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -66,7 +66,33 @@ sealed trait ConfigCursor {
...
+    } else {
+      val ve = value.valueType() match {
+        case ConfigValueType.OBJECT =>
+          val ll =
+            value.asInstanceOf[ConfigObject].asScala.foldLeft[Either[List[String], List[(Int, ConfigValue)]]](Right(Nil)) {
+              case (acc, (str, v)) =>
+                (acc, Try(str.toInt)) match {
+                  case (Right(a), Success(b)) => Right(a :+ (b -> v))
+                  case (Left(keys), Failure(_)) => Left(keys :+ str)
+                  case (_, Failure(_)) => Left(List(str))
```
nigredo-tori: From the [HOCON Spec](https://github.com/lightbend/config/blob/master/HOCON.md#conversion-of-numerically-indexed-objects-to-arrays):
> the conversion should ignore any keys which do not parse as positive integers.

Though the existing implementation treats this case as an error... I'd like to get the input from one of the maintainers on this.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -66,7 +66,33 @@ sealed trait ConfigCursor {
...
+                  case (Left(keys), _) => Left(keys)
+                }
+            }
+
+          ll.left.map(CannotConvertObjectToList)
+            .right.map(l => ConfigValueFactory.fromIterable(l.sortBy(_._1).map(_._2).asJava))
```
nigredo-tori: From the [HOCON Spec](https://github.com/lightbend/config/blob/master/HOCON.md#conversion-of-numerically-indexed-objects-to-arrays):

> the conversion should not occur if the object is empty or has no keys which parse as positive integers.

But, again, the existing implementation doesn't follow this.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -66,7 +66,33 @@ sealed trait ConfigCursor {
...
+      failed(KeyNotFound.forKeys(path, Set()))
+    } else {
+      val ve = value.valueType() match {
+        case ConfigValueType.OBJECT =>
+          val ll =
+            value.asInstanceOf[ConfigObject].asScala.foldLeft[Either[List[String], List[(Int, ConfigValue)]]](Right(Nil)) {
+              case (acc, (str, v)) =>
+                (acc, Try(str.toInt)) match {
+                  case (Right(a), Success(b)) => Right(a :+ (b -> v))
+                  case (Left(keys), Failure(_)) => Left(keys :+ str)
```
nigredo-tori: `:+` has linear complexity. Since we don't need to preserve the order here (`ConfigObject` is unordered), it's better to use `str :: keys`. Same above.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -66,7 +66,33 @@ sealed trait ConfigCursor {
    *         otherwise.
    */
   def asListCursor: Either[ConfigReaderFailures, ConfigListCursor] =
-    castOrFail(LIST, Set(LIST), _.asInstanceOf[ConfigList]).right.map(ConfigListCursor(_, pathElems))
+    if (isUndefined) {
+      failed(KeyNotFound.forKeys(path, Set()))
+    } else {
+      val ve = value.valueType() match {
+        case ConfigValueType.OBJECT =>
```
nigredo-tori: This subtly changes the behavior of `asCollectionCursor` - it will now yield `Left` for array-like objects!

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -66,7 +66,33 @@ sealed trait ConfigCursor {
...
+              case (acc, (str, v)) =>
+                (acc, Try(str.toInt)) match {
+                  case (Right(a), Success(b)) => Right(a :+ (b -> v))
+                  case (Left(keys), Failure(_)) => Left(keys :+ str)
+                  case (_, Failure(_)) => Left(List(str))
+                  case (Left(keys), _) => Left(keys)
+                }
+            }
+
+          ll.left.map(CannotConvertObjectToList)
```
nigredo-tori: Should probably sort the keys alphabetically for nicer error message.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -66,7 +66,33 @@ sealed trait ConfigCursor {
...
+            .right.map(l => ConfigValueFactory.fromIterable(l.sortBy(_._1).map(_._2).asJava))
+        case ConfigValueType.LIST =>
+          Right(value.asInstanceOf[ConfigList])
+        case other =>
+          Left(WrongType(other, Set(LIST, OBJECT)))
+      }
+
+      ve.left.flatMap(failed)
```
nigredo-tori: `ve.left.flatMap` looks slightly unusual for me, especially since we can't change from `Left` to `Right` here... How about `scopeFailure(ve)` instead?

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: Agreed. Fixed that in 6fb1c520eff4b7333f3df2fef50a47764e6c4c7c.

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: Makes sense. I've sorted them alphabetically in ee993eb969ef01a53b1603ae96d8a5e07d80af7b.

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: `scopeFailure` is indeed clearer. I've started using it in 1bf6c75547bec73a9f1d2d56f080d1e6f8be0104.

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: I'm inclined to follow the HOCON spec on this. I think we can be opinionated in our readers and writers but strive to follow the spec on cursors. @ruippeixotog and @leifwickland: what do you think?

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: Same as above. I prefer following the HOCON spec on cursors, but I'll wait for @ruippeixotog and @leifwickland's input.

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: How attached are we on keeping the `asCollectionCursor` method? If a config value can be seen as both a list and an object, then its return type is a bit confusing. I propose removing it from the `ConfigCursor` trait and replicate this logic where we need it internally (currently only on the derivation of tuple instances). @ruippeixotog and @leifwickland: what do you think?

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: I agree. In fact, we have a lot of HOCON idiosyncrasies like this throughout our readers that should be concentrated on a lower-level layer like `ConfigCursor`. In the same way, we have some HOCON idiosyncrasies that we are not following because we try to be strict with the config value types we're given.

The HOCON spec has a whole section about [Automatic type conversions](https://github.com/lightbend/config/blob/master/HOCON.md#automatic-type-conversions), and it's implemented by Typesafe Config's [DefaultTransformer](https://github.com/lightbend/config/blob/master/config/src/main/java/com/typesafe/config/impl/DefaultTransformer.java) class. Since the class is package-private, I suggest we write our own implementation (probably using `Either` for errors) and start using it on every typed access through `ConfigCursor`. What do you think, @jcazevedo?

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: If we do the `DefaultTransformer` transformations under the hood, we don't need an `asCollectionCursor` config anymore. That said, I'd prefer to keep the method deprecated in our API if we can implement the old behavior here.

core/src/main/scala/pureconfig/error/FailureReason.scala
```diff
@@ -158,3 +158,18 @@ final case class WrongSizeString(expected: Int, found: Int) extends FailureReaso
 final case class NoValidCoproductChoiceFound(value: ConfigValue) extends FailureReason {
   def description = s"No valid coproduct choice found for '${value.render(ConfigRenderOptions.concise())}'."
 }
+
+/**
+ * A failure reason given when a ConfigObject can't be converted to a ConfigList.
+ *
+ * @param failingKeys the list of keys whose conversion to integer failed
+ */
+final case class CannotConvertObjectToList(failingKeys: List[String]) extends FailureReason {
```
ruippeixotog: Do you think it's worth it to add a specific `CannotConvert`-like class for the conversion between objects and lists? We've been using `CannotConvert` for all type conversion errors throughout the codebase... I don't know if there are other cases with conversion-specific data to provide like this one.

HOCON clearly fuzzies the separation of various types in their AST with their auto-conversions. As a stretchy interpretation, I'd be tempted to consider the type of an object with numeric keys as both a list and an object and use `WrongType` here by saying that we were expecting a list but got an object.

core/src/test/scala/pureconfig/ConfigCursorSuite.scala
```diff
@@ -39,24 +39,33 @@ class ConfigCursorSuite extends BaseSuite {
...
+      Right(ConfigListCursor(conf("""["a", "b"]""").asInstanceOf[ConfigList], defaultPath))
+
+    cursor("{ 10: a, 3: b }").asListCursor shouldBe
+      Right(ConfigListCursor(conf("""["b", "a"]""").asInstanceOf[ConfigList], defaultPath))
   }
 
   it should "allow being casted to a list of cursors in a safe way" in {
     cursor("abc").asList should failWith(
-      WrongType(ConfigValueType.STRING, Set(ConfigValueType.LIST)), defaultPathStr)
+      WrongType(ConfigValueType.STRING, Set(ConfigValueType.LIST, ConfigValueType.OBJECT)), defaultPathStr)
```
ruippeixotog: It seems strange to provide an error saying that we were expecting a `LIST` or an `OBJECT` when we used `asList`. I do not have a better idea, but it is a bit misleading for those who do not know these auto-conversions in HOCON (many people, I believe).

This is related to my comment above regarding typing. We are trying to make strong AST typing and auto-conversions coexist peacefully, which is bound to lead to strange decisions in the error reporting and handling aspect.

ruippeixotog (changes requested):

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: I like that idea. I'll update the PR accordingly to support the type conversions `DefaultTransformer` provides in `ConfigCursor`.

core/src/main/scala/pureconfig/error/FailureReason.scala
jcazevedo: I personally never use `FailureReason`s directly, but I think it's a good idea to provide as much actionable information as possible when we fail reading something. I haven't gone through all usages of `CannotConvert` in the codebase, but most of them build upon exception messages, from which we can't extract much.

That being said, I'm also OK with using `CannotConvert` with a nice message.

core/src/test/scala/pureconfig/ConfigCursorSuite.scala
jcazevedo: I also had some doubts about this. I suppose it's OK to always say we were expecting a `LIST` when the conversion failed. It's more informative and less confusing for people who are not aware of the automatic conversions.

jcazevedo:
I've updated this PR to apply, at the `ConfigCursor` level, the same automatic type conversions HOCON allows. This changes some things:

- `ConfigCursor`s now support the `asBoolean`, `asLong`, `asInt`, `asShort`, `asDouble` and `asFloat` methods, that raise `WrongType` failures if the types don't match or if there's no automatic type conversion allowed;
- Basic readers that relied on conversions from `String` and raised `CannotConvert` failures now raise `WrongType` when the `ConfigValue` type is incompatible;
- Deprecates the `asCollectionCursor` method, since now a `ConfigObject` can in some cases be automatically converted to a `ConfigList` and its behavior can be a bit unpredictable.

jcazevedo:
With the new automatic conversions in place, I also took the opportunity to remove the private `stringToEitherConvert`, which blindly converted every `ConfigValue` to a `String` and make the `ConvertHelpers.fromString` use the `asString` method of `ConfigCursor`. This improves the errors of readers using the `fromString` helper by raising `WrongType` if the `ConfigValue` can't be converted to a `String`.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -173,20 +228,13 @@ sealed trait ConfigCursor {
...
-    } else if (!acceptedConfigTypes.contains(value.valueType)) {
-      failed(WrongType(value.valueType, Set(expectedType)))
     } else {
-      Try(cast(value)) match {
-        case Success(v) => Right(v)
-        case Failure(ex) =>
-          // this should never happen, this is just a safety net
-          failed(CannotConvert(value.toString, expectedType.toString, ex.toString))
-      }
+      scopeFailure(ConfigCursor.transform(value, expectedType))
```
ruippeixotog: Suggestion: instead of always wrapping calls in `scopeFailure`, you can wrap the whole expression once.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -57,7 +57,61 @@ sealed trait ConfigCursor {
...
+    castOrFail(BOOLEAN, v => Right(v.unwrapped.asInstanceOf[Boolean]))
+
+  /**
+   * Casts this cursor to a long.
+   *
+   * @return a `Right` with the long value pointed to by this cursor if the cast can be done, `Left` with a list of
+   *         failures otherwise.
+   */
+  def asLong: Either[ConfigReaderFailures, Long] =
+    castOrFail(NUMBER, { v => (ConvertHelpers.catchReadError[Long](_.toLong) _)(v.unwrapped.toString) })
```
ruippeixotog: For all numeric conversion methods, I'd rather see strategies that didn't involve converting the number to a string just to parse it again.

What do you think of relying on the internal knowledge that Typesafe Config only has `ConfigInt`, `ConfigLong` and `ConfigDouble` implementations and pattern-matching `v.unwrapped` on the primitives of those three cases instead? We can also try to do something with `Number` (since a `castOrFail(NUMBER, _)` is guaranteed to return a `ConfigNumber`), although it's not as clean.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -201,6 +249,69 @@ object ConfigCursor {
...
+      case (ConfigValueType.NUMBER | ConfigValueType.BOOLEAN, ConfigValueType.STRING) =>
+        Right(ConfigValueFactory.fromAnyRef(configValue.unwrapped.toString))
+
+      case (ConfigValueType.OBJECT, ConfigValueType.LIST) =>
+        val obj = configValue.asInstanceOf[ConfigObject].asScala
+        val ll = obj.foldLeft[Either[List[String], List[(Int, ConfigValue)]]](Right(Nil)) {
+          case (acc, (str, v)) =>
+            (acc, Try(str.toInt)) match {
+              case (Right(a), Success(b)) => Right((b -> v) :: a)
+              case (Left(keys), Failure(_)) => Left(str :: keys)
```
ruippeixotog: If I understand this well, `acc` never becomes a `Left` and even if it did, are not doing anything with the invalid keys. Can we simplify this?

(I didn't know HOCON implementations should ignore non-integer keys in this case. What a strange behavior...)

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -272,6 +383,9 @@ case class ConfigListCursor(value: ConfigList, pathElems: List[String], offset:
    */
   def list: List[ConfigCursor] =
     value.asScala.toList.drop(offset).zipWithIndex.map { case (cv, idx) => ConfigCursor(cv, indexKey(idx) :: pathElems) }
+
+  // Avoid resetting the offset when using ConfigCursor's implementation.
+  override def asListCursor: Either[ConfigReaderFailures, ConfigListCursor] = Right(this)
```
ruippeixotog, line 392: Well spotted 👍 For efficiency reasons, can you override `asObjectCursor` on `ConfigObjectCursor`, too? Also, can you add a test covering this offset bug?

core/src/main/scala/pureconfig/ConvertHelpers.scala
```diff
@@ -30,18 +29,6 @@ trait ConvertHelpers {
     case Failure(e) => Left(ExceptionThrown(e))
   }
 
-  private[pureconfig] def stringToEitherConvert[T](fromF: String => Either[FailureReason, T]): ConfigCursor => Either[ConfigReaderFailures, T] = { cur =>
```
ruippeixotog, line 33: Good thing we got rid of one more helper method from this object :)

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -121,6 +175,7 @@ sealed trait ConfigCursor {
    * @return a `Right` with this cursor as a list or object cursor if the cast can be done, `Left` with a list of
    *         failures otherwise.
    */
+  @deprecated("Use `asListCursor` and/or `asCollectionCursor` instead", "0.10.0")
```
ruippeixotog: You meant `asListCursor and/or asObjectCursor` here, right?

ruippeixotog (changes requested):

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: Makes sense. I've wrapped the whole expression in `scopeFailure` in bac6db4.

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: Makes sense. I've ended up using `Number` and playing along with its `*Value()` methods in 98499f5. This has the added benefit of allowing numbers represented in the decimal system but with no fractional part to be read as `Int`, `Long` or `Short`.

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: I ended up naively changing the old code (in which we used the keys) to fully support the HOCON specification. This can indeed be simplified. I did that in d6f0b84.

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: Sure. I've overridden `asObjectCursor` in 7f37c88 and added a test covering the offset bug in 4823cab.

core/src/main/scala/pureconfig/ConfigCursor.scala
jcazevedo: Yes, of course. Fixed in 9f33a80.

ruippeixotog (approved):
Thanks for this, @jcazevedo, LGTM!

core/src/main/scala/pureconfig/BasicReaders.scala
```diff
@@ -23,28 +23,41 @@ import pureconfig.error._
...
+
+    cur.asString.right.flatMap(s => cur.scopeFailure(asStringReader(s)))
+      .left.flatMap(_ => cur.asFloat)
+  })
+
+  implicit val intConfigReader = ConfigReader.fromCursor(_.asInt)
+
+  implicit val longConfigReader = ConfigReader.fromCursor(_.asLong)
+
+  implicit val shortConfigReader = ConfigReader.fromCursor(_.asShort)
```
leifwickland, line 60: If I read the code correctly, it appears that this change loses the special case error message for an empty input string. That's a little disappointing.

leifwickland (approved):
Good idea, @jcazevedo. LGTM.

I left one minor quibble. Do what you'd like with that.

core/src/main/scala/pureconfig/BasicReaders.scala
jcazevedo: That's true. With this change, I made an effort to avoid exposing the automatic type conversions in the failure reasons. 

I think the `EmptyStringFound` reason should only be raised when reading types that expect their underlying `ConfigValue` to be a `ConfigString`. Numeric types expect the value to a be a `ConfigNumber`, even though the HOCON specification allows certain `ConfigString`s to be automatically converted to `ConfigNumber`.

leifwickland: OK!

PR #416, opened 2018-09-27
Update cats to 1.4.0

kubukoz:


kubukoz:
I just realised cats doesn't support Scala 2.10 since 1.3.0, so we'll need to remove 2.10 from the versions this module is published to...

modules/cats/build.sbt
```diff
@@ -1,8 +1,10 @@
 name := "pureconfig-cats"
 
+val catsVersion = "1.4.0"
+
 libraryDependencies ++= Seq(
-  "org.typelevel" %% "cats-core" % "1.1.0",
-  "org.typelevel" %% "cats-laws" % "1.1.0" % "test",
+  "org.typelevel" %% "cats-core" % catsVersion,
```
ruippeixotog: Nitpick: can you leave the versions defined in-place as before? Since it is only used twice near one another, I think it becomes more readable if we just write the same thing twice.

ruippeixotog (commented):
Hi @kubukoz, thanks for the PR! In fact, #412 will bring the end of PureConfig for Scala 2.10, so this module won't need to be defined differently. This looks good to me, but let's wait until #412 is merged so that we can just update the version.

kubukoz:
Good to know! I'll get back to this one after #412 is closed then.

modules/cats/build.sbt
kubukoz: Sure, if that works better for you I'll inline :)

kubukoz:
I see #412 was merged so I'll get back to this one :)

kubukoz:
Should be good to go.

build.sbt
```diff
@@ -44,7 +44,7 @@ def module(proj: Project) = proj.
   settings(commonSettings, tutTargetDirectory := baseDirectory.value)
 
 lazy val akka = module(project) in file("modules/akka")
-lazy val cats = module(project) in file("modules/cats")
+lazy val cats = (module(project) in file("modules/cats"))
```
leifwickland: Nitpick: why this change? It appears to be unnecessary and inconsistent with the surrounding code.

leifwickland (approved):
Thanks for the contribution, @kubukoz!

build.sbt
kubukoz: Yeah, it's a leftover from when I was removing a 2.10 build. I'm rolling it back

jcazevedo (approved):
Thanks for this, @kubukoz!

PR #417, opened 2018-09-27
Update cats-effect, fs2

kubukoz:
Currently at
- cats-effect v1.0.0
- fs2 v1.0.0
- http4s v0.19.0-M3 (waiting for a release compatible with fs2 v1.0.0)

WIP because I doubt we want to merge incompatible/unstable dependencies 🗡 

(I'll update the tut tests around tomorrow)

### DIFF
--- a/modules/cats-effect/build.sbt
+++ b/modules/cats-effect/build.sbt
@@ -1,7 +1,7 @@
 name := "pureconfig-cats-effect"
 
 libraryDependencies ++= Seq(
-  "org.typelevel" %% "cats-effect" % "0.10.1")
+  "org.typelevel" %% "cats-effect" % "1.0.0")
 
 developers := List(
   Developer("keirlawson", "Keir Lawson", "keirlawson@gmail.com", url("https://github.com/keirlawson")))
--- a/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
+++ b/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
@@ -19,9 +19,9 @@ package object catseffect {
 
   private def configToF[F[_], A](getConfig: () => Either[ConfigReaderFailures, A])(implicit F: Sync[F], reader: Derivation[ConfigReader[A]], ct: ClassTag[A]): F[A] = {
     val delayedLoad = F.delay {
-      getConfig().leftMap(ConfigReaderException[A])
+      getConfig().leftMap[Throwable](ConfigReaderException[A])
     }
-    delayedLoad.flatMap(F.fromEither)
+    delayedLoad.rethrow
   }
 
   /**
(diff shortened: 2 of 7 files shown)


kubukoz:
Seeing that http4s 0.19 might not be out for some time, maybe it'll be a good idea to separate it into another PR... I'll ask around and see when we can expect a release.

ruippeixotog:
Hi @kubukoz, we'll probably release a new version of PureConfig in the next days. Do you want to do the last steps to take this PR out of WIP? :)

kubukoz:
Hi, i'm afraid one way or another we'll have binary compatibility issues or an unstable version of http4s. Http4s has been released at 0.19.0, but it was a release made mistakenly instead of 0.19.0-M4, and they'll most likely break compat by 0.19.1. if we release this without the http4s changes, the old version will be incompatible with the cats-effect and fs2 versions we update here :/

ruippeixotog:
Let's leave http4s out of the update then. In my opinion, modules must be treated independently and if some module is ready to be updated while others aren't, we should update what we can regardless of whether the two modules work frequently together.

People using only cats-effect and/or fs2 will still benefit from the upgrade, as well as http4s users that don't rely on cats-effect and fs2 modules. http4s users using other modules can keep using PureConfig 0.9 until a stable http4s version is released. What do you think?

kubukoz:
Sure, sounds like a plan. So the http4s module won't be released in this round?

ruippeixotog:
We always release a new version for all modules at once, but the http4s module will still target the old version.

kubukoz:
So it'll use incompatible dependencies, I imagine

ruippeixotog:
I'm not sure if I'm understanding the issue, but if you mean that the next version of `pureconfig-http4s` won't be usable with the next version of `pureconfig-fs2` or `pureconfig-cats-effect`, that's right, those will be incompatible. We do not guarantee dependency compatibility between different PureConfig modules, we only guarantee compatibility of each module with the latest PureConfig version.

(Let's put things into perspective here: the http4s module has literally one reader, and a trivial one at that. I'm sure people won't find too much trouble to write their own `Uri` readers in the meantime 🙂)

kubukoz:
Fair enough. I didn't look that much at the code in the module ;) I'll follow up with the changes soon!

kubukoz:
@ruippeixotog I reverted the http4s change, rebased & squashed. Sorry for stalling. We should be ready to go with this now :)

leifwickland:
@kubukoz It looks like the build is broken due to failures when running tut for fs2. See https://travis-ci.org/pureconfig/pureconfig/jobs/442571569#L1488.

kubukoz:
I'm on it.

ruippeixotog (approved):

ruippeixotog:
Thanks for doing this, @kubukoz! 🙂

kubukoz:
It's a pleasure :D

PR #418, opened 2018-09-29
Generalize AllowMissingKeys behavior to writers

ruippeixotog:
Currently, we have a pluggable system for readers that can handle missing (undefined) keys. On the writing side we currently aren't doing the same, and instead we statically check if writers are instances of `OptionConfigWriter` (a private internal class).

This PR renames `AllowMissingKeys` to `ReadsMissingKeys` and creates a new `WritesMissingKeys` trait that provides the mechanisms to allow any custom type to have option-like semantics on write.

### DIFF
--- a/core/src/main/scala/pureconfig/CollectionReaders.scala
+++ b/core/src/main/scala/pureconfig/CollectionReaders.scala
@@ -8,26 +8,26 @@ import pureconfig.ConvertHelpers._
 import pureconfig.error._
 
 /**
- * The default behavior of ConfigReaders that are implicitly derived in PureConfig is to raise a
- * KeyNotFoundException when a required key is missing. Mixing in this trait to a ConfigReader
- * allows customizing this behavior. When a key is missing, but the ConfigReader of the given
- * type extends this trait, the `from` method of the ConfigReader is called with null.
+ * A marker trait signaling that a `ConfigReader` accepts missing (undefined) values.
+ *
+ * The standard behavior of `ConfigReader`s that expect required keys in config objects is to return a `KeyNotFound`
+ * failure when one or more of them are missing. Mixing in this trait into the key's `ConfigReader` signals that a
+ * value is missing for the key, the `ConfigReader` can be called with an cursor in the "undefined" state.
  */
-trait AllowMissingKey { self: ConfigReader[_] => }
+trait ReadsMissingKeys { this: ConfigReader[_] => }
 
 /**
  * Trait containing `ConfigReader` instances for collection types.
  */
 trait CollectionReaders {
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/pureconfig/CollectionReaders.scala
```diff
@@ -8,26 +8,26 @@ import pureconfig.ConvertHelpers._
...
 
 /**
- * The default behavior of ConfigReaders that are implicitly derived in PureConfig is to raise a
- * KeyNotFoundException when a required key is missing. Mixing in this trait to a ConfigReader
- * allows customizing this behavior. When a key is missing, but the ConfigReader of the given
- * type extends this trait, the `from` method of the ConfigReader is called with null.
+ * A marker trait signaling that a `ConfigReader` accepts missing (undefined) values.
+ *
+ * The standard behavior of `ConfigReader`s that expect required keys in config objects is to return a `KeyNotFound`
+ * failure when one or more of them are missing. Mixing in this trait into the key's `ConfigReader` signals that a
```
jcazevedo: I think there's an `if` missing: `Mixing in this trait into the key's `ConfigReader` signals that if a value is missing...`.

core/src/main/scala/pureconfig/CollectionReaders.scala
```diff
@@ -8,26 +8,26 @@ import pureconfig.ConvertHelpers._
...
+ * value is missing for the key, the `ConfigReader` can be called with an cursor in the "undefined" state.
```
jcazevedo: Replace `an cursor` with `a cursor`.

core/src/main/scala/pureconfig/CollectionWriters.scala
```diff
@@ -5,12 +5,30 @@ import scala.language.higherKinds
 
 import com.typesafe.config._
 
+/**
+ * A trait signaling that a `ConfigWriter` can write missing (undefined) values.
+ *
+ * `ConfigWriter`s always produce a valid `ConfigValue` with their `to` method. This trait adds an extra `toOpt`
+ * method that parent writers and use in order to decide whether or not they should write a value using this writer.
```
jcazevedo: Replace `writers and use` by `writers can use`.

jcazevedo (changes requested):
Looks good! I just left comments in some typos in the documentation.

Even though the behavior of these traits is tested in the specs that test the derivation logic, I think it's a good idea to add an explicit test for the `WritesMissingKeys` trait, using a custom `ConfigWriter`.

ruippeixotog:
@jcazevedo I must have been half-asleep when I wrote those comments 😅 Everything should be fine now, and I also added the test you requested.

jcazevedo (approved):
Looks good to me!

PR #420, opened 2018-09-30
Add `ConfigReader.forProductN` and `ConfigWriter.forProductN` methods

jcazevedo:
This PR adds the `ConfigReader.forProductN` and `ConfigWriter.forProductN` methods. I believe these methods are useful for creating `ConfigReader`s and `ConfigWriter`s for products without using the `pureconfig-generic` module.


core/src/main/scala/pureconfig/ConfigReader.scala
```diff
@@ -118,7 +118,7 @@ trait ConfigReader[A] {
 /**
  * Provides methods to create [[ConfigReader]] instances.
  */
-object ConfigReader extends BasicReaders with CollectionReaders with ExportedReaders {
+object ConfigReader extends BasicReaders with CollectionReaders with ProductReaders with ExportedReaders {
```
leifwickland, line 121: I like the idea a lot, but I'd like to see unit tests before we approve it.

project/Boilerplate.scala
```diff
@@ -0,0 +1,192 @@
...
+      val resultCombination = synVals.foldLeft("baseF") { case (curr, v) =>
+        s"ConvertHelpers.combineResults($curr, ${v}Result)(_(_))"
+      }
+
+      block"""
+        |package pureconfig
+        |
+        |import pureconfig.error._
+        |
+        |private[pureconfig] trait ProductReaders {
```
ruippeixotog: Why the need to be private? Same thing for writers.

project/Boilerplate.scala
```diff
@@ -0,0 +1,192 @@
...
+        -  ): ConfigReader[B] = new ConfigReader[B] {
+        -    def from(cur: ConfigCursor): Either[ConfigReaderFailures, B] =
+        -      cur.asObjectCursor.right.flatMap { objCur =>
+                 $results
+        -        val baseF: Either[ConfigReaderFailures, $curriedTypes => B] = Right(${if (arity != 1) "f.curried" else "f"})
+        -        $resultCombination
+        -      }
+        -  }
+        |}
+      """
```
ruippeixotog: Can you generate an `object ProductReaders extends ProductReaders` here, just like we do with other reader instance traits? Same thing for writers.

core/src/main/scala/pureconfig/ConfigReader.scala
ruippeixotog: The unit tests are generated the same way as the files themselves. While this can lead to a "who watches the watchmen" situation, the tests can actually be seen in the Travis log.

ruippeixotog (changes requested):
This is a great improvement to make the core package more self-sufficient!

I'm not really fond of having a copy of a not-so-trivial boilerplate generator in our project. I understand that this was the simplest solution, but did you check if there wasn't a proper SBT plugin to do this? If there is a suitable alternative I'd take the time to integrate it with PureConfig and avoid maintaining build code.

project/Boilerplate.scala
jcazevedo: My idea was having these traits providing methods as if they were defined in the `ConfigReader` and `ConfigWriter` objects. That's why I made them private and didn't provide a public object extending the trait. There's no functional need for this to be private other than hiding it from the outside.

jcazevedo:
> I'm not really fond of having a copy of a not-so-trivial boilerplate generator in our project. I understand that this was the simplest solution, but did you check if there wasn't a proper SBT plugin to do this?

I had a look at [sbt-boilerplate](https://github.com/sbt/sbt-boilerplate) but I didn't like its template language. The argument against maintaining build code is strong, though, so I'll conjure up something using it.

core/src/main/scala/pureconfig/ConfigReader.scala
leifwickland: Sorry I missed that.

leifwickland (approved):

jcazevedo:
I changed the implementation in this PR to make use of [sbt-boilerplate](https://github.com/sbt/sbt-boilerplate).

core/src/main/scala/pureconfig/ConfigReader.scala
jcazevedo: The tests are also cumbersome to write by hand without a generator. In 5161a2a, I ended up adding some extra cases to test not only the successful path but also that we're producing the correct failures in `ConfigReader`s built using the `forProductN` methods.

project/Boilerplate.scala
jcazevedo: I didn't have a strong feeling towards having this private, so I ended up dropping it.

core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
```diff
@@ -0,0 +1,33 @@
...
+
+trait ProductReaders {
+  // Special case because f can't be curried.
+  final def forProduct1[B, A0](keyA0: String)(f: A0 => B)(implicit
+    readerA0: ConfigReader[A0]
+  ): ConfigReader[B] = new ConfigReader[B] {
+    def from(cur: ConfigCursor): Either[ConfigReaderFailures, B] =
+      cur.asObjectCursor.right.flatMap { objCur =>
+        val a0Result: Either[ConfigReaderFailures, A0 => B] = Right(f)
+        val a1Result = ConvertHelpers.combineResults(a0Result, if (readerA0.isInstanceOf[ReadsMissingKeys]) readerA0.from(objCur.atKeyOrUndefined(keyA0)) else objCur.atKey(keyA0).right.flatMap(readerA0.from))(_(_))
```
ruippeixotog: Can you extract this logic to a helper function? While this is only used twice in the template it will generate a huge amount of very similar calls.

core/src/main/boilerplate/pureconfig/ProductWriters.scala.template
```diff
@@ -0,0 +1,30 @@
...
+    }
+  }
+
+  [2..22#final def forProduct1[B, [#A0#]]([#keyA0: String#])(f: B => Product1[[#A0#]])(implicit
+    [#writerA0: ConfigWriter[A0]#]
+  ): ConfigWriter[B] = new ConfigWriter[B] {
+    def to(a: B): ConfigValue = {
+      val values = f(a)
+      val baseConf = ConfigFactory.empty()
+      baseConf.[#withValue(keyA0, writerA0.to(values._1))#.].root()
```
ruippeixotog: Writers must also check for the behavior of `WritesMissingKeys` now.

ruippeixotog (changes requested):
LGTM, sbt-boilerplate ended up being a nice choice!

The coverage decreased substantially in this PR, and I think it's because we are not testing the behavior of product readers with optional (`AllowMissingKeys`) values. Can you add a test template covering that? Same thing for writers, once you add the logic defined by `WritesMissingKeys`.

core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
jcazevedo: Makes sense. I've created a helper function in fd1f7c3.

core/src/main/boilerplate/pureconfig/ProductWriters.scala.template
jcazevedo: Of course. I forgot about that. Added that extra verification in b02d479.

jcazevedo:
> The coverage decreased substantially in this PR, and I think it's because we are not testing the behavior of product readers with optional (AllowMissingKeys) values. Can you add a test template covering that? Same thing for writers, once you add the logic defined by WritesMissingKeys.

The coverage for the `ProductReaders` trait was indeed low. I've added two new tests (for each arity) validating the correct behavior in the presence of readers and writers that extend those traits.

ruippeixotog (approved):
LGTM, thanks for all the changes!

jcazevedo:
I'm going to merge this and update the documentation and changelog in a follow up PR.

core/src/main/scala/pureconfig/ConfigReader.scala
leifwickland: @jcazevedo,

Thanks for switching to sbt-boilerplate. I find having separate `.template` files easier to understand.

PR #421, opened 2018-10-01
Target Scala 2.12 in webpage documentation

### LINKED ISSUE #419: Migrate website documentation to Scala 2.12
Currently, we enforce that the code snippets in our website run in all Scala versions we support (i.e. Scala 2.11 and Scala 2.12). This means that code will always work for any user, but it also means that sometimes the examples are more complex than they need to, with the most prominent example being the proliferation of `.right` calls needed before Scala 2.12.

Since the majority of our users use Scala 2.12 and, when not multi-versioned, documentation pages typically use the latest versions of the related libraries/languages, it may make sense to make our documentation target Scala 2.12.

What do you think, @jcazevedo and @leifwickland?

ruippeixotog:
Closes #419.

### DIFF
--- a/docs/build.sbt
+++ b/docs/build.sbt
@@ -1,0 +1,1 @@
+crossScalaVersions ~= { oldVersions => oldVersions.filter(_.startsWith("2.12")) }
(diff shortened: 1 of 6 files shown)


jcazevedo (approved):
LGTM

leifwickland (approved):
Thanks, @ruippeixotog! LGTM 👍

PR #423, opened 2018-10-06
Update CHANGELOG

ruippeixotog:


CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
...
+    `loadConfig`);
+  - The `AllowMissingKeys` trait was renamed to `ReadsMissingKeys`;
+  - Dropped support for Scala 2.10.
+
+- New features
+  - The auto-derivation features of PureConfig, powered by shapeless, were extracted to a separate `pureconfig-generic`
+    module, while `pureconfig-core` was left with only the absolute minimum for PureConfig to be useful. `pureconfig`
+    will continue to be published as a Maven artifact aggregating the two aforementioned artifacts;
+  - Users have now more control over reader and writer derivation. See the
+    [docs](https://pureconfig.github.io/docs) for more information;
```
ruippeixotog, line 15: I'll update this link to point to the correct page later. If I do it now, the build will fail since the `non-automatic-derivation.html` page doesn't exist yet.

leifwickland (approved):
Thanks, @ruippeixotog.

CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
+### 0.10.0 (unreleased)
+
+- Breaking changes
+  - Auto derivation of readers and writers for case classes and sealed traits is now disabled by default. Now users need
+    to import `pureconfig.generic.auto._` everywhere a config is loaded or written (e.g. in files with calls to
+    `loadConfig`);
+  - The `AllowMissingKeys` trait was renamed to `ReadsMissingKeys`;
+  - Dropped support for Scala 2.10.
```
jcazevedo, line 8: We should also mention here that the primitive readers for `Boolean`, `Double`, `Float`, `Int`, `Long` and `Short` no longer return the `EmptyStringFound` in the presence of empty strings, returning `WrongType` instead.

jcazevedo (changes requested):

CHANGELOG.md
ruippeixotog: Do you categorize that as a breaking change? I don't think it was actually documented anywhere and we already changed many times before the errors that we provided without informing users. As such, I'm not sure where to put it in the CHANGELOG, if at all.

jcazevedo: I think it is a breaking change, but it might not be above the threshold of breaking changes we want to report in the CHANGELOG. If there was no documented or expected behavior for those readers, then I think it's ok to not include it.

jcazevedo (approved):

PR #424, opened 2018-10-09
Add Scaladoc for `forProduct` methods

ruippeixotog:
Adds Scaladoc for the new methods so that basic documentation is available for them. Even though it was possible to generate the `@param` entries for the `keyN` parameters, I opted not to do that since I think that it would actually make the Scaladoc pages less readable.

### DIFF
--- a/core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
+++ b/core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
@@ -34,8 +34,8 @@ trait ProductReaders {
         a1Result
       }
   }
-  [2..22#
-  /**
+
+  [2..22#/**
    * Builds a `ConfigReader` for an object created from the values of 1 keys.
    *
    * @param f the function converting the tuple of values to the target object
--- a/core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
+++ b/core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
@@ -15,6 +15,14 @@ trait ProductReaders {
       else
         cursor.atKey(key).right.flatMap(reader.from))(_(_))
 
+  /**
+   * Builds a `ConfigReader` for an object created from the value of 1 key.
+   *
+   * @param f the function converting the value to the target object
+   * @tparam B the type of the target object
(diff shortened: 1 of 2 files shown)


core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
```diff
@@ -26,8 +34,16 @@ trait ProductReaders {
         a1Result
       }
   }
-
-  [2..22#final def forProduct1[B, [#A0#]]([#keyA0: String#])(f: ([#A0#]) => B)(implicit
+  [2..22#
+  /**
```
jcazevedo: Nitpick: Can you move this to the line above and add an extra line before the `[2..22#...`? Just so that the methods only have a blank line between them in the generated code. The same thing applies for `ProductWriters`.

jcazevedo (approved):
LGTM! Thanks for this! I just left a comment on a small nitpick. Feel free to ignore it.

core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
```diff
@@ -26,8 +34,16 @@ trait ProductReaders {
...
+   * Builds a `ConfigReader` for an object created from the values of 1 keys.
+   *
+   * @param f the function converting the tuple of values to the target object
```
leifwickland: Nitpick: I may be wrong, but I don't think it's a tuple so much as an argument list. Were it a tuple, it'd have to be written as `f: ((A0, A1)) => B`

I suggest wording it as `@param f the function constructing the target object from the read config values`

leifwickland (approved):

core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
ruippeixotog: Indeed, they're not tuples. Corrected it in 766a323.

core/src/main/boilerplate/pureconfig/ProductReaders.scala.template
ruippeixotog: Sure, fixed in 3076681.

PR #425, opened 2018-10-09
Perform an implicit search in the `Derivation` materialization instead of relying on `implicitly`

jcazevedo:
This PR changes the materialization of `Derivation` instances to perform an implicit search instead of relying on `implicitly`. Additionally, this makes use of `Derivation` in the `apply` methods of `ConfigReader`, `ConfigWriter` and `ConfigConvert`.

This solves a very weird bug that made reader and writer instances for `Option` derived as coproduct readers and writers when there was a `Derivation` at the top level of the implicit search and we were using full automatic derivation. For example:

```scala
import pureconfig.ConfigConvert

object Main extends App {
  case class Foo(bar: Option[Map[String, Int]])
  val foo = Foo(Some(Map("a" -> 2)))

  val writer1 = implicitly[ConfigConvert[Foo]]
  val writer2 = ConfigConvert[Foo]

  println(writer1.to(foo))
  // SimpleConfigObject({"bar":{"type":"some","value":{"a":2}}})
  println(writer2.to(foo))
  // SimpleConfigObject({"bar":{"a":2}})
}
```

Unfortunately, I can't replicate this issue in the context of a spec and can't understand why this change fixes the bug. 😞 

Incidentally, this changes the error message when an implicit is not found and we're using `Derivation.materializeDerivation` explicitly.

### DIFF
--- a/core/src/main/scala/pureconfig/syntax/package.scala
+++ b/core/src/main/scala/pureconfig/syntax/package.scala
@@ -6,8 +6,8 @@ import pureconfig.error.{ ConfigReaderException, ConfigReaderFailures }
 import scala.reflect.ClassTag
 
 package object syntax {
-  implicit class PimpedAny[T](val any: T) extends AnyVal {
-    def toConfig(implicit writer: ConfigWriter[T]): ConfigValue = writer.to(any)
+  implicit class RichAny[T](val any: T) extends AnyVal {
+    def toConfig(implicit writer: Derivation[ConfigWriter[T]]): ConfigValue = writer.value.to(any)
   }
 
   private def getResultOrThrow[Config](failuresOrResult: Either[ConfigReaderFailures, Config])(implicit ct: ClassTag[Config]): Config = {
@@ -17,17 +17,17 @@ package object syntax {
     }
   }
 
-  implicit class PimpedConfigValue(val conf: ConfigValue) extends AnyVal {
+  implicit class RichConfigValue(val conf: ConfigValue) extends AnyVal {
     def to[T](implicit reader: Derivation[ConfigReader[T]]): Either[ConfigReaderFailures, T] = reader.value.from(conf)
     def toOrThrow[T](implicit reader: Derivation[ConfigReader[T]], cl: ClassTag[T]): T = getResultOrThrow(reader.value.from(conf))(cl)
   }
 
-  implicit class PimpedConfigCursor(val cur: ConfigCursor) extends AnyVal {
+  implicit class RichConfigCursor(val cur: ConfigCursor) extends AnyVal {
     def to[T](implicit reader: Derivation[ConfigReader[T]]): Either[ConfigReaderFailures, T] = reader.value.from(cur)
     def toOrThrow[T](implicit reader: Derivation[ConfigReader[T]], cl: ClassTag[T]): T = getResultOrThrow(reader.value.from(cur))(cl)
   }
 
-  implicit class PimpedConfig(val conf: TypesafeConfig) extends AnyVal {
+  implicit class RichConfig(val conf: TypesafeConfig) extends AnyVal {
     def to[T](implicit reader: Derivation[ConfigReader[T]]): Either[ConfigReaderFailures, T] = conf.root().to[T]
     def toOrThrow[T](implicit reader: Derivation[ConfigReader[T]], cl: ClassTag[T]): T = getResultOrThrow(conf.root().to[T])(cl)
   }
--- a/core/src/main/scala/pureconfig/syntax/package.scala
+++ b/core/src/main/scala/pureconfig/syntax/package.scala
@@ -6,7 +6,7 @@ import pureconfig.error.{ ConfigReaderException, ConfigReaderFailures }
 import scala.reflect.ClassTag
 
 package object syntax {
-  implicit class RichAny[T](val any: T) extends AnyVal {
+  implicit class AnyWriterOps[T](val any: T) extends AnyVal {
     def toConfig(implicit writer: Derivation[ConfigWriter[T]]): ConfigValue = writer.value.to(any)
   }
 
@@ -17,17 +17,17 @@ package object syntax {
     }
   }
 
-  implicit class RichConfigValue(val conf: ConfigValue) extends AnyVal {
+  implicit class ConfigValueReaderOps(val conf: ConfigValue) extends AnyVal {
     def to[T](implicit reader: Derivation[ConfigReader[T]]): Either[ConfigReaderFailures, T] = reader.value.from(conf)
     def toOrThrow[T](implicit reader: Derivation[ConfigReader[T]], cl: ClassTag[T]): T = getResultOrThrow(reader.value.from(conf))(cl)
   }
 
-  implicit class RichConfigCursor(val cur: ConfigCursor) extends AnyVal {
+  implicit class ConfigCursorReaderOps(val cur: ConfigCursor) extends AnyVal {
     def to[T](implicit reader: Derivation[ConfigReader[T]]): Either[ConfigReaderFailures, T] = reader.value.from(cur)
     def toOrThrow[T](implicit reader: Derivation[ConfigReader[T]], cl: ClassTag[T]): T = getResultOrThrow(reader.value.from(cur))(cl)
   }
 
-  implicit class RichConfig(val conf: TypesafeConfig) extends AnyVal {
+  implicit class ConfigReaderOps(val conf: TypesafeConfig) extends AnyVal {
     def to[T](implicit reader: Derivation[ConfigReader[T]]): Either[ConfigReaderFailures, T] = conf.root().to[T]
     def toOrThrow[T](implicit reader: Derivation[ConfigReader[T]], cl: ClassTag[T]): T = getResultOrThrow(conf.root().to[T])(cl)
   }
--- a/macros/src/main/scala/pureconfig/Derivation.scala
+++ b/macros/src/main/scala/pureconfig/Derivation.scala
@@ -50,9 +50,14 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
 
     // check the `-Xmacro-settings:materialize-derivations` scalac flag
     if (!isDerivationEnabled || !isHeadImplicitADerivation) {
-      // when not present, simply render the base `Derivation` constructor with `implcitly[A]`. This results in the same
-      // behavior as without `Derivation` (apart from the extra wrapper).
-      q"_root_.pureconfig.Derivation.Successful(_root_.scala.Predef.implicitly[${weakTypeOf[A]}])"
+      // when not present, start an implicit search for `A` and place it inside a `Derivation.Successful` if the search
+      // succeeds.
+      c.inferImplicitValue(weakTypeOf[A]) match {
+        case EmptyTree =>
+          c.abort(c.enclosingPosition, s"could not derive ${prettyPrintType(weakTypeOf[A])}")
+        case t =>
+          q"_root_.pureconfig.Derivation.Successful($t)"
+      }
 
     } else {
       // if `isRootDerivation` is `false`, then this is a `Derivation` triggered inside another `Derivation`
--- a/macros/src/main/scala/pureconfig/Derivation.scala
+++ b/macros/src/main/scala/pureconfig/Derivation.scala
@@ -55,6 +55,18 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
       firstMacroCall.fold(false)(_.openImplicits.isEmpty)
     }
 
+    // Determine what to do if an implicit search fails. If we're in the context of an implicit search, we want to set
+    // the provided message in the @implicitNotFound annotation. If we're inside an explicit call to
+    // `materializeDerivation` we want to abort with the message
+    val onFailedImplicitSearch: String => Nothing = if (isMaterializationExplicitCall)
+      c.abort(c.enclosingPosition, _)
+    else
+      msg => {
+        setImplicitNotFound(msg)
+        // cause the implicit to fail materializing - the message is ignored
+        c.abort(c.enclosingPosition, "")
+      }
+
     // check the `-Xmacro-settings:materialize-derivations` scalac flag and make sure we're not in an explicit
     // materialization call
     if ((!isDerivationEnabled || !isHeadImplicitADerivation) && !isMaterializationExplicitCall) {
@@ -73,7 +85,7 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
       else
         c.openImplicits.count(_.pre =:= typeOf[Derivation.type]) == 1
 
-      if (isRootDerivation) materializeRootDerivation(weakTypeOf[A], isMaterializationExplicitCall)
+      if (isRootDerivation) materializeRootDerivation(weakTypeOf[A], onFailedImplicitSearch)
       else materializeInnerDerivation(weakTypeOf[A])
     }
   }
@@ -96,20 +108,13 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
   // found error with a basic message. If an implicit is found, it still needs to search the tree found in order to
   // check if some inner derivations materialized a `Derivation.failed`. If that's the case, it collects those failures
   // and prints a nice message.
-  private[this] def materializeRootDerivation(typ: Type, isMaterializationExplicitCall: Boolean): Tree = {
+  private[this] def materializeRootDerivation(typ: Type, failImplicitSearch: String => Nothing): Tree = {
     inferImplicitValueCompat(typ) match {
       case EmptyTree =>
         // failed to find an implicit at the root level of the derivation; set a generic implicitNotFound message
         // without further information
         val implicitNotFoundMsg = buildImplicitNotFound(typ, Nil)
-        if (isMaterializationExplicitCall)
-          c.abort(c.enclosingPosition, implicitNotFoundMsg)
-        else {
-          setImplicitNotFound(implicitNotFoundMsg)
-
-          // cause the implicit to fail materializing - the message is ignored
-          c.abort(c.enclosingPosition, "")
-        }
+        failImplicitSearch(implicitNotFoundMsg)
 
       case value =>
         // collect the failed derivations in the built implicit tree
@@ -121,14 +126,7 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
           // if there are failures, that means one of the inner implicits was not found - set a message with details
           // about the paths failed
           val implicitNotFoundMsg = buildImplicitNotFound(typ, failed)
-          if (isMaterializationExplicitCall)
-            c.abort(c.enclosingPosition, implicitNotFoundMsg)
-          else {
-            setImplicitNotFound(implicitNotFoundMsg)
-
-            // cause the implicit to fail materializing - the message is ignored
-            c.abort(c.enclosingPosition, "")
-          }
+          failImplicitSearch(implicitNotFoundMsg)
         }
     }
   }
--- a/macros/src/main/scala/pureconfig/Derivation.scala
+++ b/macros/src/main/scala/pureconfig/Derivation.scala
@@ -52,7 +52,7 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
     // messages
     val isMaterializationExplicitCall = {
       val firstMacroCall = c.enclosingMacros.reverse.find(ctx => ctx.prefix.tree.tpe =:= ctx.typeOf[Derivation.type])
-      firstMacroCall.fold(false)(_.openImplicits.isEmpty)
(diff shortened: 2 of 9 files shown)


jcazevedo:
Some updates on this:

- I was forced to rely on `inferImplicitValueCompat` for this to work with Scala 2.11.*. This involved me changing its behaviour to take into account `cc.openImplicits` being empty.
- It seems like this issue already manifested itself within [tut](https://github.com/pureconfig/pureconfig/pull/425/files#diff-2cf47a37433a5e6249d09ed7c2b1c6c0R48) but somehow went unnoticed. Unfortunately, I can't replicate this particular case in the context of a spec. 😞

core/src/main/scala/pureconfig/syntax/package.scala
```diff
@@ -6,8 +6,8 @@ import pureconfig.error.{ ConfigReaderException, ConfigReaderFailures }
 import scala.reflect.ClassTag
 
 package object syntax {
-  implicit class PimpedAny[T](val any: T) extends AnyVal {
-    def toConfig(implicit writer: ConfigWriter[T]): ConfigValue = writer.to(any)
+  implicit class RichAny[T](val any: T) extends AnyVal {
```
ruippeixotog: Since you are changing the names of the extension classes, can you rename them to names less likely to clash with other libraries? Importing unrelated extension methods with the same class leads to errors.

macros/src/main/scala/pureconfig/Derivation.scala
```diff
@@ -50,9 +50,15 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
...
     if (!isDerivationEnabled || !isHeadImplicitADerivation) {
-      // when not present, simply render the base `Derivation` constructor with `implcitly[A]`. This results in the same
-      // behavior as without `Derivation` (apart from the extra wrapper).
-      q"_root_.pureconfig.Derivation.Successful(_root_.scala.Predef.implicitly[${weakTypeOf[A]}])"
+      // when not present, start an implicit search for `A` and place it inside a `Derivation.Successful` if the search
+      // succeeds.
+      val tpe = weakTypeOf[A]
+      inferImplicitValueCompat(tpe) match {
+        case EmptyTree =>
+          c.abort(c.enclosingPosition, s"could not derive ${prettyPrintType(tpe)}")
```
ruippeixotog: As we discussed in private, I find it strange that `materializeDerivation` yields a custom, but not "derivation-like" error when called explicitly.

In the past, we decided to make `materializeDerivation` work just like calling `implicitly` when called explicitly, leading to the strange error we're having. I think that the most expectable behavior is to make `materializeDerivation` do its custom logic when called explicitly, regardless of whether the compiler flag is enabled and yielding always the hierarchical "implicit not found" message.

ruippeixotog (changes requested):

core/src/main/scala/pureconfig/syntax/package.scala
jcazevedo: Sure. I've renamed them in faec13d. I think they're less likely to clash now.

macros/src/main/scala/pureconfig/Derivation.scala
jcazevedo: I agree! I think I got `materializeDerivation` to yield the proper derivation-like message in fe5fa43 and 977e0db.

macros/src/main/scala/pureconfig/Derivation.scala
```diff
@@ -48,17 +48,32 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
...
-    // check the `-Xmacro-settings:materialize-derivations` scalac flag
-    if (!isDerivationEnabled || !isHeadImplicitADerivation) {
-      // when not present, simply render the base `Derivation` constructor with `implcitly[A]`. This results in the same
-      // behavior as without `Derivation` (apart from the extra wrapper).
-      q"_root_.pureconfig.Derivation.Successful(_root_.scala.Predef.implicitly[${weakTypeOf[A]}])"
+    // check if the materialization was called explicitly, in which case we want to have the nicer compiler error
+    // messages
+    val isMaterializationExplicitCall = {
+      val firstMacroCall = c.enclosingMacros.reverse.find(c => c.prefix.tree.tpe =:= c.typeOf[Derivation.type])
+      firstMacroCall.fold(false)(c => !c.openImplicits.exists(_.pre =:= c.typeOf[Derivation.type]))
```
ruippeixotog: Can you rename the lambda parameter to something other than `c`, which can be confused with the outer context?

Also, I think it's enough to check here that `c.openImplicits.nonEmpty`, since once we are in _any_ implicit search, we know that methods could not have been called explicitly, right?

macros/src/main/scala/pureconfig/Derivation.scala
```diff
@@ -81,15 +96,20 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
   // found error with a basic message. If an implicit is found, it still needs to search the tree found in order to
   // check if some inner derivations materialized a `Derivation.failed`. If that's the case, it collects those failures
   // and prints a nice message.
-  private[this] def materializeRootDerivation(typ: Type): Tree = {
+  private[this] def materializeRootDerivation(typ: Type, isMaterializationExplicitCall: Boolean): Tree = {
```
ruippeixotog: I think the code would become more elegant if you received here a `failImplicitSearch: String => Nothing` and have the `materializeDerivation` pass the correct strategy, instead of having implicit-vs-explicit logic on both sides.

macros/src/main/scala/pureconfig/derivation/MacroCompat.scala
```diff
@@ -26,7 +26,10 @@ trait MacroCompat {
   // reported up to Scala 2.12.2. See https://github.com/scala/bug/issues/10398 for more information.
   def inferImplicitValueCompat(typ: Type): Tree = {
     val cc = c.asInstanceOf[scala.reflect.macros.contexts.Context]
-    val enclosingTree = cc.openImplicits.head.tree.asInstanceOf[cc.universe.analyzer.global.Tree]
+    val enclosingTree =
+      cc.openImplicits.headOption.map(_.tree)
+        .orElse(cc.enclosingMacros.lastOption.map(_.macroApplication))
```
ruippeixotog, line 31: Hmm, I'm not familiar enough with the compiler internals anymore to predict the consequences of setting the enclosing tree as this expression... I hope everything works fine 😅

ruippeixotog (changes requested):

macros/src/main/scala/pureconfig/derivation/MacroCompat.scala
jcazevedo: Without this change, the build fails on Scala 2.11.12 (see [here](https://travis-ci.org/pureconfig/pureconfig/jobs/441141332)). It seemed as if implicits for types with type parameters weren't being expanded when in an explicit call to `materializeDerivation`. I got to this mostly by trial and error, and I'm also not sure of this change. 😞

macros/src/main/scala/pureconfig/Derivation.scala
jcazevedo: Yes. Checking that `openImplicits` is empty in the first derivation is sufficient. I've fixed that in dd7c368.

macros/src/main/scala/pureconfig/Derivation.scala
jcazevedo: I agree that it becomes more elegant! I've refactored this in 92cc149.

macros/src/main/scala/pureconfig/Derivation.scala
```diff
@@ -48,17 +48,44 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
...
-    // check the `-Xmacro-settings:materialize-derivations` scalac flag
-    if (!isDerivationEnabled || !isHeadImplicitADerivation) {
-      // when not present, simply render the base `Derivation` constructor with `implcitly[A]`. This results in the same
-      // behavior as without `Derivation` (apart from the extra wrapper).
-      q"_root_.pureconfig.Derivation.Successful(_root_.scala.Predef.implicitly[${weakTypeOf[A]}])"
+    // check if the materialization was called explicitly, in which case we want to have the nicer compiler error
+    // messages
+    val isMaterializationExplicitCall = {
+      val firstMacroCall = c.enclosingMacros.reverse.find(ctx => ctx.prefix.tree.tpe =:= ctx.typeOf[Derivation.type])
+      firstMacroCall.fold(false)(_.openImplicits.isEmpty)
```
ruippeixotog: `.fold(false)(f)` is equal to `.exists(f)`.

macros/src/main/scala/pureconfig/Derivation.scala
```diff
@@ -48,17 +48,44 @@ class DerivationMacros(val c: whitebox.Context) extends LazyContextParser with M
...
+      inferImplicitValueCompat(tpe) match {
+        case EmptyTree => c.abort(c.enclosingPosition, "")
+        case t => q"_root_.pureconfig.Derivation.Successful[${tpe}]($t)"
+      }
 
     } else {
       // if `isRootDerivation` is `false`, then this is a `Derivation` triggered inside another `Derivation`
-      val isRootDerivation = c.openImplicits.count(_.pre =:= typeOf[Derivation.type]) == 1
+      val isRootDerivation = if (isMaterializationExplicitCall)
+        c.enclosingMacros.count(c => c.prefix.tree.tpe =:= c.typeOf[Derivation.type]) == 1
```
ruippeixotog: Doesn't this check work for both explicit and implicit calls? I used `openImplicits` before because I was only considering implicit calls.

macros/src/main/scala/pureconfig/derivation/MacroCompat.scala
ruippeixotog: Let's keep this then, the change makes sense to me and it doesn't seem very dangerous :)

ruippeixotog (changes requested):
LGTM, just take a look at these last suggestions! Would you also mind checking that PureConfig compiles fine if you [deactivate derivations](https://github.com/pureconfig/pureconfig/blob/master/build.sbt#L81)?

macros/src/main/scala/pureconfig/Derivation.scala
jcazevedo: I've improved that expression in bfc162a.

macros/src/main/scala/pureconfig/Derivation.scala
jcazevedo: I think we still need to have different checks. `enclosingMacros` is more like a stack trace for macros whereas `openImplicits` only shows implicit candidates currently being considered.

Consider the following example:

```scala
class C(x: Int, s: String)
case class B(c: C)
case class A(b: B)
implicitly[ConfigReader[A]]
```

By the time we're attempting to derive a `ConfigReader` for `B`, there will be two enclosing macros matching the check: the one for the derivation of `ConfigReader[A]` and another for the derivation of `ConfigReader[B]`. In this case, there will be only one open implicit: the one that attempts to derive a `MapShapedReader` for `B`. We want to consider it as a root derivation to avoid deriving a `Failed` that is not collected by another derivation.

jcazevedo:
> Would you also mind checking that PureConfig compiles fine if you [deactivate derivations](https://github.com/pureconfig/pureconfig/blob/master/build.sbt#L81)?

PureConfig compiles fine with derivations disabled (obviously excluding the `DerivationSuite`).

macros/src/main/scala/pureconfig/Derivation.scala
ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

ruippeixotog: Hmm, I'm not sure I'm getting it... I don't think we ever want to consider two root derivations in the same implicit chain - only the outermost makes sense, since it's the only one who'll have access to the full AST.

In the case you mentioned it seems to be that there should be only one root derivation - the implicit search for `ConfigReader[A]`. If the search for a `ConfigReader[B]` fails it should still emit a `Failed` object, which would still be processed by the root derivation. Am I missing something? I have not looked into this code for a long time...

jcazevedo: I think that's only true for implicit searches that start on a derivation. In this case, `ConfigReader[A]` isn't wrapped in `Derivation`, so the entrypoint for the macro will be on `Derivation[ConfigReader[B]]`.

I think this exposes another underlying issue, though. I guess we were assuming that `isHeadImplicitADerivation` would be `false` in those cases, but it seems like the only way for it to be `false` is when `materializeDerivation` is called explicitly.

ruippeixotog: Oh, I see what you mean; but `Derivation` is not supposed to work at all if the root implicit search is not of a `Derivation`.

Our macro relies on that because it sets the `@implicitNotFound` annotation on `Derivation` in order to show the custom message. Setting it on other classes would not be good since it is a "global" operation (i.e. other implicit searches in the same compilation unit would be affected by that) and I also believe that users shouldn't expect a custom message to appear if they are not explicitly requesting it.

core/src/main/scala/pureconfig/syntax/package.scala
```diff
@@ -6,8 +6,8 @@ import pureconfig.error.{ ConfigReaderException, ConfigReaderFailures }
 import scala.reflect.ClassTag
 
 package object syntax {
-  implicit class PimpedAny[T](val any: T) extends AnyVal {
-    def toConfig(implicit writer: ConfigWriter[T]): ConfigValue = writer.to(any)
+  implicit class AnyWriterOps[T](val any: T) extends AnyVal {
```
leifwickland, line 9: Thanks for doing this.

leifwickland:
I'm hanging back on looking at this one until @ruippeixotog is happy. Let me know if you want me to look before then. I don't really have an opinion on the matter generally.

macros/src/main/scala/pureconfig/Derivation.scala
jcazevedo: > Oh, I see what you mean; but `Derivation` is not supposed to work at all if the root implicit search is not of a `Derivation`.

That's why I think we have a bug in our `isHeadImplicitADerivation` check. I'll investigate further later today.

jcazevedo: I think I finally found the bug! `openImplicits` is a stack, so the first implicit in the chain is not at the head, but rather at the tail of the list. Fixing the `isHeadImplicitADerivation` check allows us to only rely on the enclosing macros to check if we're in the root derivation, as you suggested.

Commit 99439e5 implements this fix.

ruippeixotog: Great job debugging this!

ruippeixotog (approved):
LGTM, awesome work! :)

jcazevedo:
@ruippeixotog: Thanks for the help with the macro code!

leifwickland (approved):
Thanks for working through this, @jcazevedo.

PR #429, opened 2018-10-21
Add `forProductN` helpers to the docs

jcazevedo:
This PR adds references to the `forProductN` helpers in the docs. I used them to simplify the implementation of method 2 in the page about supporting complex types. I also mentioned them in the _Manual_ section of the _Non-Automatic Derivation_ page, as a convenient way to create readers and writers for case class-like types.

### DIFF
--- a/docs/src/main/tut/docs/non-automatic-derivation.md
+++ b/docs/src/main/tut/docs/non-automatic-derivation.md
@@ -76,7 +76,8 @@ pureconfig.loadConfig[Person](conf)
 ### Manual
 
 When case class and sealed trait derivation is not needed or wanted, we can simply not import anything and define our
-reader using any of ways explained in [Supporting New Types](supporting-new-types.html):
+reader using any of ways explained in [Supporting New Types](supporting-new-types.html). The `forProductN` helper
+methods are also convenient for creating readers and writers for case class-like types without generic derivation:
 
 ```tut:invisible:reset
 import com.typesafe.config.ConfigFactory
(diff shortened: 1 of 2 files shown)


ruippeixotog (approved):
LGTM!

docs/src/main/tut/docs/non-automatic-derivation.md
```diff
@@ -76,7 +76,8 @@ pureconfig.loadConfig[Person](conf)
 ### Manual
 
 When case class and sealed trait derivation is not needed or wanted, we can simply not import anything and define our
-reader using any of ways explained in [Supporting New Types](supporting-new-types.html):
+reader using any of ways explained in [Supporting New Types](supporting-new-types.html). The `forProductN` helper
```
ruippeixotog, line 79: Since you are already using `forProductN` in Supporting New Types, you should probably remove "also".

ruippeixotog, line 79: Since you are already using `forProductN` in Supporting New Types, you should probably remove "also".

docs/src/main/tut/docs/non-automatic-derivation.md
```diff
@@ -89,13 +90,7 @@ val conf = ConfigFactory.parseString("{ name: John, surname: Doe }")
...
 import pureconfig._
 
-implicit val personReader = ConfigReader.fromCursor[Person] { cur =>
-  for {
-    objCur <- cur.asObjectCursor
-    name <- objCur.atKey("name").right.flatMap(_.asString)
-    surname <- objCur.atKey("surname").right.flatMap(_.asString)
-  } yield Person(name, surname)
-}
+implicit val personReader = ConfigReader.forProduct2("name", "surname")((name, surname) => Person(name, surname))
```
ruippeixotog: Can you either use the shorthand `Person(_, _)` for the second argument or split this line into two?

ruippeixotog (changes requested):

docs/src/main/tut/docs/non-automatic-derivation.md
jcazevedo: Removed the redundant "also" in 6171cd6.

docs/src/main/tut/docs/non-automatic-derivation.md
jcazevedo: I've simplified the definition of `personReader` in e75495a.

ruippeixotog (approved):
LGTM!

The CI integration seems to be stuck, probably due to the recent GitHub issues. I'd suggest amending your last commit by changing your commit message or something else in order to force another build.

kubukoz:
Looks green, can we merge?

leifwickland (approved):
Thanks for doing this!

PR #432, opened 2018-10-31
Add tests cases to the `ConfigCursor` cast methods

beatriz:
This PR adds test cases to some cast methods of `ConfigCursor` that weren't covered, including error cases. I just wasn't able to get a number value that gets a `CannotConvert` error on the `asDouble` cast.


tests/src/test/scala/pureconfig/ConfigCursorSuite.scala
```diff
@@ -37,6 +37,72 @@ class ConfigCursorSuite extends BaseSuite {
       WrongType(ConfigValueType.OBJECT, Set(ConfigValueType.STRING)), defaultPathStr)
   }
 
+  it should "allow being cast to boolean in a safe way" in {
+    cursor("true").asBoolean shouldBe Right(true)
+    cursor("false").asBoolean shouldBe Right(false)
+
+    cursor("abc").asBoolean should failWith(
+      WrongType(ConfigValueType.STRING, Set(ConfigValueType.BOOLEAN)))
+    cursor("1").asBoolean should failWith(
```
leifwickland, line 46: @mbeatrizmagalhaes The unit test is [failing](https://travis-ci.org/pureconfig/pureconfig/jobs/449084988) here because it's actually finding `ConfigValueType.NUMBER` for `"1"`.

ruippeixotog:
@mbeatrizmagalhaes good job increasing the test coverage! :)

> I just wasn't able to get a number value that gets a `CannotConvert` error on the `asDouble` cast

You should be able to find one, but it seems Scala/JVM does a weird thing with number comparisons... Wasn't expecting this:

screen shot 2018-10-31 at 15 19 08

I guess that when Scala tries to compare a `Double` with a `Long`, it converts the `Long` to a `Double` just as we do, which leads to [this condition](https://github.com/pureconfig/pureconfig/blob/master/core/src/main/scala/pureconfig/ConfigCursor.scala#L115) always being true. However, in fact, this isn't correct, as `21412415121234567L` in a config would round to `2.1412415121234568E16` when read with `asDouble`.

@jcazevedo, this seems a bug with our method of checking for safe casts :( A better check would probably be to do the full round-trip (i.e. for casting a number of type `X` to `Y`, doing `n.YValue().toX == n`). That ensures the equivalence between the config value and the casted value.

jcazevedo:
@mbeatrizmagalhaes: Thanks for this!

> @jcazevedo, this seems a bug with our method of checking for safe casts :( A better check would probably be to do the full round-trip (i.e. for casting a number of type `X` to `Y`, doing `n.YValue().toX == n`). That ensures the equivalence between the config value and the casted value.

Right! I didn't anticipate that. :(

What you suggested should work, but unfortunately I think we need to deal with `Double`s, `Int`s and `Long`s (which seem to be the possible return types of an `unwrapped` call to `ConfigNumber`) separately, instead of relying on the interface provided by `Number`.

ruippeixotog:
I also think we can't use `Number` anymore... but I guess there is no other way, and since Typesafe Config is pretty stable we should have no problem assuming there can be only those three types of numbers.

beatriz:
I'm sorry for the slow slow turnaround on this one! I've improved the conversions to rely on the specific types instead of `Number` on 7f2ae4b (and 88d42dc) and added a test for failing double conversion on 7e3641b.

jcazevedo (approved):
LGTM! Thanks for doing this!

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -76,7 +76,9 @@ sealed trait ConfigCursor {
    */
   def asLong: ConfigReader.Result[Long] =
     castOrFail(NUMBER, _.unwrapped match {
-      case i: java.lang.Number if i.longValue() == i => Right(i.longValue())
+      case i: java.lang.Long => Right(i)
+      case i: java.lang.Integer if i.longValue().toInt == i => Right(i.longValue())
```
ruippeixotog: `Long` is a strict superset of `Integer`, so you don't need the condition here.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -124,7 +132,9 @@ sealed trait ConfigCursor {
    */
   def asFloat: ConfigReader.Result[Float] =
     castOrFail(NUMBER, _.unwrapped match {
-      case i: java.lang.Number if i.floatValue() == i => Right(i.floatValue())
+      case i: java.lang.Long if i.floatValue().toLong == i => Right(i.floatValue())
+      case i: java.lang.Integer if i.floatValue().toInt == i => Right(i.floatValue())
+      case i: java.lang.Double if i.floatValue().toDouble == i => Right(i.floatValue())
```
ruippeixotog: Unfortunately, this technique doesn't quite work for floating point values - or at least it doesn't work as you'd expect. Many values like 0.3 will fail this check simply because floating point cannot represent them perfectly (regardless of the value's precision) and as such you'll always incur a loss in precision when converting `Double` to `Float`:

```
scala> 0.3.floatValue.toDouble
res2: Double = 0.30000001192092896
```

My opinion here is that we should allow conversion between floating point types at all times, even if they incur a loss in precision - it better reflects the continuous nature of decimal numbers. What do you think?

ruippeixotog (changes requested):

core/src/main/scala/pureconfig/ConfigCursor.scala
beatriz: Again, sorry for the slow turnaround... Done on b842abf

core/src/main/scala/pureconfig/ConfigCursor.scala
beatriz: I agree. Removed the condition and changed the test to accommodate to this on ee1bf11.

ruippeixotog (approved):
LGTM, thanks for taking the time to fix this! :)

PR #434, opened 2018-11-01
Move `scala-compiler` and `scala-reflect` dependencies to the `provided` scope

### LINKED ISSUE #433: pureconfig pulls in dependencies to scala-compiler, scala-reflect
Hey,

I noticed today that when I add a dependency to pureconfig, I also get a transitive dependency to scala-compiler and scala-reflect. My understanding is that these are only needed at compile-time for the generic macros and not at run-time. Other libraries seem to solve this problem by depending on these in the `Provided` scope, see e. g. [circe](https://github.com/circe/circe/blob/f38f73cfecbe25b514fb45ef434d3c26693000b9/build.sbt#L210) or [shapeless](https://github.com/milessabin/shapeless/blob/1185bc22692bd5c085b158b732c931d1091c5c62/build.sbt#L244). I've tried making this change in the pureconfig generic module but failed for reasons I don't quite understand, maybe you can figure it out…

Anyway, scala-compiler and scala-reflect 2.12.7 together currently weigh 13.5MB, which is more than 20 times larger than pureconfig.

jcazevedo:
This PR moves the `scala-compiler` and `scala-reflect` dependencies to the `provided` scope, which should fix #433.

### DIFF
--- a/macros/build.sbt
+++ b/macros/build.sbt
@@ -3,8 +3,8 @@ import Dependencies._
 name := "pureconfig-macros"
 
 libraryDependencies ++= Seq(
-  "org.scala-lang" % "scala-compiler" % scalaVersion.value,
-  "org.scala-lang" % "scala-reflect" % scalaVersion.value)
+  "org.scala-lang" % "scala-compiler" % scalaVersion.value % Provided,
+  "org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided)
 
 developers := List(
   Developer("ruippeixotog", "Rui Gonçalves", "ruippeixotog@gmail.com", url("https://github.com/ruippeixotog")))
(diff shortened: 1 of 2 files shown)


ruippeixotog (commented):
Makes sense :) I expect this to work perfectly, but just for precaution, can you just test publishing the project locally and using it in a test project with derivations enabled?

jcazevedo:
> Makes sense :) I expect this to work perfectly, but just for precaution, can you just test publishing the project locally and using it in a test project with derivations enabled?

I've tried it in a test project and things work as expected.

ruippeixotog (approved):

leifwickland (approved):
Thanks for doing this and especially for running the test to ensure that it works, @jcazevedo.

PR #435, opened 2018-11-06
simplify implementation of fs2 module

mberndt123:
The fs2 module looks pointlessly complicated to me. When you read a `Config` from a stream, it's going to require memory proportional to the size of the input anyway to store the `Config` object, hence there's no point in parsing it in a streaming fashion. So I wrote this implementation which is much simpler, places fewer constraints on the `F[_]` type parameter (can be used with e. g. SyncIO), doesn't require an `ExecutionContext` and doesn't mess around with threads.

### DIFF
--- a/modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
+++ b/modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
@@ -20,8 +20,8 @@ package object fs2 {
    * @return The returned action will complete with `A` if it is possible to create an instance of type
    *         `A` from the configuration stream, or fail with a ConfigReaderException which in turn contains
    *         details on why it isn't possible
-    *         It can also raise any exception that the stream can raise, as well as any exception that
-    *         ConfigFactory.parseString might throw. 
+   *         It can also raise any exception that the stream can raise, as well as any exception that
+   *         ConfigFactory.parseString might throw.
    */
   def streamConfig[F[_], A](configStream: Stream[F, Byte])(
     implicit
--- a/modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
+++ b/modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
@@ -21,8 +21,7 @@ package object fs2 {
    * @return The returned action will complete with `A` if it is possible to create an instance of type
    *         `A` from the configuration stream, or fail with a ConfigReaderException which in turn contains
    *         details on why it isn't possible
-   *         It can also raise any exception that the stream can raise, as well as any exception that
-   *         ConfigFactory.parseString might throw.
+   *         It can also raise any exception that the stream can raise.
    */
   def streamConfig[F[_], A](configStream: Stream[F, Byte])(
     implicit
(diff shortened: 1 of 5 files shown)


leifwickland:
@keirlawson Do you have thoughts on this proposed change?

keirlawson (approved):
LGTM I think my original implementation was probably over-engineered :)

modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
```diff
@@ -1,55 +1,37 @@
...
-
   /**
    * Load a configuration of type `A` from the given byte stream.
    *
    * @param configStream a stream of bytes representing the contents of a configuration file
    * @return The returned action will complete with `A` if it is possible to create an instance of type
    *         `A` from the configuration stream, or fail with a ConfigReaderException which in turn contains
    *         details on why it isn't possible
+   *         It can also raise any exception that the stream can raise, as well as any exception that
+   *         ConfigFactory.parseString might throw.
```
ruippeixotog: This isn't very nice, since PureConfig entrypoints handle malformed configs gracefully. You may want to take a look at [ConfigFactoryWrapper](https://github.com/pureconfig/pureconfig/blob/master/core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala), the wrapper we use to call `ConfigFactory` methods in a safe way. Feel free to add a `parseString` method there and use it in this module.

ruippeixotog (changes requested):
Looks like a nice simplification, thank you @mberndt123!

modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
mberndt123: I have made the change regarding the ConfigFactory method. However there's not really anything that can be done about the errors that the stream can raise.

ruippeixotog: Thanks, looks good! Can you just add tests for both parsing errors and `loadConfig` errors, so that this behavior is not accidentally changed in future PRs?

mberndt123: Short answer: no, I won't waste my time writing tests for behaviours I didn't change.

I wrote a new, simpler implementation because I like simple code. Take it or leave it…

ruippeixotog: Sure. Thanks anyway.

ruippeixotog (approved):

mberndt123:
@ruippeixotog any reason this isn't merged yet?

leifwickland (approved):
Thanks, @keirlawson.

leifwickland:
@mberndt123,

Thanks again for the contribution. Would you like to send a PR to add yourself to the [pureconfig contributors list](https://github.com/pureconfig/pureconfig/blob/master/CONTRIBUTORS)?

Also, PureConfig 0.10.1 has been published with your FS2 change.

mberndt123:
Yeah, thanks for letting me know. I think I'm fine without that, but thanks for asking.

PR #437, opened 2018-11-26
Add namespace to `loadConfigFromFiles`

kiara-riley:
This seems to work fine in my tests.


core/src/main/scala/pureconfig/package.scala
```diff
@@ -286,8 +286,10 @@ package object pureconfig {
    * otherwise they would yield a failure (a `Left` value).
    *
    * @param files Files ordered in decreasing priority containing part or all of a `Config`
+   * @param failOnReadError Where to return an error if any files fail to read
+   * @param namespace the base namespace from which the configuration should be load
    */
-  def loadConfigFromFiles[Config](files: Traversable[Path], failOnReadError: Boolean = false)(implicit reader: Derivation[ConfigReader[Config]]): Either[ConfigReaderFailures, Config] = {
+  def loadConfigFromFiles[Config](files: Traversable[Path], failOnReadError: Boolean = false, namespace: String = "")(implicit reader: Derivation[ConfigReader[Config]]): Either[ConfigReaderFailures, Config] = {
```
kiara-riley, line 292: I can also copy this method so we don't have a default value in here if preferred.

jcazevedo: Thanks for your PR, @kraigmckernan!

I don't mind the default value, but I'd rather have the function declarations consistent in this package. Do we have a strong case for preferring overloading instead of default parameter values? /cc @ruippeixotog @leifwickland

ruippeixotog: I don't see any problem with default parameters in this case 👍

leifwickland (approved):
Thanks, @kraigmckernan!

core/src/main/scala/pureconfig/package.scala
leifwickland: I'm good with it.

jcazevedo (approved):
Thanks for the PR, @kraigmckernan!

leifwickland:
@kraigmckernan,

 Thanks again for the contribution. Would you like to send a PR to add yourself to the [pureconfig contributors list](https://github.com/pureconfig/pureconfig/blob/master/CONTRIBUTORS)?

PR #438, opened 2018-11-27
Fix broken and redirecting links

leifwickland:
Fix broken links mentioned by @jcranky in gitter. Also, generally replace `typesafehub` with `lightbend` in URLs to avoid redirects.


ruippeixotog (approved):
Looks good, thanks for the update, @leifwickland!

jcazevedo (approved):
Thanks for this, @leifwickland!

jcazevedo:
We're having an issue with the microsite job in Travis because the most recent `html-proofer` version changed its flags. I've created #439 to address that.

jcazevedo:
It seems like the issue was not related to the flags (`--allow_hash_href` is still supported, apparently) but instead related to the way `htmlproofer` started to handle backslashes in its arguments (we were previously wrongly introducing them there). #439 is still relevant, though.

leifwickland:
@jcazevedo @ruippeixotog I ended up adding a `--url-ignore "/search.maven.org/"` to `htmlproofer` because search.maven.org, which is what maven-badges redirects to, does a diabolical thing to trick spiders: it returns a 404 for every page with a body that uses JS to load the actual page.

If you're OK with that, please say so or just merge the PR.

PR #439, opened 2018-11-29
Use specific `html-proofer` version in Travis

jcazevedo:
This PR modifies the Travis pipeline to install a specific `html-proofer` version. The `--allow_hash_href` flag changed to `--allow-hash-href` in the latest version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
       script:
         - sbt makeMicrosite
         - htmlproofer \
-            --allow_hash_href \
+            --allow-hash-href \
             --url-swap "https\://github\.com/pureconfig/pureconfig/tree/master:file\://$(pwd)" \
             docs/target/site
       after_success: ignore
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,9 +39,9 @@ jobs:
         - gem install sass jekyll:3.2.1 html-proofer:3.9.3
       script:
         - sbt makeMicrosite
-        - htmlproofer \
-            --allow-hash-href \
-            --url-swap "https\://github\.com/pureconfig/pureconfig/tree/master:file\://$(pwd)" \
+        - htmlproofer
+            --allow-hash-href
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -36,12 +36,12 @@ jobs:
...
         - rvm use 2.4 --install --fuzzy
         - gem update --system
-        - gem install sass jekyll:3.2.1 html-proofer
+        - gem install sass jekyll:3.2.1 html-proofer:3.9.3
       script:
         - sbt makeMicrosite
-        - htmlproofer \
-            --allow_hash_href \
-            --url-swap "https\://github\.com/pureconfig/pureconfig/tree/master:file\://$(pwd)" \
+        - htmlproofer
```
leifwickland, line 42: Stupid question: why are the backslashes not required? I thought something like `\` or `>` had to be done to concatenate many lines into one string. (YAML generally leaves me confused and uncomfortable and I'm mostly ignorant of its intricacies.)

jcazevedo: In the [Plain Flow Scalar Style](http://yaml.org/spec/1.2/spec.html#id2788859), in which a string is not surrounded by neither single nor double quotes, the `\` character doesn't have any special meaning. The [Double-Quoted Style](http://yaml.org/spec/1.2/spec.html#id2787109) uses a single `\` to indicate where a line ends, but it is also not necessary if lines are to be trimmed and concatenated with a single space.

ruippeixotog:
Thank you, @jcazevedo! Does this work only together with #438?

leifwickland (approved):

leifwickland:
Building on what @ruippeixotog said, it seems like we need to merge #438 into this (or vice versa) so that the broken link and the broken validator are fixed in one PR.

leifwickland:
I merged `fix-html-proofer` into #438. It's building here:
https://travis-ci.org/pureconfig/pureconfig/jobs/461512787

jcazevedo:
Thanks for merging this into #438, @leifwickland. I'll close this one.

leifwickland:
@jcazevedo Thank for the fix!

PR #442, opened 2018-12-22
 Create a ReaderResult[A] alias for Either[ConfigReaderFailures, A]

ruippeixotog:
We use `Either[ConfigReaderFailures, A]` heavily throughout our code as the default type for reading results. Writing it every time is cumbersome and forces an extra import (`pureconfig.error._`). Since we use it very heavily as a monad/applicative, we ended up creating helper methods in `ConvertHelpers`.

In my opinion, that's enough motivation to create a type alias for it, together with a companion object containing useful constructors and helpers. I updated the documentation and added two short references (one of them right in the Quick Start guide / README) so that people new to PureConfig are aware the alias is just an `Either`.

### DIFF
--- a/core/src/main/scala/pureconfig/ConfigReader.scala
+++ b/core/src/main/scala/pureconfig/ConfigReader.scala
@@ -6,7 +6,7 @@ import scala.util.Try
 import com.typesafe.config.ConfigValue
 import pureconfig.ConfigReader._
 import pureconfig.ConvertHelpers._
-import pureconfig.error.FailureReason
+import pureconfig.error.{ ConfigReaderFailures, FailureReason }
 
 /**
  * Trait for objects capable of reading objects of a given type from `ConfigValues`.
@@ -21,7 +21,7 @@ trait ConfigReader[A] {
    * @param cur The cursor from which the config should be loaded
    * @return either a list of failures or an object of type `A`
    */
-  def from(cur: ConfigCursor): ReaderResult[A]
+  def from(cur: ConfigCursor): ConfigReader.Result[A]
 
   /**
    * Convert the given configuration into an instance of `A` if possible.
@@ -29,7 +29,7 @@ trait ConfigReader[A] {
    * @param config The configuration from which the config should be loaded
    * @return either a list of failures or an object of type `A`
    */
-  def from(config: ConfigValue): ReaderResult[A] =
+  def from(config: ConfigValue): ConfigReader.Result[A] =
     from(ConfigCursor(config, Nil))
 
   /**
@@ -120,6 +120,13 @@ trait ConfigReader[A] {
  */
 object ConfigReader extends BasicReaders with CollectionReaders with ProductReaders with ExportedReaders {
 
+  /**
+   * The type of most config PureConfig reading methods.
+   *
+   * @tparam A the type of the result
+   */
+  type Result[A] = Either[ConfigReaderFailures, A]
+
   def apply[A](implicit reader: Derivation[ConfigReader[A]]): ConfigReader[A] = reader.value
 
   /**
@@ -129,7 +136,7 @@ object ConfigReader extends BasicReaders with CollectionReaders with ProductRead
    * @tparam A the type of the objects readable by the returned reader
    * @return a `ConfigReader` for reading objects of type `A` using `fromF`.
    */
-  def fromCursor[A](fromF: ConfigCursor => ReaderResult[A]) = new ConfigReader[A] {
+  def fromCursor[A](fromF: ConfigCursor => ConfigReader.Result[A]) = new ConfigReader[A] {
     def from(cur: ConfigCursor) = fromF(cur)
   }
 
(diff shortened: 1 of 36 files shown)


core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
```diff
@@ -16,37 +16,37 @@ object ConfigFactoryWrapper {
...
+  def parseFile(path: Path): ReaderResult[Config] =
     unsafeToEither(ConfigFactory.parseFile(path.toFile, strictSettings), Some(path))
 
   /** Utility methods that parse a file and then calls `ConfigFactory.load` */
-  def loadFile(path: Path): Either[ConfigReaderFailures, Config] =
+  def loadFile(path: Path): ReaderResult[Config] =
     parseFile(path).right.flatMap(rawConfig => unsafeToEither(ConfigFactory.load(rawConfig)))
 
-  private def unsafeToEither[A](f: => A, path: Option[Path] = None): Either[ConfigReaderFailures, A] = {
+  private def unsafeToEither[A](f: => A, path: Option[Path] = None): ReaderResult[A] = {
```
jcazevedo: This should now be named `unsafeToReaderResult`.

jcazevedo (changes requested):
I like this, but I think `ReaderResult` would relate better to `ConfigReader` if it was defined in the companion object of `ConfigReader`. What do you think of having it named only `Result`, but defined in the companion object of `ConfigReader`, so that that we would refer to it as `ConfigReader.Result`?

ruippeixotog:
I guess it would be more organized, even though it increases the verbosity of the type... changed it in 8f68262 👍

core/src/main/scala/pureconfig/package.scala
```diff
@@ -4,25 +4,47 @@
...
 import pureconfig.backend.ConfigFactoryWrapper._
 import pureconfig.backend.PathUtil._
 import pureconfig.error._
 
 package object pureconfig {
 
+  /**
+   * Object containing useful constructors and utility methods for `ReaderResult`s.
+   */
+  object ReaderResult {
```
jcazevedo: Shouldn't this also be moved to `ConfigReader.Result`?

ruippeixotog: 👍 cb45d1f

core/src/main/scala/pureconfig/ConfigReader.scala
```diff
@@ -120,6 +120,35 @@ trait ConfigReader[A] {
...
 
+  /**
+   * The type of most config PureConfig reading methods.
+   *
+   * @tparam A the type of the result
+   */
+  type Result[A] = Either[ConfigReaderFailures, A]
+
+  /**
+   * Object containing useful constructors and utility methods for `ReaderResult`s.
```
jcazevedo:
```suggestion
   * Object containing useful constructors and utility methods for `ConfigReader.Result`s.
```

core/src/main/scala/pureconfig/ConfigReader.scala
```diff
@@ -120,6 +120,35 @@ trait ConfigReader[A] {
...
+   */
+  object Result {
+
+    /**
+     * Merges two `ReaderResult`s using a given function.
```
jcazevedo:
```suggestion
     * Merges two `ConfigReader.Result`s using a given function.
```

core/src/main/scala/pureconfig/ConfigReader.scala
```diff
@@ -120,6 +120,35 @@ trait ConfigReader[A] {
...
+    def zipWith[A, B, C](first: ConfigReader.Result[A], second: ConfigReader.Result[B])(f: (A, B) => C): ConfigReader.Result[C] =
+      (first, second) match {
+        case (Right(a), Right(b)) => Right(f(a, b))
+        case (Left(aFailures), Left(bFailures)) => Left(aFailures ++ bFailures)
+        case (_, l: Left[_, _]) => l.asInstanceOf[Left[ConfigReaderFailures, Nothing]]
+        case (l: Left[_, _], _) => l.asInstanceOf[Left[ConfigReaderFailures, Nothing]]
+      }
+
+    /**
+     * Returns a `ReaderResult` containing a single failure.
```
jcazevedo:
```suggestion
     * Returns a `ConfigReader.Result` containing a single failure.
```

jcazevedo (changes requested):
Looks good! I just pointed out some minor things in the comments. Sorry for the nitpicking.

ruippeixotog:
Ended up renaming them to `Result` in bc52b34, as they are local to `ConfigReader`.

jcazevedo (approved):
> Ended up renaming them to `Result` in bc52b34, as they are local to `ConfigReader`.

Makes sense!

LGTM!

ruippeixotog:
@leifwickland when you have the time please take a look at this. Should be quick to review even though it touches a lot of files :)

PR #443, opened 2018-12-22
 FluentConfigCursor

ruippeixotog:
Creates a `FluentConfigCursor`, a version of `ConfigCursor` with a more fluent interface, focused on config navigation instead of error handling. This is optimized for cases where we want to easily navigate through deep structures and we don't want to worry about doing for-comprehensions or handling `Either` valued each step of the way.

Even though I have a preference for this kind of API, I added this as an alternative for the existing cursors so that we don't break existing code. With time, we can choose to change the default cursor and eventually deprecate and delete the old API. The new API can easily be used by calling `.fluent` on an existing cursor, so it should still be very convenient.

I didn't update the documentation this time. I'll do it later, before we release the next version :)

Depends on #442.


core/src/main/scala/pureconfig/FluentConfigCursor.scala
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Casts this cursor to a `ConfigListCursor` and maps each element to a result. This method tries to map all
+   * elements, combining failures from all of them if more than one exists.
+   *
+   * @param f the function used to map elements
+   * @tparam A the result type of the elements
+   * @return a `Right` with the list obtained by mapping all elements of the list pointed to by this cursor if all
+   *         casts and mappings can be done, `Left` with a list of failures otherwise.
+   */
+  def mapList[A](f: ConfigCursor => ConfigReader.Result[A]) =
```
jcazevedo: Just for consistency with the other public methods in this class:

```suggestion
  def mapList[A](f: ConfigCursor => ConfigReader.Result[A]): ConfigReader.Result[List[A]] =
```

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
```diff
@@ -29,19 +30,13 @@ package object yaml {
           case (k: String, v) => aux(v).right.map { v: AnyRef => k -> v }
           case (k, _) => Left(ConfigReaderFailures(NonStringKeyFound(k.toString, k.getClass.getSimpleName)))
         }
-        entries
-          .foldLeft(Right(Map.empty): ConfigReader.Result[Map[String, AnyRef]])(ConfigReader.Result.zipWith(_, _)(_ + _))
-          .right.map(_.asJava)
+        ConfigReader.Result.sequence[(String, AnyRef), mutable.Iterable](entries).right.map(_.toMap.asJava)
```
jcazevedo: In Scala 2.11, `Either` doesn't extend `Product` nor `Serializable`, so the type of `entries` is inferred as `Product with Serializable with Either[ConfigReaderFailures, (String, AnyRef)]`. Since `mutable.Iterable[A]` is invariant in type `A`, this fails to compile. I think we can either explicitly type the `sequence` method in `Iterable` instead (which is covariant in type `A`) or explicitly type the `entries` variable as `mutable.Iterable[ConfigReader.Result[(String, AnyRef)]]`.

jcazevedo (changes requested):
This looks very good! Thanks, @ruippeixotog!

ruippeixotog:
@jcazevedo @leifwickland can you please take another look at this?

jcazevedo (approved):
Sorry for the slow turnaround time. Looks good!

PR #444, opened 2019-01-12
Scalaz module

ChernikovP:
Added `scalaz` module. Currently, it mostly mimics `cats` module but with some extra functionality.


modules/scalaz/src/main/tut/README.md
```diff
@@ -0,0 +1,147 @@
...
+import pureconfig.module.scalaz._
+import scalaz.{==>>, IList, ISet, Maybe, NonEmptyList}
+import scalaz.std.anyVal.intInstance
+import scalaz.std.string._
+
+case class ScalazConfig(
+  numberLst: IList[Int],
+  numberSet: ISet[Int],
+  numberNel: NonEmptyList[Int],
+  numberMap: ==>>[String, Int],
```
ruippeixotog: Curiosity, isn't `String ==>> Int` the most common syntax here?

modules/scalaz/src/main/scala/pureconfig/module/scalaz/instances/package.scala
```diff
@@ -0,0 +1,57 @@
+package pureconfig.module.scalaz
+
+import com.typesafe.config.ConfigValue
+import pureconfig.{ ConfigConvert, ConfigReader, ConfigWriter }
+import pureconfig.error.{ ConfigReaderFailure, ConfigReaderFailures, FailureReason }
+
+import scalaz.{ Contravariant, Equal, InvariantFunctor, MonadError, Semigroup, Show }
+
+package object instances {
```
ruippeixotog, line 13: Can you add some simple Scaladoc in all the objects? Copying the description from the cats module is fine.

ruippeixotog (changes requested):

modules/scalaz/src/main/tut/README.md
ChernikovP: I think, it's the matter of being used to write `Map[A, B]`, but, anyway, `A ==>> B` looks more natural. Tried to replace it across all module in b1868cd.

modules/scalaz/src/main/scala/pureconfig/module/scalaz/instances/package.scala
ChernikovP: Is this enough - c3543de?

modules/scalaz/build.sbt
```diff
@@ -0,0 +1,16 @@
+name := "pureconfig-scalaz"
+
+libraryDependencies ++= Seq(
+  "org.scalaz" %% "scalaz-core" % "7.2.27",
+  "org.scalaz" %% "scalaz-scalacheck-binding" % "7.2.27-scalacheck-1.13"
```
ruippeixotog: Shouldn't this be a `% "test"` dependency?

ruippeixotog (changes requested):
Other than this minor thing I found now, LGTM!

modules/scalaz/build.sbt
ChernikovP: Yep, missed that. FIxed in ef62dd9.

ruippeixotog (approved):

leifwickland:
@ChernikovP, 
Thanks for your contribution. Please consider sending a PR to add yourself to the [pureconfig contributors list](https://github.com/pureconfig/pureconfig/blob/master/CONTRIBUTORS).

PR #445, opened 2019-01-13
[Cats module] Added support for Foldable Alternative (specifically Chain) and NonEm…

yangzai:
…ptyChain.


modules/cats/build.sbt
```diff
@@ -2,7 +2,8 @@ name := "pureconfig-cats"
 
 libraryDependencies ++= Seq(
   "org.typelevel" %% "cats-core" % "1.4.0",
-  "org.typelevel" %% "cats-laws" % "1.4.0" % "test")
+  "org.typelevel" %% "cats-laws" % "1.4.0" % "test",
+  Dependencies.shapeless % Provided)
```
ruippeixotog: Can you please avoid including Shapeless as a dependency, even if `Provided`? We try to reduce both core and module dependencies to just the essentials so that people don't get unwanted dependency conflicts.

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
```diff
@@ -16,13 +16,17 @@ class CatsSuite extends BaseSuite with ConfigConvertChecks {
...
+  case class NumChain(numbers: NonEmptyChain[Int])
 
   checkReadWrite[Numbers](parseString(s"""{ numbers: [1,2,3] }""").root() → Numbers(NonEmptyList(1, List(2, 3))))
   checkReadWrite[NumVec](parseString(s"""{ numbers: [1,2,3] }""").root() → NumVec(NonEmptyVector(1, Vector(2, 3))))
   checkReadWrite[NumSet](parseString(s"""{ numbers: [1,2,3] }""").root() → NumSet(NonEmptySet(1, SortedSet(2, 3))))
   checkReadWrite[NumMap](parseString(s"""{ 
                                            numbers {"1": 1, "2": 2, "3": 3 }
                                          }""").root() → NumMap(NonEmptyMap(("1", 1), SortedMap("2" → 2, "3" → 3))))
+  // Chain seems to break referential transparency unless prepend/append order is the same
+  checkReadWrite[NumChain](parseString(s"""{ numbers: [1,2,3] }""").root() →
```
ruippeixotog: It's a shame that equality on `Chain` depends on the order of insertion :/ An alternative to hand-crafting the instance in a special way would be to define an implicit instance of `org.scalactic.Equality[NonEmptyChain[Int]]` that compares chains as traversables. I think it would be a more elegant solution here; what do you think?

ruippeixotog (changes requested):
Thanks for the contribution, @yangzai!

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
yangzai: Hi @ruippeixotog , thanks for your feedback. I was having no issues when I try defining an `Equality` for `Chain`, but for `NEC[Int]` there were type erasure issues from its underlying dependent type of `NonEmptyChainImpl.Type`, so I ended up defining an `Equality` for `NumChain` instead.

Also, the equality issue should be fixed from Cats 1.6. https://github.com/typelevel/cats/pull/2690

Also related, https://github.com/typelevel/cats/pull/522 . But it uses `Equivalence` instead of `Equality`. Perhaps the helpers in `ConfigConvertChecks` could depend on `Equivalence` instead.

modules/cats/build.sbt
yangzai: Hi @ruippeixotog , The ConfigReader/Writer I defined depends on `Shapeless` to prevent it from overriding the implicits already defined for `TraversableOnce` and `Option`. I could make it less generic and only work for `Chain`, or I could define a mini `Shapeless` like this:
```scala
// Derived from Shapeless (https://github.com/milessabin/shapeless)
object MiniShapeless {
  def unexpected: Nothing = sys.error("Unexpected invocation")

  type ¬[T] = T => Nothing
  type ¬¬[T] = ¬[¬[T]]
  type ∧[T, U] = T with U
  type ∨[T, U] = ¬[¬[T] ∧ ¬[U]]

  @scala.annotation.implicitNotFound("${A} must not be a subtype of ${B}")
  trait <:!<[A, B] extends Serializable

  implicit def nsub[A, B]: A <:!< B = new <:!<[A, B] {}
  implicit def nsubAmbig1[A, B >: A]: A <:!< B = unexpected
  implicit def nsubAmbig2[A, B >: A]: A <:!< B = unexpected
}
```

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
ruippeixotog: I don't really have a preference between `Equivalance` and `Equality`.

Semantically it seems to make more sense to define convert tests on equality because we are not really doing computation that can take approximate results; in the case of `Chain`, I'd consider the two chains in question equal, not equivalent. Nonetheless, I'd be ok with using `Equivalence` if it helped somehow :)

modules/cats/build.sbt
ruippeixotog: Hmm, I see what you mean. I strongly want to avoid a shapeless dependency here just for this use case, but an alternative solution to defining a `MiniShapeless` would be to use our mechanism to export low-priority readers. If you define an implicit instance of `Exported[ConfigReader[A]]` instead of a `ConfigReader[A]`, that instance will still be used when a `ConfigReader[A]` is requested but only if an instance of `ConfigReader[A]` is not found.

I'd like not to abuse `Exported` in our codebase, but this seems like a valid use case for `Exported`, right @jcazevedo and/or @leifwickland?

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
yangzai: `Equivalence` is not necessary (at least for this PR) as I've already resolved it in https://github.com/pureconfig/pureconfig/pull/445/commits/11240b001be0ec7f943c6791787d9960feff986f with an instance of `Equality[NumChain]` (`NumChain` wraps `NEC[Int]`, kind of like a `ClassTag`) instead of `Equality[NEC]`, unless the latter is preferred. (And it is probably redundant from Cats 1.6 anyway.)

That said I still think that it is better to switch over eventually, perhaps in another PR. ["Scalactic provides both Equality and Equivalence because the Any in Equality can sometimes make things painful."](https://www.artima.com/docs-scalatest-3.0.0-SNAP1/org/scalactic/Equality.html) , and in this case, together with how `NEC` is implemented, causes the type erasure. Plus you will be able to get `cats.Eq` into scalactic for free.

modules/cats/build.sbt
yangzai: Will have a look at `Exported`.

yangzai: @ruippeixotog All change requests resolved unless there are objections to the use of `Exported`.

ruippeixotog (approved):
LGTM!

PR #447, opened 2019-01-25
Fix scaladoc links

nigredo-tori:
Since `core` and `generic` modules were split, there is no scaladoc for the `pureconfig` package - which is just a dummy with dependencies on `pureconfig-core` and `pureconfig-generic`. So the scaladoc links and shields, which used `latest` available documentation, pointed to the outdated (0.9.2) artifact. This MR changes those links to point to the `pureconfig-core` scaladoc.


ruippeixotog (approved):
Well spotted, thanks!

PR #448, opened 2019-01-29
Update cats-effect, fs2 and http4s

bardurdam:
- bump dependencies for cats-effect, fs2 and http4s modules

### DIFF
--- a/modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
+++ b/modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
@@ -2,6 +2,7 @@ package pureconfig.module
 
 import java.nio.charset.StandardCharsets.UTF_8
 
+import scala.collection.compat._
 import scala.language.higherKinds
 import scala.reflect.ClassTag
 import _root_.fs2.{ Stream, text }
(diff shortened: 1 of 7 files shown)


modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
```diff
@@ -2,6 +2,7 @@ package pureconfig.module
 
 import java.nio.charset.StandardCharsets.UTF_8
 
+import scala.collection.compat._
```
ruippeixotog, line 5: Why is this needed now?

ruippeixotog (commented):
LGTM!

modules/fs2/src/main/scala/pureconfig/module/fs2/package.scala
bardurdam: Since version 1.0.3, fs2 is being cross built to Scala 2.13.0-M5. For this they use the `scala.collection.compat` library that provides some of the new APIs from Scala 2.13.

Specifically the `to` combinator on `Stream` was changed from 
`def to[C[_]](implicit cbf: CanBuildFrom[Nothing, O, C[O]]): G[C[O]]` 
to 
`def to[C[_]](implicit f: Factory[O, C[O]]): G[C[O]]`

Hence we need an implicit `collection.compat.Factory` instead of implicit `scala.collection.generic.CanBuildFrom`. 

`Factory` is defined as a type alias as follows: 
`type Factory[-A, +C] <: CanBuildFrom[Nothing, A, C]`

ruippeixotog (approved):

PR #451, opened 2019-02-25
Update dependency versions

ruippeixotog:


leifwickland (approved):
Thanks, Rui.

ruippeixotog:
I had to downgrade tut and microsites due to a weird incompatibility that I couldn't reproduce locally. I'll take care of those two later in another PR.

Issue #455, opened 2019-03-16
pureconfig.Derivation[pureconfig.ConfigReader


charmby:
run the example
case class Config(dirwatch: DirWatchConfig)
    case class DirWatchConfig(path: Path, filter: String, email: EmailConfig)
    case class EmailConfig(host: String, port: Int, message: String, recipients: Set[Email], sender: Email)

    val config = loadConfigOrThrow[Config]

i get this error:
Error:(34, 35) could not find implicit value for parameter reader:pureconfig.Derivation[pureconfig.ConfigReader

ruippeixotog:
Hi @charmby! Our CI builds [compile the example project](https://github.com/pureconfig/pureconfig/blob/master/.travis.yml#L15) for all supported SBT versions, so it shouldn't be possible for that to happen.

Did you compile the example project as is, or have you copied the code to another project? If it's the latter, can you provide a minimal example project where you're experiencing the issue?

charmby:
hi, @ruippeixotog , Thank you for your reply. As a maven project, i run this example in IDEA. Import this project in my pom like this
```
  <dependency>
            <groupId>com.github.pureconfig</groupId>
            <artifactId>pureconfig_2.12</artifactId>
            <version>0.10.0</version>
        </dependency>

        <dependency>
            <groupId>org.scalamacros</groupId>
            <artifactId>paradise_2.11.0-M7</artifactId>
            <version>2.0.0-M8</version>
        </dependency>

        <dependency>
            <groupId>eu.timepit</groupId>
            <artifactId>refined-pureconfig_2.12</artifactId>
            <version>0.9.4</version>
        </dependency> 
```
coding for that
case class Person(name: String, surname: String)
object Test extends App {
    val conf = ConfigFactory.parseString("{ name: John, surname: Doe }")
    import pureconfig.generic.semiauto._
    implicit val personReader = deriveReader[Person]
    val res: Either[ConfigReaderFailures, Person] = pureconfig.loadConfig[Person](conf)
}

dvirf1:
It happens to me as well. It also happens with 0.10.1 and 0.10.2.
It does not happen with 0.9.2.
I'm cleaning the `target` directory between checks.

ruippeixotog:
[...]

PR #456, opened 2019-03-19
adding a friendly message when the implicit for Derivation is not found

dvirf1:
Adding a friendly message when the implicit for Derivation is not found.
See #455

### DIFF
--- a/macros/src/main/scala/pureconfig/Derivation.scala
+++ b/macros/src/main/scala/pureconfig/Derivation.scala
@@ -1,11 +1,17 @@
 package pureconfig
 
+import scala.annotation.implicitNotFound
 import scala.collection.mutable
 import scala.language.experimental.macros
 import scala.reflect.macros.whitebox
 
 import pureconfig.derivation._
 
+@implicitNotFound("""Cannot find an implicit Derivation for ${A}.
+If your application does not define a Derivation elsewhere,
+consider using PureConfig's auto derivation by defining the following:
+
+import pureconfig.generic.auto._""")
 sealed trait Derivation[A] {
   def value: A
 }
--- a/macros/src/main/scala/pureconfig/Derivation.scala
+++ b/macros/src/main/scala/pureconfig/Derivation.scala
@@ -7,11 +7,8 @@ import scala.reflect.macros.whitebox
 
 import pureconfig.derivation._
 
-@implicitNotFound("""Cannot find an implicit Derivation for ${A}.
-If your application does not define a Derivation elsewhere,
-consider using PureConfig's auto derivation by defining the following:
-
-import pureconfig.generic.auto._""")
+@implicitNotFound("""Cannot find an implicit instance of ${A}.
+If you are trying to read or write a case class or sealed trait consider using PureConfig's auto derivation by adding `import pureconfig.generic.auto._`""")
 sealed trait Derivation[A] {
   def value: A
 }


macros/src/main/scala/pureconfig/Derivation.scala
```diff
@@ -1,11 +1,17 @@
...
+import scala.annotation.implicitNotFound
 import scala.collection.mutable
 import scala.language.experimental.macros
 import scala.reflect.macros.whitebox
 
 import pureconfig.derivation._
 
+@implicitNotFound("""Cannot find an implicit Derivation for ${A}.
+If your application does not define a Derivation elsewhere,
+consider using PureConfig's auto derivation by defining the following:
```
leifwickland: What do you think about making the message describe _where_ the `import` needs to be added? A common problem seems to be that users fail to add it in the scope of the `loadConfig*` call.

dvirf1: so something like:
>"""Cannot find an implicit Derivation for ${A}.
If your application does not define a Derivation elsewhere,
consider using PureConfig's auto derivation by defining the following:
>
>import pureconfig.generic.auto._
>
>at the scope of your pureconfig.loadConfig[${A}] call"""

?

leifwickland: Yeah, that's roughly what I was thinking. I suggest you wait until @ruippeixotog takes a look before making any changes.

jcazevedo:
The `Derivation` macro is used to provide better "implicit not found" messages in the presence of the compiler option `-Xmacro-settings:materialize-derivations` (see [this entry](https://pureconfig.github.io/docs/faq.html#how-do-i-debug-implicit-not-found-errors) of the FAQ for more details). It is not expected for application code to define instances of `Derivation` (only of `ConfigReader`s and `ConfigWriter`s). As such, this message can be a bit misleading.

Also, the `Derivation` macro works by setting the `@implicitNotFound` annotation in the `Derivation` type at compile time, so I don't think it will play well with an annotation already set.

dvirf1:
so should I close this PR?

ruippeixotog:
Thanks for the PR @dvirf1, nice to see people taking action when they see something that can be improved :)

I am not completely opposed to the idea of adding this message in @implicitNotFound, particularly after 0.10 where we introduced a breaking change that affects practically all people. However, as @jcazevedo explained, `Derivation` is not a typical type class.

There are two important points here:
- As @jcazevedo mentioned, `Derivation` should be transparent to the user. A user does not (and cannot) create instances of `Derivation[A]` - instead, `Derivation[A]` instances are automatically generated for available implicit instances of `A`. If there was to be a message, it should be clear that `A` is the implicit type missing (e.g. "Could not find an implicit instance of ${A}");
- Right now, `Derivation` is generic enough to be applicable to every type class, not just `ConfigReader` and `ConfigWriter`. However, the proposed message makes that assumption. Should we take the practical approach or try to keep `Derivation` agnostic of its type parameter?

@jcazevedo @leifwickland what do you think?

leifwickland:
My inclination is to take the practical approach and try to help the user by telling them what they need to know to succeed most quickly.

jcazevedo:
What if, when materializing `Derivation[A]` and an implicit search for `A` fails, we check if `A` has an `@implicitNotFound` annotation and promote that annotation to the `Derivation` type?

It'd keep `Derivation` generic while allowing us to have helpful implicit not found messages in the `ConfigReader` and `ConfigWriter` type classes, both for users whose entry point is one of the `loadConfig*` methods and those who just trigger an implicit search for `ConfigReader[A]` or `ConfigWriter[A]`.

ruippeixotog:
Did you mean to close this PR, @leifwickland? Or did you press the wrong button?

@dvirf1 I'd say that apart from cleaning the references to derivation the idea looks fine. I'd write something in the lines of:

> Cannot find an implicit instance of ${A}. If you're trying to read or write a case class or sealed trait consider using PureConfig's auto derivation by importing `pureconfig.generic.auto._`.

I think the import location is not needed because people will assume the suggestion to be where the "implicit not found" error happened (as in, the file/line with which the error message is presented). I'd also include the "case class or sealed trait" remark as those are the only cases covered by `auto`. Feel free to adapt the message as you see fit, though :)

ruippeixotog:
> What if, when materializing Derivation[A] and an implicit search for A fails, we check if A has an @implicitNotFound annotation and promote that annotation to the Derivation type?

That could work now that we don't support 2.10 anymore, but I'd really like to keep `Derivation` just a pass-through implicit search when it is deactivated. We rely on undocumented compiler APIs to override `@implicitNotFound` in classes and I'd hate if a new Scala version ended up breaking PureConfig even when derivations are disabled.

dvirf1:
@ruippeixotog so should I add this message from [your comment](https://github.com/pureconfig/pureconfig/pull/456#issuecomment-475044295)? I'd be happy to do this

leifwickland:
I'm so sorry that I closed the PR. I just hit the wrong button. My apologies.

leifwickland:
I'm so sorry I closed the PR. Of course I hit the wrong button.

ruippeixotog:
> @ruippeixotog so should I add this message from [your comment](https://github.com/pureconfig/pureconfig/pull/456#issuecomment-475044295)? I'd be happy to do this

Yes, I believe the benefits outweigh the drawbacks 👍

dvirf1:
@ruippeixotog I added it with a small change that allows to copy-paste the `import pureconfig.generic.auto._` part

dvirf1:
@ruippeixotog just making sure you saw the notification

ruippeixotog (approved):
Thanks again!

PR #457, opened 2019-04-11
Add a Cron4s config writer

alonsodomin:


modules/cron4s/src/main/scala/pureconfig/module/cron4s/package.scala
```diff
@@ -2,14 +2,17 @@ package pureconfig.module
...
 package object cron4s {
 
-  implicit val cronExprReader: ConfigReader[CronExpr] =
+  implicit val cronExprConfigReader: ConfigReader[CronExpr] =
     ConfigReader.fromString(str =>
       Cron.parse(str).fold(
         err => Left(CannotConvert(str, "CronExpr", err.getMessage)),
         cronExpr => Right(cronExpr)))
+
+  implicit val cronExprConfigWriter: ConfigWriter[CronExpr] =
```
ruippeixotog: Can you combine the reader and writer parts in a `ConfigConvert`? You should be able to do that easily with `ConfigConvert.viaString` (see [pureconfig-joda](https://github.com/pureconfig/pureconfig/blob/master/modules/joda/src/main/scala/pureconfig/module/joda/package.scala) for examples).

ruippeixotog (changes requested):
Thanks for the contribution, @alonsodomin!

modules/cron4s/src/main/scala/pureconfig/module/cron4s/package.scala
alonsodomin: @ruippeixotog thanks to the pointer to `ConfigConvert`, wasn't aware of its existance. Should be sorted now.

jcazevedo (approved):
Looks good. Thanks for this, @alonsodomin!

ruippeixotog (approved):

leifwickland:
@alonsodomin Thanks for your contribution. Please consider sending a PR to add yourself to the [pureconfig contributors list](https://github.com/pureconfig/pureconfig/blob/master/CONTRIBUTORS).

PR #461, opened 2019-04-16
Update http4s to 0.20.0

kubukoz:


leifwickland:
I can't remember what we've done in the past about building against RC versions of dependencies. It feels a little weird.

kubukoz:
It's more stable than the milestones anyway :) and http4s 0.19.0 wasn't a stable release, it was a mistag of 0.19.0-M4 or something like that.

leifwickland:
Ah, that's a fair point.

Ross just said:
> So an RC2 for people to try out the recent changes and final on Monday if
no showstoppers?

If the final is coming Monday, we could merge this now, but I somewhat
doubt that pureconfig will have a release before Monday. I expect waiting
for final might be the most efficient approach.

kubukoz:
Sure, let's wait and I'll update this when 0.20 is there :)

kubukoz:
@leifwickland let's do it! :)

leifwickland (approved):
Thanks, @kubukoz!

kubukoz:
aaaaaand we're green

PR #462, opened 2019-04-17
Additional scalaz data structures

ChernikovP:
Attempt to add `DList`, `Dequeue` and `Heap` data structures to `scalaz` module.

### DIFF
--- a/modules/scalaz/src/main/tut/README.md
+++ b/modules/scalaz/src/main/tut/README.md
@@ -18,8 +18,8 @@ libraryDependencies += "com.github.pureconfig" %% "pureconfig-scalaz" % "0.10.2"
 
 The following `scalaz` data structures are supported:
 
-* `IList`, `ISet`, `Maybe`, `NonEmptyList` and `==>>`
-* `Order[A]` should also be in scope, when you're relying on either `ConfigReader[A ==>> B]` or `ConfigReader[ISet[A]]`.
+* `DList`, `Dequeue`, `IList`, `ISet`, `Heap`, `Maybe`, `NonEmptyList` and `==>>`
+* `Order[A]` should also be in scope, when you're relying on either `ConfigReader[ISet[A]]`, `ConfigReader[Heap[A]]` or `ConfigReader[A ==>> B]`.
 For example, if your `ISet` instance contains `String` values then `Order[String]` can be imported via `scalaz.std.string._`
 
 Here is an usage example:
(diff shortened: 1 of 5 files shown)


ChernikovP:
CI failed due to `README` diff:

```
-// res0: pureconfig.ConfigReader.Result[ScalazConfig] = Right(ScalazConfig(scalaz.DList@284be06d,FullDequeue(OneAnd(1,[]),1,OneAnd(3,[2]),2),[1,2,3],<heap>,Bin(2,Bin(1,Tip(),Tip()),Bin(3,Tip(),Tip())),Just(1),NonEmpty[1,2,3],Bin(three,3,Bin(one,1,Tip,Tip),Bin(two,2,Tip,Tip))))
+// res0: pureconfig.ConfigReader.Result[ScalazConfig] = Right(ScalazConfig(scalaz.DList@53409a9b,FullDequeue(OneAnd(1,[]),1,OneAnd(3,[2]),2),[1,2,3],<heap>,Bin(2,Bin(1,Tip(),Tip()),Bin(3,Tip(),Tip())),Just(1),NonEmpty[1,2,3],Bin(three,3,Bin(one,1,Tip,Tip),Bin(two,2,Tip,Tip))))
```

Unfortunately, this is because `DList` is not a case class and hasn't  overridden `toString` method.

I'm not sure whether it makes sense to modify `.travis.yml` to have `git diff --exit-code ':(exclude)modules/scalaz/README.md';` instead of `git diff --exit-code;`, but otherwise every subsequent build will fail.

Or even more simpler solution is to exclude `DList` example from `README`.

modules/scalaz/src/test/scala/pureconfig/module/scalaz/equality/package.scala
```diff
@@ -0,0 +1,22 @@
+package pureconfig.module.scalaz
+
+import org.scalactic.Equality
+
+import scalaz.{ DList, Heap }
+
+package object equality {
+
+  implicit def dListEquality[A]: Equality[DList[A]] = new Equality[DList[A]] {
```
ruippeixotog, line 9: Can you add a quick comment here and in the instance below explaining why you need to override the default equality?

modules/scalaz/src/main/tut/README.md
```diff
@@ -29,27 +29,33 @@ import com.typesafe.config.ConfigFactory.parseString
...
-  numberMaybe: Maybe[Int]
+  numberMap: String ==>> Int
 )
 ```
 
 We can read a `ScalazConfig` like:
 ```tut:book
 val scalazConf = parseString("""{
-  number-lst: [1,2,3],
+  number-dlist: [1,2,3],
```
ruippeixotog, line 51: Regarding your error, I'd actually leave a single list-like structure here and delete the others. There is little value in showing all the data structures we support if the mechanism to encode and read them is always the same.

ruippeixotog (changes requested):
Thanks for adding this, @ChernikovP!

leifwickland (approved):
I approve in general and leave getting the specifics right to @ruippeixotog.

Thanks for the contribution, @ChernikovP!

ruippeixotog:

PR #464, opened 2019-04-17
Correct typos and eliminate unnecessary code in cats module

ChernikovP:


modules/cats/src/main/scala/pureconfig/module/Cats.scala
```diff
@@ -1,15 +0,0 @@
-package pureconfig.module
-
-import pureconfig.error.FailureReason
-
-object Cats {
```
ChernikovP, line 5: It duplicates `EmptyTraversableFound` from package object. Does it make sense to simply remove it or it's better to mark deprecated since , let's say,  `0.11.0`?

ruippeixotog: Well spotted! Since it's not being used anywhere in the code it should be safe to delete.

ruippeixotog (approved):
Thanks for tidying this up!

PR #466, opened 2019-04-21
Add methods to derive readers, writers and full converters for enumerations

jcazevedo:
This PR adds methods to semi-automatically derive readers, writers and full converters for enumerations encoded as sealed families of case objects.

This is particularly helpful in a semi-automatic approach, because if we opted for an `EnumCoproductHint` to derive a reader/writer for a sealed trait, we still had to provide readers/writers for the subclasses, even though they were left unused. 

This PR also deprecates `EnumCoproductHint`. Even in a fully automatic approach, I think it's preferable to rely on these new methods than deriving unnecessary readers/writers for subclasses.


modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
```diff
@@ -0,0 +1,52 @@
+package pureconfig.generic
+
+import pureconfig.ConfigReader.Result
+import pureconfig.error.NoValidCoproductChoiceFound
+import pureconfig.{ ConfigCursor, ConfigReader }
+import shapeless._
+import shapeless.labelled._
+
+trait EnumerationConfigReaderBuilder[A] {
+  def build(transformName: String => String): ConfigReader[A]
```
ruippeixotog, line 16: Instead of parameterizing this with `transformName`, have you considered instead using `contramapConfig` or `contramapCursor` on the resulting reader to map the config strings before they are read?

That way you would have the naming logic separate from the derivation and could create a simpler `trait EnumerationConfigReader[A] extends ConfigReader[A]` trait instead of creating a completely new type class.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
```diff
@@ -0,0 +1,52 @@
...
+      def build(transformName: String => String): ConfigReader[CNil] =
+        new ConfigReader[CNil] {
+          def from(cur: ConfigCursor): Result[CNil] = cur.failed(NoValidCoproductChoiceFound(cur.value))
+        }
+    }
+
+  implicit def deriveEnumerationReaderBuilderCCons[K <: Symbol, H, T <: Coproduct](
+    implicit
+    vName: Witness.Aux[K],
+    hGen: LabelledGeneric.Aux[H, HNil],
```
ruippeixotog, line 31: Nice advantage, getting a compile-time guarantee that this is a well-formed enumeration.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigWriterBuilder.scala
```diff
@@ -0,0 +1,50 @@
+package pureconfig.generic
+
+import com.typesafe.config.{ ConfigValue, ConfigValueFactory }
+import pureconfig.ConfigWriter
+import shapeless._
+import shapeless.labelled._
+
+trait EnumerationConfigWriterBuilder[A] {
+  def build(transformName: String => String): ConfigWriter[A]
```
ruippeixotog, line 15: Same things here with `mapConfig` (we probably want to create a `mapCursor` for this?).

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigWriterBuilder.scala
```diff
@@ -0,0 +1,50 @@
...
+}
+
+object EnumerationConfigWriterBuilder {
+  implicit val deriveEnumerationWriterBuilderCNil: EnumerationConfigWriterBuilder[CNil] =
+    new EnumerationConfigWriterBuilder[CNil] {
+      def build(transformName: String => String): ConfigWriter[CNil] =
+        new ConfigWriter[CNil] {
+          def to(a: CNil): ConfigValue = ConfigValueFactory.fromAnyRef(null)
```
ruippeixotog: This is never supposed to happen, right? If that's the case, I'd throw an `IllegalStateException` with an appropriate message here to better signal a bug in our code. Otherwise, this may pass like intended behavior and it would be harder for users to figure out what's happening.

(I noticed now that `cNilWriter` does the same thing as here. Maybe we can update both?)

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
```diff
@@ -13,4 +13,32 @@ object semiauto {
 
   final def deriveConvert[A](implicit reader: Lazy[DerivedConfigReader[A]], writer: Lazy[DerivedConfigWriter[A]]): ConfigConvert[A] =
     ConfigConvert.fromReaderAndWriter(Derivation.Successful(reader.value), Derivation.Successful(writer.value))
+
+  final def deriveEnumerationReader[A](
+    implicit
+    readerBuilder: Lazy[EnumerationConfigReaderBuilder[A]]): ConfigReader[A] =
+    deriveEnumerationReader(_.toLowerCase)
```
ruippeixotog: Question for later: what do you think of doing a breaking change to make (PascalCase, KebabCase) the default transformation to use for coproduct hints and enumerations?

I believe that the only reason the default is `.toLowerCase` was because that's what `spray-json-shapeless` used, but it doesn't seem to be a popular one. Technically it's also ambiguous, as classes like `MySQL`, `MySql` and `Mysql` all map to the same config value. In product hints we are already assuming Scala code to be pascal case and Typesafe configs to be kebab case, so maybe we can make this consistent across the library?

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
```diff
@@ -13,4 +13,32 @@ object semiauto {
...
+    implicit
+    writerBuilder: Lazy[EnumerationConfigWriterBuilder[A]]): ConfigWriter[A] =
+    deriveEnumerationWriter(_.toLowerCase)
+  final def deriveEnumerationConvert[A](
+    implicit
+    readerBuilder: Lazy[EnumerationConfigReaderBuilder[A]],
+    writerBuilder: Lazy[EnumerationConfigWriterBuilder[A]]): ConfigConvert[A] =
+    deriveEnumerationConvert(_.toLowerCase)
+
+  final def deriveEnumerationReader[A](transformName: String => String)(
```
ruippeixotog, line 49: I'm not very fond of polluting the semiauto API with variable-argument alternatives. The approach we took with the `pureconfig` package object was already shown many times not to be scalable :/

Hints win here on this specific case in terms of flexibility. With them you can customize an "unlimited" number of aspects while providing only a single instance as an implicit instance. Adding "hint" type classes for enumerations may be a bit overkill and may also be optimizing prematurely, though. Upgrading this to use a  magnet pattern here may be a reasonable solution.

Can you see any alternative to having an overloaded method here? Otherwise we might be ok as it is for now...

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
```diff
@@ -0,0 +1,52 @@
+package pureconfig.generic
+
+import pureconfig.ConfigReader.Result
+import pureconfig.error.NoValidCoproductChoiceFound
+import pureconfig.{ ConfigCursor, ConfigReader }
+import shapeless._
+import shapeless.labelled._
+
+trait EnumerationConfigReaderBuilder[A] {
```
ruippeixotog, line 15: Can you add Scaladoc to every new type class and their respective methods?

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigWriterBuilder.scala
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+
+  implicit def deriveEnumerationWriterBuilder[A, Repr <: Coproduct](
+    implicit
+    gen: LabelledGeneric.Aux[A, Repr],
+    reprWriterBuilder: EnumerationConfigWriterBuilder[Repr]): EnumerationConfigWriterBuilder[A] =
+    new EnumerationConfigWriterBuilder[A] {
+      def build(transformName: String => String): ConfigWriter[A] = {
+        lazy val reprWriter = reprWriterBuilder.build(transformName)
```
ruippeixotog: You can use `ConfigWriter#contramap` to simplify this.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+
+  implicit def deriveEnumerationReaderBuilder[A, Repr <: Coproduct](
+    implicit
+    gen: LabelledGeneric.Aux[A, Repr],
+    reprReaderBuilder: EnumerationConfigReaderBuilder[Repr]): EnumerationConfigReaderBuilder[A] =
+    new EnumerationConfigReaderBuilder[A] {
+      def build(transformName: String => String): ConfigReader[A] = {
+        lazy val reprReader = reprReaderBuilder.build(transformName)
```
ruippeixotog: You can use `ConfigReader#map` to simplify this.

tests/src/test/scala/pureconfig/DerivationModesSuite.scala
```diff
@@ -56,6 +61,78 @@ class DerivationModesSuite extends BaseSuite {
...
+    implicit val colorWriter = deriveEnumerationWriter[Color]
+
+    ConfigWriter[Color].to(RainyBlue) shouldEqual ConfigValueFactory.fromAnyRef("rainyblue")
+    ConfigWriter[Color].to(SunnyYellow) shouldEqual ConfigValueFactory.fromAnyRef("sunnyyellow")
+  }
+
+  it should "provide methods to derive full converters for enumerations encoded as sealed traits" in {
+    import pureconfig.generic.semiauto._
+
+    import pureconfig.generic.semiauto._
```
ruippeixotog: Duplicate import here.

docs/src/main/tut/docs/overriding-behavior-for-sealed-families.md
```diff
@@ -63,20 +63,21 @@ Then load the config:
 loadConfig[AnimalConf](ConfigFactory.parseString("{ type: Bird, can-fly: true }"))
 ```
 
-With a `CoproductHint` you can even opt not to use any extra field at all. If you encode enumerations using sealed
-traits, you can just write the lowercase name of the class by using an `EnumCoproductHint`. For example, if we create
-an enumeration for seasons:
+If you encode enumerations using sealed traits of case objects, you can use the
+`deriveEnumeration(Reader|Writer|Convert)` methods from the
```
ruippeixotog: Last time we reviewed the documentation on our website, we decided to focus it on the config reading use case (arguably the most important one) and have a dedicated "Config Writers" section to let users know that PureConfig is also able to the reverse operation. It makes the documentation simpler for newcomers and avoids the need for us to always enumerate reader/writer and `ConfigReader`/`ConfigWriter`/`ConfigConvert` on concepts that are clearly extensible to the writing part.

With that in mind, I'd avoid talking about writers here and keep the examples using `deriveEnumerationReader`.

ruippeixotog (changes requested):
Great work, @jcazevedo! I love how we now check at compile-time whether we are trying to derive an enum reader for non-compatible sealed families.

One question: the issue you mention about using coproduct hints with semiauto derivation is not specific to enumerations, right? For other sealed families you also need to derive readers for all subclasses even though you may not them, right?

docs/src/main/tut/docs/overriding-behavior-for-sealed-families.md
```diff
@@ -86,3 +87,21 @@ We can load seasons by specifying them by class name:
...
+By default, enumerations will be encoded as strings with the lowercase name of
+the class, but that behavior can be overridden by specifying a different
+transformation function.
+
+```tut:silent
+sealed trait Color
+case object RainyBlue extends Color
+case object SunnyYellow extends Color
+
+implicit val colorReader: ConfigReader[Color] = deriveEnumerationReader[Color](ConfigFieldMapping(PascalCase, KebabCase))
```
leifwickland: That's nice. Kudos.

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
leifwickland: That lowercase default is an annoyance to me. I'd forgotten about it last week and spent a few minutes wondering why nothing worked as I expected.

I don't love the idea of a breaking change but moving in the direction of a more consistent API with less surprising defaults seems like a good reason for it.

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
leifwickland: It's be a cheat but we could make the to lower the default argument to this version and get rid of the no arg version.

leifwickland (approved):
I don't have a strong feeling about Rui's suggestions.

Thanks for this improvement.

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
leifwickland: After thinking about it more, if we're going to make a change change to the naming conventions for enums, I think doing it when this API is introduced would be preferable to waiting until afterward and then breaking this API. My vote is that if we're going to switch, we should do in this PR.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
jcazevedo: I might be overlooking something, but doesn't that require us to have two transformation functions: one from type name to config string, and vice versa?

I'd rather have this configurable with a single transformation function. It's the way we're doing with fields in `ProductHint`s, and not all transformation functions are easily invertible. It's also more error-prone because one can easily supply two functions for a converter that are not the inverse of one another.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigWriterBuilder.scala
jcazevedo: Yes, it should never happen. I copied this behavior from `cNilWriter`, but I agree that throwing an `IllegalStateException` is a better idea. I've done that in 6d88116.

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
jcazevedo: I also prefer having pascal case to kebab case as the default transformation. I agree with @leifwickland that this a good opportunity to do it. I've done it in 2fdbf7a.

I'm not sure of how to put this in the changelog, though. Should we put it as a breaking change, or not mention it at all, since it's the behavior of a new method?

leifwickland: I suggest that the changelog mention that the old way is deprecated and mention in the new features section that the new approach has a different naming behavior than the old.

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
jcazevedo: I think both the magnet pattern and the default argument are good alternatives to avoid the overloading. I went this way to be able to call `deriveEnumerationReader[T]` without the extra `()`, for consistency with the `deriveReader[T]` method calls. We won't be able to do it with default arguments, and I also think that using the magnet pattern would force us to have the extra `()` for the default case (not sure about this).

I think it's premature to add an extra hint for enumerations.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
jcazevedo: I added some Scaladoc in a26cebd.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
jcazevedo: Right! Improved it in 5e3d45e.

tests/src/test/scala/pureconfig/DerivationModesSuite.scala
jcazevedo: Import removed in fc3ad4a.

docs/src/main/tut/docs/overriding-behavior-for-sealed-families.md
jcazevedo: Makes sense. I've changed the documentation to only mention the reading use case in b5d0c6a.

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
ruippeixotog: @jcazevedo if we are to change the lowercase default it should not be only for enumeration readers, but for other coproduct hints too - otherwise we'll be encoding the same data (class names) differently in different hints, which is worse than being inconsistent between product and coproduct hints. Doing that requires "real" breaking changes, though :)

jcazevedo:
> One question: the issue you mention about using coproduct hints with semiauto derivation is not specific to enumerations, right? For other sealed families you also need to derive readers for all subclasses even though you may not them, right?

Yes. It is just more obvious that you don't need them with `EnumCoproductHint`.

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
jcazevedo: I'm OK with introducing that breaking change for the next version. I suppose we can improve the `NoValidCoproductChoiceFound` failure to show more information when deriving a reader with `FieldCoproductHint`. Showing the value found in the discriminating field together with a mention of the changed behavior should be helpful for users to migrate.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
ruippeixotog: As discussed offline, while I'm still not fond of mixing naming logic with derivation code, I can't see an alternative since we only have the set of valid alternatives at derivation time - the thing we'd need if we wanted to keep using a single function. Let's keep this as it is then.

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
ruippeixotog: Cool! Can you do that change on this PR, to avoid having two different conventions when we merge this? :)

jcazevedo: Sure. I'll update this tomorrow.

jcazevedo: I changed the default transformation of `FieldCoproductHint` to `PascalCase` -> `kebab-case` in 1735d05. 

I introduced a new `UnexpectedValueForFieldCoproductHint` failure reason, in order to improve error messages when using `FieldCoproductHint`. Incidentally, I moved the responsibility of raising `ConfigReaderFailures` when the disambiguation fails to `CoproductHint`. I think this also adds some extra flexibility to implementations of `CoproductHint`, but let me know what you think.

Sorry for the delay!

modules/generic/src/main/scala/pureconfig/generic/CoproductHint.scala
```diff
@@ -54,18 +66,20 @@ trait CoproductHint[T] {
  * This hint will cause derived `ConfigConvert` instance to fail to convert configs to objects if the object has a
  * field with the same name as the disambiguation key.
  *
- * By default, the field value written is the class or coproduct option name converted to lower case. This mapping can
+ * By default, the field value written is the class or coproduct option name converted to kebab case. This mapping can
  * be changed by overriding the method `fieldValue` of this class.
  */
 class FieldCoproductHint[T](key: String) extends CoproductHint[T] {
 
+  private val fieldTransformation: String => String = ConfigFieldMapping(PascalCase, KebabCase)
```
ruippeixotog: Can we move this to the companion object of `FieldCoproductHint`, maybe as `defaultMapping` for consistency with the value's class name? `FieldCoproductHint` instances are created on demand for every sealed trait involved in auto-derivation, there is no need to make the class heavier here.

modules/generic/src/main/scala/pureconfig/generic/CoproductHint.scala
```diff
@@ -46,6 +48,16 @@ trait CoproductHint[T] {
...
   def tryNextOnFail(name: String): Boolean
+
+  /**
+   * Returns a failed `ConfigReader` result scoped into the context of a `ConfigCursor` at the sealed family option,
+   * representing the failure to disambiguate the config value.
+   *
+   * @param cur a `ConfigCursor` at the sealed family option
+   * @return a failed `ConfigReader` result scoped into the context of a `ConfigCursor`.
+   */
+  def failToDisambiguate(cur: ConfigCursor): ConfigReader.Result[CNil] =
```
ruippeixotog: I like the idea of allowing users to customize this :) Two remarks here:
- I'd prefer if we didn't expose `CNil` here, as we've been keeping hints shapeless-free and this result type seems unnecessarily complex. Since we always require a failed result, why don't we simply ask for a `FailureReason` or a `ConfigReaderFailure`?
- The name `failToDisambiguate` is misleading here (and "disambiguate" is probably overused in the Scaladoc of this trait). It implies that we somewhat identify when a config value is ambiguous (i.e. when we're not sure whether it refers to class X or Y), while that isn't the case (e.g. if some hint implementation always returns `Right(Some(x))` every value is ambiguous but we won't ever call this method). This method is only triggered when we find no suitable option; what do you think of naming it `noOptionFound` instead?

modules/generic/src/main/scala/pureconfig/generic/error/UnexpectedValueForFieldCoproductHint.scala
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * A failure reason given when an unknown value was found in the discriminating field of a config value, when using a
+ * `FieldCoproductHint`.
+ *
+ * @param value the value found in the discriminating field of a config value
+ */
+final case class UnexpectedValueForFieldCoproductHint(value: ConfigValue) extends FailureReason {
+  def description =
+    s"Unexpected value ${value.render(ConfigRenderOptions.concise())} found. You might have a misconfigured " +
+      "FieldCoproductHint. Note that the default transformation of FieldCoproductHint changed from converting to " +
```
ruippeixotog: "You might have a misconfigured FieldCoproductHint" may be a bit confusing for people that never used coproduct hints before. Do you think we could make the message more beginner-friendly by avoiding mentioning `FieldCoproductHint`, simply explaining the naming convention that changed and pointing out to https://pureconfig.github.io/docs/overriding-behavior-for-sealed-families.html or to the CHANGELOG for more information?

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+### 0.10.3 (unreleased)
+
+- Breaking changes
+  - The default transformation in `FieldCoproductHint` changed from converting class names to lower case to converting
+    them to kebab case (e.g. `ConfigOption` is converted to `config-option`). The old behavior can be retained by
+    putting in scope an `implicit def coproductHint[T] = new FieldCoproductHint[T]("type") { override def
+    fieldValue(name: String): String = name.toLowerCase }`;
+
+- New features
```
ruippeixotog, line 9: We should probably add here the ability to configure the error message in coproduct hints, even though it's probably only important for us and for the few people that needed to implement their own hints.

modules/generic/src/main/scala/pureconfig/generic/error/UnexpectedValueForFieldCoproductHint.scala
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * A failure reason given when an unknown value was found in the discriminating field of a config value, when using a
+ * `FieldCoproductHint`.
+ *
+ * @param value the value found in the discriminating field of a config value
+ */
+final case class UnexpectedValueForFieldCoproductHint(value: ConfigValue) extends FailureReason {
+  def description =
+    s"Unexpected value ${value.render(ConfigRenderOptions.concise())} found. You might have a misconfigured " +
```
leifwickland: I love this error message. Great job.

Do you think it'd be worth adding the type of the ADT/sealed family?

ruippeixotog (changes requested):

modules/generic/src/main/scala/pureconfig/generic/CoproductHint.scala
jcazevedo: Makes sense. I moved the default transformation to the companion object of `FieldCoproductHint` in 94d2401.

modules/generic/src/main/scala/pureconfig/generic/CoproductHint.scala
jcazevedo: I agree with both points. I prefer asking for `ConfigReaderFailures` because it allows implementations to return more than one failure. I did those changes in 8da8f67.

modules/generic/src/main/scala/pureconfig/generic/error/UnexpectedValueForFieldCoproductHint.scala
jcazevedo: I suppose mentioning `FieldCoproductHint`s makes the message a bit confusing, yes. I've changed it in 2bdaad1.

CHANGELOG.md
jcazevedo: I think we should add it, yes. Done in be94389.

ruippeixotog (approved):
LGTM, thanks for addressing all the comments!

modules/generic/src/main/scala/pureconfig/generic/error/UnexpectedValueForFieldCoproductHint.scala
jcazevedo: Sorry! I missed this message.

I think we can add the type of the sealed family here. We need to be careful, though, because this error will be scoped under the `type` field (or whatever the `key` of `FieldCoproductHint` is), so we need to be explicit that we're not trying to read that type from that field.

If we opt to add the type of the sealed family to this `FailureReason`, I think we should also do it for `NoValidCoproductChoiceFound`. I think doing so will complicate a bit the current implementation for deriving enumerations because we currently don't know the original type when we're deriving an `EnumerationConfigReaderBuilder` for the generic representation. Taking that into account, I'd rather leave that for another PR to avoid this getting too big.

leifwickland: No worries. Feel free to decide it's not worth the bother to add ultimately.

PR #470, opened 2019-04-29
 Add support for Scala 2.13.0-RC1

### LINKED ISSUE #467: Support Scala 2.13-RC1?
Would be helpful :)

ruippeixotog:
Starts compiling and publishing `pureconfig-core`, `pureconfig-generic` and `pureconfig` (bundle project) for Scala 2.13.0-RC1. Incidentally, this PR also cleans up linter flags, making them detect more cases.

~I had to disable coverage publishing as the scoverage SBT plugin isn't yet published for Scala 2.13-RC1. This shouldn't be a problem since we don't rely much on coverage data anyway, but I'm tracking their progress and I'll reenable it again once everything is published on their side.~

Closes #467.


core/src/main/scala/pureconfig/CollectionWriters.scala
```diff
@@ -41,7 +41,7 @@ trait CollectionWriters {
 
   implicit def mapWriter[T](implicit configConvert: Derivation[ConfigWriter[T]]) = new ConfigWriter[Map[String, T]] {
     override def to(keyVals: Map[String, T]): ConfigValue = {
-      ConfigValueFactory.fromMap(keyVals.mapValues(configConvert.value.to).asJava)
+      ConfigValueFactory.fromMap(keyVals.mapValues(configConvert.value.to).toMap.asJava)
```
leifwickland, line 44: Just out curiosity, why is the `.toMap` helpful?

core/src/main/scala/pureconfig/ConfigReader.scala
```diff
@@ -138,8 +137,8 @@ object ConfigReader extends BasicReaders with CollectionReaders with ProductRead
     /**
      * Sequences a collection of `Result`s into a `Result` of a collection.
      */
-    def sequence[A, CC[X] <: TraversableOnce[X]](rs: CC[ConfigReader.Result[A]])(implicit cbf: CanBuildFrom[CC[A], A, CC[A]]): ConfigReader.Result[CC[A]] = {
-      rs.foldLeft[ConfigReader.Result[mutable.Builder[A, CC[A]]]](Right(cbf())) {
+    def sequence[A, CC[X] <: TraversableOnce[X]](rs: CC[ConfigReader.Result[A]])(implicit cbf: FactoryCompat[A, CC[A]]): ConfigReader.Result[CC[A]] = {
```
leifwickland, line 140: Kudos. This turn out really well.

leifwickland (approved):

leifwickland:
Thanks for doing this, @ruippeixotog!

core/src/main/scala/pureconfig/CollectionWriters.scala
ruippeixotog: In 2.13 I get this error:

```
[error] /Users/rui/Documents/Dev/own/pureconfig/core/src/main/scala/pureconfig/CollectionWriters.scala:44:76: type mismatch;
[error]  found   : Iterable[(String, com.typesafe.config.ConfigValue)]
[error]  required: java.util.Map[String, ?]
[error]  Note: implicit method mapWriter is not applicable here because it comes after the application point and it lacks an explicit result type
[error]       ConfigValueFactory.fromMap(keyVals.mapValues(configConvert.value.to).asJava)
```

I believe that `mapValues` now returns an `Iterable` instead of a `Map`.

leifwickland: Oh, you're right: 2.13 [changes the return type of `mapValues`](https://www.scala-lang.org/api/2.13.0-RC1/scala/collection/Map.html#mapValues[W](f:V=%3EW):scala.collection.MapView[K,W]) to a `MapView`. Interesting. 

Sorry I didn't think to look before asking.

build.sbt
```diff
@@ -82,12 +81,14 @@ lazy val commonSettings = Seq(
...
-    case _ => allVersionLintFlags
-  }),
+  scalacOptions ++= allVersionLintFlags ++ {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some(v) => lintFlags.getOrElse(v, Nil)
+      case _ => Nil
+    }
+  },
 
+  scalacOptions in Test ~= { _.filterNot(_.contains("-Ywarn-unused")) },
```
jcazevedo, line 91: I think that disabling this for tests allows us to promote the `DerivationModesSuite2_12` spec to all Scala versions.

ruippeixotog: Good thing you remembered that. Merged the two files in 1c9eddf

jcazevedo (approved):

PR #473, opened 2019-05-15
Add circe module

moradology:
This PR adds a very simple circe module which allows reading/writing of `io.circe.Json` instances as segments of a configuration.

Note: The implicit search in `tut` didn't seem to like my `Decoder` without wrapping things up in an object (I assume as a result of REPL limitations). As a result, the second example's code looks a bit strange

### DIFF
--- a/modules/circe/README.md
+++ b/modules/circe/README.md
@@ -1,0 +1,52 @@
+# Circe module for PureConfig
+
+Adds support for [Circe](https://circe.github.io/circe/) `Json` to PureConfig.
+
+## Add pureconfig-circe to your project
+
+In addition to [core pureconfig](https://github.com/pureconfig/pureconfig), you'll need:
+
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-circe" % "0.11.0"
+```
+
+## Example
+
+Imports to be used below:
+
+```scala
+import io.circe._
+import com.typesafe.config.ConfigFactory
+import pureconfig._
+import pureconfig.generic.auto._
+import pureconfig.module.circe._
+import pureconfig.error.CannotConvert
+```
+
+### Reading Json  values directly
+
+Imagine a class, `UserConfig`, that has a `Json` field:
+
+```scala
+case class UserConfig(username: String, age: Int, custom: Json)
+```
+
+A `UserConfig` can be read like this:
+
+```scala
+val conf = ConfigFactory.parseString("""{
+  username = nathan
+  age = 31
+  custom = {
+    favoriteFood = pizza
+  }
+}""")
+// conf: com.typesafe.config.Config = Config(SimpleConfigObject({"age":31,"custom":{"favoriteFood":"pizza"},"username":"nathan"}))
+
+loadConfig[UserConfig](conf)
+// res0: pureconfig.ConfigReader.Result[UserConfig] =
+// Right(UserConfig(nathan,31,{
+//   "favoriteFood" : "pizza"
+// }))
+```
+
--- a/modules/circe/README.md
+++ b/modules/circe/README.md
@@ -20,10 +20,9 @@ import com.typesafe.config.ConfigFactory
 import pureconfig._
 import pureconfig.generic.auto._
 import pureconfig.module.circe._
-import pureconfig.error.CannotConvert
 ```
 
-### Reading Json  values directly
+### Reading Json values directly
 
 Imagine a class, `UserConfig`, that has a `Json` field:
 
--- a/modules/circe/build.sbt
+++ b/modules/circe/build.sbt
@@ -1,0 +1,16 @@
+name := "pureconfig-circe"
+
+libraryDependencies ++= Seq(
+  "io.circe" %% "circe-core" % "0.11.1",
+  "io.circe" %% "circe-literal" % "0.11.1" % Test,
+  "org.typelevel" %% "jawn-parser" % "0.14.2" % Test
+)
+
+developers := List(
+  Developer("moradology", "Nathan Zimmerman", "npzimmerman@gmail.com", url("https://github.com/moradology")))
+
+osgiSettings
+
+OsgiKeys.exportPackage := Seq("pureconfig.module.circe.*")
+OsgiKeys.privatePackage := Seq()
+OsgiKeys.importPackage := Seq(s"""scala.*;version="[${scalaBinaryVersion.value}.0,${scalaBinaryVersion.value}.50)"""", "*")
--- a/modules/circe/src/main/scala/pureconfig/module/circe/package.scala
+++ b/modules/circe/src/main/scala/pureconfig/module/circe/package.scala
@@ -1,0 +1,55 @@
+package pureconfig.module
+
+import pureconfig.{ ConfigWriter, ConfigReader }
+import com.typesafe.config._
+import io.circe._
+import com.typesafe.config._
+
+import scala.collection.JavaConverters._
+
+package object circe {
+  private def cvToJson(cv: ConfigValue): Json = {
+    cv.valueType match {
+      case ConfigValueType.NULL =>
+        Json.Null
+      case ConfigValueType.BOOLEAN =>
+        Json.fromBoolean(cv.unwrapped.asInstanceOf[Boolean])
+      case ConfigValueType.NUMBER =>
+        cv.unwrapped match {
+          case i: java.lang.Number if i.longValue() == i =>
+            Json.fromLong(i.longValue())
+          case d: java.lang.Number =>
+            Json.fromDoubleOrNull(d.doubleValue())
+        }
+      case ConfigValueType.STRING =>
+        Json.fromString(cv.unwrapped.asInstanceOf[String])
+      case ConfigValueType.LIST => {
+        val iter = cv.asInstanceOf[ConfigList].asScala.map(cvToJson _)
+        Json.fromValues(iter)
+      }
+      case ConfigValueType.OBJECT => {
+        val jsonMap = cv.asInstanceOf[ConfigObject].asScala.mapValues(cvToJson _).toMap
+        val jsonObj = JsonObject.fromMap(jsonMap)
+        Json.fromJsonObject(jsonObj)
+      }
+    }
+  }
+
+  private def jsonToCv(json: Json): ConfigValue = {
+    json.fold(
+      ConfigValueFactory.fromAnyRef(null),
+      bool => ConfigValueFactory.fromAnyRef(bool),
+      jnum => jnum.toLong match {
+        case Some(long) => ConfigValueFactory.fromAnyRef(long)
+        case None => ConfigValueFactory.fromAnyRef(jnum.toDouble)
+      },
+      str => ConfigValueFactory.fromAnyRef(str),
+      arr => ConfigValueFactory.fromIterable(arr.map(jsonToCv).asJava),
+      obj => ConfigValueFactory.fromMap(obj.toMap.mapValues(jsonToCv).asJava))
+  }
+
+  implicit val circeJsonReader: ConfigReader[Json] =
+    ConfigReader[ConfigValue].map(cvToJson)
+  implicit val circeJsonWriter: ConfigWriter[Json] =
+    ConfigWriter[ConfigValue].contramap(jsonToCv)
+}
--- a/modules/circe/src/main/tut/README.md
+++ b/modules/circe/src/main/tut/README.md
@@ -20,10 +20,9 @@ import com.typesafe.config.ConfigFactory
 import pureconfig._
 import pureconfig.generic.auto._
 import pureconfig.module.circe._
-import pureconfig.error.CannotConvert
 ```
 
-### Reading Json  values directly
+### Reading Json values directly
 
 Imagine a class, `UserConfig`, that has a `Json` field:
 
--- a/modules/circe/src/main/tut/README.md
+++ b/modules/circe/src/main/tut/README.md
@@ -1,0 +1,47 @@
+# Circe module for PureConfig
+
+Adds support for [Circe](https://circe.github.io/circe/) `Json` to PureConfig.
+
+## Add pureconfig-circe to your project
+
+In addition to [core pureconfig](https://github.com/pureconfig/pureconfig), you'll need:
+
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-circe" % "0.11.0"
+```
+
+## Example
+
+Imports to be used below:
+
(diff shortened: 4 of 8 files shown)


modules/circe/src/main/tut/README.md
```diff
@@ -0,0 +1,92 @@
...
+piggyback on `ConfigReader[Json]`:
+
+```tut:book
+// Ignore this object; it is here to help tut find implicits
+object DecodingExample {
+  implicit val myDecoder: Decoder[MySpecialClass] = new Decoder[MySpecialClass] {
+    final def apply(c: HCursor): Decoder.Result[MySpecialClass] =
+      (for {
+        foo <- c.downField("foo").as[String]
+        bar <- c.downField("bar").as[Int]
```
jcazevedo: Since `Either` only became right-biased in Scala 2.12.x, we should add `.right` to these in order for this to compile in Scala 2.11.x.

modules/circe/src/main/tut/README.md
```diff
@@ -0,0 +1,92 @@
...
+}""")
+
+loadConfig[UserConfig](conf)
+```
+
+### Using with circe decoders
+
+If you've already defined a circe decoder for the type you want to parse
+from configuration, the `Json` `ConfigWriter` can be used to define a
+`ConfigWriter` for your type:
```
jcazevedo: This should mention `ConfigReader` to be consistent with the example below.

modules/circe/src/test/scala/pureconfig/circe/CirceSuite.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import io.circe._
+import io.circe.literal._
+import org.scalatest._
+import com.typesafe.config.{ ConfigFactory, ConfigValue }
+import pureconfig._
+import pureconfig.syntax._
+import pureconfig.generic.auto._
+
+class JavaxSuite extends FlatSpec with Matchers with EitherValues {
```
jcazevedo: We should name this spec accordingly. Also can you move this file to the folder that mimics the package structure (i.e. `/modules/circe/src/test/scala/pureconfig/module/circe/CirceSuite.scala`)?

modules/circe/src/test/scala/pureconfig/circe/CirceSuite.scala
```diff
@@ -0,0 +1,36 @@
...
+    }
+  """
+  val config = ConfigFactory.parseString(confString)
+
+  it should "be able to read a config as circe json" in {
+    config.to[JsonConf].right.value shouldEqual JsonConf(confJson)
+  }
+
+  it should "be able to write a config as circe json" in {
+    ConfigWriter[JsonConf].to(JsonConf(confJson)) shouldEqual config.to[ConfigValue].right.value
```
jcazevedo: We shouldn't need to make `Config` go through a `ConfigWriter` of `ConfigValue`. Comparing to `config.root()` should be enough:

```scala
ConfigWriter[JsonConf].to(JsonConf(confJson)) shouldEqual config.root()
```

modules/circe/src/main/tut/README.md
```diff
@@ -0,0 +1,92 @@
...
+  val requiredField: Int
+}
+```
+
+If you want to parse instances of `MySpecialClass` directly, just
+piggyback on `ConfigReader[Json]`:
+
+```tut:book
+// Ignore this object; it is here to help tut find implicits
+object DecodingExample {
```
jcazevedo: Is this object really necessary? I seem to be able to make `tut` happy without it.

modules/circe/src/main/tut/README.md
```diff
@@ -0,0 +1,92 @@
...
+          val requiredField: Int = bar
+          val optionalField: String = foo
+        }
+      }) match {
+        case success@Right(_) =>
+          success
+        case fail@Left(_) =>
+          Right(new MySpecialClass { val requiredField: Int = 42 })
+      }
+  }
```
jcazevedo: This sample `Decoder` is slightly convoluted. Also, I don't find the use case of producing a `ConfigReader` from a `Decoder` very compelling. Converting to an intermediate format is inefficient and makes you lose some information relevant for error reporting. I'm not sure if we should mention this in the README of the module.

modules/circe/src/main/scala/pureconfig/module/circe/package.scala
```diff
@@ -0,0 +1,39 @@
...
+import io.circe.parser._
+import com.typesafe.config.{ ConfigValueFactory, ConfigValue, ConfigRenderOptions }
+
+import scala.collection.JavaConverters._
+
+package object circe {
+  def cvToJson(cv: ConfigValue): Either[FailureReason, Json] = {
+    val renderOptions = ConfigRenderOptions.concise().setJson(true)
+    val jsonString = cv.render(renderOptions)
+    parse(jsonString) match {
```
jcazevedo: I prefer relying on `ConfigValue`'s `valueType()` to build a `Json` instead of converting the `ConfigValue` to a string with the JSON representation and going through Circe's parser. It would avoid this returning an `Either` (we should always be able to produce a `Json` from a `ConfigValue`) and would also remove the need for the `circe-parser` dependency in this module.

modules/circe/build.sbt
```diff
@@ -0,0 +1,16 @@
+name := "pureconfig-circe"
+
+libraryDependencies ++= Seq(
+  "io.circe" %% "circe-core" % "0.10.0",
```
jcazevedo: We can use `0.11.1`.

modules/circe/src/main/tut/README.md
```diff
@@ -0,0 +1,92 @@
+# Circe module for PureConfig
+
+Adds support for [Circe](https://circe.github.io/circe/) `Json` to PureConfig.
+
+## Add pureconfig-circe to your project
+
+In addition to [core pureconfig](https://github.com/pureconfig/pureconfig), you'll need:
+
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-circe" % "0.10.2"
```
jcazevedo: We should update this to `0.11.0` to be aligned with the current version of PureConfig.

modules/circe/src/main/scala/pureconfig/module/circe/package.scala
```diff
@@ -0,0 +1,39 @@
...
+      jnum => jnum.toLong match {
+        case Some(long) => ConfigValueFactory.fromAnyRef(long)
+        case None => ConfigValueFactory.fromAnyRef(jnum.toDouble)
+      },
+      str => ConfigValueFactory.fromAnyRef(str),
+      arr => ConfigValueFactory.fromIterable(arr.map(jsonToCv).asJava),
+      obj => ConfigValueFactory.fromMap(obj.toMap.mapValues(jsonToCv).asJava))
+  }
+
+  implicit def circeJsonReader: ConfigReader[Json] =
```
jcazevedo: This and the `circeJsonWriter` can be `implicit val`s.

jcazevedo (changes requested):
Thanks for your contribution, @moradology! Please take a look at my comments below.

modules/circe/src/main/tut/README.md
moradology: Yeah, the example is not super motivated. The reason I found this useful was that I had a somewhat complex recursive structure and that I'd written a `Decoder` for and didn't really want to rewrite that (performance of serialization/deserialization wasn't a major concern in my use case).

If this example will generally mislead, I'm happy to remove it

modules/circe/src/main/tut/README.md
moradology: It may not be - I could have misinterpreted the compiler's unhappiness with an earlier use of `???` for the `Decoder` implementation

modules/circe/src/main/scala/pureconfig/module/circe/package.scala
moradology: Point of clarification, as I'm having some difficulty seeing how you avoid using the `circe` parser. Let's say I have `LIST` from `ConfigValue.valueType`. Calling `ConfigValue.unwrapped` on such a structure produces a `List[Object]`, where the children `Object`s have *already* been unwrapped. Digging around the API, it wasn't obvious to me how to avoid this (any of the methods I could find that would have allowed me to handle recursion myself were hidden away in private classes)

Is there something I'm overlooking?

jcazevedo: I think we need to explicitly cast `ConfigValue` to `ConfigList` when the `valueType` is `LIST`. `ConfigList` then lets you handle each individual element as a `ConfigValue`. It should be reasonably safe to do so in those cases, and it's something that we already resorted to in PureConfig's core.

modules/circe/src/main/scala/pureconfig/module/circe/package.scala
```diff
@@ -0,0 +1,57 @@
...
+      case ConfigValueType.BOOLEAN =>
+        Json.fromBoolean(cv.unwrapped.asInstanceOf[Boolean])
+      case ConfigValueType.NUMBER =>
+        cv.unwrapped match {
+          case i: java.lang.Number if i.longValue() == i =>
+            Json.fromLong(i.longValue())
+          case d: java.lang.Number if !(d.doubleValue().isNaN) =>
+            Json.fromDouble(d.doubleValue()).get
+          case _ => // Non-integral, non-real; following javascript's JSON.stringify in gives null
+            Json.Null
```
moradology: I'm sure there are strong opinions about this one way or the other but I decided to just follow JS behavior here as a first pass

moradology:
I think this is ready to be looked at once more; the only thing I suspect will need some input on is the handling of the `NaN` as `Json.Null` called out above

jcazevedo:
Sorry for the lack of activity in this, @moradology. I'll take another look at the PR during this week.

moradology:
No problem - everyone's busy with their own stuff. I just wanted to bring it to your attention in case you forgot (and anyway, this project has been very welcoming/responsive compared to most)

modules/circe/src/main/scala/pureconfig/module/circe/package.scala
jcazevedo: I'm OK with `Json.Null` here. It is consistent with what the default `Encoder` for `Double` does in circe. However, we also need to take infinity into account. I think we can remove this fallback and use `Json.fromDoubleOrNull` instead.

modules/circe/src/test/scala/pureconfig/circe/CirceSuite.scala
```diff
@@ -0,0 +1,36 @@
+package pureconfig.module.circe
```
jcazevedo: Can you please move this to `modules/circe/src/test/scala/pureconfig/module/circe/CirceSuite.scala`?

modules/circe/src/main/scala/pureconfig/module/circe/package.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import pureconfig.{ ConfigWriter, ConfigReader }
+import com.typesafe.config._
+import io.circe._
+import com.typesafe.config._
+
+import scala.collection.JavaConverters._
+
+package object circe {
+  def cvToJson(cv: ConfigValue): Json = {
```
jcazevedo: This can probably be made private.

modules/circe/src/main/scala/pureconfig/module/circe/package.scala
```diff
@@ -0,0 +1,57 @@
...
+      }
+      case ConfigValueType.OBJECT => {
+        val jsonMap = cv.asInstanceOf[ConfigObject].asScala.mapValues(cvToJson _).toMap
+        val jsonObj = JsonObject.fromMap(jsonMap)
+        Json.fromJsonObject(jsonObj)
+      }
+    }
+  }
+
+  def jsonToCv(json: Json): ConfigValue = {
```
jcazevedo: This can probably be made private.

jcazevedo (changes requested):
@moradology: Sorry for the delay in reviewing this. I just left comments on some minor issues below.

jcazevedo (approved):
@moradology: Thanks for going through the suggested changes! This looks good to me. Let's just wait for another approval from either @ruippeixotog or @leifwickland before merging.

modules/circe/src/main/tut/README.md
```diff
@@ -0,0 +1,47 @@
...
+```tut:silent
+import io.circe._
+import com.typesafe.config.ConfigFactory
+import pureconfig._
+import pureconfig.generic.auto._
+import pureconfig.module.circe._
+import pureconfig.error.CannotConvert
+```
+
+### Reading Json  values directly
```
ruippeixotog: Nit: can you remove the double space?

modules/circe/src/main/tut/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Imports to be used below:
+
+```tut:silent
+import io.circe._
+import com.typesafe.config.ConfigFactory
+import pureconfig._
+import pureconfig.generic.auto._
+import pureconfig.module.circe._
+import pureconfig.error.CannotConvert
```
ruippeixotog: You don't seem to be using this anywhere.

ruippeixotog (changes requested):
LGTM, just a few nitpicks. Thanks for doing this!

ruippeixotog:
One more thing, can you add an entry for pureconfig-circe on [docs/src/main/tut/docs/library-integrations.md](https://github.com/pureconfig/pureconfig/blob/master/docs/src/main/tut/docs/library-integrations.md)?

moradology:
Thanks for having such a strong process for working with open source contributors (and sorry for the delay in getting back to this)

modules/circe/README.md
```diff
@@ -20,10 +20,9 @@ import com.typesafe.config.ConfigFactory
 import pureconfig._
 import pureconfig.generic.auto._
 import pureconfig.module.circe._
-import pureconfig.error.CannotConvert
```
jcazevedo: You also need to fix this in the [tut source](https://github.com/pureconfig/pureconfig/blob/c9bc53861f5581b9b1d917f7d360d7b7ecb04fa5/modules/circe/src/main/tut/README.md).

jcazevedo (changes requested):

modules/circe/README.md
moradology: Rebased with this fix

jcazevedo (approved):

ruippeixotog (approved):

PR #474, opened 2019-05-15
Remove JDK version requirement

ruippeixotog:
This is old code that had the goal of preventing usage of JDK 7. There's no need to have such a requirement and it's a [problem for the community build](https://github.com/scala/community-builds/pull/906#issuecomment-492482343).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -100,12 +100,6 @@ lazy val commonSettings = Seq(
     .setPreference(DoubleIndentConstructorArguments, true)
     .setPreference(SpacesAroundMultiImports, true),
 
-  initialize := {
-    val required = "1.8"
-    val current = sys.props("java.specification.version")
-    assert(current == required, s"Unsupported JDK: java.specification.version $current != $required")
-  },
-
   autoAPIMappings := true,
 
   publishMavenStyle := true,


build.sbt
```diff
@@ -100,12 +100,6 @@ lazy val commonSettings = Seq(
     .setPreference(DoubleIndentConstructorArguments, true)
     .setPreference(SpacesAroundMultiImports, true),
 
-  initialize := {
-    val required = "1.8"
-    val current = sys.props("java.specification.version")
-    assert(current == required, s"Unsupported JDK: java.specification.version $current != $required")
```
leifwickland, line 106: Do you think it'd be worth keeping this but trying to ensure `current >= 1.8`? I recall sbt having version parsing magic that should make it easy. I'm happy to just cut it too.

leifwickland (approved):

build.sbt
ruippeixotog: We could keep it, but I believe the value it brings now is very reduced (I guess that people using JDK <= 1.8 already expect some projects not to work, judging by the number of incompatible dependencies). On the other hand, doing what you say would increase the complexity and size of our build code instead of decreasing it. If you're ok with it, I'd leave the PR as it is now.

leifwickland: I'm absolutely OK with it.

PR #478, opened 2019-05-26
Remove gif from README file

### LINKED ISSUE #463: misleading example on the GIF
Following the exact instructions on the GIF doesn't take us to the end result but a few errors occurred.
https://camo.githubusercontent.com/cd550455fd9584214104b64071a02e443317a4d8/687474703a2f2f692e696d6775722e636f6d2f503673646130362e676966

Issue 1. Since the scalaVersion in the example is 2.12.1, I think the libraryDependencies needs to be fixed
as-is: libraryDependencies += "com.github.pureconfig" %% "pureconfig"
to-be: libraryDependencies += "com.github.pureconfig" %% "pureconfig" % "0.10.2"

Issue 2. According to this issue, https://github.com/pureconfig/pureconfig/issues/441#issuecomment-444106292, we need to always import ```pureconfig.generic.auto._``` before using this method `loadConfigOrTHrow` with ConfigReader implicitly.

With those two patches above I get the result all right.

jcazevedo:
This PR removes the gif from the README file, as discussed in #463.

Closes #463.


ruippeixotog (changes requested):
You also need to remove it from [docs/src/main/tut/index.md](https://github.com/pureconfig/pureconfig/blob/master/docs/src/main/tut/index.md). Otherwise LGTM :)

jcazevedo:
> You also need to remove it from docs/src/main/tut/index.md.

I forgot about the microsite. Removed it in a8fc5d4.

ruippeixotog (approved):

PR #484, opened 2019-06-11
Upgrade scala version to 2.13 in cross compilation

### LINKED ISSUE #483: Support Scala 2.13
Since Scala 2.13 is released recently it would be awesome if this library would cross compile to Scala 2.13 as well. 

Am willing to give it a go :)

hugo-vrijswijk:
Upgrade Scala version to 2.13 so it can be cross compiled.

Closes #483


jcazevedo:
We can update `pureconfig-core` and `pureconfig-generic` to Scala 2.13.0. The scoverage SBT plugin isn't published for 2.13.0 yet, but it should be OK to disable coverage tests for the time being.

Some of the external libraries we integrate with are not yet published for Scala 2.13.0, so we can't cross-compile for all modules yet.

hugo-vrijswijk:
Ah, I see. There is a PR open for [scoverage](https://github.com/scoverage/scalac-scoverage-plugin/pull/260). I can change the cross compilation for 2.13 to be back to only a couple modules?

jcazevedo:
> I can change the cross compilation for 2.13 to be back to only a couple modules?

I think it's preferable, yes. You can also disable the publishing of coverage information while scoverage/scalac-scoverage-plugin#260 isn't merged.

hugo-vrijswijk:
Thanks for the pointer. I've also removed the onScala213 logic from `Dependencies.scala` as `scalacheck-shapeless` is now published for all needed versions and as far as I can tell it's not needed anymore

ruippeixotog (changes requested):
Thanks for this PR, @hugo-vrijswijk! From a quick look at the build results, the only thing that seems to be missing is renaming `scala-2.13.0-RC1` to `scala-2.13` in `core/src/main`.

hugo-vrijswijk:
Looks like I finally got CI to pass my changes. It's too bad I couldn't get the other modules to work with 2.13. Maybe that'll be different in a couple weeks time once all dependencies are upgraded. 

If this is merged, a squash commit is probably best since I made a lot of (now) useless commits

project/Dependencies.scala
```diff
@@ -3,30 +3,32 @@ import sbt.Keys._
 
 object Dependencies {
 
-  private[this] def onScala213(onScala213: String, onOthers: String) = Def.setting {
-    CrossVersion.partialVersion(scalaVersion.value) match {
-      case Some((2, 13)) => onScala213
-      case _ => onOthers
+  private[this] def onScala213(onScala213: String, onOthers: String) =
+    Def.setting {
```
ruippeixotog: Since you can't remove this utility method, can you make sure the style remains the same as before? Same for all the other reverted changes.

ruippeixotog (changes requested):
> It's too bad I couldn't get the other modules to work with 2.13

I imagine it will take quite some time before we're able to compile all our modules against Scala 2.13. I'd actually advocate for doing that in a separate PR and keeping this one focused on core modules, so don't worry about not being able to update them :)

project/Dependencies.scala
hugo-vrijswijk: Done :)

ruippeixotog (approved):
Thanks!

PR #511, opened 2019-06-28
Upgrade to sbt 1.3.0-RC3

leifwickland:


ruippeixotog (commented):
I thought I had commented on this PR, but it seems I didn't 😮

This LGTM, but is there any reason in particular why we are updating to an RC version?

leifwickland:
@ruippeixotog I proposed upgrading to an RC because it appeared to work and thus to provide evidence to the community that the RC was solid. I'm in no rush to upgrade to an RC if you'd rather not.

ruippeixotog (approved):
@leifwickland good point :)

It seems that Travis is experiencing some problems currently.

jcazevedo:
I merged this branch with `master` since the build in Travis was still running with `oraclejdk8`.

PR #523, opened 2019-07-27
Fix documentation example

### LINKED ISSUE #522: Failing UnexpectedValueForFieldCoproductHint
The sample its self fails

sealed trait MyAdt {}

case class AdtA(a: String) extends MyAdt
case class AdtB(b: Int) extends MyAdt
final case class Port(value: Int) extends AnyVal
case class MyClass(
                    boolean: Boolean,
                    port: Port,
                    adt: MyAdt,
                    list: List[Double],
                    map: Map[String, String],
                    option: Option[String])

object Executor {
  def main(args: Array[String]): Unit = {
    println(pureconfig.loadConfig[MyClass])
  }
}

Left(ConfigReaderFailures(ConvertFailure(UnexpectedValueForFieldCoproductHint(Quoted("adtb")),Some(ConfigValueLocation(file:/home/kneuron/IdeaProjects/Aqua/target/scala-2.12/classes/application.conf,4)),adt.type),List()))

ruippeixotog:
Fixes #522.

GitHub's README is one of the only parts that is not compiled to ensure the examples are correct. The website has [a copy of this example](https://pureconfig.github.io/docs), but tut changed the output of this example to a failure without us noticing it :( Before we published the website we had this warning implicitly, because every change that tut made had to be committed - now we don't. I'll add a CI check to detect changes in the website later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ scala:
   - 2.12.8
   - 2.13.0
 jdk:
-  - oraclejdk8
+  - openjdk8
 env:
   # marker environment variable to make the build matrix more readable in the UI
   - JOB=test
--- a/docs/src/main/resources/application.conf
+++ b/docs/src/main/resources/application.conf
@@ -1,7 +1,7 @@
 boolean = true
 port = 8080
 adt {
-  type = "adtb"
+  type = "adt-b"
   b = 1
 }
 list = ["1", "20%"]


ruippeixotog:
Seems like our Travis project was moved to Xenial dist, which doesn't support `oraclejdk8` (https://travis-ci.community/t/expected-feature-release-number-in-range-of-9-to-12-but-got-8-installing-oraclejdk8/1345). Updated .travis.yml to use `openjdk8`.

jcazevedo (approved):

PR #524, opened 2019-07-27
Add API support for specifying namespace when loading YAML files

YuvalItzchakov:
This PR adds support for specifying a namespace when loading YAML files, which brings it on par with `loadConfig` for loading standard HOCON files.

Adding the following to the API surface:

```scala
loadYaml(content: String, namespace: String)
loadYaml(path: Path, namespace: String)
loadYamlOrThrow(content: String, namespace: String)
loadYamlOrThrow(path: Path, namespace: String)
```

I thought about additionally adding support for multi-doc YAML files, but since the schema between YAMLs inside the same doc can vary, I outputted out of that.


YuvalItzchakov:
Not sure why the addition in the README is causing the build to fail:

```
-We can also load a particular namespace inside the YAML file:
-
-```scala
-loadYaml[List[Person]](yamlFile, "children")
-// res2: pureconfig.ConfigReader.Result[List[Person]] = Right(List(Person(Sarah,7,List()), Person(Andy,10,List())))
-```
The command "if [[ $TRAVIS_SCALA_VERSION =~ ^2\.12.* ]]; then git diff --exit-code; fi" exited with 1.
```

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
```diff
@@ -17,30 +12,51 @@ import pureconfig._
 import pureconfig.error._
 import pureconfig.module.yaml.error.{ NonStringKeyFound, UnsupportedYamlType }
 
+import scala.collection.JavaConverters._
```
ruippeixotog: Can you please revert all code style changes on this file? I'd prefer if we keep the changes on each pull request focused on the essential.

modules/yaml/src/test/scala/pureconfig/module/yaml/YamlApiSuite.scala
```diff
@@ -4,12 +4,13 @@ import java.net.URLDecoder
 import java.nio.file.{ Files, Path, Paths }
 
 import com.typesafe.config.ConfigValue
+import org.scalatest.EitherValues
```
ruippeixotog, line 7: Ditto on this one.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -32,15 +32,15 @@ package object pureconfig {
...
+    try getValue(ConfigCursor(confValue, Nil), splitPath(namespace)) catch {
       case ex: ConfigException => ConfigReader.Result.fail(ThrowableFailure(ex, ConfigValueLocation(ex.origin())))
     }
   }
 
   // loads a value from a config in a given namespace. All `loadConfig` methods _must_ use this method to get correct
   // namespace handling, both in the values to load and in the error messages.
-  private[this] def loadValue[A](conf: TypesafeConfig, namespace: String)(implicit reader: Derivation[ConfigReader[A]]): ConfigReader.Result[A] = {
-    getValue(conf, namespace, reader.value.isInstanceOf[ReadsMissingKeys]).right.flatMap(reader.value.from)
+  private[pureconfig] def loadValue[A](configValue: ConfigValue, namespace: String)(implicit reader: Derivation[ConfigReader[A]]): ConfigReader.Result[A] = {
```
ruippeixotog, line 42: There may be some small complications with you accepting a non-object here. What would happen if you tried `loadYaml[Option[Int]](pathToYaml, "foo")`, where `pathToYaml` points to a YAML file whose root value is not an object (e.g. an array)?

(regardless of the result, can you add a test for this corner case?)

ruippeixotog (changes requested):
Thank you very much for submitting this, @YuvalItzchakov! And kudos for including tests, too :)

> Not sure why the addition in the README is causing the build to fail

We use [tut](https://github.com/tpolecat/tut) to generate documentation. Can you please add that example to `modules/yaml/src/main/tut/README.md` instead and then run `sbt yaml/tut`?

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
YuvalItzchakov: Of course. I thought this was the result of running code format on compile.

modules/yaml/src/test/scala/pureconfig/module/yaml/YamlApiSuite.scala
YuvalItzchakov: `EitherValues` is something I added to work with `right.value` instead of `right.get` to get more meaningful error messages. Would you want me to refrain from using it?

core/src/main/scala/pureconfig/package.scala
YuvalItzchakov: @ruippeixotog You're right. This would fail with `WrongType(LIST, Set(OBJECT))`. Since `getValue` does `asObjectCursor` inside the method itself, thus should only accept a `Config`. Perhaps inverting the responsibility such that `loadYaml` overloads should validate the passed YAML is a `ConfigObject`, otherwise fail with a more indicative error, not sure which exactly. WDYT?

One suggestion would be:

```scala
yamlObjToConfigValue(yamlObj).right.flatMap {
  case configObj: ConfigObject => getValue(configObj.toConfig, namespace, reader.isInstanceOf[ReadsMissingKeys]).right.flatMap(reader.value.from(_))
  case configValue => Left(ConfigReaderFailures(ConvertFailure(WrongType(configValue.valueType(), Set(ConfigValueType.OBJECT)), None, "")))
}
```

Although I'm not particularly sure this falls under a `ConvertFailure` type of error, more like `WrongInput`

ruippeixotog: Oh ok, `WrongType(LIST, Set(OBJECT))` is exactly the error I'd expect here, I misread the code  :) In my example we're expecting an object as root (in order to resolve access "foo") but we found a list, so it's OK.

Can you just add these corner cases as tests to make the expected behavior explicit?

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
```diff
@@ -86,14 +86,28 @@ package object yaml {
...
+  def loadYaml[Config](path: Path, namespace: String)(implicit reader: Derivation[ConfigReader[Config]]): ConfigReader.Result[Config] = {
     handleYamlErrors(Some(path)) {
       using(Files.newBufferedReader(path)) { ioReader =>
         // we are using `SafeConstructor` in order to avoid creating custom Java instances, leaking the PureConfig
         // abstraction over SnakeYAML
         val yamlObj = new Yaml(new SafeConstructor()).load[AnyRef](ioReader)
 
         yamlObjToConfigValue(yamlObj).right.flatMap { cv =>
-          reader.value.from(ConfigCursor(cv, Nil))
+          getValue(cv, namespace, reader.isInstanceOf[ReadsMissingKeys])
```
ruippeixotog: This has the same implementation as `pureconfig.loadValue`.

modules/yaml/src/main/scala/pureconfig/module/yaml/package.scala
```diff
@@ -107,14 +121,19 @@ package object yaml {
...
+  def loadYaml[Config](content: String, namespace: String)(implicit reader: Derivation[ConfigReader[Config]]): ConfigReader.Result[Config] = {
     handleYamlErrors(None) {
       // we are using `SafeConstructor` in order to avoid creating custom Java instances, leaking the PureConfig
       // abstraction over SnakeYAML
       val yamlObj = new Yaml(new SafeConstructor()).load[AnyRef](content)
 
       yamlObjToConfigValue(yamlObj).right.flatMap { cv =>
-        reader.value.from(ConfigCursor(cv, Nil))
-      }
+          getValue(cv, namespace, reader.isInstanceOf[ReadsMissingKeys]).right
```
ruippeixotog: This has the same implementation as `pureconfig.loadValue`.

modules/yaml/src/test/scala/pureconfig/module/yaml/YamlApiSuite.scala
ruippeixotog: If you're using it, sounds good!

ruippeixotog (changes requested):

ruippeixotog:
About the latest Travis failure: if you compile the code using `sbt compile` it should reformat the code according to scalariform rules; you may have forgotten to commit the formatting changes after you reverted your custom ones :) Sorry for the overly cautious build.

YuvalItzchakov:
@ruippeixotog LMK if there any additional changes I need to make :)

ruippeixotog (approved):
LGTM, thanks for addressing all the comments, @YuvalItzchakov!

As a heads up, I'm working towards a generalized solution to manage and combine config sources as part of #531. I think it will be provide some neat new features for YAML loading, but it will probably mean a change in API. I'll release a version with your changes in the next hours/days so you'll use them as a temporary measure :)

YuvalItzchakov:
Thanks! LMK if you need any help reworking the API :)

PR #525, opened 2019-07-27
Create script to check the website build against the version in master

ruippeixotog:
Creates a new Travis job, `diff_website`, that checks whether the output of `sbt makeMicrosite` changed from the base branch (master) to this one. The goal of this is to prevent cases like #522, where we ended up generating wrong documentation (like the current state of https://pureconfig.github.io/docs) without any warning.

I wanted this to be a warning and not a build failure, as there are quite a number legitimate changes (like updating sbt-microsites or tut, as well as breaking/presentation changes) that would fail a build without any way of making it pass. Since neither GitHub nor Travis have a concept of "passed, but with warnings" in checks/builds, I ended up creating a bot user and making it comment on the PR when the check fails.

Note: the comment below is an example of a message sent by Travis. It refers to a previous version of the PR where I did a change that broke the documentation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ script:
 
 jobs:
   include:
-    - env: JOB=microsite
+    - env: JOB=build_website
       scala: 2.12.8
       addons:
         apt:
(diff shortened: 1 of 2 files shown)


leifwickland (approved):
That's neat. Nice work!

PR #537, opened 2019-08-08
Compile against Scala 2.12.9 and build some modules for Scala 2.13

jcazevedo:
This PR enables compilation against Scala 2.12.9 and builds some modules for Scala 2.13. 

There's a bug in tut (tpolecat/tut#246) that makes comments fail the compilation step in Scala 2.13. As such, I've enabled tut only for 2.12 in the CI. This is also preventing us from building the microsite in Scala 2.13. 😢

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: scala
 
 scala:
   - 2.11.12
-  - 2.12.8
+  - 2.12.9
   - 2.13.0
 jdk:
   - openjdk8
@@ -33,7 +33,7 @@ after_success:
 jobs:
   include:
     - env: JOB=build_website
-      scala: 2.12.8
+      scala: 2.12.9
       addons:
         apt:
           packages:
@@ -55,7 +55,7 @@ jobs:
       after_success: ignore
     - env: JOB=diff_website
       if: type = pull_request AND branch = master
-      scala: 2.12.8
+      scala: 2.12.9
       install:
         - rvm use 2.4 --install --fuzzy
         - gem update --system
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,15 +15,18 @@ script:
   - >
     sbt coverage
     "++$TRAVIS_SCALA_VERSION test"
-    "++$TRAVIS_SCALA_VERSION tut"
     "++$TRAVIS_SCALA_VERSION doc"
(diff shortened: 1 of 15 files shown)


leifwickland (approved):
@jcazevedo Thanks for moving pureconfig a step forward.

example/build.sbt
```diff
@@ -2,12 +2,12 @@ name := "example"
 
 version := "1.0"
 
-scalaVersion := "2.12.8"
```
ruippeixotog, line 5: Currently it's fair to say that 2.12 is still our main version. Did you make 2.13 the main one here for any reason in particular?

.travis.yml
```diff
@@ -15,15 +15,18 @@ script:
...
     "++$TRAVIS_SCALA_VERSION publishLocal"
 
   # Compile example project
   - (cd example; sbt ++$TRAVIS_SCALA_VERSION test)
 
-  # check if there are no changes after `tut` runs
+  # Run tut and check if there are no changes in docs.
+  #
+  # We're running tut only in Scala 2.12 because it fails the compile step in
+  # the presence of comments in 2.13: https://github.com/tpolecat/tut/issues/246
```
ruippeixotog, line 27: We should consider switching to mdoc, as their engine seems much less amenable to weird edge cases. I tried to do it a few months ago but I found some blockers; maybe it's worth visiting that again.

ruippeixotog (approved):
LGTM, just take a look at the comment about the example project

.travis.yml
jcazevedo: I agree, although it won't help us here right now because mdoc is not yet available for Scala 2.13.

example/build.sbt
jcazevedo: At the time I wishfully thought I could update the documentation to Scala 2.13, which would enable us to move other CI steps to Scala 2.13 and make 2.13 more "defaultish". I agree 2.12 is a better main version for now. I'll change it.

jcazevedo: Made 2.12 the main version of the example project in f38abfb.

.travis.yml
leifwickland: Another good reason to move to mdoc eventually is that is that Rob is deprecating tut:

https://github.com/tpolecat/tut/issues/246#issuecomment-519943218

PR #541, opened 2019-08-11
Refactor config loading API

### LINKED ISSUE #531: RFC: Rethink config loading API
At the moment, the API for loading configs is lacking:
1. It has a lot of wrappers and overloads for different scenarios. This is exacerbated  by the fact that these are duplicated in the `catseffect` module.
1. It doesn't cover all the cases that `ConfigFactory` supports (loading from `URL`s, specifying classloaders, etc).
1. It is not very flexible. E.g. there is no way to load a config without the default `application.conf` and `reference.conf` fallbacks.
1. It is not very transparent. E.g. to find out that the `load*` function loads the `reference.conf` resource, I have to dig down until I find that it's based on a `ConfigFactory.load*` function, and then read the javadoc for that. This can be mitigated by properly documenting each function, but that is inconvenient due to point 1.

Most of this stems from the fact that we use the high-level `ConfigFactory.load*` functions.

I propose we add an API for building custom `Config` loading chains, backed by more composable `ConfigFactory.parse*` functions. Something like this:
```scala
// Some ADT - implementation pending
sealed ConfigSource

// `ConfigSource` should provide a combinator-style micro-DSL.
val source: ConfigSource = ConfigSource.file(myConfigFile)
  // reference.conf etc.. The classloader is optional.
  .withFallback(ConfigSource.defaultFallbacks(classloader = Some(myClassLoader)))
  .withNamespace("myApp")

// No more overloads - we only need a single wrapper.
val config = loadConfig[IO, MyConfig](source)

// Mixing custom files with custom resources - not possible with the current API.
val testSource = ConfigSource.file(myTestConfigFile)
  // Optional sources are explicit.
  .optional
  .withFallback(ConfigSource.resource("test-reference.conf", classloader = Some(myClassLoader)))
  .withFallback(ConfigSource.defaultFallbacks(classloader = Some(myClassLoader)))
  .withNamespace("myApp")
```

ruippeixotog:
This pull request creates a `ConfigSource` trait that should become the PureConfig entrypoint for non-default setups. The design draws heavily from @nigredo-tori's RFC at #531.

Things to note:
- I deprecated all methods in `pureconfig` except `loadConfig[A]` and `loadConfigOrThrow[A]` (no arg versions). Part of the motivation for this change was to mitigate the combinatorial explosion of methods there, and because of that I think we should push users to use `ConfigSource` and its combinators for anything different than `ConfigFactory.load()`;
- Because of the changes here, I ended up dropping support for doing `loadConfig[Option[A]]` and returning `Right(None)` when the path didn't exist. Because namespace navigation (`.at(namespace)`) and `ConfigReader` usage (`.load[A]`) are now separated, supporting `ReadsMissingValues` there would introduce a lot of complexity for a use case I believe to be rarely needed. This doesn't affect the behavior of `Option` values inside configs, only the cases where we load an `Option` as a root;
- I haven't migrated pureconfig-yaml to take advantage of this API yet. I plan to do it in another PR soon, before we release the next version;
- I'll also update the documentation in the website before the next release.

This PR is better reviewed commit-by-commit due to the huge number of touched files.

Closes #531 (even though custom classloader support, one of the requested features, is not here yet, the main goal of making the API easily extensible should be considered done)


core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,87 @@
...
+
+class ConfigObjectSource(readConfig: => Result[Config]) extends ConfigSource {
+  private lazy val config: Result[Config] = readConfig
+  lazy val value: Result[ConfigObject] = config.right.map(_.resolve.root)
+
+  def withFallback(cs: ConfigObjectSource): ConfigObjectSource =
+    new ConfigObjectSource(Result.zipWith(config, cs.config)(_.withFallback(_)))
+
+  def withOptionalFallback(cs: ConfigObjectSource): ConfigObjectSource =
+    withFallback(cs.recoverWith({ case _ => Right(ConfigFactory.empty) }))
```
nigredo-tori: This is not very intuitive. I think that if the config file is invalid (rather than missing), most of the users would want to report the errors. So we should handle those two conditions separately.

Also, I think we should factor out a more fundamental primitive:
```scala
def optional: ConfigObjectSource =
  cs.recoverWith({ case _: CannotRead => Right(ConfigFactory.empty) })

def withOptionalFallback(cs: ConfigObjectSource): ConfigObjectSource =
  withFallback(cs.optional)
```

core/src/main/scala/pureconfig/package.scala
```diff
@@ -78,7 +46,7 @@ package object pureconfig {
    *         isn't possible
    */
   def loadConfig[Config](path: Path)(implicit reader: Derivation[ConfigReader[Config]]): ConfigReader.Result[Config] =
-    loadConfig(path, "")
+    ConfigSource.file(path).load[Config]
```
nigredo-tori: This changes the behavior, since it doesn't load `reference.conf`. The same goes for a lot of the functions in this file.

core/src/main/scala/pureconfig/package.scala
```diff
@@ -290,16 +239,7 @@ package object pureconfig {
...
-      .map {
-        case Left(failures) if failures.toList.exists(_.isInstanceOf[CannotReadFile]) && !failOnReadError =>
-          Right(ConfigFactory.empty())
-        case conf => conf
-      }
-      .foldLeft[ConfigReader.Result[TypesafeConfig]](Right(ConfigFactory.empty())) {
-        case (c1, c2) => ConfigReader.Result.zipWith(c1, c2)(_.withFallback(_))
-      }
-      .right.flatMap { conf => loadConfig[Config](conf.resolve, namespace) }
+    files.map(ConfigSource.file).foldLeft(ConfigSource.empty)(filesReduceFunc(failOnReadError)).at(namespace).load[Config]
```
nigredo-tori: <s> This changes the behavior:
1. It fails if the first file doesn't exist.
1. It doesn't load `reference.conf`.

If we had an `optional` combinator, we could write:
```scala
files.map(ConfigSource.file(_).optional)
  .foldLeft(ConfigSource.empty)(_.withFallback(_))
  .withFallback(ConfigSource.defaultReference)
  .at(namespace)
  .load[Config]
```
</s>

Scratch all that, your version works as expected (except for `reference.conf`), though it might be better to express the same logic like this (assuming we have the `.optional` combinator):
```scala
files.map(ConfigSource.file(_))
  .map(cs => if (failOnReadError) cs else cs.optional)
  .foldLeft(ConfigSource.empty)(_.withFallback(_))
  .withFallback(ConfigSource.defaultReference)
  .at(namespace)
  .load[Config]
```

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -56,7 +60,7 @@ package object catseffect {
    *         details on why it isn't possible
    */
   def loadConfigF[F[_], A](path: Path)(implicit F: Sync[F], reader: Derivation[ConfigReader[A]], ct: ClassTag[A]): F[A] =
-    loadConfigF[F, A](path, defaultNameSpace)
+    ConfigSource.file(path).loadF[F, A]
```
nigredo-tori: Same as in `core` - this doesn't load `reference.conf`.

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,284 @@
...
+ * All config loading methods are lazy and defer resolution of references until needed.
+ */
+trait ConfigSource {
+
+  /**
+   * Retrieves a `ConfigValue` from this source.
+   *
+   * @return a `ConfigValue` retrieved from this source.
+   */
+  def value: Result[ConfigValue]
```
nigredo-tori: I think that both the name and the signature here should reflect that it performs side effects:
```scala
def getValue(): Result[Config]
```

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,284 @@
...
+  }
+}
+
+/**
+ * A `ConfigSource` which is guaranteed to generate config objects (maps) as root values.
+ *
+ * @param config the thunk to generate a `Config` instance. This parameter won't be memoized so it
+ *               can be used with dynamic sources (e.g. URLs)
+ */
+class ConfigObjectSource(config: => Result[Config]) extends ConfigSource {
```
nigredo-tori: We might want to make this `final` (and the constructor `private`) so we don't leak the implementation, and have the ability to add new methods later in a binary compatible way.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Fair enough. I'll remove `withOptionalFallback` altogether, as it doesn't seem to provide a lot of benefits if `optional` exists.

core/src/main/scala/pureconfig/package.scala
ruippeixotog: Well spotted. I'll create a test case to make sure that we don't miss this again and then I'll fix it.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: I added empty parenthesis to the signature and opted to keep the function name the same (mostly a personal preference; I'm not very fond of `get` prefixes in methods as it doesn't convey a lot of additional information).

ruippeixotog:
@nigredo-tori thanks for your thorough review! I believe I addressed all of your comments. Let me know if you find anything else.

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,299 @@
...
+   *
+   * This method is provided here to support use cases that previously depended on
+   * `ConfigFactory.load(config)`. Creating a custom source by merging the layers manually is
+   * usually recommended as it makes the config priorities more transparent.
+   *
+   * @param appSource the source providing the application config
+   * @return a `ConfigObjectSource` for the default loading process in Typesafe Config with a
+   *         custom application config source.
+   */
+  def applicationConf(appSource: ConfigObjectSource): ConfigObjectSource =
```
nigredo-tori: We need a better name - this actually includes *everything but* `application.conf` :smile:. How about `withDefaultReferenceAndOverrides`?

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,299 @@
...
+    systemProperties.withFallback(appSource).withFallback(defaultReference)
```
nigredo-tori: This should use `defaultOverrides` instead of `systemProperties` in case `ConfigFactory` includes some other overrides in the future.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: The goal of this method is to specify your application config (not `application.conf`) in the argument, but I agree that I don't exactly excel at naming things 🙃 `withDefaultReferenceAndOverrides` is certainly more descriptive than my name, but when I name methods as `withX` I expect to provide `X` as the argument (like in `withFallback`), while here it goes the other way around - we wrap something else in `X`.

@jcazevedo, @leifwickland, do you have any suggestion?

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Added `defaultOverrides` on b3937d2. I also clarified the behavior of the default sources in 3b0a9af (in particular, the fact that they follow Typesafe Config's behavior of ignoring missing files).

core/src/main/scala/pureconfig/ConfigSource.scala
leifwickland: You didn't ask if I had a good suggestion, so I'll offer my bad ones:
- `sandwichWithDefaults`
- `wrappedInDefaults`
- `addPreAndPostDefaults`
- `addDefaultReferenceAndOverrides`

If it weren't for the fact that we need this method to tell a simpler upgrade story, I'd say we should omit it since nobody will guess what it does without reading the source.

ruippeixotog: Well, it's not off the table to delete this method and recommend users to migrate to `defaultOverrides.withFallback(ConfigSource.file(path)).withFallback(defaultReference)`. We're already deprecating `loadConfigFromFiles` without an alternative and the fact that we're having trouble naming it and "nobody will guess what it does without reading the source" makes me think this should never be provided in the first place.

leifwickland: I'm not sure that providing this method as the default suggestion in the deprecation notes is what most people expect. My guess is that most users would be surprised to learn that when they tell pureconfig to load a file, it's also looking at multiple other sources.

I'd lean toward not having this method, but since I've never wanted the pre-and-post-defaults behavior I'm definitely not the target market, and thus won't express a strong opinion.

leifwickland (commented):
In general I like this approach and it feels like this makes the API a lot more scalable to enable other sources of config than typesafe config.

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,309 @@
...
+   * Creates a `ConfigObjectSource` from a `FluentConfigCursor`.
+   *
+   * @param cur the cursor to be provided by this source
+   * @return a `ConfigObjectSource` providing the given cursor.
+   */
+  def fromCursor(cur: FluentConfigCursor): ConfigObjectSource =
+    new ConfigObjectSource(cur.asObjectCursor.right.map(_.value.toConfig))
+}
+
+object ConfigSource {
```
leifwickland, line 164: It seems like this object's methods are almost entirely tied to typesafe config. Do you think it's worth trying to rename or partition or comment it in such a way that it leaves semantic room for other sources of config?

core/src/main/scala/pureconfig/ConfigSource.scala
nigredo-tori: > I name methods as `withX` I expect to provide `X` as the argument (like in `withFallback`), while here it goes the other way around - we wrap something else in `X`.

Certainly, `withDefaultReferenceAndOverrides` would work better as a part of the fluent API:
```scala
ConfigSource.defaultApplication
  .withDefaultReferenceAndOverrides
```
We can also split that into two orthogonal methods:
```scala
ConfigSource.defaultApplication
  .withDefaultReference
  .withDefaultOverrides
```

> My guess is that most users would be surprised to learn that when they tell pureconfig to load a file, it's also looking at multiple other sources.

That speaks more about the lack of transparency in the old API (and the `ConfigFactory` API it's based on). I don't think it's a bad thing to get it all out in the open.

> I'd lean toward not having this method

I'm *for* having this method, or some variation of it. This is the default config loading pattern (as [mentioned](https://github.com/lightbend/config/blob/558c1e22f53c7f165daaf3e1e792399087d79f93/HOCON.md#conventional-configuration-files-for-jvm-apps) in the HOCON spec), so it makes sense to provide a cookie-cutter implementation for anyone switching over from the other HOCON libraries.

ruippeixotog: I don't like the idea of polluting the `ConfigSource` trait with Typesafe Config specific defaults, let's keep Typesafe's opinionated config stack at the edge of the API :)

If we are to add this: since the "full" default process is called `.default`, what do you think of naming this `defaultWithCustomAppConf`?

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: I can add some Scaladoc to explain this 👍 It makes sense that we have a lot of Typesafe Config specific methods here since it's part of the HOCON spec and useful for migration, but I intend to add a `YamlConfigSource` later in pureconfig-yaml. One awesome thing about this refactoring is that people using YAML will be able to mix sources, including `reference.conf`, into their YAML application files.

core/src/main/scala/pureconfig/ConfigSource.scala
nigredo-tori: > `defaultWithCustomAppConf`

This has the same problem that `applicationConf` has: it focuses more on what this function doesn't provide.

After this discussion, I'm starting to lean towards one of @leifwickland's suggestions for the function name. `sandwichWithDefaults` (or `defaultSandwich` to match other functions) is memorable and descriptive, albeit unorthodox.

jcazevedo: What if we remove this method and modify the signature of `ConfigSource.default` to become `def default(appSource: ConfigObjectSource = defaultApplication)`? It'd map to the `ConfigFactory.load` API and save us from an extra method that's only there to provide compatibility with the HOCON spec.

nigredo-tori: @jcazevedo, that works, too. However, in my (limited) experience, default parameter values tend to be a pain where binary compatibility is concerned. We can achieve the same with overloading:
```scala
def default(appSource: ConfigObjectSource) = ???
def default(): ConfigSource = default(defaultApplication)
```

ruippeixotog: > This has the same problem that `applicationConf` has: it focuses more on what this function doesn't provide.

While I agree that `applicationConf` is very unclear on what it does, I don't see any problem with `defaultWithCustomAppConf`. I don't want to specify in any way what this function actually provides, as its purpose (and the only reason this function exists) is purely to delegate the loading strategy to Typesafe Config. The source of truth of what it actually does should be Typesafe Config's documentation, as there's nothing here preventing them from changing how loading works in a later version (which may not involve sandwiching).

I'm equally fine with having a `default(appSource: ConfigObjectSource)` (even though I don't see how it's any different from `defaultWithCustomAppConf`), so I'll go with that one since everyone agrees with it 🙂

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Done in 9e82848.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Done in 98c9032.

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,320 @@
...
+    ConfigSource.fromCursor(fluentCursor().at(PathUtil.splitPath(namespace).map(p => p: PathSegment): _*))
+
+  /**
+   * Loads a configuration of type `A` from this source.
+   *
+   * @tparam A the type of the config to be loaded
+   * @return A `Right` with the configuration if it is possible to create an instance of type
+   *         `A` from this source, a `Failure` with details on why it isn't possible otherwise
+   */
+  def load[A](implicit reader: Derivation[ConfigReader[A]]): Result[A] =
```
leifwickland, line 66: It occurs to me that about the only thing I don't like about this new approach is the aesthetics feel a little funny. (Feel free to file the rest of this comment in the appropriate rubbish bin for feelings from folks who don't really contribute.)

I think I'd rather see `load` and `loadOrThrow` move to the `package object pureconfig` and take a `ConfigSource` as a argument.

`load[Config](ConfigSource.applicationConf(ConfigSource.file(path)))`
_feels_ nicer than
`ConfigSource.applicationConf(ConfigSource.file(path)).load[Config]`

I'd also deprecate `pureconfig.loadConfig` in favor of `load(ConfigSource.default)`. 

Feel free to ignore this.

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,320 @@
...
+ * A representation of a source from which `ConfigValue`s can be loaded, such as a file or a URL.
+ *
+ * A source allows users to load configs from this source as any type for which a `ConfigReader` is
+ * available. Raw configs can also be retrieved as a `ConfigValue`, a `ConfigCursor` or a
+ * `FluentConfigCursor`. Before using any of the loading methods described, Users can opt to focus
+ * on a specific part of a config by specifying a namespace.
+ *
+ * All config loading methods are lazy and defer resolution of references until needed.
+ */
+trait ConfigSource {
```
leifwickland, line 25: (Here's another aesthetics comment. Feel free to ignore.)

Reading code which uses `ConfigSource` makes me wish the name of this `trait` didn't include `Config`; I wish it were just `Source`.  `I don't think the `Config` bit helps clarify or disambiguate in a serious way. I think it's unlikely that `scala.io.Source` is going to be used in the same context frequently.

```
    ConfigSource.applicationConf(
      files.map(ConfigSource.file(_).optional)
        .foldLeft(ConfigSource.empty)(_.withFallback(_)))
      .loadOrThrow[Config]
```

feels better to me as

```
    Source.applicationConf(
      files.map(Source.file(_).optional)
        .foldLeft(Source.empty)(_.withFallback(_)))
      .loadOrThrow[Config]
```

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,320 @@
...
+ * config formats or data sources).
+ */
+object ConfigSource {
+
+  /**
+   * A config source for the default loading process in Typesafe Config. Typesafe Config stacks
+   * `reference.conf` resources provided by libraries, application configs (by default
+   * `application.conf` in resources) and system property overrides, resolves them and merges them
+   * into a single config. This source is equivalent to
+   * `defaultOverrides.withFallback(defaultApplication).withDefault(defaultReference)`.
```
jcazevedo: This should be `withFallback`:

```suggestion
   * `defaultOverrides.withFallback(defaultApplication).withFallback(defaultReference)`.
```

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,320 @@
...
+  val defaultReference = ConfigObjectSource(ConfigFactoryWrapper.defaultReference())
+
+  /**
+   * A config source for the default reference config in Typesafe Config (`reference.conf`
+   * resources provided by libraries) before being resolved. This can be used as an alternative
+   * to `defaultReference` for use cases that require `reference.conf` to depend on
+   * `application.conf`. Like Typesafe Config, it provides an empty object if `reference.conf`
+   * files are not found.
+   */
+  val defaultReferenceUnresolved = resources("reference.conf").optional
```
jcazevedo, line 215: I don't think this works as intended. Calling `value()` on this will force a resolve, which would fail with an exception in case some substitutions can't be resolved. From the name of the variable, I would expect to get back an unresolved `Config`.

Since the `withFallback` method of `ConfigObjectResource` forces a resolve, this also won't work well for the use case described in the Scaladoc. If we have an unresolved substitution in `reference.conf` that is only capable of being resolved after a merge with `application.conf`, a call to `ConfigSource.defaultApplication.withFallback(ConfigSource.defaultReferenceUnresolved)` will force `ConfigSource.defaultReferenceUnresolved` to be resolved, which, at that point, will fail.

If we want to provide the flexibility for users to control when substitutions are applied, I think we should provide an explicit `resolve` method in `ConfigSource` that produces a configuration tree with substitutions applied.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -79,16 +80,53 @@ final case class ThrowableFailure(throwable: Throwable, location: Option[ConfigV
  * @param path the file system path of the file that couldn't be read
  * @param reason an optional exception thrown when trying to read the file
  */
-final case class CannotReadFile(path: Path, reason: Option[Throwable]) extends ConfigReaderFailure {
+trait CannotRead extends ConfigReaderFailure {
```
jcazevedo, line 81: Can you update the Scaladoc of this trait to better reflect its purpose?

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,320 @@
...
+/**
+ * A `ConfigSource` which is guaranteed to generate config objects (maps) as root values.
+ *
+ * @param config the thunk to generate a `Config` instance. This parameter won't be memoized so it
+ *               can be used with dynamic sources (e.g. URLs)
+ */
+final class ConfigObjectSource private (config: => Result[Config]) extends ConfigSource {
+
+  def value(): Result[ConfigObject] =
+    config.right.map(_.resolve.root)
```
jcazevedo: We should create a wrapper for `resolve` to avoid having `ConfigException`s escape the context of `Result`.

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,320 @@
...
+   * @return a config source that provides configs read from a file.
+   */
+  def file(file: File) = ConfigObjectSource(ConfigFactoryWrapper.parseFile(file))
+
+  /**
+   * Returns a config source that provides configs read from a URL. The URL can either point to a
+   * local file or to a remote HTTP location.
+   *
+   * @param url the URL
+   * @return a config source that provides configs read from a file.
```
jcazevedo:
```suggestion
   * @return a config source that provides configs read from a URL.
```

core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
```diff
@@ -23,21 +25,57 @@ object ConfigFactoryWrapper {
...
+      onIOFailure = Some(CannotReadFile(file.toPath, _)))
+
   /** @see `com.typesafe.config.ConfigFactory.parseFile()` */
   def parseFile(path: Path): ConfigReader.Result[Config] =
-    unsafeToReaderResult(ConfigFactory.parseFile(path.toFile, strictSettings), Some(path))
+    unsafeToReaderResult(
+      ConfigFactory.parseFile(path.toFile, strictSettings),
+      onIOFailure = Some(CannotReadFile(path, _)))
+
+  /** @see `com.typesafe.config.ConfigFactory.parseFile()` */
```
jcazevedo:
```suggestion
  /** @see `com.typesafe.config.ConfigFactory.parseResources()` */
```

core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
```diff
@@ -23,21 +25,57 @@ object ConfigFactoryWrapper {
...
+  def parseResources(resource: String): ConfigReader.Result[Config] =
+    unsafeToReaderResult(
+      ConfigFactory.parseResources(resource, strictSettings),
+      onIOFailure = Some(CannotReadResource(resource, _)))
+
+  /** @see `com.typesafe.config.ConfigFactory.parseFile()` */
```
jcazevedo:
```suggestion
  /** @see `com.typesafe.config.ConfigFactory.parseURL()` */
```

jcazevedo (changes requested):
@ruippeixotog: Thanks for your work on this!

I left some comments below. In an effort to make this API more transparent, I think we should also allow users of `ConfigSource` to decide when substitutions should be resolved.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Makes sense, and I'll add a test case for this, too.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Good catch 👍 I'll add test cases for `defaultReferenceUnresolved` and fix the issue.

We can add a method to explicitly resolve configs, sounds like it would be useful in some cases. However, I'm still fond of the approach "keep all configs unresolved by default, force resolve them when you call and load* method". Most users aren't even aware that such a complex system of variable resolution exists, and people just calling `ConfigSource.file(myFile).load[Conf]` should have this simply working.

I agree we should move the auto-resolution part to `load` and `loadOrThrow`, to make it happen on the outer layer of the API. However, this still means that `ConfigSource.defaultReferenceUnresolved.load[A]` *will* resolve the config before it tries to apply the `ConfigReader`, which seems reasonable - at this point we are creating the config tree for the whole purpose of reading it, so having undefined references at this point doesn't seem desirable in any case.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Your comments are always welcome and they are contributions as good as any other :)

As a fan of fluent interfaces, I actually prefer the readability of chaining `load` after building a source (skipping the extra set of parentheses and keeping the sequence of operations readable left-to-right, top-to-bottom). I'm not too strong about it though - and there's also the possibility of defining both approaches, but I think that it would be one case where having multiple ways of doing the same thing might be confusing. What do you think? What's your preference, @jcazevedo?

I have a somewhat stronger opinion against having static methods or classes with very general names. `pureconfig` is typically imported as `import pureconfig._`; having a method called simply `load` can easily cause name clashes with other modules, besides the fact that `load` is not a descriptive name at all for client code (as opposed to `loadConfig`, that clearly indicates what we are loading).

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: I don't like this for the same reason as naming `load` above. While we can make a case for this not being confused with `scala.io.Source`, if every library decided to add their own `Source` class things would get very messy. I also don't think it differs much from all the other `Config*` classes (cursor, reader, writer).

core/src/main/scala/pureconfig/ConfigSource.scala
jcazevedo: > As a fan of fluent interfaces, I actually prefer the readability of chaining `load` after building a source (skipping the extra set of parentheses and keeping the sequence of operations readable left-to-right, top-to-bottom). I'm not too strong about it though - and there's also the possibility of defining both approaches, but I think that it would be one case where having multiple ways of doing the same thing might be confusing. What do you think? What's your preference, @jcazevedo?

I prefer the fluent interface. 

With the change proposed by this pull request, I also wouldn't mind deprecating the no arguments versions of `pureconfig.load` and `pureconfig.loadConfigOrThrow` and keep the "entrypoint" methods in the `ConfigSource` interface only. Since part of the motivation of this pull request was to avoid the explosion of methods to load configurations, I think it'd simplify things if we didn't have those methods in the `pureconfig` package.

core/src/main/scala/pureconfig/ConfigSource.scala
jcazevedo: I also prefer `ConfigSource` for consistency with the other types in PureConfig, and because there are other `Source`s in Scala libraries in the wild (there's at least `scala.io.Source`, `Source` in Akka Streams and `Source` in Scalameta)

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Added test in 966053f, fixed in 182d456.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Fixed the existing test case in 966053f (I had the two config lines swapped) and fixed it in 74d4ef2, but not in the way we discussed. After delving a bit more into Typesafe Config's API, I noticed that:
- Only `Config` has a `resolve()` method, not `ConfigValue`;
- The [Javadoc](https://static.javadoc.io/com.typesafe/config/1.3.4/com/typesafe/config/Config.html#resolve--) of `resolve()` says that "a given Config must be resolved before using it to retrieve config values". Technically this is not entirely true as we can pass non-default `ConfigResolveOptions` that allow unresolved configs, but it implies that `ConfigValue`s are not supposed to be unresolved at any point;
- People can easily convert between `Config` and `ConfigObject`, but forcing resolution to be on `Config` kind of discourages people from ignoring that "`resolve()` should be invoked on root config objects, rather than on a subtree", as the Javadoc says.

I ended up exposing a `config()` method returning an unresolved config and left the current behavior of `value()` of forcing resolution. I added appropriate Scaladoc to warn users of the semantics.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: I left them there because my guess is that they are the main and only entrypoints for the great majority of our users and `loadConfig[A]` feels much simpler and to the point than `ConfigSource.default.load[A]`. Like the design around config resolution, I feel we should design to give as much flexibility as possible to power users, but I'd still like to spare casual users of the nitty-gritty details - chances are that most of them will never need to know about `ConfigSource` or most of the other awesome PureConfig features. If we keep to the rule of keeping only no-arg `loadConfig` and `loadConfigOrThrow` in the object, there won't be any explosion of methods.

That said, I'm not too strong about these two as with the other design decisions. I'll happily deprecate `load` and `loadConfigOrThrow` if you think it will simplify the API 🙂 @jcazevedo, @leifwickland, do you still prefer them to be deprecated?

leifwickland: I'd lean toward deprecating the no-argument `loadConfig` so that there aren't two substantially different ways to load a config.

jcazevedo: I also lean towards deprecating them and only recommend using the `ConfigSource` way to load a config.

modules/cats-effect/src/test/scala/pureconfig/module/catseffect/CatsEffectSuite.scala
```diff
@@ -96,18 +95,4 @@ class CatsEffectSuite extends BaseSuite {
 
     pipeInput.available should be > 0
   }
-
-  "loadConfigFromFilesF" should "combine multiple config files" in {
-    val paths = NonEmptyList.of(getPath("multiple/one.properties"), getPath("multiple/two.properties"))
```
jcazevedo, line 101: Since we're removing these examples due to deprecating the `loadCOnfigFromFilesF` method, we can remove the `multiple/one.properties` and `multiple/two.properties` files from the test resources.

tests/src/test/scala/pureconfig/ConfigSourceSuite.scala
```diff
@@ -0,0 +1,205 @@
...
+import pureconfig.error._
+import pureconfig.generic.auto._
+
+class ConfigSourceSuite extends BaseSuite {
+
+  behavior of "ConfigSource"
+
+  it should "load from reference.conf by default" in {
+    case class Conf(d: Double, i: Int, s: String)
+    ConfigSource.default.load[Conf] shouldBe Right(Conf(0D, 0, "default"))
```
jcazevedo: Can we also add an `application.conf` file to the test resources to show that it takes precedence over the `reference.conf` in `ConfigSource.default`?

tests/src/test/scala/pureconfig/ConfigSourceSuite.scala
```diff
@@ -0,0 +1,205 @@
...
+    val resolve2 = ConfigSource.resources("conf/configSource/resolve2.conf")
+
+    resolve1.withFallback(resolve2).load[Conf] shouldBe Right(Conf("myhost1.example.com"))
+  }
+
+  it should "fail safely when a substitution can't be resolved" in {
+    case class Conf(hostOverride: String)
+    val resolve1 = ConfigSource.resources("conf/configSource/resolve1.conf")
+
+    resolve1.load[Conf] should failLike {
```
jcazevedo: Can you also add an example showing that calling `value()` on `resolve1` would fail similarly?

core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -0,0 +1,330 @@
...
+    new ConfigObjectSource(() => conf)
+
+  /**
+   * Creates a `ConfigObjectSource` from a `ConfigObjectCursor`.
+   *
+   * @param cur the cursor to be provided by this source
+   * @return a `ConfigObjectSource` providing the given cursor.
+   */
+  def fromCursor(cur: ConfigObjectCursor): ConfigObjectSource =
+    new ConfigObjectSource(() => Right(cur.value.toConfig))
```
jcazevedo: Unfortunately, materializing a `ConfigCursor` to a `ConfigValue` makes us lose the path to the value, which is useful for locating failures. This ends up changing the failure values of `ConfigSource#at` from what we previously had in the `loadConfig` methods scoped by a namespace. I think it might make sense to have a `ConfigSource` implementation that keeps its internal representation as a `ConfigCursor` for as long it can. I'd also not allow the `withFallback` method on `ConfigSource`s created from a cursor.

jcazevedo (changes requested):

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Got it 👍 Done in 0fac76a

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: It's tricky to get the semantics right for this. Config sources should always be treated as root configs in their loading pipelines, regardless of where they come from - partly because of the Typesafe Config substitution semantics and partly for simplicity.

I don't think it's correct to show the full path in errors when you decide to navigate in a cursor and then use it as a source, as the namespace for failures would then be different from the namespace used for variable resolution. This is also why I decided not to provide a method promoting a `ConfigSource` to a `ConfigObjectSource` (as you can navigate using `.at`).

This makes me think that I probably shouldn't allow creating config sources of any kind from cursors of any kind (except internally). What do you think?

tests/src/test/scala/pureconfig/ConfigSourceSuite.scala
ruippeixotog: Done.

tests/src/test/scala/pureconfig/ConfigSourceSuite.scala
ruippeixotog: 👍

core/src/main/scala/pureconfig/ConfigSource.scala
jcazevedo: > This makes me think that I probably shouldn't allow creating config sources of any kind from cursors of any kind (except internally). What do you think?

It'd help in this, yes. 

I think the biggest source of confusion here comes from allowing the "merging" of `ConfigSource`s created from external sources with `ConfigSource`s created from cursors. Since the "merging" is only allowed via the `withFallback` method of `ConfigObjectSource`, it might make sense to only remove the possibility of creating `ConfigObjectSource`s from cursors.

One other thing that we should probably do is override the `cursor()` and `fluentCursor()` methods in `ConfigSource.fromCursor`, so that history is kept in multiple `.at` calls.

ruippeixotog: TBH, I'm leaning more towards removing `fromCursor` even from `ConfigSource`. Even though I cannot think of any case where `ConfigSource.fromCursor` might cause inconsistencies with variable resolution, I think that we should discourage building sources and out of previously loaded configs - sources and cursors share some common functionalities like navigation and I'd rather "force" a single way of doing things my making the cursor->source path harder to do.

I cannot think of any use case where users might need this, but if people _really_ need it I don't mind having them converting their `ConfigCursor` to a `Config` manually to create a source.

leifwickland:
I'd support removing from Cursor.

core/src/main/scala/pureconfig/ConfigSource.scala
jcazevedo: > I cannot think of any use case where users might need this, but if people really need it I don't mind having them converting their `ConfigCursor` to a `Config` manually to create a source.

That seems reasonable, yes.

tests/src/test/scala/pureconfig/ConfigSourceSuite.scala
```diff
@@ -0,0 +1,199 @@
...
+    source.at("foo.bar.h").load[Float] should failWith(KeyNotFound("h", Set.empty), "foo.bar")
+    source.at("bar.foo.f").load[Float] should failWith(KeyNotFound("bar", Set.empty), "")
+    source.at("foo.baz.f").load[Float] should failWith(
+      WrongType(ConfigValueType.NUMBER, Set(ConfigValueType.OBJECT)), "foo.baz")
+
+    source.at("foo.bar.f").load[Option[Float]] shouldBe Right(Some(1.0F))
+    source.at("foo.bar.h").load[Option[Float]] should failWith(KeyNotFound("h", Set.empty), "foo.bar")
+    source.at("bar.foo.f").load[Option[Float]] should failWith(KeyNotFound("bar", Set.empty), "")
+    source.at("foo.baz.f").load[Option[Float]] should failWith(
+      WrongType(ConfigValueType.NUMBER, Set(ConfigValueType.OBJECT)), "foo.baz")
```
jcazevedo, line 51: Can you also add a similar test that uses multiple `.at` calls (e.g. `source.at("foo").at("bar").at("f")`) and checks that failures have the appropriate path?

jcazevedo (changes requested):
I just commented below about a test example that I think we should add, mainly to help avoid a regression. Everything else looks great! Thanks again for your work on this!

tests/src/test/scala/pureconfig/ConfigSourceSuite.scala
ruippeixotog: 1b594a0

jcazevedo (approved):

ruippeixotog:
@nigredo-tori are you ok with how `ConfigSource` looks right now?

nigredo-tori:
@ruippeixotog, looks great to me.

PR #545, opened 2019-08-15
Add key name to the error message on UnknownKey

adampauls:
It's nice to know for e.g. misspellings which key is unknown.

### DIFF
--- a/core/src/main/scala/pureconfig/error/FailureReason.scala
+++ b/core/src/main/scala/pureconfig/error/FailureReason.scala
@@ -89,7 +89,7 @@ object KeyNotFound {
  * @param key the unknown key
  */
 final case class UnknownKey(key: String) extends FailureReason {
-  def description = s"Unknown key."
+  def description = s"Unknown key: $key"
 }
 
 /**


ruippeixotog:
Hi @adampauls! `UnknownKey`'s description, like all others, are designed to provide friendly messages in reader exceptions:

```
scala> import pureconfig._

scala> import pureconfig.generic.auto._

scala> import com.typesafe.config.ConfigFactory

scala> val conf = ConfigFactory.parseString("a = 3, b = 4")
conf: com.typesafe.config.Config = Config(SimpleConfigObject({"a":3,"b":4}))

scala> case class Conf(a: Int)
defined class Conf

scala> implicit def productHint[A] = generic.ProductHint[A](allowUnknownKeys = false)
productHint: [A]=> pureconfig.generic.ProductHint[A]

scala> loadConfigOrThrow[Conf](conf)
pureconfig.error.ConfigReaderException: Cannot convert configuration to a Conf. Failures are:
  at 'b':
    - Unknown key.

  at pureconfig.ConfigSource.loadOrThrow(ConfigSource.scala:79)
  at pureconfig.ConfigSource.loadOrThrow$(ConfigSource.scala:76)
  at pureconfig.ConfigObjectSource.loadOrThrow(ConfigSource.scala:90)
  at pureconfig.package$.loadConfigOrThrow(package.scala:153)
  ... 36 elided
```

As you can see, failures are already grouped by their key, so there is no need to add it to the description. On the other hand, when you are handling the errors programatically you have direct access to `key` as a case class field. Can you explain a bit better your use case for this?

adampauls:
Ah, my apologies. We have a customized version of the parser and my error was in not passing the correct location to the ConvertFailure. I was doing `ConvertFailure(UnknownKey(k), cursor)` where `cursor` point to the node in the config tree that contains `k` but shouldn't, but the code is expecting `ConvertFailure(UnknownKey(k), cursor.atKey(k).right.get)`, is that correct? 

It is a little confusing that the class takes the key as a parameter but does not print it in its description, but I now understand why it's set up this way. Thanks!

Update: I see that I can use the `ConfigObjectCursor.map` to loop over the keys and their cursors.

ruippeixotog:
Just for a small note, you can simplify `ConvertFailure(UnknownKey(k), cursor.atKey(k).right.get)` to `cursor.atKey(k).failureFor(UnknownKey(k))` (if you want a `ConfigReaderFailure`) or `cursor.atKey(k).failed(UnknownKey(k))` (if you want a more common `ConfigReader.Result[A]`).

PR #561, opened 2019-09-07
Support for Scala 2.13 in cron4s

alonsodomin:
`cron4s` supports Scala 2.13 in the 0.6 version series, but at the same time drops support for Scala 2.11. This change makes this module available for all the cross Scala versions at the cost of choosing a different version of the dependency.

Other option would be dropping 2.11 support in this module.

ps: `cron4s` 0.6 will be source compatible with 0.5.

### DIFF
--- a/modules/cron4s/build.sbt
+++ b/modules/cron4s/build.sbt
@@ -1,13 +1,8 @@
 name := "pureconfig-cron4s"
 
-libraryDependencies += {
-  val cron4sVersion = CrossVersion.partialVersion(scalaVersion.value) match {
-    case Some((2, n)) if n < 12 => "0.5.0"
-    case _                      => "0.6.0"
-  }
-  
-  "com.github.alonsodomin.cron4s" %% "cron4s-core" % cron4sVersion
-}
+crossScalaVersions := List("2.13.1", "2.12.9")
+
+libraryDependencies += "com.github.alonsodomin.cron4s" %% "cron4s-core" % "0.6.0"
 
 developers := List(
   Developer("bardurdam", "Bárður Viberg Dam", "bardurdam@gmail.com", url("https://github.com/bardurdam")))
--- a/modules/cron4s/build.sbt
+++ b/modules/cron4s/build.sbt
@@ -1,9 +1,13 @@
 name := "pureconfig-cron4s"
 
-crossScalaVersions ~= { _.filterNot(_.startsWith("2.13")) }
-
-libraryDependencies ++= Seq(
-  "com.github.alonsodomin.cron4s" %% "cron4s-core" % "0.5.0")
+libraryDependencies += {
+  val cron4sVersion = CrossVersion.partialVersion(scalaVersion.value) match {
+    case Some((2, n)) if n < 12 => "0.5.0"
+    case _                      => "0.6.0-M2"
+  }
+  
+  "com.github.alonsodomin.cron4s" %% "cron4s-core" % cron4sVersion
+}
 
 developers := List(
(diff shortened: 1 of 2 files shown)


jcazevedo:
Thanks for the pull request! 

I'm OK with supporting different Cron4s versions in different Scala versions. However, we'd rather only publish modules for stable releases. Can we wait for a stable release of Cron4s 0.6.0 before starting publishing the module for Scala 2.13?

alonsodomin:
yes of course, do you prefer to close this and reopen later when the stable version is available? - it basically depends on final versions from `cats`, `circe`, etc.

jcazevedo:
We can do as you prefer. I don't mind keeping this open for visibility and update it once a final version is available.

jcazevedo:
@alonsodomin: Now that Cron4s 0.6.0 was released, would you like to update this PR to use it?

modules/cron4s/src/test/scala/pureconfig/module/cron4s/Cron4sSuite.scala
```diff
@@ -24,7 +24,7 @@ class Cron4sSuite extends BaseSuite {
 
     val errors = ConfigReaderFailures(
       ConvertFailure(
-        CannotConvert("10-65 * * * * *", "CronExpr", "Expected '\" \" | 0-59' at position 2 but found '\"-65 * * * * *\"'"),
+        CannotConvert("10-65 * * * * *", "CronExpr", "blank expected at position 3 but found '-'"),
```
ruippeixotog, line 27: Unfortunately this will fail for Scala 2.11, which still uses the old version :( You can choose to either not check the error message or drop Scala 2.11 support.

ruippeixotog (changes requested):

modules/cron4s/src/test/scala/pureconfig/module/cron4s/Cron4sSuite.scala
alonsodomin: what about dropping 2.11 support? not sure if there are any big changes in the core of `pureconfig` but in regards to cron4s, it could be argued that users could use the previous module version if they need to run with 2.11...

ruippeixotog: Yes, I tend to agree. While we'll support Scala 2.11 in core for some more time, I think it wouldn't be critical for `pureconfig-cron4s` to stop being updated.

modules/cron4s/build.sbt
```diff
@@ -1,9 +1,8 @@
 name := "pureconfig-cron4s"
 
-crossScalaVersions ~= { _.filterNot(_.startsWith("2.13")) }
+crossScalaVersions := List("2.13.1", "2.12.9")
```
jcazevedo: Just to make it easier to define versions in a single location, can you change this to filter scala versions starting with 2.11?

```suggestion
crossScalaVersions ~= { _.filterNot(_.startsWith("2.11")) }
```

jcazevedo (changes requested):

ruippeixotog (approved):
LGTM!

@alonsodomin Unrelated to this PR: the cron4s webpage seems to be having problems, something which is detected by our website build (https://travis-ci.org/pureconfig/pureconfig/jobs/590773745). Are you going to fix it soon? :)

alonsodomin:
I’ve been linking the GH Pages domain to one of my own today. I’m waiting on the Let’s Encrypt certificate, shall be ready in a day or two.

I wonder how that could affect the the other domain... 🤔

jcazevedo (approved):

PR #562, opened 2019-09-07
Fix bugs and improve ConfigSource#recoverWith API

ruippeixotog:
Fixes a bug where `.recoverWith` crashes with an exception if the provided partial function is not defined for the failures to be handled. It also makes it more consistent by requesting a `ConfigObjectSource` as fall back instead of an `Either[ConfigReaderFailures, Config]`.


jcazevedo (approved):
Looks good. Nice catch!

PR #563, opened 2019-09-07
Add website section for ConfigSource

ruippeixotog:
Depends on #562.

### DIFF
--- a/docs/src/main/tut/docs/loading-a-config.md
+++ b/docs/src/main/tut/docs/loading-a-config.md
@@ -5,7 +5,7 @@ title: Loading a Config
 
 ## {{page.title}}
 
-In the Quick Start section we used `ConfigSource.default.load[MyClass]` to read a config from an `application.conf`
+In [Quick Start](index.html) we used `ConfigSource.default.load[MyClass]` to read a config from an `application.conf`
 resource and convert it to a case class. `ConfigSource.default` is an instance of `ConfigSource` - a trait representing
 sources from which we can load configuration data. This `ConfigSource` in particular reads and builds a config according
 to Typesafe Config's [standard behavior](https://github.com/lightbend/config#standard-behavior), which means it can be
@@ -18,7 +18,7 @@ The `ConfigSource` companion object defines many other ready-to-use sources, lik
 - `ConfigSource.url` - reads a config from an URL;
 - `ConfigSource.string` - reads a literal config from a string.
 
-After you have a config source you can read your config using several methods:
+After you have a config source you can load your config using several methods:
 
 ```tut:silent
 import pureconfig._
@@ -81,7 +81,7 @@ otherAppSource.optional.withFallback(defaultsSource).load[Conf]
 You also have the option to use an alternative source in case your primary source can't be read by using `.recoverWith`:
 
 ```tut:book
-otherAppSource.recoverWith { case _ => appSource }.load[Conf]
+otherAppSource.recoverWith { case _ => defaultsSource }.load[Conf]
 ```
 
 ### Loading a Config in a Path
--- a/docs/src/main/tut/docs/loading-a-config.md
+++ b/docs/src/main/tut/docs/loading-a-config.md
@@ -1,0 +1,117 @@
+---
+layout: docs
+title: Loading a Config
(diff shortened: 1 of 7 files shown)


docs/src/main/tut/docs/loading-a-config.md
```diff
@@ -0,0 +1,117 @@
+---
+layout: docs
+title: Loading a Config
+---
+
+## {{page.title}}
+
+In the Quick Start section we used `ConfigSource.default.load[MyClass]` to read a config from an `application.conf`
```
jcazevedo: Can we add a link to the Quick Start section here?

docs/src/main/tut/docs/loading-a-config.md
```diff
@@ -0,0 +1,117 @@
...
+used as a replacement for `ConfigFactory.load` in codebases already using Typesafe Config.
+
+The `ConfigSource` companion object defines many other ready-to-use sources, like:
+
+- `ConfigSource.file` - reads a config from a file in a file system;
+- `ConfigSource.resources` - reads a config from resources in your classpath or packaged application;
+- `ConfigSource.url` - reads a config from an URL;
+- `ConfigSource.string` - reads a literal config from a string.
+
+After you have a config source you can read your config using several methods:
```
jcazevedo: Since we already used `read` before with a slightly different meaning, I think that changing this to `load` makes things clearer:

```suggestion
After you have a config source you can load your config using several methods:
```

docs/src/main/tut/docs/loading-a-config.md
```diff
@@ -0,0 +1,117 @@
...
+```tut:book
+otherAppSource.withFallback(defaultsSource).load[Conf]
+
+otherAppSource.optional.withFallback(defaultsSource).load[Conf]
+```
+
+You also have the option to use an alternative source in case your primary source can't be read by using `.recoverWith`:
+
+```tut:book
+otherAppSource.recoverWith { case _ => appSource }.load[Conf]
```
jcazevedo: I prefer using a successful scenario here. We can show more usages of `recoverWith`, but I think it's clearer if the first one is a successful case.

```suggestion
otherAppSource.recoverWith { case _ => defaultsSource }.load[Conf]
```

jcazevedo (changes requested):
There are some other pages in the website docs that we should update:

* There are mentions to `loadConfig*` methods in the ["Quick Start"](https://github.com/pureconfig/pureconfig/blob/56de256ec1bfd28245c0a1000ff28bb7cc91f616/docs/src/main/tut/docs/index.md) page. Now that we have a whole section on loading configs, we can point users there for more details. It also adds some continuity to the documentation.
* The ["Supporting New Types"](https://github.com/pureconfig/pureconfig/blob/56de256ec1bfd28245c0a1000ff28bb7cc91f616/docs/src/main/tut/docs/supporting-new-types.md) page mentions `loadConfig` before a code block that uses `load` on a `ConfigSource`.
* The ["Error Handling"](https://github.com/pureconfig/pureconfig/blob/56de256ec1bfd28245c0a1000ff28bb7cc91f616/docs/src/main/tut/docs/error-handling.md) page mentions `loadConfig` and `loadConfigOrThrow` before a code block that showcases `loadOrThrow`.
* The ["Non-Automatic Derivation"](https://github.com/pureconfig/pureconfig/blob/56de256ec1bfd28245c0a1000ff28bb7cc91f616/docs/src/main/tut/docs/non-automatic-derivation.md) page mentions `loadConfig` before a code block that uses `load` on `ConfigSource`.

docs/src/main/tut/docs/loading-a-config.md
ruippeixotog: Sure. I tried to distinguish the two actions as you said, but this one must have slipped.

docs/src/main/tut/docs/loading-a-config.md
ruippeixotog: I intended it to be a successful scenario, but I actually didn't notice the error. Thanks!

jcazevedo (approved):
Looks great! Thanks for this!

PR #579, opened 2019-09-16
Cross-build cats-effect module for 2.13

ghost:


.sbtopts
```diff
@@ -0,0 +1,4 @@
+-J-Xss4m
```
ruippeixotog: Is there a way to define these only on the cats-effect compilation SBT task?

ghost: I don't believe so. Those need to be set before sbt initialises anyway I guess.

jcazevedo: @dom-nguyen: Do we need to have these settings? I'm able to compile the project locally without them (I'm using [paulp/sbt-extras](https://github.com/paulp/sbt-extras) which provides more sensible defaults), and Travis also seems happy without them (see https://travis-ci.org/pureconfig/pureconfig/builds/585745699).

ghost: Hmm yeah I got an failure earlier in Travis wrt out of Metaspace, but now it's fine. Removed the sbtopts now.

Didn't know about sbt-extras, thanks for that!

jcazevedo (approved):
@dom-nguyen: Thanks for this!

.sbtopts
ruippeixotog: FYI, Travis also uses sbt-extras 👍

PR #580, opened 2019-09-16
Build pureconfig-cats for Scala 2.13

jcazevedo:
This PR starts building `pureconfig-cats` for Scala 2.13.

I had to introduce a dependency on a milestone version of `discipline-scalatest`, which I think is OK since it's only being used in tests. Unfortunately, that dependency depends on a snapshot version of ScalaTest, so we can't use our `BaseSuite` and `ConfigConvertChecks` directly in `CatsSuite`. For the time being, I copied the relevant methods to `CatsSuite`.

### DIFF
--- a/modules/cats/build.sbt
+++ b/modules/cats/build.sbt
@@ -1,11 +1,9 @@
 name := "pureconfig-cats"
 
-crossScalaVersions ~= { _.filterNot(_.startsWith("2.13")) }
-
 libraryDependencies ++= Seq(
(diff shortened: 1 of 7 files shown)


jcazevedo:
I've updated ScalaCheck to 1.14.1 here. This ends up superseding #495 and #586.

ruippeixotog (approved):
I don't really like the idea of depending on shapshots (prod or test), but since `cats` is an important module I guess we can do it right away...

PR #581, opened 2019-09-16
Build pureconfig-circe for Scala 2.13

jcazevedo:
This PR starts building `pureconfig-circe` for Scala 2.13.

Circe is no longer available for Scala 2.11 since 0.11.1, so we have to stay there for that Scala version.

### DIFF
--- a/modules/circe/build.sbt
+++ b/modules/circe/build.sbt
@@ -1,16 +1,11 @@
+import Utilities._
+
 name := "pureconfig-circe"
 
-def onScala211(onScala211: String, onOthers: String) = Def.setting {
-  CrossVersion.partialVersion(scalaVersion.value) match {
-    case Some((2, 11)) => onScala211
-    case _ => onOthers
-  }
-}
-
 libraryDependencies ++= Seq(
-  "io.circe"      %% "circe-core"     % onScala211("0.11.1", "0.12.1").value,
-  "io.circe"      %% "circe-literal"  % onScala211("0.11.1", "0.12.1").value % Test,
-  "org.typelevel" %% "jawn-parser"    % "0.14.2" % Test
+  "io.circe"      %% "circe-core"    % forScalaVersions { case (2, 11) => "0.11.1"; case _ => "0.12.1" }.value,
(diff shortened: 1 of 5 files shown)


modules/circe/build.sbt
```diff
@@ -1,11 +1,16 @@
 name := "pureconfig-circe"
 
-crossScalaVersions ~= { _.filterNot(_.startsWith("2.13")) }
+def onScala211(onScala211: String, onOthers: String) = Def.setting {
```
ruippeixotog: Can we generalize this function a bit and move it to `project/`? It's a pretty good utility and we've been using it frequently, so I guess we'll need it again later in other places.

ruippeixotog (changes requested):

modules/circe/build.sbt
jcazevedo: Can we generalize it and move it if/when we need it? With #580 we no longer need to have different versions of ScalaCheck for different Scala versions so this becomes the only module requiring something like this. Project definitions tend to become bloated, so I'd rather keep the main `build.sbt` and whatever we have under the `project/` folder as simple as possible.

ruippeixotog: I'm ok with adding this later 👍

It's more a question of when, not if. There are some projects that have a policy of maintaining only the latest two Scala versions (`http4s`, for which we also have a module, is one of them). In general, we can't reasonably expect that we'll always drop Scala versions before every other dependency we have.

I wouldn't find very bad to have general-purpose utilities like this in a `.scala` file in `project/`. Right now our build is fairly clean and because we already have a `Dependencies.scala` with it it wouldn't be any more bloated than what it is right now. Fun fact: I already added this function before for Scala 2.10, then removed it and later added the Scala 2.13 one we have now. We already went through this 🙂 

Accepting the PR to unblock this - feel free to merge if you still prefer to add it later :)

ruippeixotog (approved):

modules/circe/build.sbt
jcazevedo: Fair enough. I've added a `forScalaVersions` utility function in 3cc99c7. We can also use it in other locations of the build definition not related to dependencies.

PR #585, opened 2019-09-17
Create ConfigSource for YAML

### LINKED ISSUE #573: Create ConfigSource for YAML
v0.12.0 saw the introduction of `ConfigSource`s, but `pureconfig-yaml` still needs to be migrated to the new API.

ruippeixotog:
Converts `pureconfig.module.yaml` to use the ConfigSource API.

Closes #573.

### DIFF
--- a/modules/yaml/src/main/scala/pureconfig/module/yaml/YamlConfigSource.scala
+++ b/modules/yaml/src/main/scala/pureconfig/module/yaml/YamlConfigSource.scala
@@ -18,12 +18,22 @@ import pureconfig.error._
 import pureconfig.module.yaml.error.{ NonStringKeyFound, UnsupportedYamlType }
 import pureconfig.{ ConfigObjectSource, ConfigSource }
 
+/**
+ * A `ConfigSource` that reads configs from YAML documents in a stream, file or string.
+ *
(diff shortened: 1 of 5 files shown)


leifwickland (approved):
Thanks, @ruippeixotog!

modules/yaml/src/main/scala/pureconfig/module/yaml/YamlConfigSource.scala
```diff
@@ -0,0 +1,174 @@
...
+   * Returns a new source that produces a multi-document YAML read by this source as a config list.
+   *
+   * @return a new source that produces a multi-document YAML read by this source as a config list.
+   */
+  def multiDoc: ConfigSource = new ConfigSource {
+    def value(): Result[ConfigValue] = {
+      usingReader { reader =>
+        loader.loadAll(reader).asScala
+          .map(yamlObjToConfigValue)
+          .foldRight(Right(Nil): Result[List[ConfigValue]])(Result.zipWith(_, _)(_ :: _))
```
jcazevedo, line 64: Similarly to `fail`, we should probably also add a smart constructor for successful `Result`s.

jcazevedo (approved):
Looks good. Thanks for this!

Issue #572, opened 2019-09-13
Create initial version of Magnolia module


ruippeixotog:
[Magnolia](https://github.com/propensive/magnolia) is an alternative to shapeless to derive type classes for sealed traits and case classes. Many people use semiauto derivation instead of auto derivation because of the compilation times incurred by shapeless on auto-derivation. Magnolia is supposed to be better in this regard.

We can create an initial `pureconfig-generic-magnolia` that serves as POC. It would be a replacement for `pureconfig-generic` on simple case class and sealed trait derivations, but it does not have to be feature par in any sense - in particular hints might be watered down or even non-existent.

The discussion in #396 might be a good starting point for this.

PR #594, opened 2019-09-24
Move hints and related errors to a separate module

ruippeixotog:
As a preparation for #572, this PR moves `CoproductHint`, `ProductHint` and related failures to a separate internal module so that they can be shared by all generic derivation libs.


modules/generic/src/main/scala/pureconfig/generic/EnumCoproductHint.scala
```diff
@@ -0,0 +1,38 @@
+package pureconfig.generic
+
+import com.typesafe.config.{ ConfigFactory, ConfigObject, ConfigValue, ConfigValueType }
+import pureconfig._
+import pureconfig.error._
+import pureconfig.syntax._
```
jcazevedo, line 6: Nitpick:

```suggestion
import pureconfig.syntax._

```

jcazevedo (approved):

Issue #595, opened 2019-09-24
Improve SBT project structure


ruippeixotog:
Currently, our modules are organized in a complex and tricky way:
- We have test utilities (testkits) mixed with test suites and we have module tests depend on both of them;
- Even though we define `generic` as a module, it has a special status since all tests and tut code depend on it - making it impossible for `pureconfig-generic` to have tests using `BaseSuite` and for us to generalize project definitions in `build.sbt`;
- All tests depend heavily on generic derivation, even when we're only testing unrelated `pureconfig-core` classes;
- The effort to have an alternative derivation module (#572) will create more special modules.

Ideally, I see our project having a structure like this:
- All documentation (including module-specific docs) in a single `docs` project that is rendered on the website (this is also required by #574);
- A `testkit` project containing only test utilities. This could be elevated later to a public (published) module, if we decide to get it mature enough;
- Test suites live in their related projects, like `core/src/test` or `modules/generic/src/test`.

What do you think about this proposal, @jcazevedo and @leifwickland?

jcazevedo:
I like that! 

I'd also like to entertain the idea of moving the modules not related to type class derivation to their own repository under the PureConfig GitHub organization. Some advantages that I see with that approach:

* We could have different maintainers for those modules;
* It'd make the build for the main PureConfig repository faster and simpler;
* It'd allow their documentation to be kept separated from the core documentation;
* The modules could have an independent release cycle.

ruippeixotog:
I don't like at all projects with a multi-repo organization for closely related modules 🙁 Generally, we'd have all the problems for which a monorepo is usually chosen:
- It would make it much harder to keep modules up-to-date with the latest PureConfig version, and abandoned modules wouldn't get updated anymore (it's not reasonable for us to be going to N repos to make sure they are up-to-date);
- It's way harder for people changing pureconfig-core to understand if the changes will negatively affect other modules;
- We'd need to be way more disciplined with binary compatibility, or risk modules needing to publish their latest versions for multiple PureConfig versions;
- Finding correct versions for each module would be hellish for users (I like how currently you are guaranteed that choosing any X.Y.Z version of PureConfig will get you a compatible set of modules to work with);
- Modules are currently very standardized (tut REDAMEs, packages, tests) and I think that us having control over this results in a better experience for users.

Regarding your points:

[...]

PR #596, opened 2019-09-24
Refactor test dependency definitions

ruippeixotog:
Part of #595, this PR splits `tests` files into test utilities (`tests/src/main`, temporarily) and test suites (`tests/src/test`). It also makes module tests depend only on the utilities part (only the `joda` module posed some problems).


ruippeixotog:
Tests are still failing because of the broken cron4s link. Maybe it's worth changing it to point to the GitHub page.

jcazevedo (approved):

jcazevedo:
> Maybe it's worth changing it to point to the GitHub page.

Yes. Let's do it for now.

Issue #576, opened 2019-09-13
sbt-release tries to publish artefacts twice


ruippeixotog:
For the past few releases, sbt-release has been failing with an error message saying I was trying to overwrite an artefact in Sonatype and that was not allowed. This suggests that our project configuration is trying to publish the same artefact twice.

This is not related with the recent update to `sbt-sonatype`; it has been happening for some time now, I was just too lazy to fix it and finished the process manually instead 😞 It would be good to fix this for real.

ruippeixotog:
I didn't have this problem when releasing v0.12.2, so I guess #601 fixed this!

PR #601, opened 2019-09-28
Make some improvements to the build definition

jcazevedo:
This PR makes some improvements to our build definition:

* Removes an extra Scala 2.13.0 version from `crossScalaVersions` in the `bundle` and `tests` projects;
* Prevent us from having an extra Scala version in the root project defined by sbt (which I think was what was causing #576), by reusing the Scala version settings used in subprojects;
* ~~Start compiling against Scala 2.13.1.~~


jcazevedo:
I have reverted the update to Scala 2.13.1 due to scoverage/sbt-scoverage#295. I think it's preferable to wait for an sbt-coverage update than adding an exception to the CI build in this case.

build.sbt
```diff
@@ -39,7 +39,7 @@ lazy val docs = (project in file("docs")).
 def module(proj: Project) = proj.
   enablePlugins(SbtOsgi, TutPlugin).
   dependsOn(core).
-  dependsOn(tests % "test->test"). // In order to reuse thDerivationSuite scalacheck generators
+  dependsOn(tests % "test->test"). // In order to reuse the DerivationSuite scalacheck generators
```
ruippeixotog: This will clash with #596, submitted a few days ago. Can we merge that one first?

build.sbt
```diff
@@ -179,6 +180,10 @@ lazy val lintFlags = {
   }
 }
 
+// Use common settings for Scala versions in the root project
+commonScalaVersionSettings
+crossScalaVersions ~= { versions => Seq(versions.head) }
```
ruippeixotog: This seems to make sense to me, but what exactly is your theory on how this might be causing #576? And Is there any problem in leaving `crossScalaVersions` defined here like in other projects? Currently we use `commonScalaVersionSettings` in a quite funky way (we extracted that to a different method to use it twice, where one of the times we "discard" one of the two settings).

ruippeixotog: Also, can you test `release` (the steps before publishing) to make sure this doesn't break the current setup?

ruippeixotog (changes requested):
LGTM overall!

build.sbt
jcazevedo: > This seems to make sense to me, but what exactly is your theory on how this might be causing #576?

I couldn't replicate #576 when 0.12.1 was released, and I noticed that sbt 1.3.2 was built for Scala 2.12.10, which is the same Scala 2.12 version we use. If I roll back to sbt 1.3.0 (which was built for Scala 2.12.8), the `scalaVersion` for the root project will be set to 2.12.8, and every command prefixed with `+` will run for all versions in `crossScalaVersions` plus the `scalaVersion` of the root project (it's easy to test with a `+compile` or `+test`). My assumption was that `+publishSigned` was similarly affected, so we'd try to publish two artifacts for Scala 2.12.

> And Is there any problem in leaving `crossScalaVersions` defined here like in other projects?

Unfortunately, if `crossScalaVersions` of the root project is defined with more than one Scala version, strict aggregation doesn't seem to work properly (trying `++ 2.13.0 test` will fail in modules that shouldn't be built for Scala 2.13.0). There was a report of a similar issue in sbt/sbt#3698 after sbt/sbt#3995.

We might be able use less `crossScalaVersions` shenanigans if we define our Scala versions somewhere accessible by the root project.

build.sbt
jcazevedo: Sure!

build.sbt
jcazevedo: > We might be able use less `crossScalaVersions` shenanigans if we define our Scala versions somewhere accessible by the root project.

6980018 improves things a bit.

jcazevedo: > Also, can you test `release` (the steps before publishing) to make sure this doesn't break the current setup?

I tried with a custom release process that publishes things locally and things seem to work as expected.

build.sbt
```diff
@@ -73,8 +74,8 @@ lazy val commonSettings = Seq(
     Developer("ruippeixotog", "Rui Gonçalves", "ruippeixotog@gmail.com", url("https://github.com/ruippeixotog")),
     Developer("derekmorr", "Derek Morr", "morr.derek@gmail.com", url("https://github.com/derekmorr"))),
 
+  crossScalaVersions := Seq(scala212, scala213, scala211),
   scalaVersion := crossScalaVersions.value.head,
```
ruippeixotog: Nit: since the first version of `crossScalaVersions` is not guaranteed to be the "main" one anymore, can we define `scalaVersion := scala212` and sort `crossScalaVersions`? We can define a `mainScala` in `Dependency.Versions` if we really want to define it a single place, but I don't think it's necessary.

ruippeixotog (approved):

build.sbt
jcazevedo: Sure. Done that in cc447ee.

PR #602, opened 2019-09-28
Enable `sonatypeBundleRelease`

jcazevedo:
sbt-sonatype introduced the `sonatypeBundleUpload` command in [version 3.1](https://github.com/xerial/sbt-sonatype/blob/master/ReleaseNotes.md#31) which speeds up the upload of multiple artifacts to Sonatype. `sonatypeBundleRelease` was introduced in [version 3.4](https://github.com/xerial/sbt-sonatype/blob/master/ReleaseNotes.md#34) and is an alias for common bundle commands.

I haven't tried out yet, but since we upload a few dozen artifacts to Sonatype on each release, we should benefit from this.


ruippeixotog (approved):
This sounds awesome, publishing all the modules currently takes an awful amount of time 😢

LGTM, but please test doing a `release` (replacing `sonatypeBundleRelease` with `;sonatypePrepare;sonatypeUploadBundle`) before merging, to check if everything works as expected. Thank you!

jcazevedo:
@ruippeixotog: There was one thing missing (which I added in 71edc90). We need to define the organization in the root project so that the bundle is properly created. I was able to create and upload the bundle to sonatype using `;sonatypePrepare;sonatypeBundleUpload`.

PR #609, opened 2019-10-03
Create Magnolia module

### LINKED ISSUE #572: Create initial version of Magnolia module

ruippeixotog:
Creates a module for case class and sealed family derivation powered by Magnolia. Most `pureconfig-generic` features were ported into this module, which should make life easy for anyone willing to try it.

About the differences in behavior listed in the README:

- _When loading case classes and multiple fields cause failures, only the first error is reported_
  - We can either wait for a response or solution in https://github.com/propensive/magnolia/issues/181 or use Magnolia's `rawConstruct`, which will result in much uglier derivation code
- _Value classes that are not case classes are not supported (other value classes have the same reading behavior)_
  - Not likely to be fixed, as Magnolia seems to target only case classes
- _The Scala compiler may emit a warning unless `scala.language.higherKinds` is imported_
  - This probably happens because the generated macro code uses higher kinds. It's unlikely that we can change that, but maybe the Magnolia macro could import this internally?

I don't think any of these differences will be a huge deal for users.

I also had to copy many tests from `tests` to this module so that I could provide appropriate test coverage for the ported features. I'm not happy at all with doing that, but sharing code here would probably mean:
- Playing with macros to dynamically choose import statements. We can't just extract tests to a trait as the auto-derivation methods don't share a common signature and testing hints require readers not to be hardcoded;
- Creating yet another test utils project that also imports `pureconfig-generic-base`.

I plan to find a solution for test organization and build utilities to test auto-derivation at some point, but it would probably take too much time for the value it would bring right now.

Depends on #607. Closes #572. This is definitely easier to review commit by commit.

### DIFF
--- a/docs/src/main/tut/docs/library-integrations.md
+++ b/docs/src/main/tut/docs/library-integrations.md
@@ -21,7 +21,7 @@ The core of PureConfig eschews unnecessary dependencies. Separate modules exist 
 - [`pureconfig-http4s`](https://github.com/pureconfig/pureconfig/tree/master/modules/http4s) provides converters for [Http4s](http://http4s.org/) types;
 - [`pureconfig-javax`](https://github.com/pureconfig/pureconfig/tree/master/modules/javax) provides converters for classes in `javax` packages;
 - [`pureconfig-joda`](https://github.com/pureconfig/pureconfig/tree/master/modules/joda) provides configurable converters for [Joda-Time](http://www.joda.org/joda-time/) types;
-- [`pureconfig-magnolia`](https://github.com/pureconfig/pureconfig/tree/master/modules/magnolia) provides an alternative way to derive case classes and sealed traits using [Magnolia](https://propensive.com/opensource/magnolia);
+- [`pureconfig-magnolia`](https://github.com/pureconfig/pureconfig/tree/master/modules/magnolia) provides an alternative way to derive readers and writers for case classes and sealed traits using [Magnolia](https://propensive.com/opensource/magnolia);
 - [`pureconfig-scala-xml`](https://github.com/pureconfig/pureconfig/tree/master/modules/scala-xml) provides support for XML via [scala-xml](https://github.com/scala/scala-xml);
 - [`pureconfig-scalaz`](https://github.com/pureconfig/pureconfig/tree/master/modules/scalaz) provides converters for [Scalaz](https://github.com/scalaz/scalaz) data structures and Scalaz typeclass instances;
 - [`pureconfig-squants`](https://github.com/pureconfig/pureconfig/tree/master/modules/squants) provides converters for [Squants](http://www.squants.com/)'s beautiful types representing units of measure;
--- a/docs/src/main/tut/docs/library-integrations.md
+++ b/docs/src/main/tut/docs/library-integrations.md
@@ -21,6 +21,7 @@ The core of PureConfig eschews unnecessary dependencies. Separate modules exist 
 - [`pureconfig-http4s`](https://github.com/pureconfig/pureconfig/tree/master/modules/http4s) provides converters for [Http4s](http://http4s.org/) types;
 - [`pureconfig-javax`](https://github.com/pureconfig/pureconfig/tree/master/modules/javax) provides converters for classes in `javax` packages;
 - [`pureconfig-joda`](https://github.com/pureconfig/pureconfig/tree/master/modules/joda) provides configurable converters for [Joda-Time](http://www.joda.org/joda-time/) types;
+- [`pureconfig-magnolia`](https://github.com/pureconfig/pureconfig/tree/master/modules/magnolia) provides an alternative way to derive case classes and sealed traits using [Magnolia](https://propensive.com/opensource/magnolia);
 - [`pureconfig-scala-xml`](https://github.com/pureconfig/pureconfig/tree/master/modules/scala-xml) provides support for XML via [scala-xml](https://github.com/scala/scala-xml);
 - [`pureconfig-scalaz`](https://github.com/pureconfig/pureconfig/tree/master/modules/scalaz) provides converters for [Scalaz](https://github.com/scalaz/scalaz) data structures and Scalaz typeclass instances;
 - [`pureconfig-squants`](https://github.com/pureconfig/pureconfig/tree/master/modules/squants) provides converters for [Squants](http://www.squants.com/)'s beautiful types representing units of measure;
--- a/modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
+++ b/modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
@@ -16,7 +16,7 @@ object MagnoliaConfigReader {
     else if (ctx.isValueClass) combineValueClass(ctx)
     else combineCaseClass(ctx)
 
-  def combineCaseClass[A](ctx: CaseClass[ConfigReader, A])(implicit hint: ProductHint[A]): ConfigReader[A] = new ConfigReader[A] {
+  private def combineCaseClass[A](ctx: CaseClass[ConfigReader, A])(implicit hint: ProductHint[A]): ConfigReader[A] = new ConfigReader[A] {
     def from(cur: ConfigCursor): Result[A] = {
       cur.asObjectCursor.flatMap { objCur =>
         val res = ctx.constructMonadic[Result, Param[ConfigReader, A]#PType] { param =>
@@ -24,7 +24,7 @@ object MagnoliaConfigReader {
           objCur.atKeyOrUndefined(keyStr) match {
             case keyCur if keyCur.isUndefined =>
               param.default match {
-                case Some(defaultValue) if hint.useDefaultArgs => Right(defaultValue): Result[param.PType]
+                case Some(defaultValue) if hint.useDefaultArgs => Right(defaultValue)
                 case _ if param.typeclass.isInstanceOf[ReadsMissingKeys] => param.typeclass.from(keyCur)
                 case _ => cur.failed(KeyNotFound.forKeys(keyStr, objCur.keys))
               }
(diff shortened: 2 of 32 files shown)


ruippeixotog:
It occurred to me that I didn't implement semiauto enumeration converters. I'll add them in a future PR in order not to make this one more difficult to review than it already is.

docs/src/main/tut/docs/library-integrations.md
```diff
@@ -21,6 +21,7 @@ The core of PureConfig eschews unnecessary dependencies. Separate modules exist
 - [`pureconfig-http4s`](https://github.com/pureconfig/pureconfig/tree/master/modules/http4s) provides converters for [Http4s](http://http4s.org/) types;
 - [`pureconfig-javax`](https://github.com/pureconfig/pureconfig/tree/master/modules/javax) provides converters for classes in `javax` packages;
 - [`pureconfig-joda`](https://github.com/pureconfig/pureconfig/tree/master/modules/joda) provides configurable converters for [Joda-Time](http://www.joda.org/joda-time/) types;
+- [`pureconfig-magnolia`](https://github.com/pureconfig/pureconfig/tree/master/modules/magnolia) provides an alternative way to derive case classes and sealed traits using [Magnolia](https://propensive.com/opensource/magnolia);
```
jcazevedo: I think that saying `...an alternative way to derive readers and writers for case classes and sealed traits...` is preferable.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  def combineTuple[A](ctx: CaseClass[ConfigReader, A]): ConfigReader[A] = new ConfigReader[A] {
+    def from(cur: ConfigCursor): Result[A] = {
+      val collCur = cur.asListCursor.map(Right.apply).left.flatMap(failure =>
+        cur.asObjectCursor.map(Left.apply).left.map(_ => failure))
+
+      collCur.flatMap {
+        case Left(_) => combineCaseClass(ctx).from(cur)
```
jcazevedo: This `combineCaseClass` is relying on the default `ProductHint` implementation. While it's very unlikely that we'll ever change it in a way that would break this, I think it'd be safer to provide an explicit `ProductHint` that we know won't mess with the field names.

Also, I think it's preferable to cache the call to `combineCaseClass(ctx)` outside the `from` method, to avoid creating a new `ConfigReader` on every `from` call.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  def combineCaseClass[A](ctx: CaseClass[ConfigReader, A])(implicit hint: ProductHint[A]): ConfigReader[A] = new ConfigReader[A] {
+    def from(cur: ConfigCursor): Result[A] = {
+      cur.asObjectCursor.flatMap { objCur =>
+        val res = ctx.constructMonadic[Result, Param[ConfigReader, A]#PType] { param =>
+          val keyStr = hint.configKey(param.label)
+          objCur.atKeyOrUndefined(keyStr) match {
+            case keyCur if keyCur.isUndefined =>
+              param.default match {
+                case Some(defaultValue) if hint.useDefaultArgs => Right(defaultValue): Result[param.PType]
```
jcazevedo: Do we need to provide the type hint here?

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -0,0 +1,94 @@
...
+ * An object containing Magnolia `combine` and `dispatch` methods to generate `ConfigReader` instances.
+ */
+object MagnoliaConfigReader {
+
+  def combine[A](ctx: CaseClass[ConfigReader, A])(implicit hint: ProductHint[A]): ConfigReader[A] =
+    if (ctx.typeName.full.startsWith("scala.Tuple")) combineTuple(ctx)
+    else if (ctx.isValueClass) combineValueClass(ctx)
+    else combineCaseClass(ctx)
+
+  def combineCaseClass[A](ctx: CaseClass[ConfigReader, A])(implicit hint: ProductHint[A]): ConfigReader[A] = new ConfigReader[A] {
```
jcazevedo: I think we can make all these `combine*` methods private and only leave the `combine` and `dispatch` methods public.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
jcazevedo: I suppose we can also use the cursor on the `Left` instead of `cur`. It'd avoid an extra unnecessary conversion to a `ConfigObjectCursor`.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigWriter.scala
```diff
@@ -0,0 +1,52 @@
...
+ * An object containing Magnolia `combine` and `dispatch` methods to generate `ConfigWriter` instances.
+ */
+object MagnoliaConfigWriter {
+
+  def combine[A](ctx: CaseClass[ConfigWriter, A])(implicit hint: ProductHint[A]): ConfigWriter[A] =
+    if (ctx.typeName.full.startsWith("scala.Tuple")) combineTuple(ctx)
+    else if (ctx.isValueClass) combineValueClass(ctx)
+    else combineCaseClass(ctx)
+
+  def combineCaseClass[A](ctx: CaseClass[ConfigWriter, A])(implicit hint: ProductHint[A]): ConfigWriter[A] = new ConfigWriter[A] {
```
jcazevedo: As with the `MagnoliaConfigReader` object, I think we can make all these `combine*` methods private and only expose `combine` and `dispatch`.

jcazevedo (changes requested):
@ruippeixotog: Excellent work! The derivation code has much less boilerplate than its Shapeless counterpart.

It's very unfortunate that we currently don't have a way to easily accumulate errors using Magnolia.

Regarding the package organization, and taking into account that we don't seem to be able to have single `auto` and `semiauto` packages, have you considered having things the other way around? Instead of `reader.auto`, `reader.semiauto`, `writer.auto` and `writer.semiauto`, having `auto.reader`, `auto.writer`, `semiauto.reader` and `semiauto.writer`? At least from a semiauto perspective, it feels nicer to have a single `semiauto._` import and `reader.derive` and `writer.derive` calls.

I'm not sure if it's used frequently, but we provide a `deriveConvert` method in `pureconfig-generic`'s `semiauto` package. Have you considered porting it to the Magnolia module?

jcazevedo:
I want to try this in a project before getting it merged in. I haven't done it yet.

docs/src/main/tut/docs/library-integrations.md
ruippeixotog: Ups, missed a word. Will fix!

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
ruippeixotog: > This `combineCaseClass` is relying on the default `ProductHint` implementation. While it's very unlikely that we'll ever change it in a way that would break this, I think it'd be safer to provide an explicit `ProductHint` that we know won't mess with the field names.

Good point, I didn't think of that. The current generic derivation behavior is to fall back to reading the tuple exactly like you would read a case class (including hints), so the matching behavior here is actually to also propagate `ProductHint` from `combine` to `combineTuple`.

> Also, I think it's preferable to cache the call to combineCaseClass(ctx) outside the from method, to avoid creating a new ConfigReader on every from call.

This is a micro optimization where we're sacrificing extra persistent memory for a minimal increase in performance. We're trading a simple object instantiation (the only thing `combineCaseClass` does) to have a type class with an extra field in its object, something I don't usually consider worth it.

> I suppose we can also use the cursor on the Left instead of cur. It'd avoid an extra unnecessary conversion to a ConfigObjectCursor.

Sounds good :)

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
ruippeixotog: Looks like we don't, I may have needed this as I was developing and now it's not necessary.

ruippeixotog:
> Regarding the package organization, and taking into account that we don't seem to be able to have single `auto` and `semiauto` packages, have you considered having things the other way around? Instead of `reader.auto`, `reader.semiauto`, `writer.auto` and `writer.semiauto`, having `auto.reader`, `auto.writer`, `semiauto.reader` and `semiauto.writer`? At least from a semiauto perspective, it feels nicer to have a single `semiauto._` import and `reader.derive` and `writer.derive` calls.

Good idea, I didn't consider that but I also prefer the other way around :) I'll leave the method names unchanged though - `derive` has a lot of potential to cause name clashes within the rest of the module and with other dependencies. For example, users wouldn't be able to import both `auto.reader._` and `auto.writer._` without having to shadow one of them.

> I'm not sure if it's used frequently, but we provide a `deriveConvert` method in `pureconfig-generic`'s `semiauto` package. Have you considered porting it to the Magnolia module?

I did, but it's not straightforward to implement it. It would likely need to be generated by a macro, since I cannot call `deriveReader[A]` and `deriveWriter[A]` in normal functions when `A` is a generic parameter. It works on the shapeless version because we created a side type class (`DerivedConfigReader`) which we can require as parameter, something that doesn't work with Magnolia (we cannot configure the output type class to be different than the ones it requires for product fields / coproduct options).

jcazevedo (approved):
I just tried using this in a project that was using `pureconfig-generic`'s semiauto derivation (only for readers) and things seem to work well. I also got around 10% decrease in compile time using the Magnolia module.

ruippeixotog:
@jcazevedo that's great, thanks for testing this!

PR #612, opened 2019-10-08
Extract `ConfigReaderException` pretty printing logic to `ConfigReaderFailures`

### LINKED ISSUE #385: Move `ConfigReaderFailures` pretty printing logic away from `ConfigReaderException`
Here's my problem: I want to load a config in a safe way (without throwing exceptions) and log the errors in a pretty error messages.

Right now, it appears that the two best ways to do this are:
* Use `loadConfig[Config](config, "field")` and write my own `ConfigReaderFailures` pretty printer;
* Use `Try(loadConfigOrThrow[Config](config, "field"))` and log the exception message;

I think that the pretty printing logic could be moved to the `ConfigReaderFailures`.

It might also be nice to add a `Show` implementation to the cats module (or maybe not, not sure if this is out of the scope of that module).

JD557:
Extracts the bulk of the `ConfigReaderException` pretty printing logic to a `prettyPrint` method on `ConfigReaderFailures` (with configurable spacing).

Closes #385

### DIFF
--- a/core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
+++ b/core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
@@ -18,8 +18,8 @@ case class ConfigReaderFailures(head: ConfigReaderFailure, tail: List[ConfigRead
   def ++(that: ConfigReaderFailures): ConfigReaderFailures =
     new ConfigReaderFailures(head, tail ++ that.toList)
 
-  def prettyPrint(identLevel: Int = 0, tabSize: Int = 2): String = {
-    def tabs(n: Int): String = " " * ((identLevel + n) * tabSize)
+  def prettyPrint(identLevel: Int = 0, identSize: Int = 2): String = {
+    def tabs(n: Int): String = " " * ((identLevel + n) * identSize)
     def descriptionWithLocation(failure: ConfigReaderFailure, ident: Int): String = {
       val failureLines = failure.description.split("\n")
       (failure.location.fold(s"${tabs(ident)}- ${failureLines.head}")(f => s"${tabs(ident)}- ${f.description} ${failureLines.head}") ::
--- a/core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
+++ b/core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
@@ -3,6 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package pureconfig.error
 
+import scala.collection.mutable
+
 /**
  * A non-empty list of ConfigReader failures
  */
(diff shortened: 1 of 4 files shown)


core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
```diff
@@ -15,6 +17,37 @@ case class ConfigReaderFailures(head: ConfigReaderFailure, tail: List[ConfigRead
 
   def ++(that: ConfigReaderFailures): ConfigReaderFailures =
     new ConfigReaderFailures(head, tail ++ that.toList)
+
+  def prettyPrint(identLevel: Int = 0, tabSize: Int = 2): String = {
```
leifwickland: Nitpick: I'd prefer to rename `tabSize` to `indentSize`.

leifwickland (approved):
@JD557 Thanks for doing this. It's a change I've wanted for a while.

Did you check if it'd make sense to add any tests now that `prettyPrint` is accessible?

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
JD557: Renamed in 5f420d0

JD557:
>Did you check if it'd make sense to add any tests now that prettyPrint is accessible?

I think that most pretty printing logic is already tested by the `ConfigReaderExceptionSuite`.

Nevertheless, I added some basic tests in 023f219 to a new `ConfigReaderFailuresPrettyPrintSuite` that simply checks if the identation is correctly applied.

leifwickland:
Thanks for adding the test @JD557. We'll wait for another maintainer to approve before merging.

ruippeixotog (commented):
@JD557 looks good, thanks for doing this!

> I think that most pretty printing logic is already tested by the ConfigReaderExceptionSuite

Nit: because we've moved this logic to another class, I'd prefer that you also moved the pretty printing logic tests to `ConfigReaderFailuresPrettyPrintSuite`, as we're not really testing the exception anymore. What do you think?

jcazevedo (approved):
@JD557: Looks good! Thanks for this!

I agree with @ruippeixotog about moving the tests related to the pretty printing logic from `ConfigReaderExceptionSuite` to `ConfigReaderFailuresPrettyPrintSuite`.

In the scope of #385, it'd also be nice to have access to the string representation (with location) of individual `ConfigReaderFailure`s, but that can be left for future work.

JD557:
>Nit: because we've moved this logic to another class, I'd prefer that you also moved the pretty printing logic tests to ConfigReaderFailuresPrettyPrintSuite, as we're not really testing the exception anymore. What do you think?

Done in 6aef1c9 (I think I moved everything).

I left the first pretty print test `ConfigReaderExceptionSuite` on purpose, to check that the exception message is actually using the pretty printer.

ruippeixotog (approved):
Excellent, thanks again!

PR #613, opened 2019-10-10
Avoid unnecessarily removing keys from `ConfigObject`s in derived readers

jcazevedo:
This PR restores a behavior we had in the past and that is consistent with what we have commented. Somehow we changed it and were left with this very unnecessary if statement.


ruippeixotog (approved):
Huh, looks like [it's my fault](https://github.com/pureconfig/pureconfig/commit/326153ca8aa235a1c9702a61134a243114d29477#diff-79555c13fdfc48f437c5dbcfd26fc475L128) 😅

PR #614, opened 2019-10-10
Add classLoader parameter in ConfigSource.resources

### LINKED ISSUE #599: Allow passing ClassLoaders to the ConfigSource constructors that can use it
This was a part of #531, but didn't make it into #541.

`ConfigFactory.parseResources` and its derivatives that don't accept a `ClassLoader` use `Thread.getContextClassLoader` inside, which doesn't mix at all with the threadpool-heavy execution model that the Cats-Effect ecosystem provides, and with the `ClassLoader` wizardry that SBT does by default.

Case in point: https://gitter.im/melrief/pureconfig?at=5d8ddd7573dd29627ebe5ab2.

ruippeixotog:
Closes #599.

### DIFF
--- a/core/src/main/scala/pureconfig/ConfigSource.scala
+++ b/core/src/main/scala/pureconfig/ConfigSource.scala
@@ -277,7 +277,8 @@ object ConfigSource {
    * @param name the resource name
    * @return a config source that provides configs read from JVM resource files.
    */
-  def resources(name: String) = ConfigObjectSource(ConfigFactoryWrapper.parseResources(name))
+  def resources(name: String, classLoader: ClassLoader = null) =
+    ConfigObjectSource(ConfigFactoryWrapper.parseResources(name, classLoader))
 
   /**
    * Returns a config source that provides a config parsed from a string.
--- a/core/src/main/scala/pureconfig/ConfigSource.scala
+++ b/core/src/main/scala/pureconfig/ConfigSource.scala
@@ -272,12 +272,25 @@ object ConfigSource {
 
   /**
    * Returns a config source that provides configs read from JVM resource files. If multiple files
-   * are found, they are merged in no specific order.
+   * are found, they are merged in no specific order. This method uses Typesafe Config's default
+   * class loader (`Thread.currentThread().getContextClassLoader()`).
    *
    * @param name the resource name
    * @return a config source that provides configs read from JVM resource files.
    */
-  def resources(name: String, classLoader: ClassLoader = null) =
+  def resources(name: String) =
+    ConfigObjectSource(ConfigFactoryWrapper.parseResources(name, null))
+
+  /**
+   * Returns a config source that provides configs read from JVM resource files. If multiple files
+   * are found, they are merged in no specific order. The given class loader will be used to look
+   * for resources.
+   *
+   * @param name the resource name
+   * @param classLoader the class loader to use to look for resources
+   * @return a config source that provides configs read from JVM resource files.
+   */
+  def resources(name: String, classLoader: ClassLoader) =
     ConfigObjectSource(ConfigFactoryWrapper.parseResources(name, classLoader))
 
   /**
--- a/core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
+++ b/core/src/main/scala/pureconfig/backend/ConfigFactoryWrapper.scala
@@ -61,9 +61,9 @@ object ConfigFactoryWrapper {
       onIOFailure = Some(CannotReadFile(path, _)))
 
   /** @see `com.typesafe.config.ConfigFactory.parseResources()` */
-  def parseResources(resource: String): ConfigReader.Result[Config] =
+  def parseResources(resource: String, classLoader: ClassLoader = null): ConfigReader.Result[Config] =
     unsafeToReaderResult(
-      ConfigFactory.parseResources(resource, strictSettings),
+      ConfigFactory.parseResources(resource, strictSettings.setClassLoader(classLoader)),
       onIOFailure = Some(CannotReadResource(resource, _)))
 
   /** @see `com.typesafe.config.ConfigFactory.parseURL()` */


core/src/main/scala/pureconfig/ConfigSource.scala
```diff
@@ -277,7 +277,8 @@ object ConfigSource {
    * @param name the resource name
    * @return a config source that provides configs read from JVM resource files.
    */
-  def resources(name: String) = ConfigObjectSource(ConfigFactoryWrapper.parseResources(name))
+  def resources(name: String, classLoader: ClassLoader = null) =
```
nigredo-tori: We should probably make this an `Option[ClassLoader]`. Same goes for `ConfigFactoryWrapper.parseResources`.

nigredo-tori:
Note that this breaks binary compatibility.

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: I can change this 👍 I also need to add Scaladoc for this new parameter.

Out of curiosity, is there any reason other than personal API preference? `null` class loaders are [explicitly supported](https://github.com/lightbend/config/blob/master/config/src/main/java/com/typesafe/config/ConfigParseOptions.java#L212-L221) by `ConfigParseOptions` public API. I personally don't see any problem with a default value of `null` either and I find that wrapping default arguments in `Some` is cumbersome (as `None` is the natural default in most of them).

ruippeixotog:
> Note that this breaks binary compatibility.

Maintaining binary compatibility is not something we have as a goal when releasing new versions, currently 🙂

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: Alternatively, if it feels like a nicer API to you, I can overload this method so that we don't have an explicit `null` default parameter.

nigredo-tori: @ruippeixotog, `Option` requires the programmer to explicitly handle the empty value, while with `null` *almost anything* can be empty, and you can have a `NullPointerException` in *any method call*. So the general consensus in the Scala community is to convert everything nullable (that we mostly get from the Java APIs) to `Option` as soon as possible, and forget about [the billion-dollar mistake](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/) in the rest of your code. Even in Java we now have `Optional` for this, although it will be a long while before that catches on.

As for overloading... Both overloading and default arguments are slightly inconvenient (see the [WartRemover checks](https://www.wartremover.org/doc/warts.html)), but to me personally both are acceptable. However, `ConfigFactory.parseResources` has two optional parameters (`ClassLoader` and `ConfigParseOptions`), so overloading promises to become even messier in the future.

ruippeixotog: You gave me a generic explanation of the advantages of using `Option` - one that throughout my years programming in Scala I had the chance to both read and explain to others many times 🙂 However, let's focus on the specific case we have here: this `classLoader` parameter is passed directly to `ConfigFactoryWrapper` (a thin wrapper to `ConfigFactory`) as `ConfigParseOptions`, which explicitly supports `null` as a default value. We are consuming `null` values, not providing them; library users will never have to deal with them and on our side we only deal with it in a very controlled scope. On the other hand, as I library user I'd rather not to have to wrap every single optional parameter in a `Some` - it always felt artificial to me, as default arguments already introduce an optional component into the API. 

About overloading / default arguments: I agree that both are inconvenient. The approach I prefer in these cases is to define my own `Options` class and have it as a single arg. However, as I mentioned in the issue, we probably don't want any other field from `ConfigParseOptions` as an opiton here as many of those change semantics quite a bit. Let's consider that we'll have a single argument for the time being.

I'll change this to overloading so that no `null`s are present in the public API at all (even if I consider this a non-issue).

nigredo-tori: > We are consuming null values, not providing them; library users will never have to deal with them and on our side we only deal with it in a very controlled scope.

In general, when we have a method that does something different when one of its arguments is `null`, we're asking of the downstream code to be aware of that. Default arguments can mask the issue, but they never get rid of it entirely. Here, the code that always passes a known class loader, or never passes a class loader at all won't need to spell that out, but anything that tries to pass an optional class loader would need to do `.orNull`. That would also make running without a class loader less discoverable, since it would not be obvious from the function signature.

Even more generally, using `Option` is a known and long-standing convention. I don't say every API must adhere to it (although most should), but those that do need to do so **consistently**. It may seem OK to "cut corners" with `null` in one particular case, but in doing so you break the consistency - now the user has to check every value in your API for nullability, and pray that that is mentioned somewhere in the ScalaDoc. I'm looking at you, `ConfigCursor.value`.

Basically, I have zero-tolerance policy for `null` in my projects (enforced via WartRemover), and I feel that for this convention to be effective we need to fully adhere to it at least in our APIs.

nigredo-tori (approved):

core/src/main/scala/pureconfig/ConfigSource.scala
ruippeixotog: > but anything that tries to pass an optional class loader would need to do `.orNull`

I'd prefer that people specified `Thread.currentThread().getContextClassLoader()` explicitly instead of passing `null`, TBH. Class loaders were never a problem in my production projects, but it seems that they can lead to a some unexpected and undesirable results - if that happens, I'd prefer if people were fully aware of what class loader they are really using. I hope that changing to overloaded methods and the extra Scaladoc makes that intention clearer.

> now the user has to check every value in your API for nullability, and pray that that is mentioned somewhere in the ScalaDoc. I'm looking at you, ConfigCursor.value.

If you find the time, could you please open an issue listing the places where you noticed that we can return `null` values? `ConfigCursor.value` is a good one to note - it's public [for compatibility/historical reasons](https://github.com/pureconfig/pureconfig/pull/314#issuecomment-335279085), but many things changed since then and we should be in a much better position now to just deprecate and hide that field (probably with a `getValueUnsafe`-like escape hatch).

jcazevedo (approved):

PR #616, opened 2019-10-12
Skip publish of `pureconfig-generic-base`

jcazevedo:
This PR enables skipping the publish of `pureconfig-generic-base`.


jcazevedo:
Nevermind. This is required.

Issue #356, opened 2018-01-16
Make FirstSuccessCoproductHint NoValidCoproductChoiceFound error user friendly


yarulan:
```scala
sealed trait Shape
case class Circle(radius: Int) extends Shape
case class Square(side: Int) extends Shape
implicit val hint = new FirstSuccessCoproductHint[Shape]
println(pureconfig.loadConfig[Shape](ConfigFactory.parseString("")))
```
The error says `ConvertFailure(NoValidCoproductChoiceFound(SimpleConfigObject({})),None,)` which doesn't explain much. It would be better if it contained all possible choices, like
```
missing radius
  or
missing side
```

ruippeixotog:
Hi @yarulan! Improving the default error messages for coproducts is a good idea, not only with `FirstSuccessCoproductHint` but also with other hints. Unfortunately, right now it's not trivial to obtain the information you mentioned when the error is created, but a small refactoring to allow that would be nice.

PR #629, opened 2019-10-28
Make `ProductHint`s and `CoproductHint`s more general

jcazevedo:
This PR proposes a refactoring to `ProductHint`s and `CoproductHint`s in order to make them more general. The idea behind this is to allow better customization of derived readers and writers, regardless of the derivation code used to produce them. Incidentally, this moves a lot of repeated code between shapeless-based and magnolia-based derivation to the hints. This also enables #356 (which I leave for a future PR).

This is obviously not binary compatible with older versions of PureConfig. However, only custom implementations of `ProductHint` (not created via `ProductHint.apply`) and `CoproductHint` need to be migrated.

This also modifies the derivation of readers for case classes in `pureconfig-magnolia` to accumulate errors, at the expense of using the unsafe `rawConstruct`. I ended up doing so because the `constructMonadic` interface of Magnolia doesn't play very well with the proposed `ProductHint` interface of returning both the result of reading a field and the next `ConfigObjectCursor` to use.


modules/generic-base/src/main/scala/pureconfig/generic/CoproductHint.scala
```diff
@@ -57,6 +36,19 @@ trait CoproductHint[T] {
...
+   * with its representation, possibly encoded with disambiguation information.
+   *
+   * @param writer the `ConfigWriter` in scope for type `A`
+   * @param name the name of the class or coproduct option
+   * @param value the value of the coproduct option
+   * @tparam A the type of the class or coproduct option
+   * @return the config for the sealed family or coproduct wrapped in a `Right`, or a `Left` with the failure if some
+   *         error occurred.
+   */
+  def to[A <: T](writer: ConfigWriter[A], name: String, value: A): ConfigReader.Result[ConfigValue]
```
ruippeixotog: Same thing here regarding `writer`: what are the advantages of providing `writer` and `value` here? All our three hint types call `writer.to(value)` and it seems natural to do so - if other behavior is needed it would probably be more elegant for users provided a custom writer for `A` instead.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -67,27 +52,15 @@ object MagnoliaConfigReader {
...
-          }
-        case (res, _) => res
-      }
-      res.flatMap {
-        case Some(a) => Right(a)
-        case None => Left(hint.noOptionFound(cur))
+    def from(cur: ConfigCursor): ConfigReader.Result[A] = {
+      ctx.subtypes.foldRight[ConfigReader.Result[A]](Left(hint.noOptionFound(cur))) {
+        case (subtype, rest) =>
+          hint.from(cur, subtype.typeclass, subtype.typeName.short, rest)
```
ruippeixotog: Even though `rest` is lazy this implementation does not make use of it, which means that `hint.from` will always be called with all subtypes regardless of what the hint does.

modules/generic-base/src/main/scala/pureconfig/generic/CoproductHint.scala
```diff
@@ -14,39 +14,18 @@ import pureconfig.syntax._
...
-
-  /**
-   * Defines what to do if `from` returns `Success(Some(_))` for a class or coproduct option, but its `ConfigConvert`
-   * fails to deserialize the config.
-   *
-   * @param name the name of the class or coproduct option
-   * @return `true` if the next class or coproduct option should be tried, `false` otherwise.
-   */
-  def tryNextOnFail(name: String): Boolean
+  def from[A <: T](cur: ConfigCursor, reader: ConfigReader[A], name: String, rest: => ConfigReader.Result[T]): ConfigReader.Result[T]
```
ruippeixotog: I don't like this new API very much, for two reasons:
- We're changing this to a kind of continuation-passing style, which seems like a clumsy interface for users (why do you need to use specifically that value at the end to fall back?) besides allowing more than one possible implementation for the same behavior (can we do `rest.left.flatMap(_ => /* try reading this subtype */)` instead of the canonical way? Which one is more efficient?). I always thought of derivation code as owning the control flow, the "algorithm" to try the different options.
It can prevent some duplicated logic in both derivation libs, but it did it at the cost of adding extra logic at the end of each hint; and between the two I definitely prefer repeating code in derivation.
- I don't think we need to pass subtype readers or writers to hints, in general - once more, I see this as a responsibility of derivation code, giving hints only a way to transform inputs/outputs of reading/writing. Note that if some users really need to call different readers/writers, it seems much more natural to provide custom readers/writers for that purpose.

It seems that a second motivation to change this, besides moving code from derivations to hints, might have been to get rid of that weird `tryNextOnFail` function. I like that idea, but that can be accomplished more nicely by replacing the `Option[ConfigCursor]` in the return type with a richer type that can encode `Found(cur)`, `Tentative(cur)` and `Continue`.

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
```diff
@@ -10,30 +12,80 @@ import pureconfig._
...
+   *
+   * @param writer the `ConfigWriter` currently in scope for `A`
+   * @param fieldName the name of the field of type `A` in `T`
+   * @param value the current value of `fieldName`
+   * @tparam A the type of the field to be written
+   * @return an optional key-value pair to be used in the `ConfigObject` representation of `T`. If `None`, the field is
+   *         omitted from the `ConfigObject` representation.
    */
-  def allowUnknownKeys: Boolean
+  def to[A](writer: ConfigWriter[A], fieldName: String, value: A): Option[(String, ConfigValue)]
```
ruippeixotog: Ditto: consider replacing `writer` and `value` by an `Option[ConfigValue]`.

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
```diff
@@ -10,30 +12,80 @@ import pureconfig._
...
-   * @return the key in the config object associated with the given case class field.
+   * @param cur the cursor from which to read a value
+   * @param reader the `ConfigReader` currently in scope for `A`
+   * @param fieldName the name of the field of type `A` in `T`
+   * @param default the default value for the field with name `fieldName` in `T`
+   * @tparam A the type of the field to be read
+   * @return a tuple with both the result of reading the field and the cursor to supply to subsequent reads.
    */
-  def configKey(fieldName: String): String
+  def from[A](cur: ConfigObjectCursor, reader: ConfigReader[A], fieldName: String, default: Option[A]): (ConfigReader.Result[A], ConfigObjectCursor)
```
ruippeixotog: As an alternative way of defining this, consider not providing `reader` and `default` as arguments and instead providing a richer return type (e.g. `Use(ConfigCursor)`, `UseDefault`). 

This is a good example of providing good flexibility without giving away control flow: if we were to do that here we would have included a `rest: (A, ConfigObjectCursor) => ConfigReader.Result[A]` to keep it consistent with `CoproductHint`).

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
```diff
@@ -10,30 +12,80 @@ import pureconfig._
...
 
   /**
-   * A boolean indicating if the default arguments of the case class should be used when fields are missing
+   * Returns optional failures if the provided `ConfigObjectCursor` resulted from reading all fields necessary to
```
ruippeixotog: Did you mean "Returns optional failures given the `ConfigObjectCursor`"? The sentence seems weird as it is right now.

modules/generic-base/src/main/scala/pureconfig/generic/CoproductHint.scala
ruippeixotog: I know this was like this before, but why do we return a `ConfigReader.Result` here if the write path does not handle errors? Since we are changing the API, shouldn't we drop `Result` here and throw exceptions on hints if needed? As it is it may be misleading to users.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -17,37 +18,21 @@ object MagnoliaConfigReader {
...
-        else {
-          val usedKeys = ctx.parameters.map { param => hint.configKey(param.label) }.toSet
-          val unknownKeyFailures = objCur.map.collect {
-            case (k, keyCur) if !usedKeys(k) => keyCur.failureFor(UnknownKey(k))
-          }.toList
 
-          if (unknownKeyFailures.isEmpty) res
-          else Left(ConfigReaderFailures(unknownKeyFailures.head, unknownKeyFailures.tail))
-        }
+        res.right.flatMap(values => hint.bottom(obj).fold[ConfigReader.Result[A]](Right(ctx.rawConstruct(values.result())))(Left.apply))
```
ruippeixotog: If `res` is a `Left` and `hint.bottom(obj)` is `Some`, we should return all failures and not just the ones from `res`. That's the current behavior of `pureconfig-generic` and it seems to be the one that makes more sense.

I see that the previous implementation also has this bug; can you add a test case for this?

modules/generic/src/main/scala/pureconfig/generic/MapShapedReader.scala
```diff
@@ -11,7 +10,7 @@ import shapeless.labelled.{ FieldType, field }
  * @tparam Wrapped the original type for which `Repr` is a generic sub-representation
  * @tparam Repr the generic representation
  */
-trait MapShapedReader[Wrapped, Repr] extends ConfigReader[Repr]
+trait MapShapedReader[Wrapped, Repr] extends ConfigReader[Wrapped]
```
ruippeixotog: Why did you change this parameter? `MapShapedReader` is supposed to be a `ConfigReader[Repr]`; `Wrapped` is here just to type the product hint.

ruippeixotog (changes requested):
Overall looks good @jcazevedo, thanks for doing this! I really like the overhaul of `ProductHint`, it puts it on par with `CoproductHint` while keeping a simplified implementation for users to use.

I have some concerns with generalizing excessively an API that we provide for users to extend (uncommon as it may be). Some of the changes in hints seem a bit confusing and give it too much power to change how the derivation is made - to a point where for certain implementations it may not be recognizable as a derivation anymore. I understood the intention to avoid repeating code between different derivation libs and it worked brilliantly in products, but IMO we can leave control flow in derivation libs (e.g. by avoiding things like `rest` in `CoproductHint`). I left some specific comments inline.

If you did some of these choices to enable new use cases and not just for removing duplicated code, can you provide some interesting examples of how the new capabilities could be used?

ruippeixotog:
> This also modifies the derivation of readers for case classes in pureconfig-magnolia to accumulate errors, at the expense of using the unsafe rawConstruct. I ended up doing so because the constructMonadic interface of Magnolia doesn't play very well with the proposed ProductHint interface of returning both the result of reading a field and the next ConfigObjectCursor to use.

This worries me a bit, because it seems that it won't work with the upcoming error accumulating constructor for Magnolia (https://github.com/propensive/magnolia/pull/184) - the current design forces derivations to either follow a recursive approach or use mutable variables. I'm ok with staying with `rawConstruct` if we don't find any good alternative, but if we can think of one we should probably go for it.

jcazevedo:
@ruippeixotog: Thanks for your thorough review, and sorry for taking so long to address your comments.

I revamped both `ProductHint`s and `CoproductHint`s from my initial proposal, following your suggestion of keeping the control flow in derivation code. The `from` method of `ProductHint` and `CoproductHint` now returns a `FieldHint` and a `ChoiceHint`, respectively (I'm not completely sold on these names). These hints signal intent on what derivation code should do.

> If you did some of these choices to enable new use cases and not just for removing duplicated code, can you provide some interesting examples of how the new capabilities could be used?

Part of the motivation for the `rest` parameter in `CoproductHint` was to allow access to all failures when reading a sealed family (in order to tackle #356). I changed some things in my latest proposal, while also enabling that use case. The `bottom` method of `CoproductHint` now receives a list of all attempted reads and their corresponding failures, allowing better error messages for derived coproduct readers. In order to support that, the `MapShapedReader` implementation changed so that attempts in coproduct reads are accumulated until we reach the bottom type. I suppose it's OK to do that given the "private" nature of `MapShapedReader` and its isolated use in derivation code.

modules/generic-base/src/main/scala/pureconfig/generic/CoproductHint.scala
jcazevedo: > I know this was like this before, but why do we return a `ConfigReader.Result` here if the write path does not handle errors? Since we are changing the API, shouldn't we drop `Result` here and throw exceptions on hints if needed? As it is it may be misleading to users.

I dropped the `Result` and started throwing `CoproductHintException`s in 977aacf323b3d687bd458943595a01d97c8cc7b9. A `CoproductHintException` wraps a single `FailureReason`, which is enough for our failure scenarios in hints.

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
jcazevedo: Yeah. That was weird. Improved it in 8f5bf8d.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
jcazevedo: I fIxed that in `MagnoliaConfigReader` and also add a test case in 76d35fc.

modules/generic/src/main/scala/pureconfig/generic/MapShapedReader.scala
jcazevedo: This change was partly motivated by the need to provide `ConfigReader`s of the `Wrapped` type to `CoproductHint`, but this is not relevant anymore, given the current proposal.

jcazevedo:
> This worries me a bit, because it seems that it won't work with the upcoming error accumulating constructor for Magnolia (propensive/magnolia#184) - the current design forces derivations to either follow a recursive approach or use mutable variables. I'm ok with staying with `rawConstruct` if we don't find any good alternative, but if we can think of one we should probably go for it.

With the latest changes, I was able to drop `rawConstruct` and start using `constructEither` from Magnolia. The behavior is slightly different from the shapeless implementation, since keys are only removed from the `ConfigObject` after all fields are read, whereas in the shapeless implementation they are removed after reading each field. I think both implementations should have the same end result unless the provided field mapping is non-injective, in which case other types of problems should show up first.

modules/generic-base/src/main/scala/pureconfig/generic/CoproductHint.scala
```diff
@@ -14,49 +15,38 @@ import pureconfig.syntax._
...
-   *
-   * @param cur a `ConfigCursor` at the sealed family option
-   * @return a non empty list of reader failures.
+   * @param cv the `ConfigValue` of the class or coproduct option
+   * @return the config for the sealed family or coproduct wrapped in a `Right`, or a `Left` with the failure if some error
+   *         occurred.
    */
-  def noOptionFound(cur: ConfigCursor): ConfigReaderFailures =
-    ConfigReaderFailures(cur.failureFor(NoValidCoproductChoiceFound(cur.value)))
+  def to(name: String, cv: ConfigValue): ConfigValue
```
ruippeixotog: Can we switch the two arguments back to their original positions? This would make it consistent with `from` and result in an easier migration.

modules/generic-base/src/main/scala/pureconfig/generic/CoproductHint.scala
```diff
@@ -78,34 +68,35 @@ class FieldCoproductHint[T](key: String) extends CoproductHint[T] {
...
-        WrongType(cv.valueType, Set(ConfigValueType.OBJECT)),
-        ConfigValueLocation(cv), "")))
+    } yield if (valueStr == fieldValue(name)) Use(objCur.withoutKey(key)) else Skip).right.getOrElse(Skip)
   }
 
-  def tryNextOnFail(name: String) = false
+  override def bottom(cur: ConfigCursor, attempts: List[(String, ConfigReaderFailures)]): ConfigReaderFailures =
+    attempts match {
+      case h :: _ => h._2
+      case _ => (for {
```
ruippeixotog: This structure is rather hacky; we are relying on the implementation of `from` in that we know that 1) the only way no attempts were made was if there's something wrong with the key, 2) we'll only try at most one key.

I think this is not really a case for `bottom`, but because `from` does not return an `Either` anymore we have no way of saying "an error occurred" (when I suggested moving to a richer type I did not mean getting rid of the `Result` part). What do you think of allowing `from` calls to signal an error and stopping the process? That can be either by wrapping in a `Result` (my preference, for consistency with the rest of the lib) or adding a 4th class `Error` to `ChoiceHint`.

modules/generic-base/src/main/scala/pureconfig/generic/CoproductHint.scala
```diff
@@ -78,34 +68,35 @@ class FieldCoproductHint[T](key: String) extends CoproductHint[T] {
...
+        objCur <- cur.asObjectCursor.right
+        valueCur <- objCur.atKey(key).right
+        _ <- valueCur.asString.right
+      } yield ConfigReaderFailures(valueCur.failureFor(UnexpectedValueForFieldCoproductHint(valueCur.value)))).fold(identity, identity)
```
ruippeixotog: Nit: it's really confusing to understand for comprehensions wrapped in parentheses with methods chained afterwards, as it makes for huge lines like this one and indentation is not well handled. When that happens, consider refactoring the code or using auxiliary variables.

modules/generic-base/src/main/scala/pureconfig/generic/CoproductHint.scala
```diff
@@ -78,34 +68,35 @@ class FieldCoproductHint[T](key: String) extends CoproductHint[T] {
...
+    }
 
-  override def noOptionFound(cur: ConfigCursor): ConfigReaderFailures =
-    cur.fluent.at(key).cursor.fold(
-      identity,
-      cur => ConfigReaderFailures(cur.failureFor(UnexpectedValueForFieldCoproductHint(cur.value))))
+  def to(name: String, cv: ConfigValue): ConfigValue = {
+    cv match {
+      case co: ConfigObject =>
+        if (co.containsKey(key))
```
ruippeixotog: Nit: consider merging this `if` into the pattern match above for better readability.

modules/generic-base/src/main/scala/pureconfig/generic/CoproductHint.scala
```diff
@@ -117,11 +108,42 @@ object FieldCoproductHint {
...
+  def to(name: String, cv: ConfigValue): ConfigValue =
+    cv
 }
 
 object CoproductHint {
+
+  /**
+   * A hint on what should be done when reading a given coproduct option.
+   */
+  sealed trait ChoiceHint
```
ruippeixotog: May I suggest `Action` or `Decision` instead of anything involving `Hint`? This has very little in common with `ProductHint`s and `CoproductHint`s, which may be confusing to users.

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
```diff
@@ -10,34 +13,73 @@ import pureconfig._
...
+   * Returns an optional key-value pair that should be used for the field with name `fieldName` in the `ConfigObject`
+   * representation of `T`.
+   *
+   * @param fieldName the name of the field in `T`
+   * @param value the optional serialized value of the field
+   * @return an optional key-value pair to be used in the `ConfigObject` representation of `T`. If `None`, the field is
+   *         omitted from the `ConfigObject` representation.
    */
-  def allowUnknownKeys: Boolean
+  def to(fieldName: String, value: Option[ConfigValue]): Option[(String, ConfigValue)]
```
ruippeixotog: Same thing here wrt swapping args.

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
```diff
@@ -10,34 +13,73 @@ import pureconfig._
...
 
+  /**
+   * A hint on what should be used when attempting to read a given field from a product.
+   *
+   * @param cursor the cursor to use when reading the field
+   * @param field the name of the field in the `ConfigObject` representation of the product
+   * @param remove whether the field should be removed from the object after read
+   * @param useDefault whether to use default values when reading this field if the cursor is undefined
+   */
+  case class FieldHint(cursor: ConfigCursor, field: String, remove: Boolean, useDefault: Boolean)
```
ruippeixotog: Ditto here about the name. `Action`/`Decision` still sound good to me here.

It seems strange to pass all this metadata together with the real purpose of the hint (the `cursor` to be used to read the key). I wonder if we should rethink the assumption that all field mappings are non-injective and that we have to _remove_ the config keys as they are read - this makes things a bit confusing any not parallelizable (in the sense that we cannot use the power of `Applicative` to process fields independently and join errors while keeping the same behavior).

Semantically, `!allowUnknownKeys` only requires that at the end no keys were unused. Can't we just have this `FieldHint` return a list of `usedFields` (instead of `field`+`remove`), which are either passed to `bottom` or used to filter the config object before calling `bottom`? This paves the way for hints that read fields multiple times (e.g. case classes having multiple fields with different representations of the same config value) and hints that read several config keys for a single field (e.g. reading a number and a unit to form a [coulomb](https://github.com/erikerlandson/coulomb) dimension).

Regarding `useDefault`: just like we have different actions like `Attempt` and `Skip`, I wonder if we shouldn't have a `UseDefault` or `UseOrDefault(ConfigCursor)` here for consistency? This is more a style concern, I don't see any immediate advantage.

modules/generic/src/main/scala/pureconfig/generic/MapShapedReader.scala
```diff
@@ -71,30 +67,34 @@ object MapShapedReader {
...
 
-          case Right(None) => tConfigReader.value.from(cur).right.map(s => Inr(s))
-          case l: Left[_, _] => l.asInstanceOf[ConfigReader.Result[FieldType[Name, V] :+: T]]
+        coproductHint.from(cur, vName.value.name) match {
+          case Use(optCur) =>
+            vReader.from(optCur)
+              .right.map(v => Inl(field[Name](v)))
+              .left.map(failures => coproductHint.bottom(cur, attempts :+ (vName.value.name -> failures)))
+          case Attempt(optCur) =>
+            vReader.from(optCur)
```
ruippeixotog: Can you please extract this to an auxiliary function?

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -17,37 +18,44 @@ object MagnoliaConfigReader {
...
-                case _ => cur.failed(KeyNotFound.forKeys(keyStr, objCur.keys))
-              }
-            case keyCur =>
-              param.typeclass.from(keyCur)
+        val hints = ctx.parameters.map { param => param.label -> hint.from(objCur, param.label) }.toMap
+
+        val res = ctx.constructEither[ConfigReaderFailures, Param[ConfigReader, A]#PType] { param =>
+          val fieldHint = hints(param.label)
+          lazy val reader = param.typeclass
+          lazy val keyNotFoundFailure = cur.failed(KeyNotFound.forKeys(fieldHint.field, objCur.keys))
```
ruippeixotog: Why are we creating this here instead of creating it inline on `case _ =>`?

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -17,37 +18,44 @@ object MagnoliaConfigReader {
...
-          if (unknownKeyFailures.isEmpty) res
-          else Left(ConfigReaderFailures(unknownKeyFailures.head, unknownKeyFailures.tail))
+        val filteredObj = hints.foldLeft(objCur) {
+          case (currObj, (_, FieldHint(_, field, true, _))) =>
+            currObj.withoutKey(field)
+          case (currObj, _) =>
+            currObj
         }
+
+        hint.bottom(filteredObj).fold(
```
ruippeixotog: This can be greatly simplified to `ConfigReader.Result.zipWith(res, hint.bottom(filteredObj).toLeft(()))((r, _) => r)`.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -17,37 +18,44 @@ object MagnoliaConfigReader {
...
-          val unknownKeyFailures = objCur.map.collect {
-            case (k, keyCur) if !usedKeys(k) => keyCur.failureFor(UnknownKey(k))
-          }.toList
+        }.left.map(_.reduce(_ ++ _))
 
-          if (unknownKeyFailures.isEmpty) res
-          else Left(ConfigReaderFailures(unknownKeyFailures.head, unknownKeyFailures.tail))
+        val filteredObj = hints.foldLeft(objCur) {
+          case (currObj, (_, FieldHint(_, field, true, _))) =>
+            currObj.withoutKey(field)
```
ruippeixotog: Now that we don't need any of the intermediate results and in most cases this config object will end up empty, it's more efficient and cleaner if we collect all used keys and create a single new config object with all of them filtered out. We can create such a method on `ConfigObject` to support this (probably `filterKeys`).

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -67,28 +75,33 @@ object MagnoliaConfigReader {
...
-        case (Right(None), subtype) =>
-          hint.from(cur, subtype.typeName.short) match {
-            case Right(Some(optCur)) =>
-              subtype.typeclass.from(optCur).map(Some.apply) match {
-                case Left(_) if hint.tryNextOnFail(subtype.typeName.short) => Right(None)
-                case res => res
+    def from(cur: ConfigCursor): ConfigReader.Result[A] = {
+      val (_, res) = ctx.subtypes.foldLeft[(Boolean, Either[List[(String, ConfigReaderFailures)], A])]((true, Left(Nil))) {
+        case (acc, subtype) =>
+          acc match {
```
ruippeixotog: Is there any reason why you split this pattern match in two nested ones? I find it easier to read if you have a single matching level, and if you put `case ((true, Left(attempts)), subtype) =>` as your first case you only have `case (acc, _) => acc` as your second one, just like we had before.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -67,28 +75,33 @@ object MagnoliaConfigReader {
...
+            case (false, _) => acc
+            case (true, Right(_)) => acc
+            case (true, Left(attempts)) =>
+              val typeName = subtype.typeName.short
+              hint.from(cur, typeName) match {
+                case Use(cur) =>
+                  subtype.typeclass.from(cur).fold(
```
ruippeixotog: Ditto here regarding moving to an aux function.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigWriter.scala
```diff
@@ -22,12 +21,13 @@ object MagnoliaConfigWriter {
   private def combineCaseClass[A](ctx: CaseClass[ConfigWriter, A])(implicit hint: ProductHint[A]): ConfigWriter[A] = new ConfigWriter[A] {
     def to(a: A): ConfigValue = {
       val fieldValues = ctx.parameters.map { param =>
-        param.typeclass match {
-          case tc: WritesMissingKeys[param.PType @unchecked] =>
-            tc.toOpt(param.dereference(a)).map(hint.configKey(param.label) -> _)
-          case tc =>
-            Some(hint.configKey(param.label) -> tc.to(param.dereference(a)))
+        val valueOpt = param.typeclass match {
+          case w: WritesMissingKeys[param.PType @unchecked] =>
```
ruippeixotog: Any reason why you renamed this? (just out of curiosity)

modules/generic/src/main/scala/pureconfig/generic/MapShapedReader.scala
```diff
@@ -11,7 +13,9 @@ import shapeless.labelled.{ FieldType, field }
  * @tparam Wrapped the original type for which `Repr` is a generic sub-representation
  * @tparam Repr the generic representation
  */
-trait MapShapedReader[Wrapped, Repr] extends ConfigReader[Repr]
+private[generic] trait MapShapedReader[Wrapped, Repr] {
```
ruippeixotog: If we are making this private, we should be consistent and make all other related ones, too (`Seq` and `Map` shaped readers and writers).

modules/generic/src/main/scala/pureconfig/generic/MapShapedReader.scala
```diff
@@ -48,21 +46,19 @@ object MapShapedReader {
...
-        case keyCur if keyCur.isUndefined =>
-          default.head match {
-            case Some(defaultValue) if hint.useDefaultArgs => Right(defaultValue)
-            case _ if headReader.isInstanceOf[ReadsMissingKeys] => headReader.from(keyCur)
-            case _ => cur.failed(KeyNotFound.forKeys(keyStr, cur.keys))
-          }
-        case keyCur => headReader.from(keyCur)
+      val fieldHint = hint.from(cur, fieldName)
+      lazy val reader = vFieldReader.value.value
+      lazy val keyNotFoundFailure = cur.failed[V](KeyNotFound.forKeys(fieldHint.field, cur.keys))
```
ruippeixotog: Why are we creating this here instead of creating it inline on `case _ =>`?

modules/generic-base/src/main/scala/pureconfig/generic/error/CoproductHintException.scala
```diff
@@ -0,0 +1,7 @@
+package pureconfig.generic.error
+
+import pureconfig.error.FailureReason
+
+final case class CoproductHintException(failure: FailureReason) extends RuntimeException {
```
ruippeixotog, line 10: Can you add Scaladoc to this?

ruippeixotog (changes requested):

jcazevedo:
@ruippeixotog: Sorry for taking so long to get back to this. I think I addressed all your comments, but I ended up changing some other things I wasn't very happy with.

I wasn't very happy with how the actions of `CoproductHint` were dependent on the order the coproduct options were visited. Even though both Shapeless and Magnolia guarantee a stable ordering of coproduct options, we have no guarantees that the order will be the same between them. To take this into account, I modified `CoproductHint`'s `from` method to receive the names of the coproduct options. Implementations of this method can then choose to use one of the options (with `CoproductHint.Use`) or attempt various options, in a defined order (with `Coproduct.Attempt`). 

To implement this, I introduced the `CoproductReaderOptions` type class to collect `ConfigReader`s for all coproduct options of a given type, indexed by their name. This also stops considering `ConfigReader`s for coproducts as `MapShapedReader`s, which is also more accurate. In order to be consistent in the writing side, I introduced the `CoproductConfigWriter` type class, so that `ConfigWriter`s for coproducts stop being `MapShapedWriter`s.

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
```diff
@@ -10,34 +12,101 @@ import pureconfig._
...
+    else
+      ProductHint.Use(keyCur, keyStr)
+  }
+
+  def bottom(cursor: ConfigObjectCursor, usedFields: Set[String]): Option[ConfigReaderFailures] = {
+    lazy val unknownKeys = cursor.map.toList.collect {
+      case (k, keyCur) if !usedFields.contains(k) =>
+        keyCur.failureFor(UnknownKey(k))
+    }
+    if (allowUnknownKeys || unknownKeys.isEmpty)
```
ruippeixotog: nit: can we use pattern matching on `unknownKeys` instead of `.isEmpty`/`.head`/`.tail`? I'd prefer to avoid these and lazy vals, which make the flow fuzzier.

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
```diff
@@ -10,34 +12,101 @@ import pureconfig._
...
+  /**
+   * An action to use the provided `ConfigCursor` when trying to read a given field.
+   *
+   * @param cursor the `ConfigCursor` to use when trying to read the field
+   * @param field the name of the field in the `ConfigObject` representation of the product
+   */
+  case class Use(cursor: ConfigCursor, field: String) extends Action
+
+  /**
+   * An action to either use the provided `ConfigCursor` (if it isn't null nor undefined) or fallback to the default
```
ruippeixotog:
```suggestion
   * An action to either use the provided `ConfigCursor` (if it isn't null or undefined) or fallback to the default
```

modules/generic-base/src/main/scala/pureconfig/generic/error/MissingCoproductChoice.scala
```diff
@@ -0,0 +1,12 @@
+package pureconfig.generic.error
+
+import pureconfig.error.FailureReason
+
+/**
+ * A failure reason given when a given coproduct option is missing.
+ *
+ * @param option the coproduct option that is missing
+ */
+final case class MissingCoproductChoice(option: String) extends FailureReason {
```
ruippeixotog: nit: consider renaming this `InvalidCoproductOption` or `InvalidCoproductOptionChosen` - we should be consistent with the term "coproduct option" to refer to subtypes here and I prefer "invalid" because it makes it clear that the chosen option is not supposed to be there in the first place (i.e. that this is a bug on the hint implementation).

modules/generic-base/src/main/scala/pureconfig/generic/error/MissingCoproductChoice.scala
```diff
@@ -0,0 +1,12 @@
...
+  def description = s"Missing coproduct choice '$option''"
```
ruippeixotog: I'd also add here that this is likely to be a bug on the coproduct hint implementation, as the only way this can happen is if users violate the contract of `CoproductHint#from`. As it is now the description is too broad.

modules/generic/src/main/scala/pureconfig/generic/EnumCoproductHint.scala
```diff
@@ -1,8 +1,10 @@
 package pureconfig.generic
 
-import com.typesafe.config.{ ConfigFactory, ConfigObject, ConfigValue, ConfigValueType }
+import com.typesafe.config.{ ConfigObject, ConfigValue, ConfigValueType }
 import pureconfig._
 import pureconfig.error._
+import pureconfig.generic.CoproductHint.Use
+import pureconfig.generic.error.{ CoproductHintException, NoValidCoproductChoiceFound }
```
ruippeixotog: I noticed now that we already had a `NoValidCoproductChoiceFound` failure. We can take this opportunity to rename it to "option", if you want.

modules/generic/src/main/scala/pureconfig/generic/EnumCoproductHint.scala
```diff
@@ -22,18 +24,20 @@ class EnumCoproductHint[T] extends CoproductHint[T] {
...
-    case _ => Left(ConfigReaderFailures(ConvertFailure(
-      WrongType(cv.valueType, Set(ConfigValueType.OBJECT)), ConfigValueLocation(cv), "")))
-  }
-
-  def tryNextOnFail(name: String) = false
+  def to(value: ConfigValue, name: String): ConfigValue =
+    value match {
+      case co: ConfigObject if co.isEmpty => fieldValue(name).toConfig
+      case _: ConfigObject =>
+        throw new CoproductHintException(NonEmptyObjectFound(name))
```
ruippeixotog:
```suggestion
        throw CoproductHintException(NonEmptyObjectFound(name))
```

modules/generic/src/main/scala/pureconfig/generic/EnumCoproductHint.scala
```diff
@@ -22,18 +24,20 @@ class EnumCoproductHint[T] extends CoproductHint[T] {
...
+      case cv =>
+        throw new CoproductHintException(WrongType(cv.valueType, Set(ConfigValueType.OBJECT)))
```
ruippeixotog:
```suggestion
        throw CoproductHintException(WrongType(cv.valueType, Set(ConfigValueType.OBJECT)))
```

modules/generic/src/main/scala/pureconfig/generic/MapShapedReader.scala
```diff
@@ -1,100 +1,55 @@
...
-          case l: Left[_, _] => l.asInstanceOf[ConfigReader.Result[FieldType[Name, V] :+: T]]
+    vConfigReader: Derivation[Lazy[ConfigReader[V]]],
+    tConfigReader: Lazy[MapShapedReader[Wrapped, T, U]],
+    hint: ProductHint[Wrapped]): MapShapedReader[Wrapped, FieldType[K, V] :: T, Option[V] :: U] =
+    new MapShapedReader[Wrapped, FieldType[K, V] :: T, Option[V] :: U] {
+      def from(cur: ConfigObjectCursor, default: Option[V] :: U, usedFields: Set[String]): ConfigReader.Result[FieldType[K, V] :: T] = {
+        val fieldName = key.value.name
+        val fieldAction = hint.from(cur, fieldName)
+        lazy val reader = vConfigReader.value.value
+        lazy val keyNotFoundFailure = cur.failed[V](KeyNotFound.forKeys(fieldAction.field, cur.keys))
```
ruippeixotog: Can we move this to the only `case` where it's used? It's generally harder to read lazy vals with no context of where they're going to be used.

modules/generic/src/main/scala/pureconfig/generic/DerivedConfigReader.scala
```diff
@@ -64,20 +66,42 @@ trait DerivedConfigReader1 {
...
-    cc: Lazy[MapShapedReader.WithDefaults[F, Repr, DefaultRepr]]): DerivedConfigReader[F] = new DerivedConfigReader[F] {
+    cc: Lazy[MapShapedReader[F, Repr, DefaultRepr]]): DerivedConfigReader[F] = new DerivedConfigReader[F] {
 
     override def from(cur: ConfigCursor): ConfigReader.Result[F] = {
-      cur.asObjectCursor.right.flatMap(cc.value.fromWithDefault(_, default())).right.map(gen.from)
+      cur.asObjectCursor.right.flatMap(cc.value.from(_, default(), Set.empty)).right.map(gen.from)
     }
   }
 
   final implicit def coproductReader[F, Repr <: Coproduct](
```
ruippeixotog, line 76: I'm not really sure why we chose `F` for this type parameter, but it's a strange choice as `F` is typically used for higher-kinded types...

modules/generic/src/main/scala/pureconfig/generic/DerivedConfigReader.scala
```diff
@@ -64,20 +66,42 @@ trait DerivedConfigReader1 {
...
+            case None => ConfigReader.Result.fail[F](cursor.failureFor(MissingCoproductChoice(option)))
+          }
+
+        case CoproductHint.Attempt(cursor, options, combineF) =>
+          val initial: Either[Vector[(String, ConfigReaderFailures)], F] = Left(Vector.empty)
+          val res = options.foldLeft(initial) { (curr, option) =>
+            curr.left.flatMap { currentFailures =>
+              readerFor(option) match {
+                case Some(value) => value.from(cursor).left.map(f => currentFailures :+ (option -> f))
+                case None => Left(currentFailures :+ (option -> ConfigReaderFailures(cursor.failureFor(MissingCoproductChoice(option)))))
```
ruippeixotog: nit: please add a newline somewhere in this line.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
```diff
@@ -67,27 +62,33 @@ object MagnoliaConfigReader {
...
+            case None => ConfigReader.Result.fail[A](cur.failureFor(MissingCoproductChoice(option)))
+          }
+
+        case CoproductHint.Attempt(cur, options, combineF) =>
+          val initial: Either[Vector[(String, ConfigReaderFailures)], A] = Left(Vector.empty)
+          val res = options.foldLeft(initial) { (curr, option) =>
+            curr.left.flatMap { currentFailures =>
+              readerFor(option) match {
+                case Some(value) => value.from(cur).left.map(f => currentFailures :+ (option -> f))
+                case None => Left(currentFailures :+ (option -> ConfigReaderFailures(cur.failureFor(MissingCoproductChoice(option)))))
```
ruippeixotog: nit: newline here too

ruippeixotog (approved):
LGTM overall! I left a few minor comments around, please take a look.

I'm really looking forward to merge this, after almost 7 months of backs and forths :)

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
jcazevedo: Sure. Improved that in a0b3fb8d.

modules/generic-base/src/main/scala/pureconfig/generic/ProductHint.scala
jcazevedo: Fixed in 34f3a4bb.

modules/generic-base/src/main/scala/pureconfig/generic/error/MissingCoproductChoice.scala
jcazevedo: Make sense. I've started using "coproduct option" instead of "coproduct choice" throughout the codebase in f5013e48.

modules/generic-base/src/main/scala/pureconfig/generic/error/MissingCoproductChoice.scala
jcazevedo: Made the description more specific in f5013e48.

modules/generic/src/main/scala/pureconfig/generic/EnumCoproductHint.scala
jcazevedo: Removed the unnecessary `new`s in a1c7fe6c.

modules/generic/src/main/scala/pureconfig/generic/MapShapedReader.scala
jcazevedo: Improved in 2fb7f84e.

modules/generic/src/main/scala/pureconfig/generic/DerivedConfigReader.scala
jcazevedo: Not really sure. I'd rather fix that in a different PR, though.

modules/generic/src/main/scala/pureconfig/generic/DerivedConfigReader.scala
jcazevedo: Added newline in f5013e48.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/MagnoliaConfigReader.scala
jcazevedo: Added newline in f5013e48.

ruippeixotog:
Travis is failing because html-proofer is failing to get bootstrap.min.css. I'll merge this anyway as it's unrelated to this PR. Thanks again for this refactoring, @jcazevedo!

ruippeixotog (approved):

PR #634, opened 2019-11-01
added a counterpart for config

kailuowang:


jcazevedo:
@kailuowang: Thanks for your PR!

Since a `ConfigSource` can map to values that aren't map-like, it doesn't expose a `config()` method. Only `ConfigObjectSource` does. 

We provide a `ConfigReader` for `Config`, so it's always possible to get instances of `Config` suspended in `F` using `source.loadF[F, Config]`. As such, I'm not very keen on adding this particular implementation for `Config`.

kailuowang:
sorry about the lazy PR. I was too occupied today. I noticed that I could do `loadF[F, Config]` but a quick glance of the implementation of `load` left the impression that it's more efficient to work with `config()`. Anyway thanks for the clarification.

jcazevedo:
It should be slightly more efficient to work with `config()` since you avoid creating a `ConfigCursor`, but I don't expect the difference to be significant (I haven't measured it).

PR #635, opened 2019-11-04
Avoid unnecessary casts when creating a cursor from a `ConfigObjectSource`

jcazevedo:
This PR proposes a small performance improvement when creating a cursor from a `ConfigObjectSource` by creating a specialized cursor, which avoids some unnecessary casts in its `asObjectCursor` method.


ruippeixotog (approved):
LGTM!

PR #654, opened 2019-12-05
Upgrade ConfigReader to ConfigConvert for http4s Uri

LMnet:
Minor improvement: upgraded `ConfigReader` to `ConfigConvert` for http4s `Uri`.

### DIFF
--- a/modules/http4s/src/main/scala/pureconfig/module/http4s/package.scala
+++ b/modules/http4s/src/main/scala/pureconfig/module/http4s/package.scala
@@ -1,8 +1,8 @@
 package pureconfig.module
 
 import org.http4s.Uri
-import pureconfig.ConfigReader
 import pureconfig.error.CannotConvert
+import pureconfig.{ ConfigReader, ConfigWriter }
 
 package object http4s {
 
@@ -11,4 +11,6 @@ package object http4s {
       Uri.fromString(str).fold(
         err => Left(CannotConvert(str, "Uri", err.sanitized)),
         uri => Right(uri)))
+
+  implicit val uriWriter: ConfigWriter[Uri] = ConfigWriter[String].contramap(_.renderString)
 }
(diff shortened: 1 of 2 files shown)


leifwickland (approved):
Thanks, @LMnet.

@ruippeixotog & @jcazevedo This change is not binary compat. Do you think that should have any impact on how we version the release that includes this?

LMnet:
@leifwickland oh, I completely forgot about that.

I can provide `ConfigWriter` and `ConfigConvert` without touching the `ConfigReader`, so it will binary compatible.

modules/http4s/src/main/scala/pureconfig/module/http4s/package.scala
```diff
@@ -1,8 +1,8 @@
 package pureconfig.module
 
 import org.http4s.Uri
-import pureconfig.ConfigReader
 import pureconfig.error.CannotConvert
+import pureconfig.{ConfigReader, ConfigWriter}
```
ruippeixotog: The build is failing because this line doesn't conform to the Scalariform rule (running `sbt compile` should fix this automatically). Could you please fix this?

ruippeixotog (changes requested):
> @ruippeixotog & @jcazevedo This change is not binary compat. Do you think that should have any impact on how we version the release that includes this?

We don't usually have this requirement in our core library and we are even less strict on modules, so I think this is totally fine.

jcazevedo (changes requested):
@LMnet: Thanks for your PR! Can you please add a test for round-trip reading/writing of `Uri`?

LMnet:
> @LMnet: Thanks for your PR! Can you please add a test for round-trip reading/writing of `Uri`?

@jcazevedo Done. I added a test and fixed the formatting.

jcazevedo (approved):
@LMnet: Thanks for this!

ruippeixotog (approved):

PR #668, opened 2019-12-18
Provide cats Eq[Config] and Monoid[ConfigObjectSource] instances

### LINKED ISSUE #667: Add Semigroup[ConfigSource], Monoid[ConfigObjectSource]
We already have a `Semigroup` for `ConfigValue`, so we can just as well make a corresponding one for `ConfigSource`. Maybe even think upgrade one or both of these to `Monoid`, if `ConfigSource.empty` and/or the corresponding `ConfigValue` turn out to be an identity. And we can definitely make a `Monoid[ConfigObjectSource]`.

ruippeixotog:
Closes #667.

cc @nigredo-tori

### DIFF
--- a/modules/cats/src/main/scala/pureconfig/module/cats/instances/package.scala
+++ b/modules/cats/src/main/scala/pureconfig/module/cats/instances/package.scala
@@ -1,7 +1,7 @@
 package pureconfig.module.cats
 
 import cats._
-import com.typesafe.config.ConfigValue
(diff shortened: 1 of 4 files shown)


nigredo-tori:
LGTM, but could you please add a `Monoid[Config]` while you're at it? I think the laws should hold, since a `Config` is basically a `ConfigObject`.

jcazevedo (approved):
Thanks for this!

PR #672, opened 2019-12-21
Configure GitHub Actions for CI

ruippeixotog:
Let's try out GitHub Actions as an alternative to Travis. The advantages are that have better integration with GitHub (duh!) and they are more flexible, allowing us to automate some more workflows in the future not strictly related to CI and not necessarily triggering on pushes and PRs.

In this PR, the only things not on par with our Travis build are:
- Cache dependencies: the mechanism to cache things is a bit more complex than in Travis. However, GitHub seems to be outperforming Travis anyway ([13m8s on GitHub](https://github.com/pureconfig/pureconfig/actions) vs [13m32s on Travis](https://travis-ci.org/pureconfig/pureconfig/jobs/628148882), with GitHub builds triggering earlier), so we can start without caching and iterate later if needed;
- Upload coverage to Coveralls: see TODO on workflow definition. Since we haven't disabled Travis yet we'll continue to have coverage data;
- PR comments on website changes: this would actually be easier now as GitHub provides us a token, but since jobs are shown separately in PR checks this is not needed anymore - we can simply make diff_website not push-blocking.

Some other ideas to use GitHub Actions in the future:
- Publish a snapshot JAR to Sonatype automatically each time we push to `master`;
- Publish the website to pureconfig.github.io on every tag push (risky for now, the script is not foolproof).

We can leave both GitHub Actions and Travis running for a while to evaluate whether we are good to move definitely (and whether we want it).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,116 @@
+name: CI
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        jdk: [8]
+        scala: [2.11.12, 2.12.10, 2.13.1]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.jdk }}
+
+      - name: Compile
+        run: sbt coverage "++${{ matrix.scala }} compile"
+
+      - name: Run tests
+        run: sbt coverage "++${{ matrix.scala }} test"
+
+      - name: Build Scaladoc
+        run: sbt "++${{ matrix.scala }} doc"
+
+      - name: Publish artifact locally
+        run: sbt "++${{ matrix.scala }} publishLocal"
+
+      - name: Compile example project
+        run: cd example && sbt "++${{ matrix.scala }} test"
+
+      - name: Check tut output
+        run: >
+          if [[ "${{ matrix.scala }}" =~ ^2\.12.* ]]; then
+            sbt "++${{ matrix.scala }} tut" &&
+            git diff --exit-code;
+            fi
+
+# TODO: enable this once either sbt-coveralls starts supporting GitHub Actions
+# (https://github.com/scoverage/sbt-coveralls/issues/126) or coverallsapp/github-action starts
+# supporting cobertura.xml format (https://github.com/coverallsapp/github-action/issues/30)
+#
+#      - name: Compile coverage data
+#        run: sbt ++${{ matrix.scala }} coverageAggregate
+#
+#      - name: Upload coverage data to Coveralls
+#        uses: coverallsapp/github-action@master
+#        with:
+#          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+  build_website:
+    name: Build Website
+    runs-on: ubuntu-latest
+    env:
+      NOKOGIRI_USE_SYSTEM_LIBRARIES: true
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 8
+
+      - name: Set up Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6
+
+      - name: Install dependencies
+        run: >
+          sudo apt install libcurl4-openssl-dev libxslt-dev &&
+          gem install sass jekyll:4.0.0 html-proofer:3.9.3
+
+      - name: Build website
+        run: sbt makeMicrosite
+
+      - name: Verify website
+        run: >
+          htmlproofer
+          --allow-hash-href
+          --url-swap "https\://github\.com/pureconfig/pureconfig/tree/master:file\://$(pwd)"
+          --url-ignore "/search.maven.org/"
+          docs/target/site
+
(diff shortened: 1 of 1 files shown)


ruippeixotog:
I also made the `diff_website` check available to all non-master pushes (previously it only ran on PRs). It may be of some value even if there is not a PR yet, but I can set it up like Travis if you think it's not worth it.

ruippeixotog:
@jcazevedo @leifwickland can you please take a look at this? :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,116 @@
...
+      matrix:
+        jdk: [8]
+        scala: [2.11.12, 2.12.10, 2.13.1]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      - name: Set up JDK
+        uses: actions/setup-java@v1
```
jcazevedo: It might be preferable to use [`olafurpg/setup-scala`](https://github.com/olafurpg/setup-scala) which sets up sbt from [`sbt/paulp-extras`](https://github.com/paulp/sbt-extras/).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,116 @@
...
+
+      - name: Publish artifact locally
+        run: sbt "++${{ matrix.scala }} publishLocal"
+
+      - name: Compile example project
+        run: cd example && sbt "++${{ matrix.scala }} test"
+
+      - name: Check tut output
+        run: >
+          if [[ "${{ matrix.scala }}" =~ ^2\.12.* ]]; then
```
jcazevedo: There seems to be an `if` conditional for steps (https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsif) which we should be able to use here.

jcazevedo (approved):
> PR comments on website changes: this would actually be easier now as GitHub provides us a token, but since jobs are shown separately in PR checks this is not needed anymore - we can simply make diff_website not push-blocking.

If we can't encode the concept of warnings in checks (which we don't seem to), I prefer `diff_website` as a step that only comments in PRs rather than a check. Even if we make `diff_website` not push-blocking, there will still be the indication that the check didn't pass. Due to the binary nature of checks, a non-passing one seems to indicate that something should be done to fix it, which might not always be the case for this check. Without a PR comment, it's also hard to identify what went wrong without knowing what `diff_website.sh` does.

We also seem to have lost colors in the sbt output (that's also an issue with Travis). It seems like it was introduced in [sbt 1.3.x](https://github.com/sbt/sbt/issues/4879). We can regain colors by running sbt with `sbt -Dsbt.color=always -Dsbt.supershell=false`.

.github/workflows/ci.yml
ruippeixotog: Makes sense, I knew about it but hadn't thought of using it here. Well spotted!

ruippeixotog:
> Even if we make `diff_website` not push-blocking, there will still be the indication that the check didn't pass. Due to the binary nature of checks, a non-passing one seems to indicate that something should be done to fix it, which might not always be the case for this check.

I understand the concern and agree that having a red cross on builds may convey the wrong message. However, note that we already a check like this (Coveralls), so it wouldn't be the first one. I don't have an example at hand, but I think that GitHub shows them less prominently. Given how hacky the current approach is and how annoying the comments are (particularly in PRs with multiple iterations) I'm inclined to leave it this way, but we can always consider re-adding comments later (nothing prevents us from having both approaches).

> Without a PR comment, it's also hard to identify what went wrong without knowing what `diff_website.sh` does.

PureConfig maintainers know what the check is about, while for other contributors the check is currently failing the whole build due to how Travis (and GitHub actions) handles secrets. I'd argue that this is still a strict improvement over the previous CI setup :)

ruippeixotog:
> We also seem to have lost colors in the sbt output (that's also an issue with Travis). It seems like it was introduced in [sbt 1.3.x](https://github.com/sbt/sbt/issues/4879). We can regain colors by running sbt with `sbt -Dsbt.color=always -Dsbt.supershell=false`.

I tried various values for `-Dsbt.color` and none of them seem to work :/

.github/workflows/ci.yml
```diff
@@ -0,0 +1,118 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      - name: Set up Scala
+        uses: olafurpg/setup-scala@v7
+        with:
+          java-version: ${{ matrix.jdk }}
+
+      - name: Configure SBT
+        run: echo '-Dsbt.color=true -Dsbt.supershell=false' > .sbtopts
```
jcazevedo: I think that you might need to put these options on different lines for sbt to pick them up.

jcazevedo:
> PureConfig maintainers know that the check is about, while for other contributors the check is currently failing the whole build due to how Travis (and GitHub actions) handles secrets. I'd argue that this is still a strict improvement over the previous CI setup :)

I agree that it is an improvement and I think we can go with this for now. 

I was comparing against the "ideal" scenario of the other strategy (i.e. a maximum of one comment per PR and working with forks). If we can achieve that, I'd much rather have that than a check.

.github/workflows/ci.yml
ruippeixotog: Nope, it didn't work 🙁 I don't even know if GitHub Actions support ANSI color codes at all. They already seem to highlight lines in different shades of grey, so maybe they have their own logic for it?

ruippeixotog: They clearly mention colors in https://github.com/features/actions, so the problem must be somewhere else.

Issue #485, opened 2019-06-11
Support a simpler form of manual derivation


chetanmeh:
With #420 support was added for [manual derivation][1]. This mode requires one to explicitly enumerate the field names like

```scala
implicit val personReader = ConfigReader.forProduct2("name", "surname")(Person(_, _))
```

It would be easier if we can avoid listing the field names explicitly. We can use an approach similar to one taken in [spray-json][2] which enables creating custom formats for case classes like

```scala
object MyJsonProtocol extends DefaultJsonProtocol {
  implicit val colorFormat = jsonFormat4(Color)
}
```

This would enable one to avoid using shapeless based approach with lesser effort and still benefit from not having any significant impact on compile times. 

See [discussion on gitter][3]

[1]: https://pureconfig.github.io/docs/non-automatic-derivation.html#manual
[2]: https://github.com/spray/spray-json#providing-jsonformats-for-case-classes
[3]: https://gitter.im/melrief/pureconfig?at=5cfdeb626f530d3b6168c85f

PR #673, opened 2019-12-29
Simplify manual derivation via reflection

chetanmeh:
This PR implements the approach proposed in #485. It uses reflection to discover the case class properties and then uses the [manual derivation][1] for defining the `ConfigReader`. Reflection is only used once per case class to discover the property names at time of initializing the reader.

### Usage
Define a case class `MyClass`

```scala
package example

case class MyClass(
  boolean: Boolean,
  port: Port,
  adt: MyAdt,
  list: List[Double],
  map: Map[String, String],
  option: Option[String])
```

Now define the `ConfigReader`. It can be either placed in companion object for `MyClass` or in package object

```scala
import pureconfig._
import pureconfig.module.reflect.ReflectConfigReaders._

package object example {
  implicit val myClassConfigReader: ConfigReader[MyClass] = configReader6(MyClass)
}
```

Here we use any of the `configReaderXXX` method (where xxx is based on arity of case class) to define the `ConfigReader`. Now one can simply do `source.load[MyClass]` to create `MyClass` instance from config.

Key aspects

1. Supports only case class. Does not support sealed traits, value classes and tuples
2. No dependency on any other library like Shapeless
3. Minimal impact on compilation time - As this module does not Shapeless there is no major impact on compilation timing. In our project we have seen 30%+ reduction in timing for compilation (Compilation time reduced from ~58 sec to ~38-40 secs, See apache/openwhisk#4783 for more stats)
4. Currently does not support default values. Support for this can be added later if needed

Test cases are based on tests done for `pureconfig-mangolia` module

[1]: https://pureconfig.github.io/docs/non-automatic-derivation.html#manual

### DIFF
--- a/modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigReaders.scala.template
+++ b/modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigReaders.scala.template
@@ -8,7 +8,7 @@ trait ReflectConfigReaders extends ProductReaders {
     import ReflectSupport._
     private type CR[T] = ConfigReader[T]
 [#  // Case classes with 1 parameters
-    def forProduct1[[#A1 :CR#], T <: Product: ClassTag](construct: ([#A1#]) => T)(implicit ph: ReflectProductHint[T]): ConfigReader[T] = {
+    def configReader1[[#A1 :CR#], T <: Product: ClassTag](construct: ([#A1#]) => T)(implicit ph: ReflectProductHint[T]): ConfigReader[T] = {
         val Array([#a1#]) = extractFieldNames(classTag[T])
         forProduct1([#ph.configKey(a1)#])(construct)
     }#
--- a/modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigReaders.scala.template
+++ b/modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigReaders.scala.template
@@ -1,0 +1,20 @@
+package pureconfig.module.reflect
+
+import pureconfig.{ConfigReader, ProductReaders, BasicReaders, CollectionReaders, ExportedReaders}
+
+import scala.reflect.{ClassTag, classTag}
+
+trait ReflectConfigReaders extends ProductReaders {
+    import ReflectSupport._
+    private type CR[T] = ConfigReader[T]
+[#  // Case classes with 1 parameters
+    def forProduct1[[#A1 :CR#], T <: Product: ClassTag](construct: ([#A1#]) => T)(implicit ph: ReflectProductHint[T]): ConfigReader[T] = {
+        val Array([#a1#]) = extractFieldNames(classTag[T])
+        forProduct1([#ph.configKey(a1)#])(construct)
+    }#
+
+]
+
+}
+
+object ReflectConfigReaders extends ReflectConfigReaders with BasicReaders with CollectionReaders with ExportedReaders
--- a/modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigWriters.scala.template
+++ b/modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigWriters.scala.template
@@ -8,13 +8,13 @@ trait ReflectConfigWriters extends ProductWriters {
     import ReflectSupport._
     private type CW[T] = ConfigWriter[T]
 
-    def forProduct1[A1: CW,  T <: Product: ClassTag](deconstruct: T => A1)(implicit ph: ReflectProductHint[T]): ConfigWriter[T] = {
+    def configWriter1[A1: CW,  T <: Product: ClassTag](deconstruct: T => A1)(implicit ph: ReflectProductHint[T]): ConfigWriter[T] = {
         val Array(p1) = extractFieldNames(classTag[T])
         forProduct1(ph.configKey(p1))(deconstruct)
     }
 
 [2..22#  // Case classes with 1 parameters
-    def forProduct1[[#A1 :CW#], T <: Product: ClassTag](deconstruct: T => ([#A1#]) )(implicit ph: ReflectProductHint[T]): ConfigWriter[T] = {
+    def configWriter1[[#A1 :CW#], T <: Product: ClassTag](deconstruct: T => ([#A1#]) )(implicit ph: ReflectProductHint[T]): ConfigWriter[T] = {
         val Array([#a1#]) = extractFieldNames(classTag[T])
         forProduct1([#ph.configKey(a1)#])(deconstruct)
     }#
--- a/modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigWriters.scala.template
+++ b/modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigWriters.scala.template
@@ -1,0 +1,25 @@
+package pureconfig.module.reflect
+
+import pureconfig.{ConfigWriter, ProductWriters, BasicWriters, CollectionWriters, ExportedWriters}
+
+import scala.reflect.{ClassTag, classTag}
+
+trait ReflectConfigWriters extends ProductWriters {
+    import ReflectSupport._
+    private type CW[T] = ConfigWriter[T]
+
+    def forProduct1[A1: CW,  T <: Product: ClassTag](deconstruct: T => A1)(implicit ph: ReflectProductHint[T]): ConfigWriter[T] = {
+        val Array(p1) = extractFieldNames(classTag[T])
+        forProduct1(ph.configKey(p1))(deconstruct)
+    }
+
+[2..22#  // Case classes with 1 parameters
+    def forProduct1[[#A1 :CW#], T <: Product: ClassTag](deconstruct: T => ([#A1#]) )(implicit ph: ReflectProductHint[T]): ConfigWriter[T] = {
+        val Array([#a1#]) = extractFieldNames(classTag[T])
+        forProduct1([#ph.configKey(a1)#])(deconstruct)
+    }#
+]
+
+}
+
+object ReflectConfigWriters extends ReflectConfigWriters with BasicWriters with CollectionWriters with ExportedWriters
--- a/modules/reflect/src/main/scala/pureconfig/module/reflect/ReflectProductHint.scala
+++ b/modules/reflect/src/main/scala/pureconfig/module/reflect/ReflectProductHint.scala
@@ -1,0 +1,35 @@
+package pureconfig.module.reflect
+
+import pureconfig.{ CamelCase, ConfigFieldMapping, KebabCase }
+
+/**
+ * A trait that can be implemented to customize how case classes are read from and written to a config.
+ *
+ * @tparam T the type of case class for which this hint applies
+ */
+trait ReflectProductHint[T] {
+
+  /**
+   * Returns the key in the config object associated with a given case class field.
+   *
+   * @param fieldName the case class field
+   * @return the key in the config object associated with the given case class field.
+   */
+  def configKey(fieldName: String): String
+
+}
+
+private[pureconfig] case class ReflectProductHintImpl[T](
+    fieldMapping: ConfigFieldMapping) extends ReflectProductHint[T] {
+
+  def configKey(fieldName: String) = fieldMapping(fieldName)
+}
+
+object ReflectProductHint {
+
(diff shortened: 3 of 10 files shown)


chetanmeh:
This PR needs some more test coverage. Opened this PR now to get feedback on approach taken or any other comments

nigredo-tori:
I'm opposed to this approach. The general consensus in Scala's FP ecosystem is to make sure that all instance derivation is checked at compile time. Using reflection for deriving readers/writers opens the calling code up to runtime failures. These can be caught by tests, but I don't think many people bother testing their config readers.

Since the goal here seems to be to speed up compilation times, how about implementing the same functionality using macros? A black box macro should incur significantly less overhead than the full Shapeless machinery, without sacrificing runtime safety.

Alternatively, we can try using [Magnolia](https://magnolia.work/opensource/magnolia), which advertises itself as a Shapeless alternative for instance derivation (with faster compilation times). There is an experimental [circe-magnolia](https://github.com/circe/circe-magnolia) project, from which we can borrow some ideas.

Please take the above with a grain of salt, since I'm not a maintainer of this project, and I tend to explicitly write out all the codecs anyway. Also, it's been a while since I had to write a macro, and I didn't have a chance to use Magnolia yet.

jcazevedo:
> Alternatively, we can try using [Magnolia](https://magnolia.work/opensource/magnolia), which advertises itself as a Shapeless alternative for instance derivation (with faster compilation times). There is an experimental [circe-magnolia](https://magnolia.work/opensource/magnolia) project, from which we can borrow some ideas.

We already have support for derivation using Magnolia since version 0.12.2 (https://github.com/pureconfig/pureconfig/tree/master/modules/magnolia). Replacing `pureconfig-generic` with it should only require changing the imports.

@chetanmeh: Have you experimented with it? I tried it in a project that was using semiauto derivation of readers and got around 10% reduction in compile time.

chetanmeh:
> The general consensus in Scala's FP ecosystem is to make sure that all instance derivation is checked at compile time. Using reflection for deriving readers/writers opens the calling code up to runtime failures.

Use of reflection here is minimal to determine the field names. Rest of the reader/writer derivation works in a typesafe way and is built on top of existing Manual mode reader and writer implementation. So all instance derivation still gets checked at compile time. Note that approach taken is on same lines as one used in spray-json library which is widely used

> Have you experimented with it? I tried it in a project that was using semiauto derivation of readers and got around 10% reduction in compile time.

Have not used it so far as its [status][1] is mentioned as experimental and it would be bit tricky to switch to this library given the mentioned state.

> Magnolia is currently experimental. It has been shown to work for a variety of test cases, though it has not had the same exposure to real-world datatypes that, for example, Shapeless has had

[1]: https://magnolia.work/opensource/magnolia#current-status

nigredo-tori:
> Use of reflection here is minimal to determine the field names.

It still breaks runtime safety guarantees, making the resulting codecs dangerous to use.

More precisely, the main issue I see here is that the reflective code has to try and reconstruct the Scala case class definition from its `Class` representation. The implementation has to rely on `copy` and `case class` encoding details. This makes it inherently brittle. I haven't tested this, but it looks like adding a `val` to a case class, or trying to derive codecs for a nested class, would be enough to break it (since it has to rely on `getDeclaredFields`). We also have to rely on `getDeclaredFields` returning the fields in the order they were defined in (which [is not guaranteed](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/lang/Class.html#getDeclaredFields())). And (I can't stress this enough) any issues like these will only be discovered at runtime.

nigredo-tori:
As an aside, spray-json itself seems to have been [moving](https://github.com/spray/spray-json/issues/59) towards a macro-based implementation, although that doesn't look like it had come to anything before the active development has stopped - which was around 2014, judging by the latest (last?) minor release.

chetanmeh:
> This makes it inherently brittle. I haven't tested this, but it looks like adding a val to a case class, or trying to derive codecs for a nested class, would be enough to break it (since it has to rely on getDeclaredFields). We also have to rely on getDeclaredFields returning the fields in the order they were defined in (which is not guaranteed)

Ack. Though all this time with spray-json using case classes we had never faced such issues. But then we have probably used case class in simpler way. The problem you mentioned can be fixed through approach suggested [here][1]. For ordering safety we can use approach suggested [here][2].

A macro based approach would be more safer but then they do have an [adverse impact on compile times][3]. So I am looking for a way to minimize impact on compile time and still have manageable config handling (which does not require one to list all field names explicitly)

[1]: https://github.com/spray/spray-json/pull/93#issuecomment-104585274
[2]: https://stackoverflow.com/questions/16079113/scala-2-10-reflection-how-do-i-extract-the-field-values-from-a-case-class-i-e/16079804#comment41053355_16079804
[3]: https://www.scala-lang.org/blog/2018/06/04/scalac-profiling.html#the-cost-of-implicit-macros

chetanmeh:
> A black box macro should incur significantly less overhead than the full Shapeless machinery, without sacrificing runtime safety

Thinking about it bit more may be difference is not that big as we just need to get field names. Would try to implement a macro based approach and see how much difference is there in performance.

nigredo-tori:
> The problem you mentioned can be fixed through approach suggested here. For ordering safety we can use approach suggested here.

I wasn't aware we can get the constructor parameters' names through reflection. This can make the whole thing much cleaner (and more robust). There are still theoretical ways to mess everything up - multiple constructors, scalajs, obfuscators - but for the common use case this might be good enough.

> A macro based approach would be more safer but then they do have an adverse impact on compile times.

The post you referenced mentions *whitebox* macros (which complicate typechecking) and *implicit* macros (which slow down implicit derivation, possibly exponentially). We need neither of these here. Note that `configWriter1` and its siblings are not implicit.

modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigReaders.scala.template
```diff
@@ -0,0 +1,20 @@
+package pureconfig.module.reflect
+
+import pureconfig.{ConfigReader, ProductReaders, BasicReaders, CollectionReaders, ExportedReaders}
+
+import scala.reflect.{ClassTag, classTag}
+
+trait ReflectConfigReaders extends ProductReaders {
```
ruippeixotog, line 7: Can you please use the `ProductReaders` object to access product readers? In PureConfig we usually rely as much as possible in static function methods as opposed to inheritance (we provide each set of instances as a trait just to give users the ability to build their own stacks if they want).

modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigReaders.scala.template
```diff
@@ -0,0 +1,20 @@
...
+    def configReader1[[#A1 :CR#], T <: Product: ClassTag](construct: ([#A1#]) => T)(implicit ph: ReflectProductHint[T]): ConfigReader[T] = {
+        val Array([#a1#]) = extractFieldNames(classTag[T])
+        forProduct1([#ph.configKey(a1)#])(construct)
+    }#
+
+]
+
+}
+
+object ReflectConfigReaders extends ReflectConfigReaders with BasicReaders with CollectionReaders with ExportedReaders
```
ruippeixotog, line 20: Same thing here, the object should extend just `ReflectConfigReaders`. Please add a newline, too.

modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigReaders.scala.template
```diff
@@ -0,0 +1,20 @@
...
+
+import pureconfig.{ConfigReader, ProductReaders, BasicReaders, CollectionReaders, ExportedReaders}
+
+import scala.reflect.{ClassTag, classTag}
+
+trait ReflectConfigReaders extends ProductReaders {
+    import ReflectSupport._
+    private type CR[T] = ConfigReader[T]
+[#  // Case classes with 1 parameters
+    def configReader1[[#A1 :CR#], T <: Product: ClassTag](construct: ([#A1#]) => T)(implicit ph: ReflectProductHint[T]): ConfigReader[T] = {
```
ruippeixotog, line 11: Can you mark this as `final`?

modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigReaders.scala.template
```diff
@@ -0,0 +1,20 @@
+package pureconfig.module.reflect
+
+import pureconfig.{ConfigReader, ProductReaders, BasicReaders, CollectionReaders, ExportedReaders}
+
+import scala.reflect.{ClassTag, classTag}
+
+trait ReflectConfigReaders extends ProductReaders {
+    import ReflectSupport._
+    private type CR[T] = ConfigReader[T]
```
ruippeixotog, line 9: There seems to be no real reason to create this type alias, as it's used just once in this file. Can you use it directly?

modules/reflect/src/main/boilerplate/pureconfig/module/reflect/ReflectConfigWriters.scala.template
```diff
@@ -0,0 +1,25 @@
+package pureconfig.module.reflect
+
+import pureconfig.{ConfigWriter, ProductWriters, BasicWriters, CollectionWriters, ExportedWriters}
+
+import scala.reflect.{ClassTag, classTag}
+
+trait ReflectConfigWriters extends ProductWriters {
```
ruippeixotog, line 7: Same comments apply here.

modules/reflect/src/main/scala/pureconfig/module/reflect/ReflectProductHint.scala
```diff
@@ -0,0 +1,35 @@
+package pureconfig.module.reflect
+
+import pureconfig.{ CamelCase, ConfigFieldMapping, KebabCase }
+
+/**
+ * A trait that can be implemented to customize how case classes are read from and written to a config.
+ *
+ * @tparam T the type of case class for which this hint applies
+ */
+trait ReflectProductHint[T] {
```
ruippeixotog, line 10: I am not fond of adding parallel hints, as it makes migration harder. On the other hand, using full hints may lead users to think that all its features are supported. Maybe we can add an `implicit def` building a `ReflectProductHint` from a `ProductHint`?

modules/reflect/src/main/scala/pureconfig/module/reflect/ReflectSupport.scala
```diff
@@ -0,0 +1,35 @@
...
+        import Modifier._
+        (f.getModifiers & (TRANSIENT | STATIC | 0x1000 /* SYNTHETIC*/ )) > 0
+      }
+      if (copyDefaultMethods.length != fields.length)
+        sys.error("Case class " + clazz.getName + " declares additional fields")
+      if (fields.zip(copyDefaultMethods).exists { case (f, m) => f.getType != m.getReturnType })
+        sys.error("Cannot determine field order of case class " + clazz.getName)
+      fields.map(f => unmangle(f.getName))
+    } catch {
+      case NonFatal(ex) => throw new RuntimeException("Cannot automatically determine case class field names and order " +
```
ruippeixotog, line 29: This is the main problem with this module. We definitely don't want this happening at runtime.

ruippeixotog (changes requested):
Like @nigredo-tori, my main concern with this module is the possibility that it may throw exceptions at runtime. This is generally not recommended and breaks the assumptions of exception safety we have throughout the library.

I believe we can go forward with a macro-based approach using reflection as long as it's guaranteed that:
- It doesn't throw exceptions at runtime;
- All successful compilations result in readers/writers with reasonable behavior (in particular, always returning a `Left` instead of throwing an exception is not reasonable).

But at this point, I'd suggest considering trying out `pureconfig-magnolia` as suggested above. The main reason being...

> Have not used it so far as its status is mentioned as experimental and it would be bit tricky to switch to this library given the mentioned state.

...that I would find Magnolia much more production-ready than any custom implementation here based on reflection or macros 🙂. The greatest strength of compile-time instance derivation lies precisely here: in the unlikely case that Magnolia has some issues with your project you'll find that out on the first time you compile it. I'm assuming you are already using Shapeless-based derivation, so using Magnolia would only require adding a dependency and changing some imports.

I reviewed your PR anyway (except for tests, which you said were not complete), please take a look at the comments.

ruippeixotog:
Closing this due to inactivity and because it's unlikely we'll merge it in its current form.

Issue #574, opened 2019-09-13
Migrate website and documentation to mdoc


ruippeixotog:
Tut is [being abandoned](https://github.com/tpolecat/tut/issues/246#issuecomment-519943218) in favor of [mdoc](https://github.com/scalameta/mdoc), so we should try to migrate our documentation. This would also allow us to start compiling the documentation against Scala 2.13.

I tried to do this before and I hit some roadblocks at the time (https://github.com/scalameta/mdoc/issues/155). Some things might have changed, but we'll probably need some structural changes in our repo.

PR #687, opened 2020-01-22
Start using mdoc for compiling the website

ruippeixotog:
In this PR we start using mdoc for compiling our website. We still use tut for modules' READMEs. Even though we can now compile the website against Scala 2.13 we are still using Scala 2.12 (I tested it and I'm going to change it in a later PR).

Most of the changes were made by a migration script, but I had to fix several cases due to the different semantics used by mdoc. mdoc is way faster as it doesn't run on a REPL context but on a class/object context. However, this same thing introduces restrictions like not being able to redeclare variables and value classes out of the box. With the new `reset-object` and `nest` tags provided by mdoc I was finally able to have a working version.

Part of #574.


ruippeixotog:
@jcazevedo, @leifwickland do you mind taking a look at this? Even though it's large it should be fairly easy to review.

jcazevedo:
> @jcazevedo, @leifwickland do you mind taking a look at this? Even though it's large it should be fairly easy to review.

Sorry. I'll try to catch up on some of the PRs during this week.

docs/build.sbt
```diff
@@ -1 +1 @@
-crossScalaVersions ~= { oldVersions => oldVersions.filter(_.startsWith("2.12")) }
+crossScalaVersions ~= { oldVersions => oldVersions.filterNot(_.startsWith("2.11")) }
```
jcazevedo, line 1: Not related to this change in particular, but do we need to have `crossScalaVersions` defined for the `docs` project? I assume we're always going to build the site using a specific Scala version.

jcazevedo (approved):
@ruippeixotog: Sorry for taking so long to review this. Looks great.

docs/build.sbt
ruippeixotog: We don't really need it, but even though we always publish the site using a specific version cross compiling ensures that our documentation works with both Scala 2.12 and 2.13.

jcazevedo: We should then start running `++docs/mdoc` in the CI, otherwise we will only build the documentation for Scala 2.12.

ruippeixotog: You're right, we aren't cross-compiling in CI. I think I didn't do it at the time because the deprecation warnings would make the build fail, but I manually did `++docs/mdoc` when I was migrating to make sure that was the case.

PR #692, opened 2020-01-27
http4s: add readers for client and server configuration

sideeffffect:
closes https://github.com/http4s/http4s/pull/3081

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,7 +25,10 @@ lazy val generic = (project in file("modules/generic")).
 
 lazy val macros = (project in file("macros")).
   enablePlugins(TutPlugin).
-  settings(commonSettings)
+  settings(commonSettings).
+  settings(
+    scalacOptions -= "-Ypartial-unification"
+  )
 
 lazy val tests = (project in file("tests")).
   enablePlugins(BoilerplatePlugin).
@@ -62,7 +65,7 @@ lazy val enumeratum = module(project) in file("modules/enumeratum")
 lazy val fs2 = module(project) in file("modules/fs2")
 lazy val hadoop = module(project) in file("modules/hadoop")
 lazy val http4s = module(project) in file("modules/http4s")
-lazy val http4sBlazeClient = module(project).in(file("modules/http4s-blaze-client")).withId("http4s-blaze-client")
+lazy val http4sBlazeClient = module(project).in(file("modules/http4s-blaze-client")).withId("http4s-blaze-client").dependsOn(generic)
 lazy val http4sBlazeServer = module(project).in(file("modules/http4s-blaze-server")).withId("http4s-blaze-server").dependsOn(generic)
 lazy val javax = module(project) in file("modules/javax")
 lazy val joda = module(project) in file("modules/joda")
--- a/build.sbt
+++ b/build.sbt
@@ -63,6 +63,7 @@ lazy val fs2 = module(project) in file("modules/fs2")
 lazy val hadoop = module(project) in file("modules/hadoop")
 lazy val http4s = module(project) in file("modules/http4s")
 lazy val http4sBlazeClient = module(project).in(file("modules/http4s-blaze-client")).withId("http4s-blaze-client")
+lazy val http4sBlazeServer = module(project).in(file("modules/http4s-blaze-server")).withId("http4s-blaze-server").dependsOn(generic)
 lazy val javax = module(project) in file("modules/javax")
 lazy val joda = module(project) in file("modules/joda")
 lazy val magnolia = module(project) in file("modules/magnolia") dependsOn `generic-base`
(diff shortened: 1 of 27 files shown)


sideeffffect:
@jcazevedo @ruippeixotog guys do you think you would find time to have a look at this PR anytime soon? :pray:

ruippeixotog:
Hi @sideeffffect! I'm sorry for taking so long to review this, but it was quite difficult to get to a conclusion on how to handle these new modules.

These modules are quite different from the others we have here. PureConfig modules come in a variety of forms, but most of them provide one or more of these:
- Config readers and writers for other library data types (e.g. akka, hadoop, joda);
- Instances of libraries' own type classes for `Config` or `ConfigReader` (e.g. cats, scalaz); 
- Adapters for other config formats, libraries or IO abstractions (e.g. circe, yaml, cats-effect, fs2);

They all share some commonalities - we strive to provide the thinnest possible layer so that these libraries play well with PureConfig. This comes with at least two advantages:
- They are unopinionated: we don't usually decide on specific config formats. The few readers we provide have a natural representation or allow the user to heavily customize. Other adapters are also straightforward and independent of users' code style or patterns.
- They are low maintenance: we don't have a lot of problems when we upgrade dependency versions or refactor PureConfig's core as the module layers are thin.

While this module looks good to me in general, I don't think it's a good fit to be maintained in this repository as it doesn't play well with the two criteria above. You chose a specific format for a feature-heavy HTTP client and server, something which is far from strightforward. You hardcoded supported configurable settings and their format (you even used camel case for config keys, which is not our default naming convention) and you had to create new mechanisms to accumulate config settings and transform client/server objects once they are created.

While I'm not very fond of accepting this here, if you publish these modules on your own repository I'd be more than happy to include them in our [list of external integrations](https://pureconfig.github.io/docs/library-integrations.html). What do you think?

/cc @jcazevedo and @leifwickland for a second or third opinion here.

sideeffffect:
@ruippeixotog thanks for the feedback!
I've reworked the PR to make it conceptually much more simple and more streamlined with the other modules.
Integrating http4s builders with PureConfig is a bit tricky, but there is a great value in having this integration upstream (here in PureConfig).

please let me know what you think of the new implementation or if you have any suggestions how to make it a better fit for PureConfig

jcazevedo:
I echo @ruippeixotog's comment. Introducing new types or setting opinionated defaults are things that I don't think `ConfigReader`s provided by internal PureConfig modules should do. That being said, I also don't see a problem linking to these modules as external integrations.

sideeffffect:
@jcazevedo there's probably a misunderstanding (btw are commenting on the previous `664f678`, or the current `02e037e` implementation)
I too am striving for as minimalistic, lean, zero maintenance and as unopinionated as possible
the only trouble here is that the `BlazeClientBuilder` is not a normal case class for which a ConfigReader can be derived via the usual means, so I have to work around that somehow... :man_shrugging: 
also, the default values are not opinionated at all -- those are the defaults of `BlazeClientBuilder` (there's even a test which checks that)

The current solution works around the problem by creating a custom case class `BlazeClientBuilderConfig`, which mirrors the original `BlazeClientBuilder` 1:1, but since it's a case class, deriving ConfigReader is trivial.
So are you suggesting to scrap this mirror case class altogether? Or do you think having the defaults (even if just a mirror) is a problem?

I'm no PureConfig expert, so maybe I'm missing the obvious right solution, so I'll greatly appreciate any kind of suggestion, pointers or feedback! :pray:  I think it will be really great, when we will have http4s readers here upstream in PureConfig, because on their own, they don't hold their weight to warrant a dedicated library.
Thank you :slightly_smiling_face:

aicodeguard:
### Code Review – Key Findings

**Summary:**
The review identified critical security risks related to the use of outdated `http4s-blaze` dependencies (version 0.20.17) which are subject to known CVEs. Upgrading these dependencies is strongly recommended to ensure the security of the configuration modules.

## Issues

1. **[Security]** Vulnerable Dependency Version (Client)
   - **File:** `modules/http4s-blaze-client/build.sbt`
   - **Lines:** [7](https://github.com/pureconfig/pureconfig/blob/02e037e1f68440ce2897d1c4403b92cee47953c4/modules/http4s-blaze-client/build.sbt#L7)
   - **Details:** The project pulls in `http4s-blaze-client` version `0.20.17`. This version is vulnerable to CVE-2021-21293 and potentially others.
   - **Fix:** Update the library dependency to a supported, secure version (e.g., >= 0.21.33).

2. **[Security]** Vulnerable Dependency Version (Server)
   - **File:** `modules/http4s-blaze-server/build.sbt`
   - **Lines:** [7](https://github.com/pureconfig/pureconfig/blob/02e037e1f68440ce2897d1c4403b92cee47953c4/modules/http4s-blaze-server/build.sbt#L7)
   - **Details:** The project pulls in `http4s-blaze-server` version `0.20.17`. This version is vulnerable to CVE-2021-32637 and potentially others.
   - **Fix:** Update the library dependency to a supported, secure version (e.g., >= 0.21.33).

PR #697, opened 2020-02-05
Proposal: Using blocker on the cats-effect module

BalmungSan:
AFAIK reading those configuration files are a blocking operation. As such, they should be done inside a _blocking context_, using [**Blocker**](https://typelevel.org/cats-effect/api/cats/effect/Blocker.html).

Since this is a binary breaking change, I took the liberty to also remove all the deprecated methods.
Obviously, that means this change can not be merged until you plan to release `0.13.0`

Let me know what do you think about this proposal.
I thought about asking in the **gitter** room before, but looking at the code I realized the change was easy to do, so I believed it was better just to start the proposal with the changes already done.

### DIFF
--- a/modules/cats-effect/README.md
+++ b/modules/cats-effect/README.md
@@ -23,11 +23,13 @@ To load a configuration file from a path using cats-effect's `IO`:
 import pureconfig._
 import pureconfig.generic.auto._
 import pureconfig.module.catseffect.syntax._
-import cats.effect.IO
+import cats.effect.{ Blocker, IO }
 
-case class MyConfig(somefield: Int, anotherfield: String)
+final case class MyConfig(somefield: Int, anotherfield: String)
 
-val load: IO[MyConfig] = ConfigSource.file(somePath).loadF[IO, MyConfig]
+val load: IO[MyConfig] = Blocker[IO].use { blocker =>
+  ConfigSource.file(somePath).loadF[IO, MyConfig](blocker)
+}
 ```
 
 To test that this `IO` does indeed return a `MyConfig` instance:
@@ -48,8 +50,11 @@ To create an IO that writes out a configuration file, do as follows:
 
 ```scala
 import pureconfig.module.catseffect._
-import cats.effect.IO
+import cats.effect.{ Blocker, IO }
 
 val someConfig = MyConfig(1234, "some string")
-val save: IO[Unit] = saveConfigAsPropertyFileF[IO, MyConfig](someConfig, somePath)
+
+val save: IO[Unit] = Blocker[IO].use { blocker =>
+  saveConfigAsPropertyFileF[IO, MyConfig](someConfig, somePath, blocker)
+}
 ```
--- a/modules/cats-effect/README.md
+++ b/modules/cats-effect/README.md
@@ -23,16 +23,20 @@ To load a configuration file from a path using cats-effect's `IO`:
 import pureconfig._
 import pureconfig.generic.auto._
 import pureconfig.module.catseffect.syntax._
-import cats.effect.{ Blocker, IO }
+import cats.effect.{ Blocker, ContextShift, IO }
 
-final case class MyConfig(somefield: Int, anotherfield: String)
+case class MyConfig(somefield: Int, anotherfield: String)
 
-val load: IO[MyConfig] = Blocker[IO].use { blocker =>
+def load(blocker: Blocker)(implicit cs: ContextShift[IO]): IO[MyConfig] = {
   ConfigSource.file(somePath).loadF[IO, MyConfig](blocker)
 }
 ```
 
 To test that this `IO` does indeed return a `MyConfig` instance:
+
+
+
+
 ```scala
 //Show the contents of the file
 new String(Files.readAllBytes(somePath), StandardCharsets.UTF_8)
@@ -40,7 +44,7 @@ new String(Files.readAllBytes(somePath), StandardCharsets.UTF_8)
 // somefield=1234
 // anotherfield=some string
 
-load.unsafeRunSync().equals(MyConfig(1234, "some string"))
+Blocker[IO].use(load).unsafeRunSync().equals(MyConfig(1234, "some string"))
 // res3: Boolean = true
 ```
 
@@ -48,13 +52,17 @@ load.unsafeRunSync().equals(MyConfig(1234, "some string"))
 
 To create an IO that writes out a configuration file, do as follows:
 
+
+
+
 ```scala
 import pureconfig.module.catseffect._
-import cats.effect.{ Blocker, IO }
+import pureconfig.generic.auto._
+import cats.effect.{ Blocker, ContextShift, IO }
 
 val someConfig = MyConfig(1234, "some string")
 
-val save: IO[Unit] = Blocker[IO].use { blocker =>
-  saveConfigAsPropertyFileF[IO, MyConfig](someConfig, somePath, blocker)
+def save(blocker: Blocker)(implicit cs: ContextShift[IO]): IO[Unit] = {
+  blockingSaveConfigAsPropertyFileF[IO, MyConfig](someConfig, somePath, blocker)
 }
 ```
--- a/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
+++ b/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
@@ -44,8 +44,7 @@ package object catseffect {
    *         `A` from the configuration source, or fail with a ConfigReaderException which in turn contains
    *         details on why it isn't possible
    */
-  def loadF[F[_], A](cs: ConfigSource, blocker: Blocker)
-                    (implicit F: Sync[F], csf: ContextShift[F], reader: Derivation[ConfigReader[A]], ct: ClassTag[A]): F[A] =
+  def loadF[F[_], A](cs: ConfigSource, blocker: Blocker)(implicit F: Sync[F], csf: ContextShift[F], reader: Derivation[ConfigReader[A]], ct: ClassTag[A]): F[A] =
     EitherT(blocker.delay(cs.cursor()))
       .subflatMap(reader.value.from)
       .leftMap(ConfigReaderException[A])
@@ -70,8 +69,7 @@ package object catseffect {
    *         `A` from the configuration files, or fail with a ConfigReaderException which in turn contains
    *         details on why it isn't possible
    */
-  def loadConfigF[F[_], A](blocker: Blocker)
-                          (implicit F: Sync[F], csf: ContextShift[F], reader: Derivation[ConfigReader[A]], ct: ClassTag[A]): F[A] =
+  def loadConfigF[F[_], A](blocker: Blocker)(implicit F: Sync[F], csf: ContextShift[F], reader: Derivation[ConfigReader[A]], ct: ClassTag[A]): F[A] =
     loadF[F, A](ConfigSource.default, blocker)
 
   /**
@@ -166,16 +164,13 @@ package object catseffect {
     outputPath: Path,
     blocker: Blocker,
     overrideOutputPath: Boolean = false,
-    options: ConfigRenderOptions = ConfigRenderOptions.defaults()
-  )(implicit F: Sync[F], csf: ContextShift[F], writer: Derivation[ConfigWriter[A]]): F[Unit] = {
+    options: ConfigRenderOptions = ConfigRenderOptions.defaults())(implicit F: Sync[F], csf: ContextShift[F], writer: Derivation[ConfigWriter[A]]): F[Unit] = {
     if (!overrideOutputPath && Files.isRegularFile(outputPath))
       F.raiseError(
-        new IllegalArgumentException(s"Cannot save configuration in file '$outputPath' because it already exists")
-      )
+        new IllegalArgumentException(s"Cannot save configuration in file '$outputPath' because it already exists"))
     else if (Files.isDirectory(outputPath))
       F.raiseError(
-        new IllegalArgumentException(s"Cannot save configuration in file '$outputPath' because it is a directory")
-      )
+        new IllegalArgumentException(s"Cannot save configuration in file '$outputPath' because it is a directory"))
     else
       Resource.fromAutoCloseable(F.delay(Files.newOutputStream(outputPath))).use { outputStream =>
         blockingSaveConfigToStreamF(conf, outputStream, blocker, options)
@@ -211,8 +206,7 @@ package object catseffect {
     conf: A,
     outputStream: OutputStream,
     blocker: Blocker,
-    options: ConfigRenderOptions = ConfigRenderOptions.defaults()
-  )(implicit F: Sync[F], csf: ContextShift[F], writer: Derivation[ConfigWriter[A]]): F[Unit] =
+    options: ConfigRenderOptions = ConfigRenderOptions.defaults())(implicit F: Sync[F], csf: ContextShift[F], writer: Derivation[ConfigWriter[A]]): F[Unit] =
     F.delay(writer.value.to(conf)).map { rawConf =>
       // HOCON requires UTF-8:
       // https://github.com/lightbend/config/blob/master/HOCON.md#unchanged-from-json
--- a/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
+++ b/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
@@ -70,7 +70,7 @@ package object catseffect {
    *         details on why it isn't possible
    */
   def loadConfigF[F[_], A](blocker: Blocker)(implicit F: Sync[F], csf: ContextShift[F], reader: Derivation[ConfigReader[A]], ct: ClassTag[A]): F[A] =
-    F.delay(ConfigSource.default).flatMap(cs => loadF(cs, blocker))
+    loadF(ConfigSource.default, blocker)
 
   /**
    * Load a configuration of type `A` from the standard configuration files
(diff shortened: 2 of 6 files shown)


modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -6,24 +6,19 @@ import java.nio.file.Path
...
+import cats.effect.{ Blocker, ContextShift, Sync }
 import cats.implicits._
-import com.typesafe.config.{ ConfigRenderOptions, Config => TypesafeConfig }
+import com.typesafe.config.ConfigRenderOptions
 import pureconfig._
 import pureconfig.error.ConfigReaderException
 
 package object catseffect {
 
-  @deprecated("Root will be treated as the default namespace", "0.12.0")
```
ruippeixotog, line 18: We have no fixed rule to delete deprecated methods, but we usually take way more than one breaking version to delete them. Can you please revert all deletions?

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -6,24 +6,19 @@ import java.nio.file.Path
...
-  val defaultNameSpace = ""
+  private[catseffect] type ConfReader[A] = Derivation[ConfigReader[A]]
```
ruippeixotog: We usually don't add a type alias `Derivation[ConfigReader[A]]` on a per-file basis, as it may make the code less readable (e.g. `ConfReader` is hard to distinguish from `ConfigReader`). I am not against creating an alias for this since it's used so heavily in our code base, but if we do it we'll end up doing that on a project-wide basis.

Note that you can still use the nice context bound syntax for most other type classes, you just need to keep the reader as an "explicit" implicit param.

modules/cats-effect/README.md
```diff
@@ -23,11 +23,13 @@ To load a configuration file from a path using cats-effect's `IO`:
 import pureconfig._
 import pureconfig.generic.auto._
 import pureconfig.module.catseffect.syntax._
-import cats.effect.IO
+import cats.effect.{ Blocker, IO }
 
-case class MyConfig(somefield: Int, anotherfield: String)
+final case class MyConfig(somefield: Int, anotherfield: String)
```
ruippeixotog: Is this `final` necessary to make `Blocker` work? If not, could you leave it as it was before?

ruippeixotog (changes requested):
Thanks for taking the time to write this pull request, @BalmungSan!

I don't have much context on `Blocker` and cats-effect in general, so I'll probably defer the review to someone more expert here. @nigredo-tori, do you have more context here? Do you mind reviewing this?

I have two more general comments about the PR:
- Do we really have to do a breaking change here? Can't we work out a solution where we add a new separate method and possibly deprecate the old one?
- Please avoid doing multiple unrelated changes on the same PR. This is currently hard to review because it doesn't just start using Blocker - it also deletes a lot of code and changes the existing documentation in unrelated ways. I'd prefer if you kept this PR focused on Blocker and left the other changes to other PR.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
BalmungSan: Mixing context bounds and implicits are a bad practice. But I really didn't liked the named implicits since neither of them is directly used. However, I understand the problem with the typealias, so if you prefer I can revert to named implicits and  if someday you guys decide to add the alias on the core module, then we could use context bounds.

modules/cats-effect/README.md
BalmungSan: Not really, is just that case classes have to be final _(since extending them is a super bad practice)_ I thought it was minor change, but if you prefer to leave it as it was before, no problem. It was just my OCD :sweat_smile:

BalmungSan:
I thought about removing the deprecations because that minimized the amount of code I needed to change. However, thinking about your suggestion, what about this:

I revert all the deletes, then I also deprecate the current methods. And finally I add new methods we do use blocker.

What do you think @ruippeixotog?
If you like that idea, then I have three follow up questions. 
1.What should the tests do? Test only the non deprecated or test bothm?
2. What should the documentation show? Only the recommended version using blocker or both?
3. What should be the name of the new methods? Because I really like the current names.

modules/cats-effect/README.md
```diff
@@ -23,11 +23,13 @@ To load a configuration file from a path using cats-effect's `IO`:
...
 
-val load: IO[MyConfig] = ConfigSource.file(somePath).loadF[IO, MyConfig]
+val load: IO[MyConfig] = Blocker[IO].use { blocker =>
+  ConfigSource.file(somePath).loadF[IO, MyConfig](blocker)
```
nigredo-tori, line 31: It's better to make this a `def load(blocker: Blocker)(implicit cs: ContextShift[IO]): IO[MyConfig]` - creating a `Blocker` for a single operation is almost always a mistake. We can create a `Blocker` and a `ContextShift` in the "test code" below. Same with `save` later in this file.

modules/cats-effect/src/main/tut/README.md
```diff
@@ -23,17 +23,21 @@ import java.nio.charset.StandardCharsets
...
 import pureconfig.module.catseffect.syntax._
-import cats.effect.IO
+import cats.effect.{ Blocker, IO }
 
-case class MyConfig(somefield: Int, anotherfield: String)
+final case class MyConfig(somefield: Int, anotherfield: String)
 
-val load: IO[MyConfig] = ConfigSource.file(somePath).loadF[IO, MyConfig]
+val load: IO[MyConfig] = Blocker[IO].use { blocker =>
+  ConfigSource.file(somePath).loadF[IO, MyConfig](blocker)
```
nigredo-tori, line 37: Again, we shouldn't teach people to allocate a `Blocker` for a single operation. Same with `save` later.

modules/cats-effect/src/test/scala/pureconfig/module/catseffect/CatsEffectSuite.scala
```diff
@@ -11,39 +11,41 @@ import java.nio.file.{ Path, Paths }
...
+  private implicit val ioCS = IO.contextShift(scala.concurrent.ExecutionContext.global)
 
   private def getPath(classPathPath: String): Path = {
     val resource = getClass.getClassLoader.getResource(classPathPath)
     Paths.get(resource.toURI)
   }
 
   "loadConfigF" should "run successfully when correctly formatted file is in place" in {
-    val load = loadConfigF[IO, SomeCaseClass]
+    val load = Blocker[IO].use(loadConfigF[IO, SomeCaseClass])
```
nigredo-tori: We should probably use a single `Blocker` for the whole suite. Something like this would be enough for now:
```scala
val blocker: Blocker =
  Blocker.liftExecutionContext(scala.concurrent.ExecutionContext.global)
```

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -6,24 +6,19 @@ import java.nio.file.Path
...
+  private[catseffect] type ConfWriter[A] = Derivation[ConfigWriter[A]]
 
-  def loadF[F[_], A](cs: ConfigSource)(implicit F: Sync[F], reader: Derivation[ConfigReader[A]], ct: ClassTag[A]): F[A] = {
-    val delayedLoad = F.delay {
-      cs.load[A].leftMap[Throwable](ConfigReaderException[A])
-    }
-    delayedLoad.rethrow
-  }
+  def loadF[F[_]: Sync: ContextShift, A: ConfReader: ClassTag](cs: ConfigSource, blocker: Blocker): F[A] =
+    blocker.delay(cs.load[A].leftMap[Throwable](ConfigReaderException[A])).rethrow
```
nigredo-tori: Note that reading `A` is still done in the blocking context. Ideally, this would be something like
```scala
EitherT(blocker.delay[F] { cs.cursor() })
  // back in `ContextShift`
  .subflatMap(_.to[A])
  .leftMap(ConfigReaderException[A])
  .rethrowT
```
Although, this would potentially be confusing if someone overrides `load` in their `ConfigSource` implementation. @ruippeixotog, WDYT?

nigredo-tori (changes requested):
> @nigredo-tori, do you have more context here? Do you mind reviewing this?

Not at all. I agree with the motivation behind this PR. Basically, in the `cats-effect` ecosystem it's customary to separate two kinds of processing.
1. CPU-intensive operations are handled by a thread pool with limited amount of threads (usually one per CPU core), to avoid the overhead of running two threads on one core. This thread pool is abstracted as `ContextShift`.
1. Blocking operations (the ones that can block the thread they are executed on) are executed on a separate thread pool with unbounded number of threads. This makes it really hard to inadvertently get yourself into a deadlock. This unbounded threadpool is abstracted as `Blocker`.

Since most of the `ConfigSource` implementations have to rely on blocking IO, it makes sense to keep them in `Blocker`, which is what this PR does.

> Can't we work out a solution where we add a new separate method and possibly deprecate the old one?

We should be able to. Since the new methods have differing numbers of arguments, we can keep the old ones as `@deprecated` overloads.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
BalmungSan: Yeah, I thought about the same. Ideally only the blocking part should be done inside blocker, but I wasn't sure which method should have I called instead.

Maybe, `load` and `loadOrThrow` could / should be final?

ruippeixotog: I agree with @nigredo-tori suggestion and with making `load` and `loadOrThrow` final, as they were never intended to be overridden in the first place.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
ruippeixotog: I understand it's a bad practice and I totally understand the preference for context bounds instead of named implicits. However, for the sake of keeping the PR focused on a single thing and avoiding a confusing type alias let's leave this as it was, if that's ok with you :)

modules/cats-effect/README.md
ruippeixotog: Yes, please revert it if it's ok with you. I don't have a strong opinion for or against this (even though I prefer documentation in particular to be as simple and concise as possible contrary to lib design, where we want to have all safety we can get through annotations and qualifiers). However, this PR is not the space to do it. If we want do it later, we'll do it project-wide and not just on a single module.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -111,6 +147,25 @@ package object catseffect {
...
+   * Save the given configuration into a property file
+   *
+   * @param conf The configuration to save
+   * @param outputPath Where to write the configuration
+   * @param blocker the blocking context which will be used to load the configuration.
+   * @param overrideOutputPath Override the path if it already exists
+   * @param options the config rendering options
+   * @return The return action will save out the supplied configuration upon invocation
+   */
+  def blockingSaveConfigAsPropertyFileF[F[_], A](
```
ruippeixotog, line 162: Can't we keep the same name just like we did for `loadConfigF`?

modules/cats-effect/src/main/tut/README.md
```diff
@@ -23,17 +23,21 @@ import java.nio.charset.StandardCharsets
...
 
 ```tut:silent
 import pureconfig._
 import pureconfig.module.catseffect.syntax._
-import cats.effect.IO
+import cats.effect.{ Blocker, IO }
 
-case class MyConfig(somefield: Int, anotherfield: String)
+final case class MyConfig(somefield: Int, anotherfield: String)
```
ruippeixotog: Ditto here.

ruippeixotog (changes requested):

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
BalmungSan: No we can't, scalac prints an error because two overloaded methods can not have both default arguments.

I wanted to ask about this, we have four options:

+ Remove the defaults from the deprecated method _(probably this breaks code, so not an option)_.
+ Remove the defaults from the new method, but I guess this makes less pleasant to use this new method.
+ Different names.
+ ~Do not add these two methods, since an user can easily do this: `blocker.blockOn(oldMethod)` but providing the method in the library seems nice.~ Not longer valid, since the implementations are now different.

I opted for renaming _(but I am not happy with the new names)_. What do you think?

ruippeixotog: Makes sense, I forgot about the default params. These methods are rather old and haven't been updated in a long time. We should rethink what API we want to offer on the write path, but for now I think it's fine to just rename 👍

BalmungSan:
@nigredo-tori @ruippeixotog I just pushed the changes with all the suggestions. Let me know what do you think about them.

-----

~Uhm, the test pass in my local, do you have any idea of what happened and how can I fix it?~

### Edit

Fixed!

nigredo-tori (approved):
LGTM. 👍 I'm not too happy with the `blocking*` functions' names, but there isn't much we can do about it without breaking the existing code.

nigredo-tori:
About the failing test - I can't see any errors in the logic itself. I don't have the time to personally get into debugging this at the moment, but you could try printing out the result of `loadConfigF[Config](ConfigSource.default)` and `blocker.delay { ConfigSource.default.load[Config] }` in the same environment. I fear that you might get an empty `Config` in both, which would signal a class loader issue (`ConfigSource.default` is using a thread context class loader, and we've just put that into a global execution context) . As a temporary solution, we could fork the cats-effect tests. A less invasive alternative would be to build a new `Blocker` for the test suite (via `Blocker.use`, or some other way), instead of using the global execution context. In any case, if the problem is, indeed, class loaders, we should make sure we address that at some point.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -70,7 +70,7 @@ package object catseffect {
    *         details on why it isn't possible
    */
   def loadConfigF[F[_], A](blocker: Blocker)(implicit F: Sync[F], csf: ContextShift[F], reader: Derivation[ConfigReader[A]], ct: ClassTag[A]): F[A] =
-    loadF[F, A](ConfigSource.default, blocker)
+    F.delay(ConfigSource.default).flatMap(cs => loadF(cs, blocker))
```
nigredo-tori:
```suggestion
    loadF(ConfigSource.default, blocker)
```

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -164,18 +164,28 @@ package object catseffect {
...
-    else if (Files.isDirectory(outputPath))
-      F.raiseError(
-        new IllegalArgumentException(s"Cannot save configuration in file '$outputPath' because it is a directory"))
-    else
-      Resource.fromAutoCloseable(F.delay(Files.newOutputStream(outputPath))).use { outputStream =>
-        blockingSaveConfigToStreamF(conf, outputStream, blocker, options)
-      }
-  }
+    options: ConfigRenderOptions = ConfigRenderOptions.defaults())(implicit F: Sync[F], csf: ContextShift[F], writer: Derivation[ConfigWriter[A]]): F[Unit] =
+    F.pure(overrideOutputPath).flatMap {
```
nigredo-tori: The old implementation (before the last commit) was much easier to understand - we can just wrap it in `F.suspend { ... }` to capture any side effects, and be done with it.

BalmungSan: I agree it was easier to understand, however, if you look carefully the problem is that some operations are blocking and others not. So after playing with it, I thought this was the best way.

nigredo-tori: Ah, sure, I forgot about blocking. Still, you can see that both `isRegularFile` and `isDirectory` are blocking, so we can use `blocker.delay[F] { ... }.flatten` in a similar fashion.

BalmungSan: @nigredo-tori That won't work, because that would execute the blocking operations in the compute pool when called `flatten` and would use the blocking context just to construct the inner `F` which is not a blocking operation. I am thinking on alternatives, I will let you know.

BalmungSan: @nigredo-tori Let me know what do you think of the new implementation.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -111,6 +149,39 @@ package object catseffect {
...
+    val fileIsDirectory =
+      F.raiseError(new IllegalArgumentException(s"Cannot save configuration in file '$outputPath' because it is a directory"))
+
+    val check = F.suspend {
+      if (!overrideOutputPath && Files.isRegularFile(outputPath)) fileAlreadyExists
+      else if (Files.isDirectory(outputPath)) fileIsDirectory
+      else F.unit
+    }
+
+    blocker.blockOn(check) >> Resource.fromAutoCloseable(F.delay(Files.newOutputStream(outputPath))).use { outputStream =>
```
nigredo-tori: Actually, `newOutputStream` (and  the corresponding `close`) is blocking as well. So we have to do
```scala
Resource.fromAutoCloseable(F.delay(Files.newOutputStream(outputPath)))
  .mapK(blocker.blockOnK[F])
  .use { ... }
```

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
nigredo-tori: > would execute the blocking operations in the compute pool when called flatten

I'm not quite sure what you meant here, but since we've moved on from that implementation, that doesn't really matter.

> would use the blocking context just to construct the inner F which is not a blocking operation.

I think that's splitting hairs. :smile: But I'm happy with the current implementation (except for the `newOutputStream` thing). I think it does good job with separating path validation from writing the file.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
BalmungSan: ~I am not totally sure if those two methods are blocking or not _(I believe not)_. Especially because the official example of `fromAutoClosable` doesn't involve a **Blocker** anywhere.~

I confirmed in the **cats-effect** channel that they are in fact blocking. I am proceeding to change the code.

ruippeixotog (approved):
LGTM, thank you both for the work here! @nigredo-tori, are you ok with the PR in its current state?

nigredo-tori:
@ruippeixotog, I am.

PR #699, opened 2020-02-10
Http4s 0.21.0

ChristopherDavenport:
- Is Binary Breaking
- Http4s 0.21.0 only supports 2.12/2.13


ruippeixotog (approved):
Thanks @ChristopherDavenport!

PR #703, opened 2020-02-20
Fix pureconfig-magnolia not using existing pureconfig instances in auto mode, e.g. when deriving collection instances

### LINKED ISSUE #107: Separate exported instance type and internal dispatch/combine type
This is a more principled version of #89. I found at least two cases when the typeclass intance being derived would have a type different from the typeclass used for `dispatch` and `combine`.

### 1) `ObjectEncoder[A]`/`Encoder[A]`

In circe, generic derivation provides not just an `Encoder[A]`, but a more specific `ObjectEncoder[A]`, which means that the serialization result is always going to be a json object. But for all the inner steps of the derivation, a more general `Encoder[A]` typeclass is looked for and used. This makes sense - object internals don't have to be objects themselves.

### 2) `Exported[Decoder[A]]`/`Decoder[A]`

This is how prioritization of derived instances works in circe. `auto` derives an instance of `Exported[Decoder[A]]` ([code](https://github.com/circe/circe/blob/master/modules/generic/shared/src/main/scala/io/circe/generic/AutoDerivation.scala#L18)). And there's a [low-priority implicit conversion](https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Encoder.scala#L470) `Exported[Decoder[A]] => Decoder[A]`, which allows the default instances to be selected before the derived ones.

But of course, under the hood derivation of `Exported[Decoder[A]]` combines and dispatches on just `Decoder[A]` (otherwise default instances would not be picked up).

### Possible Solution

I feel that this is not specific to circe and similar concerns might arise in other contexts.

I understand this would complicate things quite a bit, but ideally for proper circe auto derivation we'd need something like this:

```scala
object MagnoliaDecoder {

  type ExportedTypeclass[T] = Exported[Decoder[T]]
  type InternalLookup[T] = Decoder[T]

  // run outer level combine for the exported instance
  // ExportedCaseClass is the same as CaseClass, but parametersArray 
  // will already have InternalLookup instances 
  def combineExported[T](
    caseClass: ExportedCaseClass[ExportedTypeclass, InternalLookup, T]): ExportedTypeclass[T] = ???

  // same idea for the dispatch on the outer level instance
  def dispatchExported[T](
    sealedTrait: ExportedSealedTrait[ExportedTypeclass, InternalLookup, T]): ExportedTypeclass[T] = ???

  // all the internal derivation works in the same way as before - using InternalLookup

  def combine[T](caseClass:    CaseClass[InternalLookup, T]): InternalLookup[T] = ???
  def dispatch[T](sealedTrait: SealedTrait[InternalLookup, T]): InternalLookup[T] = ???

  implicit def magnoliaDecoder[T]: ExportedTypeclass[T] = macro Magnolia.gen[T]
```

WDYT, how hard is that gonna be?

P.S. Thanks for 0.8.0, semiauto seems to be working perfectly! :+1:

neko-kai:
* Also forces all magnolia derivations to run in `blackbox.Context` (after typer) for performance. Seems magnolia doesn't use whitebox capabilities anyway
* Fixes https://github.com/propensive/magnolia/issues/107 for pureconfig-magnolia
* Ported from `izumi` - https://github.com/7mind/izumi/pull/915

EDIT: /cc'ing maintainters present in `pureconfig-magnolia` PR


modules/magnolia/src/main/scala/pureconfig/module/magnolia/ExportedMagnolia.scala
```diff
@@ -0,0 +1,24 @@
...
+// Thankfully at least it doesn't check the output type of its `macroApplication`
+object ExportedMagnolia {
+  def exportedMagnolia[TC[_], A](c: blackbox.Context)(implicit t: c.WeakTypeTag[A]): c.Expr[Exported[TC[A]]] = {
+    val magnoliaTree = forcedBlackboxMagnolia[TC, A](c)
+    c.universe.reify(Exported(magnoliaTree.splice))
+  }
+
+  // if i TELL you it's whitebox, you BETTER believe me
+  def forcedBlackboxMagnolia[TC[_], A](c: blackbox.Context)(implicit t: c.WeakTypeTag[A]): c.Expr[TC[A]] = {
+    val c0 = c.asInstanceOf[whitebox.Context]
```
neko-kai: This is safe because the actual Context instance always implements both blackbox & whitebox interfaces:

```scala
abstract class Context extends scala.reflect.macros.blackbox.Context
                          with scala.reflect.macros.whitebox.Context
```

modules/magnolia/src/test/scala/pureconfig/module/magnolia/DerivationModesSuite.scala
```diff
@@ -58,4 +83,12 @@ class DerivationModesSuite extends BaseSuite {
     ConfigReader[Entity].from(conf.root) shouldBe Right(person)
     ConfigWriter[Entity].to(person) shouldBe conf.root()
   }
+
+  it should "use existing reader and writer instances when they exist" in {
```
jcazevedo, line 88: Can you also include this test in the `DerivationModesSuite` of [`pureconfig-tests`](https://github.com/pureconfig/pureconfig/blob/master/tests/src/test/scala/pureconfig/DerivationModesSuite.scala)? We're currently not reusing the base test suite for the generic derivation modules, but I think a test like this is also relevant to prevent a regression from being introduced in the shapeless-based derivation.

jcazevedo (changes requested):
@neko-kai: Thanks for the PR! I approve the usage of the `Exported` strategy. 

However, I'm not fully aware of the implications of forcing a blackbox context. Can you move that change to a different PR so that we evaluate its inclusion outside this derivation fix?

modules/magnolia/src/test/scala/pureconfig/module/magnolia/DerivationModesSuite.scala
neko-kai: @jcazevedo Included!

neko-kai:
@jcazevedo Done. Hmm, I've tried to move the `magnolia` macro itself to blackbox and while it compiles, its test suite after the change causes compiler StackOverflows... I guess current usages of `distage-extension-config`/`pureconfig-magnolia` doesn't trigger the error somehow, but this requires more thought...

jcazevedo (approved):
> Hmm, I've tried to move the `magnolia` macro itself to blackbox and while it compiles, its test suite after the change causes compiler StackOverflows...

This makes me more wary of including that change here. 😅 

This PR looks good to me as is. Let's just wait for @ruippeixotog's review before merging. Thanks again for this!

modules/magnolia/src/main/scala/pureconfig/module/magnolia/semiauto/writer.scala
```diff
@@ -1,13 +1,13 @@
...
 
-import scala.language.experimental.macros
-import scala.reflect.ClassTag
-
 import magnolia._
 import pureconfig.ConfigWriter
 import pureconfig.generic.{ CoproductHint, ProductHint }
 import pureconfig.module.magnolia.MagnoliaConfigWriter
 
+import scala.language.experimental.macros
```
ruippeixotog: nit: can we revert all import changes in files that you didn't touch? This tends to pollute the commit log for a file and makes it harder to understand what really changed at a glance.

ruippeixotog (approved):

modules/magnolia/src/main/scala/pureconfig/module/magnolia/semiauto/writer.scala
neko-kai: @ruippeixotog @jcazevedo I removed import changes and unnecessary diffs (adopting scalafmt may help ensure this doesn't happen again... actually I assumed Intellij's formatting changes were unimportant _because_ there was no scalafmt dictating the style for the repo, Sorry! 😓)

PR #733, opened 2020-04-04
replace ConfigValueLocation with ConfigOrigin

mberndt123:
Hey,

I've noticed that pureconfig doesn't report error locations properly for configurations that were parsed from Strings. When [parsing a String](https://lightbend.github.io/config/latest/api/com/typesafe/config/ConfigFactory.html#parseString-java.lang.String-com.typesafe.config.ConfigParseOptions-) with the Typesafe Config library, it's possible to pass a `ConfigParseOptions` object, and that object may contain a textual description of the origin set with [`setOriginDescription`](https://lightbend.github.io/config/latest/api/com/typesafe/config/ConfigParseOptions.html#setOriginDescription-java.lang.String-). That string should show up somewhere in configuration error messages, but the current implementation in pureconfig only considers URLs when creating a `ConfigValueLocation` from a `ConfigOrigin`, so it will not work for this case.

I think that `ConfigValueLocation` is actually not needed, because `ConfigOrigin` does everything that it does and more. So I propose to simply remove `ConfigValueLocation` altogether and just use `ConfigOrigin` instead, because the library is inextricably linked to Typesafe Config anyway and there's no need to duplicate their data structures.


mberndt123:
OK, CI seems to be failing because of documentation. Any hints on how to fix that?

jcazevedo:
> OK, CI seems to be failing because of documentation. Any hints on how to fix that?

The CI is failing because the output of the PureConfig's site changed. The check is there to catch unintended changes. It's OK to fail that check if the changes are intended, as it seems to be the case.

mberndt123:
Yes it is indeed intentional. This is an API/ABI change, and so it may break users' code. If you think it's necessary, it's probably possible to find a way to implement this that doesn't break the ABI. Let me know what you think…

mberndt123:
Yes, this is a API/ABI-breaking change… I'd love to implement it without breaking users, but the way things stand, the URL is a non-optional field of the `ConfigValueLocation` class, and the whole point here is to allow `ConfigValueLocation`s that don't have that field.

mberndt123:
@ruippeixotog, @jcazevedo, can you provide any feedback on this idea?

ruippeixotog:
@mberndt123 thanks for the PR! I'll take a look at this on the weekend.

mberndt123:
Great, looking forward for your comments!

tests/src/main/scala/pureconfig/ConfigReaderMatchers.scala
```diff
@@ -35,4 +40,8 @@ trait ConfigReaderMatchers { this: AnyFlatSpec with Matchers =>
         inside(left) { case Left(ConfigReaderFailures(failure, Nil)) => pf.valueAt(failure) }
       }
     }
+
+  val hardcodedValueLocation = ConfigOriginFactory.newSimple("hardcoded value")
```
jcazevedo: Can you add a comment here explaining where this comes from? It took me a while to figure out that this was what Typesafe Config was using for hardcoded values.

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -32,7 +32,7 @@ sealed trait ConfigCursor {
   /**
    * The file system location of the config to which this cursor points.
    */
-  def location: Option[ConfigValueLocation] = ConfigValueLocation(value)
+  def location: ConfigOrigin = if (isUndefined) ConfigOriginFactory.newSimple() else value.origin()
```
jcazevedo: I think this should still be an `Option`. I don't think that a cursor pointing to an undefined value should have an origin.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -23,7 +24,7 @@ trait ConfigReaderFailure {
   /**
    * The optional location of the failure.
    */
-  def location: Option[ConfigValueLocation]
+  def location: ConfigOrigin
```
jcazevedo: I think this should be an `Option`. We have failures that we can't relate to a given origin (the `CannotRead` ones, for example).

modules/yaml/src/main/scala/pureconfig/module/yaml/YamlConfigSource.scala
```diff
@@ -122,17 +121,17 @@ final class YamlConfigSource private (
...
       case NonFatal(e) =>
-        Result.fail(ThrowableFailure(e, None))
+        Result.fail(ThrowableFailure(e, ConfigOriginFactory.newSimple()))
     }
   }
 
   // Converts a SnakeYAML `Mark` to a `ConfigValueLocation`, provided the file path.
-  private[this] def toConfigValueLocation(path: URL, mark: Mark): ConfigValueLocation = {
-    ConfigValueLocation(path, mark.getLine + 1)
+  private[this] def toConfigValueLocation(path: URL, mark: Mark): ConfigOrigin = {
```
jcazevedo: This should be renamed to `toConfigOrigin`.

tests/src/main/scala/pureconfig/ConfigReaderMatchers.scala
```diff
@@ -35,4 +40,8 @@ trait ConfigReaderMatchers { this: AnyFlatSpec with Matchers =>
...
+  val string1ValueLocation = stringValueLocation(1)
```
jcazevedo: Can you also add a comment to this explaining its uses?

tests/src/test/scala/pureconfig/ConfigReaderExceptionSuite.scala
```diff
@@ -161,38 +162,51 @@ class ConfigReaderExceptionSuite extends AnyFlatSpec with Matchers {
...
 
     val exception = intercept[ConfigReaderException[_]] {
       conf.get("conf").toOrThrow[Conf]
     }
 
-    exception.failures.toList.toSet shouldBe Set(
-      ConvertFailure(KeyNotFound("a", Set()), Some(ConfigValueLocation(url, 1)), ""),
-      ConvertFailure(WrongType(ConfigValueType.STRING, Set(ConfigValueType.NUMBER)), Some(ConfigValueLocation(url, 3)), "c"))
+    val List(fail1, fail2) = exception.failures.toList
+    fail1 match {
```
jcazevedo: Can we also use `inside` for this?

tests/src/test/scala/pureconfig/ConfigReaderExceptionSuite.scala
```diff
@@ -161,38 +162,51 @@ class ConfigReaderExceptionSuite extends AnyFlatSpec with Matchers {
...
+      ConvertFailure(KeyNotFound("camel-case-string", Set("camelCaseString")), stringValueLocation(2), "camel-case-conf"),
+      ConvertFailure(KeyNotFound("snake-case-int", Set("snake_case_int")), stringValueLocation(10), "snake-case-conf"),
+      ConvertFailure(KeyNotFound("snake-case-string", Set("snake_case_string")), stringValueLocation(10), "snake-case-conf"))
   }
 
   it should "have failures with the proper file system location of the values that raised errors, if available" in {
     val workingDir = getClass.getResource("/").getFile
     val file = "conf/configFailureLocation/single/a.conf"
-    val url = new URL("file://" + workingDir + file)
+    val location = s"$file @ file:$workingDir$file"
```
jcazevedo: This is now unused.

jcazevedo (changes requested):
@mberndt123: Thanks for the PR!

I'm OK with using `ConfigOrigin` instead of `ConfigValueLocation`. However, I don't think we should assume that all cursors and all failures have an origin.

I'd also prefer renaming mentions of "location" to "origin", for consistency.

tests/src/main/scala/pureconfig/ConfigReaderMatchers.scala
mberndt123: I now removed this and replaced it with this:
```
val emptyConfigOrigin: Option[ConfigOrigin] =
  Some(ConfigOriginFactory.newSimple())
```
I think that makes it sufficiently obvious that this is a default value.

core/src/main/scala/pureconfig/ConfigCursor.scala
mberndt123: Well, I think there shouldn't even be a ConfigCursor that points to an undefined value, because `value` is not an `Option` and we usually pretend `null` doesn't exist in the Scala world… Anyway, I've made it optional.

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
mberndt123: OK

modules/yaml/src/main/scala/pureconfig/module/yaml/YamlConfigSource.scala
mberndt123: OK

tests/src/main/scala/pureconfig/ConfigReaderMatchers.scala
mberndt123: I've removed it, because `stringValueLocation(1)` is almost as easy to type.

tests/src/test/scala/pureconfig/ConfigReaderExceptionSuite.scala
mberndt123: Yes. The reason I hadn't done that in the first place was that `inside` uses a type parameter called `A1`, and there's a `case class` defined in the file that is also called `A1`, and that caused a compiler error message. I've now renamed the case class to `AA1`.

tests/src/test/scala/pureconfig/ConfigReaderExceptionSuite.scala
mberndt123: It is also now gone.

mberndt123:
I think I have addressed your comments, let me know what you think. Btw, the diffstat is now -313+313 :smiley:

mberndt123:
Hey @jcazevedo ,

Do you think you might get around to reviewing this over the next couple of days?

mberndt123:
Hey @jcazevedo and @ruippeixotog,

I would still like to get this merged, is there anything I can do to make this happen?

I've now updated the branch to the latest master, so there are no more conflicts.

mberndt123:
Any idea why the CI job fails on 2.13? It works on my machine :man_shrugging:

ruippeixotog (approved):
LGTM! Sorry for the long wait, @mberndt123. I'll merge this tomorrow unless there are further comments until then 👍

mberndt123:
> LGTM! Sorry for the long wait, @mberndt123. I'll merge this tomorrow unless there are further comments until then :+1: 

Awesome, thanks!

jcazevedo (approved):
LGTM

mberndt123:
:partying_face:

PR #744, opened 2020-04-18
Update to magnolia 0.15.0

neko-kai:


ruippeixotog:
Thanks for the quick fix, @neko-kai! What's the reasoning for adding `scala-reflect` as provided? Will `pureconfig-magnolia` users need to add `scala-reflect` manually to their dependencies?

neko-kai:
@ruippeixotog So that it's not required at runtime (it's not used at runtime, it's only a dependency of macros)
> Will pureconfig-magnolia users need to add scala-reflect manually to their dependencies?

No, it's required only to declare `def gen = macro Magnolia.gen`, but users don't have to declare macros to use `pureconfig-magnolia` - macros will execute without scala-reflect

mberndt123:
An exclusion should also be added for the scala-compiler jar, because magnolia pulls that in too.

ruippeixotog:
Sorry for the slow response, @neko-kai. This is quite a strange setup... Is it a common practice for macro projects to have `scala-compiler` as dependency and libraries using macros to manually exclude `scala-reflect` afterwards?

LGTM anyway, but could you please add a comment in `build.sbt` explaining why this is needed? Thank you for fixing this!

neko-kai:
@ruippeixotog No, it's not common practice, just caused by propensive not being too vigilant about `Provided` scope for compile-time only macro dependencies. The exclusion of scala-compiler here will go away once `magnolia-0.15.1` is released with [this pr](https://github.com/propensive/magnolia/pull/237) or newer mercator is released with [this pr](https://github.com/propensive/mercator/pull/20/files)

neko-kai:
I've added a comment with explanation

mberndt123:
@ruippeixotog,
The reason that magnolia pulls in scala-compiler is that magnolia depends on mercator, and mercator didn't use `% Provided` for that dependency. So you can bug propensive about the following PR, and maybe it will be fixed:
https://github.com/propensive/mercator/pull/20

ruippeixotog (approved):

ruippeixotog:
LGTM. Do you mind updating this branch with the latest master before we merge it? It seems that I am unable to do it myself for this PR.

neko-kai:
updated

PR #747, opened 2020-04-24
Add example of semiauto derivation for sealed families

leifwickland:


docs/docs/docs/non-automatic-derivation.md
```diff
@@ -75,6 +75,39 @@ We are now ready to read `Person` configs:
 ConfigSource.fromConfig(conf).load[Person]
 ```
 
+### Semi-Automatic for Sealed Families
```
ruippeixotog: nit: can you either remove this title or make it an h4? It looks like a subsection of "Semi-Automatic", not on the same level as "Automatic" and "Manual".

ruippeixotog (commented):
Thanks @leifwickland!

docs/docs/docs/non-automatic-derivation.md
leifwickland: @ruippeixotog Thanks for catching that. I meant to make it an h4 and simply missed.

ruippeixotog (approved):

PR #750, opened 2020-04-29
Apply some minor fixes to example about semi-automatic derivation of families

jcazevedo:
This fixes a typo in occupation, starts using different names for implicits to avoid shadowing and fills a paragraph at 120 chars, to be consistent with the rest of the document.


leifwickland (approved):
Thanks, @jcazevedo.

PR #756, opened 2020-05-08
be Scala 2.13.3 compatible: don't rely on Symbol#toString to make key names

SethTisue:
because `Symbol#toString` will change in Scala 2.13.3 to be `Symbol(foo)`
instead of `'foo`: https://github.com/scala/scala/pull/8933

using `.name` instead works as expected in any Scala version


jcazevedo (approved):
@SethTisue: thanks!

PR #762, opened 2020-05-20
Upgrades sttp to latest version

sebver:
The latest versions of sttp use a different artifact: https://github.com/softwaremill/sttp-model

Unfortunately the package name has changed to just `sttp` instead of `com.softwaremill.sttp`, which conflicts with the pureconfig module package name. Therefore I had to prefix the imports with `_root_`.


ruippeixotog (approved):
Thanks, @sebver!

PR #763, opened 2020-05-26
Update sttp

guizmaii:


ruippeixotog:
@guizmaii is there any reason to require `com.softwaremill.sttp.client` and not just `com.softwaremill.sttp.model`? @sebver updated sttp recently in #762 and chose to use `model` (which makes sense to me, since we just have a reader for URIs here).

guizmaii:
Hi @ruippeixotog,

sttp, for its v2, moved to the `com.softwaremill.sttp.client` namespace/organisation name. `com.softwaremill.sttp.model` is not used anymore.

ruippeixotog:
@guizmaii looking at [their GitHub page](https://github.com/softwaremill/sttp) it doesn't seem to be the case. sttp moved from `com.softwaremill.sttp` on v1 to `com.softwaremill.sttp.client` on v2, with the core data structures and models being extracted to `com.softwaremill.sttp.model` (and `client` [depending on it](https://mvnrepository.com/artifact/com.softwaremill.sttp.client/core_2.13/2.1.5)).

Since this module only depends on `model`, we should keep targeting it.

PR #766, opened 2020-05-31
Use more consistent names for type parameters

jcazevedo:
Following up on https://github.com/pureconfig/pureconfig/pull/629#discussion_r429694129, this attempts to make the naming conventions for type parameters more consistent throughout the codebase.


modules/generic/src/main/scala/pureconfig/generic/DerivedConfigReader.scala
```diff
@@ -12,24 +12,24 @@ trait DerivedConfigReader[A] extends ConfigReader[A]
 
 object DerivedConfigReader extends DerivedConfigReader1 {
 
-  implicit def anyValReader[T, U](
+  implicit def anyValReader[A, Wrapped](
```
ruippeixotog, line 17: `Wrapped` is ambiguous as it can refer equally well to the value _that is_ wrapped and to the value _after_ wrapped. Here we are using it as the type of the value when unwrapped, but we use it in the opposite way in shapeless generic type classes :/

Can you either use them consistently or rename them to a less ambiguous option?

ruippeixotog (changes requested):
LGTM overall, thanks!

modules/generic/src/main/scala/pureconfig/generic/DerivedConfigReader.scala
jcazevedo: I'd rather keep this `Wrapped` (it is at least more aligned with the `W` from shapeless' `Unwrapped`). Maybe rename the other `Wrapped` to `Original`?

jcazevedo: I've renamed `Wrapped` to `Original` in shapeless generic type classes in 7ab22c8.

ruippeixotog (approved):
Thanks again :)

PR #771, opened 2020-06-11
Use a different test suite for collection converters

jcazevedo:
Since we provide `ConfigReader`s and `ConfigWriter`s for collections in a different trait, it helps to locate their tests by using a similarly named suite.


leifwickland (approved):
Nice catch, @jcazevedo

ruippeixotog (approved):

PR #773, opened 2020-06-11
Improve error message for when no valid coproduct option is found

### LINKED ISSUE #356: Make FirstSuccessCoproductHint NoValidCoproductChoiceFound error user friendly

jcazevedo:
Closes #356.

This improves the error message for the `NoValidCoproductOptionFound` failure reason by including the failures for all the attempted options.

Incidentally, this also:
* Fixes typos on "ident" (should've been "indent");
* Moves the `indentSize` to the `description` method of failure reasons, so that we can use consistent indentation sizes from the `ConfigReaderFailures#prettyPrint` method;
* Makes sure that available coproduct options provided to the `CoproductHint#from` method are sorted.

### DIFF
--- a/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
+++ b/core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
@@ -18,8 +18,10 @@ trait ConfigReaderFailure {
 
   /**
    * A human-readable description of the failure.
+   *
+   * @param indentSize the number of characters to use for any indentation applied inside the description
    */
-  def description: String
+  def description(indentSize: Int): String
 
   /**
    * The optional origin of the failure.
@@ -38,7 +40,7 @@ trait ConfigReaderFailure {
 case class ConvertFailure(reason: FailureReason, origin: Option[ConfigOrigin], path: String)
   extends ConfigReaderFailure {
 
-  def description = reason.description
+  def description(indentSize: Int) = reason.description(indentSize)
 }
 
 object ConvertFailure {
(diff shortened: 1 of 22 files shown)


core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
```diff
@@ -18,8 +18,10 @@ trait ConfigReaderFailure {
 
   /**
    * A human-readable description of the failure.
+   *
+   * @param indentSize the number of characters to use for any indentation applied inside the description
    */
-  def description: String
+  def description(indentSize: Int): String
```
ruippeixotog: It doesn't seem ok to pollute the `ConfigReaderFailure` API for implementations to handle presentation-specific concerns. Is there really any pressing need to make indentation configurable, or can we simply go with standardizing all descriptions to use 2 spaces?

Even if we really want to make indentation configurable (which I'm not so sure of), I would still prefer to encode the description structure in the type system with some `case class Description(text: String, causes: Seq[Description])`, which a pretty printer can render accurately.

ruippeixotog (changes requested):
Thanks @jcazevedo, this is really useful :)

core/src/main/scala/pureconfig/error/ConfigReaderFailure.scala
jcazevedo: I'm OK with always going with 2 spaces, but then I'd rather disallow configuring the indentation size in the `prettyPrint` method of `ConfigReaderFailures`, so that the output is consistent. Did that in 6f10a2f.

ruippeixotog (approved):

PR #774, opened 2020-06-11
Have a more convenient apply method for `ConfigReaderFailures`

jcazevedo:
Modifies the alternative `apply` method of `ConfigReaderFailures` to make them more convenient to create.

### DIFF
--- a/core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
+++ b/core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
@@ -10,7 +10,7 @@ import scala.collection.mutable
  */
 case class ConfigReaderFailures(head: ConfigReaderFailure, tail: ConfigReaderFailure*) {
 
-  lazy val toList: List[ConfigReaderFailure] = head :: tail.toList
+  def toList: List[ConfigReaderFailure] = head :: tail.toList
 
   def +:(failure: ConfigReaderFailure): ConfigReaderFailures =
     ConfigReaderFailures(failure, this.toList: _*)
--- a/core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
+++ b/core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
@@ -52,6 +52,6 @@ case class ConfigReaderFailures(head: ConfigReaderFailure, tail: List[ConfigRead
 
 object ConfigReaderFailures {
 
-  def apply(configReaderFailure: ConfigReaderFailure): ConfigReaderFailures =
-    new ConfigReaderFailures(configReaderFailure, List.empty[ConfigReaderFailure])
+  def apply(first: ConfigReaderFailure, rest: ConfigReaderFailure*): ConfigReaderFailures =
+    new ConfigReaderFailures(first, rest.toList)
 }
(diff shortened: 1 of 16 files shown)


core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
```diff
@@ -52,6 +52,6 @@ case class ConfigReaderFailures(head: ConfigReaderFailure, tail: List[ConfigRead
 
 object ConfigReaderFailures {
 
-  def apply(configReaderFailure: ConfigReaderFailure): ConfigReaderFailures =
-    new ConfigReaderFailures(configReaderFailure, List.empty[ConfigReaderFailure])
+  def apply(first: ConfigReaderFailure, rest: ConfigReaderFailure*): ConfigReaderFailures =
```
ruippeixotog: I'm afraid the type signature of this method will be almost the same as the case class constructor (the compiler is only ok with it because this is a `Seq` after erasure, while the case class constructor is a `List`). Do you think this might create some issues in the future? The alternative is to name this anything other than `apply`.

If you think this is ok, LGTM 👍

ruippeixotog (approved):

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
jcazevedo: Good point. What do you think of having a single `apply` (see 8975dd1)? I believe part of the motivation for having both was to have more convenient pattern matches on `ConfigReaderFailures`, but I suspect that's not a frequent use case outside of tests, and the available `toList` conversion can also help with that.

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
```diff
@@ -8,15 +8,15 @@ import scala.collection.mutable
 /**
  * A non-empty list of ConfigReader failures
  */
-case class ConfigReaderFailures(head: ConfigReaderFailure, tail: List[ConfigReaderFailure]) {
+case class ConfigReaderFailures(head: ConfigReaderFailure, tail: ConfigReaderFailure*) {
 
-  def toList: List[ConfigReaderFailure] = head +: tail
+  lazy val toList: List[ConfigReaderFailure] = head :: tail.toList
```
ruippeixotog: Nit: why did you turn this into a `lazy val`? It seems that for our own use cases we always use it at most once to construct new `ConfigReaderFailures` and there's no expectation of it being memoized from the client code's perspective. Unless there's some good reason for it, `def`s are more appropriate to expose in APIs.

ruippeixotog (approved):
LGTM, please take a look at the nit below

ruippeixotog:
One thing we could also do later was to add a `toString` to `ConfigReaderFailures ` that prints all elements as a collection instead of the default representation.

core/src/main/scala/pureconfig/error/ConfigReaderFailures.scala
jcazevedo: Yes. This shouldn't have been made a `lazy val`. Reverted in 1536514.

jcazevedo:
> One thing we could also do later was to add a `toString` to `ConfigReaderFailures ` that prints all elements as a collection instead of the default representation.

Good idea! Added that in 1895af4.

leifwickland:
I think this ought to be treated as a breaking change and the version number bumped accordingly at the next release.

jcazevedo:
> I think this ought to be treated as a breaking change and the version number bumped accordingly at the next release.

Yes. Next version is definitely breaking, not only because of this but also due to #733 and #629.

Issue #769, opened 2020-06-10
missing support for typesafe config's getBytes


eyalfa:
1. missing asBytes method on the cursor (delegating to `com.typesafe.config.Config#getBytes`)
2. missing `ConfigReader' instance for com.typesafe.config.ConfigMemorySize

please lmk if you're open for PR contribution.

jcazevedo:
I was wondering if we should support `asBytes` or adapt the `as<NumericType>` methods to follow the [size in bytes format](https://github.com/lightbend/config/blob/master/HOCON.md#size-in-bytes-format) from HOCON. Even though the latter may be more convenient for implementations of `ConfigReader`s, the fact that single-character unit abbreviations map to powers of two (i.e. `1K` is 1024) can be confusing, so I think it should be preferable to have a separate `asBytes` implementation.

I believe we're open for PRs with these changes, yes.

jcazevedo:
Closed by #775.

PR #775, opened 2020-06-12
769 get bytes

eyalfa:
#769 add support to typesafe config's ConfigMemorySize

### DIFF
--- a/core/src/main/scala/pureconfig/ConfigCursor.scala
+++ b/core/src/main/scala/pureconfig/ConfigCursor.scala
@@ -83,21 +83,6 @@ sealed trait ConfigCursor {
     })
 
   /**
-   * Casts this cursor to a long representing a memory size according to the way com.typesafe.config.Config#getBytes(java.lang.String) does.
-   *
-   * @return a `Right` with the memory size value pointed to by this cursor if the cast can be done, `Left` with a list of
-   *         failures otherwise.
-   */
-  def asBytes: ConfigReader.Result[Long] = scopeFailure {
-    ConvertHelpers.tryToEither {
-      Try {
-        val wrapped = value.atKey("_")
-        wrapped.getBytes("_")
-      }
-    }
-  }
-
-  /**
    * Casts this cursor to an int.
    *
    * @return a `Right` with the int value pointed to by this cursor if the cast can be done, `Left` with a list of
--- a/core/src/main/scala/pureconfig/ConfigCursor.scala
+++ b/core/src/main/scala/pureconfig/ConfigCursor.scala
@@ -82,6 +82,12 @@ sealed trait ConfigCursor {
       case v => Left(CannotConvert(v.toString, "Long", "Unable to convert Number to Long"))
     })
 
+  /**
+   * Casts this cursor to a long representing a memory size according to the way com.typesafe.config.Config#getBytes(java.lang.String) does.
+   *
+   * @return a `Right` with the memory size value pointed to by this cursor if the cast can be done, `Left` with a list of
+   *         failures otherwise.
+   */
   def asBytes: ConfigReader.Result[Long] = scopeFailure {
     ConvertHelpers.tryToEither {
       Try {
--- a/core/src/main/scala/pureconfig/ConfigCursor.scala
+++ b/core/src/main/scala/pureconfig/ConfigCursor.scala
@@ -82,6 +82,15 @@ sealed trait ConfigCursor {
       case v => Left(CannotConvert(v.toString, "Long", "Unable to convert Number to Long"))
     })
 
+  def asBytes: ConfigReader.Result[Long] = scopeFailure {
+    ConvertHelpers.tryToEither {
+      Try {
+        val wrapped = value.atKey("_")
+        wrapped.getBytes("_")
+      }
+    }
+  }
+
   /**
    * Casts this cursor to an int.
    *
--- a/tests/src/test/scala/pureconfig/ConfigCursorSuite.scala
+++ b/tests/src/test/scala/pureconfig/ConfigCursorSuite.scala
@@ -1,7 +1,7 @@
 package pureconfig
 
 import com.typesafe.config._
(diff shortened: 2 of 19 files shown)


eyalfa:
CC: @jcazevedo

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -82,6 +82,21 @@ sealed trait ConfigCursor {
...
+   * Casts this cursor to a long representing a memory size according to the way com.typesafe.config.Config#getBytes(java.lang.String) does.
+   *
+   * @return a `Right` with the memory size value pointed to by this cursor if the cast can be done, `Left` with a list of
+   *         failures otherwise.
+   */
+  def asBytes: ConfigReader.Result[Long] = scopeFailure {
+    ConvertHelpers.tryToEither {
+      Try {
+        val wrapped = value.atKey("_")
+        wrapped.getBytes("_")
```
jcazevedo: This is a bit unfortunate, but it seems necessary...

tests/src/test/scala/pureconfig/ConfigCursorSuite.scala
```diff
@@ -60,6 +60,14 @@ class ConfigCursorSuite extends BaseSuite {
       CannotConvert("1.1", "Long", "Unable to convert Number to Long"))
   }
 
+  it should "support asBytes on longs and properly formatted strings" in {
+    cursor("30").asBytes shouldBe Right(30)
+    cursor("30b").asBytes shouldBe Right(30)
+    cursor("30k").asBytes shouldBe Right(30 * 1024)
+    cursor("30m").asBytes shouldBe Right(30 * 1024 * 1024)
+    cursor("30MB").asBytes shouldBe Right(30 * 1000 * 1000)
```
jcazevedo: Can you also add an example for a failure scenario?

jcazevedo (changes requested):

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -82,6 +82,21 @@ sealed trait ConfigCursor {
       case v => Left(CannotConvert(v.toString, "Long", "Unable to convert Number to Long"))
     })
 
+  /**
+   * Casts this cursor to a long representing a memory size according to the way com.typesafe.config.Config#getBytes(java.lang.String) does.
+   *
+   * @return a `Right` with the memory size value pointed to by this cursor if the cast can be done, `Left` with a list of
+   *         failures otherwise.
+   */
+  def asBytes: ConfigReader.Result[Long] = scopeFailure {
```
ruippeixotog: I don't think this is suited for `ConfigCursor`. It's important to distinguish here between two different levels of config reading:
1. Primitive values, like integers or strings, that we can read out of `ConfigValues` because they are internally represented that way in HOCON;
1. Non-primitive values, represented in HOCON by one or more primitives in a known format.

`ConfigCursor` deals with the lower level, while the higher level is left to `ConfigReader`s. In this particular case we seem to be adding here code that is most suited to a `ConfigReader` (this is a very opinionated and high level way of reading bytes, in fact - `Config` tries to read this as a long, possibly falls back to reading a string and returns a `Long` instead of an `Array[Byte]`).

Note that the fact that it's supported by `Config` does _not_ make it a primitive - there are several other instances there unsuited to the cursor class, like `getMemorySize`, `getMilliseconds`, `getPeriod` and much more. Their API is bloated mainly because they don't have anything like configurable `ConfigReader`s.

ruippeixotog (changes requested):
Thanks for submitting the PR, @eyalfa! Having support for this seems nice, but I don't think this is core enough to be at the `ConfigCursor` level. See my comment below for details.

If we just care about reading `ConfigMemorySize`, we should move everything to the definition of `ConfigReader[ConfigMemorySize]`, like we did with our readers for `Period` and `Duration`. If you really think there are use cases for reading bytes as a `Long` (even if it goes against PureConfig's philosophy of wrapping types to add semantic to them as opposed to overloading readers) you can create a [configurable reader](https://github.com/pureconfig/pureconfig/blob/master/core/src/main/scala/pureconfig/configurable/package.scala) that people can use when composing readers.

core/src/main/scala/pureconfig/ConfigCursor.scala
eyalfa: 👍

core/src/main/scala/pureconfig/ConfigCursor.scala
eyalfa: @jcazevedo  coming to think of it, do u think it's more appropriate to name this: `asMemorySize`, current name seems like the plural of the `asByte` method

eyalfa: 👍  since this pr also introduces a `ConfigReader` instance for `ConfigMemorySize` I'll just move this implementation there.

eyalfa:
@ruippeixotog I basically agree with your review and refactored accordingly.
p.s. I really don't think that parsing a `long` should return bytes, however mapping a long to a `contextual` class like `ConfigMemorySize` should and I believe this is what this PR now achieves.

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: The alternative would be to copy their bytes parsing code and convert it to PureConfig standards (e.g. using either along the way instead of throwing). We did this with [Duration](https://github.com/pureconfig/pureconfig/blob/master/core/src/main/scala/pureconfig/DurationUtils.scala) and [Period](https://github.com/pureconfig/pureconfig/blob/master/core/src/main/scala/pureconfig/PeriodUtils.scala). I'm ok with using Typesafe Config's implementation for now though.

ruippeixotog (approved):
LGTM, I'll wait for @jcazevedo to review though.

jcazevedo (approved):

PR #783, opened 2020-06-26
Update Scala to 2.12.12 and 2.13.3

jcazevedo:
Besides updating Scala 2.13 to 2.13.3, this also removes the installation of `libcurl4-openssl-dev` from the GitHub actions workflow. I was getting 404s when attempting to fetch it (https://github.com/pureconfig/pureconfig/runs/811501986?check_suite_focus=true#step:5:26) and it doesn't seem to be required in that environment.


leifwickland (approved):

jcazevedo:
The example project compilation is failing with a linting warning I need to give a more thorough look at. Converting this PR to draft for the time being.

jcazevedo:
I couldn't isolate the culprit for the warning yet, but it seems like it is only being triggered when using shapeless-based derivation. The warning disappears when switching to magnolia-based derivation.

I've extracted 4fe7ada to #784 to fix the CI build.

jcazevedo:
The warning is being triggered by the use of shapeless' Lazy (https://github.com/scala/bug/issues/12072). For the time being, It's probably preferable to just include `-Xlint:byname-implicit` in the example project build.

jcazevedo:
Due to scala/bug#12072, I disabled `-Xlint:byname-implicit` in the example project. 

I also took the opportunity to update to Scala 2.12.12 in this PR.

ruippeixotog (approved):
LGTM, but let's keep track of this issue and revert this once a fix is released. I wouldn't want the example project to become bloated with SRT settings just because we're strict on warnings :/

jcazevedo:
> LGTM, but let's keep track of this issue and revert this once a fix is released. I wouldn't want the example project to become bloated with SRT settings just because we're strict on warnings :/

I'll keep track of the Scala issue. I also don't mind not compiling the example project with the `-Xlint` flag. I don't think it's necessary for the main purpose of the example project.

PR #784, opened 2020-06-27
Remove unnecessary libcurl4-openssl-dev dependency from CI build

jcazevedo:
Cherry-picked from #783 to fix the CI build.


ruippeixotog (approved):
Thanks @jcazevedo!

PR #787, opened 2020-06-30
Report valid options when reading an enum fail

leifwickland:


modules/generic/src/main/scala/pureconfig/generic/EnumCoproductHint.scala
```diff
@@ -28,7 +28,10 @@ class EnumCoproductHint[A] extends CoproductHint[A] {
     cursor.asString.right.flatMap { str =>
       options.find(str == fieldValue(_)) match {
         case Some(opt) => Right(Use(cursor, opt))
-        case None => cursor.failed[CoproductHint.Action](NoValidCoproductOptionFound(cursor.value, Seq.empty))
+        case None =>
```
ruippeixotog: I'm not sure of the right path to take here, but since `EnumCoproductHint` is deprecated should we keep this file "frozen" until it finally gets deleted? I'd prefer to not add new features here (if anything, as an incentive for people to migrate).

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
```diff
@@ -12,7 +12,7 @@ object semiauto {
   final def deriveWriter[A](implicit writer: Lazy[DerivedConfigWriter[A]]): ConfigWriter[A] = writer.value
 
   final def deriveConvert[A](implicit reader: Lazy[DerivedConfigReader[A]], writer: Lazy[DerivedConfigWriter[A]]): ConfigConvert[A] =
-    ConfigConvert.fromReaderAndWriter(Derivation.Successful(reader.value), Derivation.Successful(writer.value))
+    ConfigConvert(reader.value, writer.value)
```
ruippeixotog, line 16: Nice one!

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
```diff
@@ -13,15 +15,19 @@ import shapeless.labelled._
  * @tparam A the type of objects capable of being read as an enumeration
  */
 trait EnumerationConfigReaderBuilder[A] {
```
ruippeixotog, line 15: Like other auxiliary type classes for generic derivation, we should make this trait `private[generic]` so that we can more confidently change its API.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
```diff
@@ -13,15 +15,19 @@ import shapeless.labelled._
...
 
 object EnumerationConfigReaderBuilder {
   implicit val deriveEnumerationReaderBuilderCNil: EnumerationConfigReaderBuilder[CNil] =
     new EnumerationConfigReaderBuilder[CNil] {
-      def build(transformName: String => String): ConfigReader[CNil] =
+      def build(transformName: String => String, validOptions: Set[String]): ConfigReader[CNil] =
         new ConfigReader[CNil] {
-          def from(cur: ConfigCursor): Result[CNil] = cur.failed(NoValidCoproductOptionFound(cur.value, Seq.empty))
+          def from(cur: ConfigCursor): Result[CNil] = {
+            val failure = ConvertFailure(KeyNotFound(cur.value.render(ConfigRenderOptions.concise()), validOptions), None, cur.path)
```
ruippeixotog: This doesn't sound right... Enum configs are not about (object) keys, as they do not rely on objects at all. The true error is that we couldn't convert the value to an instance of the enum class, which is more accurately expressed as `CannotConvert(value in cursor, enum trait, some reason)`.

I wonder if `NoValidCoproductOptionFound` should be used here at all. It was built for coproduct options that have their own (possibly custom) readers, but here it seems overkill as we'd always generate something like:

```
- No valid coproduct option found for 'blue'.
  Can't use coproduct option 'blue':
    - Key not found: 'blue'. You might have a misconfigured ProductHint, since...
```

'blue' is here incorrectly used as a coproduct option and as an object key. This could be more concisely described as:

```
- Cannot convert 'blue' to Color: accepted values are 'rainy-blue' and 'sunny-yellow'.
```

What do you think, @leifwickland, @jcazevedo?

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
```diff
@@ -13,15 +15,19 @@ import shapeless.labelled._
  * @tparam A the type of objects capable of being read as an enumeration
  */
 trait EnumerationConfigReaderBuilder[A] {
-  def build(transformName: String => String): ConfigReader[A]
+  def build(transformName: String => String, validOptions: Set[String]): ConfigReader[A]
```
ruippeixotog: nit: it looks like at every call this is more like `attemptedOptions` and not always the full set of `validOptions`.

ruippeixotog (changes requested):

modules/generic/src/main/scala/pureconfig/generic/EnumCoproductHint.scala
jcazevedo: I prefer not modifying this behavior, not only as an incentive for users to migrate but also because we don't have tests for this.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
jcazevedo: A `CannotConvert` failure reason seems more appropriate, yes.

modules/generic/src/main/scala/pureconfig/generic/EnumCoproductHint.scala
leifwickland: I'm OK with it either way. If @jcazevedo doesn't object I'll revert this file

Might it be time to remove this deprecated class anyway? It's been 10 months since the commit which marked that type as deprecated? That should certainly be done as a separate PR.

modules/generic/src/main/scala/pureconfig/generic/semiauto.scala
leifwickland: Thanks. It's always nice to get credit for a cleanup that I missed earlier. ;-)

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
leifwickland: To reflect your nittiness back at you, at the time `build` is called, the options haven't been "attempted" yet. I definitely don't have a better suggestion than `attemptedOptions` but it doesn't feel quite right because it gets the temporal ordering wrong. 

Do you have another suggestion.

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
```diff
@@ -31,11 +37,12 @@ object EnumerationConfigReaderBuilder {
     hGen: LabelledGeneric.Aux[H, HNil],
     tReaderBuilder: EnumerationConfigReaderBuilder[T]): EnumerationConfigReaderBuilder[FieldType[K, H] :+: T] =
     new EnumerationConfigReaderBuilder[FieldType[K, H] :+: T] {
-      def build(transformName: String => String): ConfigReader[FieldType[K, H] :+: T] = {
-        lazy val tReader = tReaderBuilder.build(transformName)
+      def build(transformName: String => String, validOptions: Set[String]): ConfigReader[FieldType[K, H] :+: T] = {
+        val transformedName = transformName(vName.value.name)
+        lazy val tReader = tReaderBuilder.build(transformName, validOptions + transformedName)
```
jcazevedo: I think this would be slightly easier to reason about if we wouldn't mix a bottom-up approach (creating a reader for the remaining coproduct alternatives) with a top-down one (supplying the options previously tried).

Do you know if there's a native shapeless way to extract the names of the types from the `Coproduct` representation `LabelledGeneric` produces?

leifwickland: @jcazevedo It looks like there's a pretty easy way to get a [list of the names](https://stackoverflow.com/a/29679342), but I don't see way to get a similar way to get a list of the instances without using a ["bottom-up" approach of destructing the `LabelleledGeneric`](https://stackoverflow.com/a/43658710). 

If you can find a way to simply get a list of the instances, I'd love to do it that way instead.

ruippeixotog: The solution that plays well with the structurally recursive nature of shapeless generics is to get rid of this parameter, start by failing with an empty set of options in `EnumerationConfigReaderBuilder[CNil]` and keep prepending options to the failure in the recursive case (i.e. changing line 45 to something like `tReader.from(cur).right.map(Inr.apply).left.map(maybeAddOptionToKeyNotFound)`). We'd need some extra boilerplate for destructuring and reconstructing the failure when we are going "back up" (probably a single pattern match is enough).

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
leifwickland: I don't think we can make this `private[generic]` because it's part of the interface of the [public `deriveEnumerationReader` method](https://github.com/pureconfig/pureconfig/blob/3bb840f0d8d3a1c2d028b3d7215271d672f5e419/modules/generic/src/main/scala/pureconfig/generic/semiauto.scala#L23). When I do make it `private[generic]`, compiling the main code succeeds (😱 🙀 😱 ) but semiauto derivation fails for `Color` in `EnumerationSuites`.

Am I missing a trick here?

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
leifwickland: As of https://github.com/pureconfig/pureconfig/pull/787/commits/b4bb7cf311cbf832d00b9c1dff51d16f9e1c952c, the error message printed when `Color` appears inside of a `Fooz` type looks like:
```
Cannot convert configuration to a pureconfig.EnumerationsSuite$Fooz. Failures are:
at 'foreground-color':
  - (e.conf: 2) Cannot convert "blue" to a Color. The valid options are:
      - rainy-blue
      - sunny-yellow
```

How do you feel about that approach?

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
leifwickland: @ruippeixotog I struggle with shapeless enough that what you're describing isn't obvious to me. Can you think of an example similar to what you described that I could refer to?

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
ruippeixotog: Looks good! 🙂

modules/generic/src/main/scala/pureconfig/generic/EnumerationConfigReaderBuilder.scala
ruippeixotog: Unfortunately I don't think we have similar code examples. What I'm suggesting is to get rid of that parameter and follow a bottom-up, inductive approach:

- **Base case**: the coproduct is empty (e.g. an empty sealed trait). The result is simply `cur.failed(NoValidEnumOptionFound(cur.value, Set.empty, enumType))`.

- **Inductive case**: we're dealing with a coproduct of type `FieldType[K, H] :+: T`. We do just like we are doing now, but we map both successful results (wrapping it in `Inr`) and failed results with type `NoValidEnumOptionFound` to add the attempted option. In code this would be replacing line 41 with something like:

```scala
case Right(_) =>
  tReader.from(cur) match {
    case Right(res) => Right(Inr(res))
    case Left(ConfigReaderFailures(NoValidEnumOptionFound(value, validOptions, enumType))) =>
      Left(ConfigReaderFailures(NoValidEnumOptionFound(value, validOptions + transformedName, enumType)))
    case Left(failures) => Left(failures)
  }.right.map(Inr.apply)
```

I didn't compile this, it can probably be simplified and some care has to be taken with pattern matching on `ConfigReaderFailures` like this, but I hope it helps explaining what I meant :)

ruippeixotog:

PR #790, opened 2020-07-03
Replace Scalariform with Scalafmt

ruippeixotog:
[Scalafmt](https://github.com/scalameta/scalafmt) is more modern, more popular and better maintained than [Scalariform](https://github.com/scala-ide/scalariform). Moreover it has integration with multiple editors and IDEs. In this PR we do the plugin change and reformat all files.

I don't have strong opinions about most style settings, but I prefer to stay as close as possible to scalafmt defaults, as they're likely closer to most other projects making the code base more approachable. Therefore, I intentionally did not try to make scalafmt behave the same as scalariform. I opted to change the defaults for `align` (despite not being default it's recognized by scalafmt to be [easier for diffs](https://scalameta.org/scalafmt/docs/configuration.html#alignpresetnone)) and for `maxColumn` (no one needs 80 chars per line anymore and it actually makes code harder to read IMO). Let me know if you feel strongly for any of the other options :)

The PR is split in two commits, in case you only care about the plugin change and not about all the reformatted code.


leifwickland (approved):
Thanks for doing this. The time has come.

I'm fine with the default formatting. I'd be fine increasing the max line length to 120, but feel free to leave it as is.

Please consider adding a step to the build process which ensures that the code is formatted before any compile takes place by calling `scalafmtCheckAll` and `scalafmtSbtCheck`.

ruippeixotog:
> I'm fine with the default formatting. I'd be fine increasing the max line length to 120, but feel free to leave it as is.

Nice, I also prefer 120 :) I think @jcazevedo is ok with it too.

> Please consider adding a step to the build process which ensures that the code is formatted before any compile takes place by calling `scalafmtCheckAll` and `scalafmtSbtCheck`.

I believe we already check this in our CI indirectly - after we run all the steps (including compiling, which automatically formats files) [we run `git diff --exit-code`](https://github.com/pureconfig/pureconfig/blob/master/.github/workflows/ci.yml#L42).

This was originally intended to catch both bad formatting and outdated tut-generated files. Do you see value in having an explicit step prior to compilation that just checks formatting? While it can be clearer for users to know what went wrong, it would prevent them from seeing compilation errors in the first round (which are arguably more important).

leifwickland:
@ruippeixotog I'd forgotten we had the `git diff` check. I have a light, but easily abandoned, preference to check the formatting before compile so that the root cause of the problem is reported directly to the user, rather than requiring the user to make an inference. Feel free to leave it as is.

.scalafmt.conf
```diff
@@ -0,0 +1,4 @@
+version = 2.5.2
```
jcazevedo: I think we can go with 2.6.2.

core/src/main/scala/pureconfig/BasicReaders.scala
```diff
@@ -76,29 +81,30 @@ trait JavaEnumReader {
...
- * Trait containing `ConfigReader` instances for classes related to file system paths and URIs.
- */
+  * Trait containing `ConfigReader` instances for classes related to file system paths and URIs.
+  */
 trait UriAndPathReaders {
 
   implicit val urlConfigReader = ConfigReader.fromNonEmptyString[URL](catchReadError(new URL(_)))
-  implicit val uuidConfigReader = ConfigReader.fromNonEmptyString[UUID](catchReadError(UUID.fromString))
+  implicit val uuidConfigReader =
+    ConfigReader.fromNonEmptyString[UUID](catchReadError(UUID.fromString))
```
jcazevedo: Can you rerun the formatter from the original state? This line should fit in 120 chars, and there are other in the same situation.

core/src/main/scala/pureconfig/BasicReaders.scala
```diff
@@ -133,17 +139,23 @@ trait DurationReaders {
...
-        case _ => Left(CannotConvert(string, "FiniteDuration",
-          s"Couldn't parse '$string' into a FiniteDuration because it's infinite."))
+        case _ =>
+          Left(
+            CannotConvert(
+              string,
+              "FiniteDuration",
+              s"Couldn't parse '$string' into a FiniteDuration because it's infinite."
+            )
+          )
```
jcazevedo, line 142: I'm OK with going with the defaults, but, for the record, I'm not a fan of this style of dangling parenthesis. 😄

project/plugins.sbt
```diff
@@ -4,7 +4,7 @@ addSbtPlugin("com.jsuereth"      % "sbt-pgp"         % "2.0.1")
 addSbtPlugin("com.typesafe.sbt"  % "sbt-osgi"        % "0.9.5")
 addSbtPlugin("io.spray"          % "sbt-boilerplate" % "0.6.1")
 addSbtPlugin("org.scalameta"     % "sbt-mdoc"        % "2.2.3")
-addSbtPlugin("org.scalariform"   % "sbt-scalariform" % "1.8.3")
+addSbtPlugin("org.scalameta"     % "sbt-scalafmt"    % "2.4.0")
```
jcazevedo, line 7: Can you also make this change, and respective code format, in the example project?

jcazevedo (changes requested):

core/src/main/scala/pureconfig/BasicReaders.scala
ruippeixotog: I also didn't use dangling parenthesis before, but now I like them a lot :) They make it much easier to understand the depth of each expression in a way that doesn't rely on indentation or on counting several closing parenthesis together at the end of a line.

ruippeixotog:
> @ruippeixotog I'd forgotten we had the git diff check. I have a light, but easily abandoned, preference to check the formatting before compile so that the root cause of the problem is reported directly to the user, rather than requiring the user to make an inference. Feel free to leave it as is.

I agree with providing more direct feedback to the user. My concern is on frustrating contributors by not notifying them of everything at the first chance - it's a bit annoying to have a build fail because of formatting issues, push a commit to fix it just to be notified later that a test is also failing or tut is not synced. We could have a parallel GitHub Actions job just for formatting, but it seems a bit overkill, since a job spins a full fledged Docker container (in this case three due to the matrix build).

Anyway, I added this extra step. I'll assume most sensible people compile the project before submitting a PR :)

jcazevedo (approved):

PR #791, opened 2020-07-04
Stop using `null` to represent missing values in cursors

ruippeixotog:
This came up several times in issues and PR discussions. On this PR we start using `Option[ConfigValue]` to hold values in `ConfigCursor`.


core/src/main/scala/pureconfig/BasicReaders.scala
```diff
@@ -165,30 +165,30 @@ trait NumericReaders {
...
 
   implicit val configValueConfigReader: ConfigReader[ConfigValue] =
-    ConfigReader.fromCursor(c => Right(c.value))
+    ConfigReader.fromCursor(_.asConfigValue)
 
   implicit val configListConfigReader: ConfigReader[ConfigList] =
-    ConfigReader.fromCursor(_.asListCursor.right.map(_.value))
+    ConfigReader.fromCursor(_.asListCursor.right.map(_.listValue))
 
+  // TODO: improve error handling
```
leifwickland: Could I trouble you to add to the comment a description of how you want to improve error handling?

ruippeixotog: Sure, done 👍

core/src/main/scala/pureconfig/ConfigCursor.scala
```diff
@@ -32,23 +32,36 @@ sealed trait ConfigCursor {
...
    *
    * @return `true` if this cursor points to a `null` value, `false` otherwise.
    */
-  def isNull: Boolean = value != null && value.unwrapped == null
+  def isNull: Boolean = valueOpt.exists(_.unwrapped == null)
+
+  /**
+   * Casts this cursor to a `ConfigValue`.
+   *
+   * @return a `Right` with the string value pointed to by this cursor if the value is defined, `Left` with a list of
```
jcazevedo: This is not necessarily a string value.

jcazevedo (approved):

core/src/main/scala/pureconfig/ConfigCursor.scala
ruippeixotog: Copy-paste error, thanks for pointing it out!

PR #792, opened 2020-07-08
Move diff_website comments to GitHub Actions

ruippeixotog:
As titled. `secrets.GITHUB_TOKEN` is available on forked repos, so this will work on every pull request (as opposed to the "manual" method used on `diff_website.sh`).

Unfortunately, I couldn't generate a direct link to the "Diff Website" job in GitHub Actions as I don't have easy access to check_run_id in the environment, but it doesn't seem very problematic (we explain how to get to the diff in the message).


jcazevedo (commented):
This is great! 

I'd also prefer if we'd remove the diff job from Travis, so as to not having a fail check there when the GitHub Actions version would just skip.

jcazevedo (approved):

PR #795, opened 2020-07-12
Add a new module for akka-http models

himanshu4141:
This PR adds support for [Akka-Http](https://doc.akka.io/docs/akka-http/current/common/http-model.html#http-model)'s `Uri` .


himanshu4141:
@ruippeixotog Hello, need a bit of help here. I am not able to figure out why website creation is failing.

jcazevedo:
@himanshu4141: The failed check indicates that this PR produces a change to the contents of PureConfig's website. It's expected if you change the documentation.

@ruippeixotog recently changed the CI step so that the build would no longer fail (#792). If you sync your branch with master, the failed check should disappear.

himanshu4141:
@jcazevedo ahh ok. The website change I guess would be the link I added to include the new module.

Thanks for the quick response. I have updated the branch. Hopefully, this should go away, making this PR ready for review.

modules/akka-http/README.md
```diff
@@ -0,0 +1,37 @@
+
+# Akka-Http module for PureConfig
```
ruippeixotog, line 2: Can you please use tut for the documentation? You should move this to `modules/akka-http/src/main/tut/README.md`, use tags as in [this example](https://github.com/pureconfig/pureconfig/blob/master/modules/akka/src/main/tut/README.md) and run `sbt tut`.

modules/akka-http/build.sbt
```diff
@@ -0,0 +1,16 @@
+name := "pureconfig-akka-http"
+
+crossScalaVersions ~= { _.filterNot(_.startsWith("2.11")) }
+
+libraryDependencies ++= Seq(
+  "com.typesafe.akka" %% "akka-actor" % "2.5.31",
```
ruippeixotog: Why is akka-actor required here?

modules/akka-http/build.sbt
```diff
@@ -0,0 +1,16 @@
...
+  "com.typesafe.akka" %% "akka-http" % "10.1.12")
+
+developers := List(
+  Developer("himanshu4141", "Himanshu Yadav", "himanshu4141@gmail.com", url("https://github.com/himanshu4141")))
+
+osgiSettings
+
+OsgiKeys.exportPackage := Seq("pureconfig.module.akkahttp.*")
+OsgiKeys.privatePackage := Seq()
+OsgiKeys.importPackage := Seq(s"""scala.*;version="[${scalaBinaryVersion.value}.0,${scalaBinaryVersion.value}.50)"""", "*")
```
ruippeixotog: Can you add a newline at the end?

modules/akka-http/src/main/scala/pureconfig/module/akkahttp/package.scala
```diff
@@ -0,0 +1,24 @@
...
+import scala.util.Try
+
+package object akkahttp {
+
+  implicit val uriReader: ConfigReader[Uri] =
+    ConfigReader.fromString(str =>
+      Try(Uri(str, ParsingMode.Strict)).fold(
+        {
+          case err: IllegalUriException => Left(CannotConvert(str, "Uri", err.info.summary))
+          case err => Left(CannotConvert(str, "Uri", err.getMessage.replace('^', " ".charAt(0)).trim))
```
ruippeixotog: Why are we parsing the message here, in the fallback exception case? If we have no information on why an exception was thrown we should use `ExceptionThrown` instead of `CannotConvert`.

modules/akka-http/src/main/scala/pureconfig/module/akkahttp/package.scala
```diff
@@ -0,0 +1,24 @@
...
+import pureconfig.{ConfigReader, ConfigWriter}
+import pureconfig.error.CannotConvert
+
+import scala.util.Try
+
+package object akkahttp {
+
+  implicit val uriReader: ConfigReader[Uri] =
+    ConfigReader.fromString(str =>
+      Try(Uri(str, ParsingMode.Strict)).fold(
```
ruippeixotog: nit: you can use `Try(...).toEither.left.map(...)` to simplify this expression, by getting rid of the success `case` and the need to wrap errors in `Left`.

ruippeixotog (changes requested):
Thanks for the contribution, @himanshu4141!

modules/akka-http/build.sbt
himanshu4141: ~It's a required dependency of akka-http.~

Actually latest version relies upon `akka-stream` not `akka-actor` as per developer [docs](https://doc.akka.io/docs/akka-http/current/introduction.html#using-akka-http). I"ll update.

modules/akka-http/README.md
himanshu4141: sure

modules/akka-http/src/main/scala/pureconfig/module/akkahttp/package.scala
himanshu4141: used `ExceptionThrown` for fallback cases.

modules/akka-http/build.sbt
```diff
@@ -0,0 +1,16 @@
+name := "pureconfig-akka-http"
+
+crossScalaVersions ~= { _.filterNot(_.startsWith("2.11")) }
+
+libraryDependencies ++= Seq(
+  "com.typesafe.akka" %% "akka-stream" % "2.5.31",
```
ruippeixotog: I see. Nonetheless, this being a library project that does not depend explicitly on any code from `akka-stream`, it's still important that we mark `akka-stream` as `provided` in order to make sure we don't bind client code to a specific version (and force them to provide their own akka-stream version as described [here](https://doc.akka.io/docs/akka-http/current/compatibility-guidelines.html#compatibility-with-akka)).
```suggestion
  "com.typesafe.akka" %% "akka-stream" % "2.5.31" % Provided,
```

ruippeixotog (changes requested):
LGTM! Please take care of this last comment and we'll merge it.

modules/akka-http/build.sbt
himanshu4141: hmm, didn't think of that. Sounds good. Thanks for the suggestion

ruippeixotog: I forgot about tut, which still needs it to run code in the documentation. I took the liberty of adding tut to the dependency scopes myself, hope it's ok :)

ruippeixotog (approved):

himanshu4141:
@ruippeixotog thanks for the merge and that last change. 👍

Issue #796, opened 2020-07-15
pureconfig-magnolia: support reading/writing case objects


neko-kai:
Currently `pureconfig-magnolia` reads/writes case objects only as empty objects instead of strings, e.g

```scala
sealed trait SealedTrait
object SealedTrait {
  case object Yes extends SealedTrait
  case object No extends SealedTrait
}
```

Becomes

```
{ No {} }
```

in config instead of

```
"No"
```

ruippeixotog:
The expected default behavior of sealed traits is to disambiguate using a `type` field, so `No` would result in `{ type: "no" }`. Can you share some minimal example in which `{ No {} }` is written or read by PureConfig?

To achieve the behavior you want you need to use `deriveEnumerationReader`. Please take a look at https://pureconfig.github.io/docs/overriding-behavior-for-sealed-families.html.

neko-kai:
> Can you share some minimal example in which { No {} } is written or read by PureConfig?

Ah, I've added a custom coproduct hint to read sealed traits in the style of `circe`, and expose auto-derivations using it in `distage-config` library:
https://github.com/7mind/izumi/blob/develop/distage/distage-extension-config/src/main/scala/izumi/distage/config/codec/PureconfigInstances.scala
https://github.com/7mind/izumi/blob/develop/distage/distage-extension-config/src/main/scala/izumi/distage/config/codec/PureconfigAutoDerive.scala

I've forgot this isn't pureconfig's default behavior.

> To achieve the behavior you want you need to use deriveEnumerationReader. Please take a look at https://pureconfig.github.io/docs/overriding-behavior-for-sealed-families.html.

That's correct, but this instance is part of `pureconfig-generic`, there isn't a way to derive an enumeration reader using only the magnolia module yet.

ruippeixotog:
Right, I forgot that we didn't have enumeration readers and writers for Magnolia. Sounds like a good feature to add!

jcazevedo:
Closed by #798.

PR #798, opened 2020-07-16
Add support for deriving readers and writers for enumerations in the Magnolia module

jcazevedo:
Following the discussion on #796, this adds the `deriveEnumerationReader` and `deriveEnumerationWriter` methods for the Magnolia module.

The derived `ConfigReader`s and `ConfigWriter`s follow the behavior of their counterparts in the generic module and may need to be revisited after #787.

With Magnolia, I didn't find a way to guarantee at compile time that we're only trying to derive enumeration readers and writers for sealed traits of objects. However, the cases in which we can derive instances of `EnumerationConfigReaderBuilder` and `EnumerationConfigWriterBuilder` for "incorrect" types look small. We seem to be able to derive it for case classes with empty parameter lists or whose parameter types are also eligible for an instance of those auxiliary type classes. I don't think this is a deal breaker.


tests/src/test/scala/pureconfig/EnumerationsSuite.scala
```diff
@@ -72,6 +72,9 @@ class EnumerationsSuite extends BaseSuite {
   }
 
   it should "not allow deriving readers, writers and full converters for enumerations encoded as sealed traits whose subclasses are not all case objects" in {
+    sealed trait Entity
+    case class Person(name: String, surname: String) extends Entity
+    case class Place(name: String, lat: Double, lon: Double) extends Entity
```
jcazevedo, line 77: This was declared here because these types weren't accessible from this suite, so the derivations below were not compiling for the wrong reasons.

modules/magnolia/src/test/scala/pureconfig/module/magnolia/EnumerationsSuite.scala
```diff
@@ -0,0 +1,58 @@
...
+import pureconfig.module.magnolia.semiauto.reader._
+import pureconfig.module.magnolia.semiauto.writer._
+
+class EnumerationsSuite extends BaseSuite {
+
+  sealed trait Color
+  case object RainyBlue extends Color
+  case object SunnyYellow extends Color
+
+  val conf = ConfigFactory.parseString("{ type: person, name: John, surname: Doe }")
```
ruippeixotog: We don't seem to be using this anywhere.

tests/src/test/scala/pureconfig/EnumerationsSuite.scala
ruippeixotog: Good catch! Can we add a message regex to `illTyped` to prevent it from happening again?

modules/magnolia/src/main/scala/pureconfig/module/magnolia/EnumerationConfigReaderBuilder.scala
```diff
@@ -0,0 +1,49 @@
...
+object EnumerationConfigReaderBuilder {
+  type Typeclass[A] = EnumerationConfigReaderBuilder[A]
+
+  def combine[A](ctx: CaseClass[EnumerationConfigReaderBuilder, A]): EnumerationConfigReaderBuilder[A] =
+    new EnumerationConfigReaderBuilder[A] {
+      def build(transformName: String => String): ConfigReader[A] =
+        new ConfigReader[A] {
+          def from(cur: ConfigCursor): Result[A] =
+            if (ctx.isObject) Right(ctx.rawConstruct(Seq.empty))
+            else cur.asConfigValue.right.flatMap(v => cur.failed(NoValidCoproductOptionFound(v, Seq.empty)))
```
ruippeixotog: How are you planning to handle valid coproduct options in errors? I know you mentioned revisiting this after #787; isn't it more convenient to do this now that we're adding a new implementation? Or do you prefer to do it on a separate PR?

tests/src/test/scala/pureconfig/EnumerationsSuite.scala
```diff
@@ -72,6 +72,9 @@ class EnumerationsSuite extends BaseSuite {
   }
 
   it should "not allow deriving readers, writers and full converters for enumerations encoded as sealed traits whose subclasses are not all case objects" in {
```
ruippeixotog, line 74: Can we have a test checking that we disallow derivation of enum readers for case classes at compile time, like this one? I'm ok with ignoring the zero-arg case class edge case, but I'd like to ensure we don't generate type classes for case classes with arguments.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/EnumerationConfigReaderBuilder.scala
ruippeixotog: Also, is `NoValidCoproductOptionFound` the right error here?

If I understood the code correctly this is the case where we find ourselves deriving an enum reader for a case class rather than an object. In this case, shouldn't we throw an exception to cause the derivation to fail at compile-time?

ruippeixotog (changes requested):

modules/magnolia/src/main/scala/pureconfig/module/magnolia/EnumerationConfigReaderBuilder.scala
jcazevedo: > Also, is `NoValidCoproductOptionFound` the right error here?
> 
> If I understood the code correctly this is the case where we find ourselves deriving an enum reader for a case class rather than an object. In this case, shouldn't we throw an exception to cause the derivation to fail at compile-time?

Unfortunately, an exception thrown here will only be visible at run time. I don't think there's a straightforward way to fail the compilation at this point. I suppose we can create a custom `FailureReason` for this, but ideally we wouldn't be deriving type classes for case classes.

modules/magnolia/src/test/scala/pureconfig/module/magnolia/EnumerationsSuite.scala
jcazevedo: It's being used in the first example of the spec. I moved its definition to an inner scope to make its use more evident in d805ae7.

tests/src/test/scala/pureconfig/EnumerationsSuite.scala
jcazevedo: Added a message regex in fe4e08b. We should add an `@implicitNotFound` for the enumeration helper type classes, but I'll leave that for another PR.

tests/src/test/scala/pureconfig/EnumerationsSuite.scala
jcazevedo: Added a copy of this example to pureconfig-magnolia's test suite in b02e5dc.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/EnumerationConfigReaderBuilder.scala
ruippeixotog: That's unexpected :/ I think we really need this error to stand out, as the behavior differs from shapeless and from the compile-time safety of PureConfig in general. I'd probably go for a different error where we can specify a reason, like `CannotConvert`. However, if you still feel we're ok `NoValidCoproductOptionFound` I'm ok with it too.

It's probably worth asking in magnolia how to make this use case possible, too.

ruippeixotog (approved):
LGTM, please take a look at my comment above.

modules/magnolia/src/main/scala/pureconfig/module/magnolia/EnumerationConfigReaderBuilder.scala
jcazevedo: I opened an issue in Magnolia (propensive/magnolia#267) to see if this use case would be possible. In propensive/magnolia#263, they introduced the possibility to only define the `dispatch` method. I suppose we could have an implementation for `dispatch` only and provide an alternative derivation method for objects. However, I think we'd still need a macro to provide an evidence that a given type is an object.

For the time being, I've changed the failure reason to a `CannotConvert`, so that we can make the underlying issue for the error more explicit.

ruippeixotog (approved):
Thank you for building this!

PR #799, opened 2020-07-16
Submit coverage data through GitHub Actions

ruippeixotog:
~Requires https://github.com/scoverage/sbt-coveralls/pull/128 to be merged and released.~

It looks like sbt-coveralls is not actively maintained anymore, by the looks of https://github.com/scoverage/sbt-coveralls/issues/126 or the repo's PRs. In order to unblock this I published the plugin myself with https://github.com/scoverage/sbt-coveralls/pull/128 and I think it's ok for us to start using it. I think we can have both Travis and GitHub actions posting coverage data to Coveralls for a while and remove Travis completely once we are confident enough that the integration works well.

If you could take a look at https://github.com/scoverage/sbt-coveralls/pull/128 as part of this PR's review I'd appreciate it :)


jcazevedo (approved):
LGTM, but I wonder if we shouldn't be looking into using an alternative to track coverage data (like [Codecov](https://codecov.io/)) due to Coveralls apparent lack of support for Scala.

ruippeixotog:
@jcazevedo I thought about that too, but I don't think there's any way to migrate historical coverage data from one service to the other and it would be a shame to have incomplete data :/ But if you want to try integrating Codecov I'd be happy to try it!

PR #805, opened 2020-07-30
Fix instructions in readme for pureconfig-magnolia

bplommer:


ruippeixotog (changes requested):
Thanks for catching this error, @bplommer! You need to perform these changes in [modules/magnolia/src/main/tut/README.md](https://github.com/pureconfig/pureconfig/blob/master/modules/magnolia/src/main/tut/README.md) and run `sbt tut` afterwards to compile the code snippets in the documentation.

jcazevedo (approved):

PR #828, opened 2020-09-16
826 bump cats version

### LINKED ISSUE #826: Upgrade cats to 2.2.0
After I bumped cats to 2.2.0, it broke the automatic derivation for ConfigReader of my case class (using pureconfig-cats). 

Anyone else suffering from this ?

saraiva132:
Closes #826

### DIFF
--- a/modules/cats-effect/build.sbt
+++ b/modules/cats-effect/build.sbt
@@ -1,7 +1,7 @@
 name := "pureconfig-cats-effect"
 
 libraryDependencies ++= Seq(
-  "org.typelevel" %% "cats-effect" % "2.0.0")
+  "org.typelevel" %% "cats-effect" % "2.2.0")
 
 developers := List(
   Developer("keirlawson", "Keir Lawson", "keirlawson@gmail.com", url("https://github.com/keirlawson")))
--- a/modules/cats-effect/build.sbt
+++ b/modules/cats-effect/build.sbt
@@ -1,13 +1,18 @@
+import Utilities._
+
 name := "pureconfig-cats-effect"
 
 libraryDependencies ++= Seq(
-  "org.typelevel" %% "cats-effect" % "2.2.0")
+  "org.typelevel" %% "cats-effect" % forScalaVersions { case (2, 11) => "2.0.0"; case _ => "2.2.0" }.value
+)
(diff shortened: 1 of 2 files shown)


leifwickland:
The challenge with upgrading to cats 2.2.0 is that it's not published for Scala 2.11. Pureconfig is still cross publishing for Scala 2.11.

It seems like we need to do a final release of Pureconfig including support for 2.11 and then discontinue support for 2.11 and upgrade cats.

@ruippeixotog @jcazevedo, does that sound right to you?

saraiva132:
@leifwickland I am making a commit to address this in similar fashion that pureconfig does with circe. For scala 2.11 maintain version 2.0.0 otherwise use version 2.2.0.

jcazevedo:
> It seems like we need to do a final release of Pureconfig including support for 2.11 and then discontinue support for 2.11 and upgrade cats.
> 
> @ruippeixotog @jcazevedo, does that sound right to you?

We still have some nonnegligible monthly downloads of PureConfig for Scala 2.11 in Sonatype. As long as keeping compatibility at least in the core and generic modules is not too cumbersome, I'd prefer keeping it.

Regarding the pureconfig-cats module, I'm OK with either dropping support for Scala 2.11 entirely (we have very few monthly downloads of the module for that version) or depending on a different Cats version for Scala 2.11, as we do for Circe. I have a slight preference towards keeping compatibility if the Cats versions we depend on remain source compatible and it's not too cumbersome to maintain it.

jcazevedo (approved):
@saraiva132: Thanks for this!

ruippeixotog (approved):
As @jcazevedo said, we shouldn't stop compiling PureConfig for Scala 2.11 just because one module dependency doesn't support Scala 2.11 - the default solution for that is to stop publishing that module for Scala 2.11, something we've done in many of our modules now.

My preference would actually be to drop support for Scala 2.11 altogether rather than using different cats versions in order to keep both the build and maintenance simple. Even with source compatibility, users might experience a few weird issues when migrating Scala versions. Moreover, having Scala 2.11 in the build might discourage people from using modern cats features.

I don't have strong feelings against it though, so I'm accepting it.

PR #832, opened 2020-09-21
Format build files according to the Scalafmt config

jcazevedo:
Applies our Scalafmt config to build files and adds an additional CI check to make sure they're properly formatted. I don't think we should automatically format these like we do for Scala files on compile, since it can become annoying in an interactive environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,12 +20,8 @@ jobs:
         with:
           java-version: ${{ matrix.jdk }}
 
-      - name: Check sbt formatting
-        if: startsWith(matrix.scala, '2.12')
-        run: sbt "++${{ matrix.scala }} scalafmtSbtCheck"
-
       - name: Check formatting
-        run: sbt "++${{ matrix.scala }} scalafmtCheckAll"
+        run: sbt "++${{ matrix.scala }} scalafmtCheckAll; scalafmtSbtCheck"
 
       - name: Compile
         run: sbt coverage "++${{ matrix.scala }} compile"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,12 @@ jobs:
         with:
           java-version: ${{ matrix.jdk }}
 
+      - name: Check sbt formatting
+        if: startsWith(matrix.scala, '2.12')
+        run: sbt "++${{ matrix.scala }} scalafmtSbtCheck"
+
       - name: Check formatting
-        run: sbt "++${{ matrix.scala }} fmtCheckAll"
+        run: sbt "++${{ matrix.scala }} scalafmtCheckAll"
 
       - name: Compile
         run: sbt coverage "++${{ matrix.scala }} compile"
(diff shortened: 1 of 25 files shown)


leifwickland (approved):
Good idea. Thanks.

build.sbt
```diff
@@ -74,40 +72,30 @@ lazy val yaml = module(project) in file("modules/yaml")
...
     Developer("leifwickland", "Leif Wickland", "leifwickland@gmail.com", url("https://github.com/leifwickland")),
     Developer("jcazevedo", "Joao Azevedo", "joao.c.azevedo@gmail.com", url("https://github.com/jcazevedo")),
     Developer("ruippeixotog", "Rui Gonçalves", "ruippeixotog@gmail.com", url("https://github.com/ruippeixotog")),
-    Developer("derekmorr", "Derek Morr", "morr.derek@gmail.com", url("https://github.com/derekmorr"))),
-
+    Developer("derekmorr", "Derek Morr", "morr.derek@gmail.com", url("https://github.com/derekmorr"))
+  ),
   crossScalaVersions := Seq(scala211, scala212, scala213),
   scalaVersion := scala212,
-
```
ruippeixotog, line 87: I didn't format this part before because scalafmt doesn't work very well with this DSL. These spaces are intended to split different sections of the config (e.g. compilation, publish, etc) just like we add empty lines between statements sometimes. I find scalafmt's output too confusing to read in this case.

Can you either annotate this section for scalafmt to ignore it or refactor the build to avoid this? (e.g. by creating multiple `val`s with related configs)

.github/workflows/ci.yml
```diff
@@ -20,6 +20,10 @@ jobs:
         with:
           java-version: ${{ matrix.jdk }}
 
+      - name: Check sbt formatting
```
ruippeixotog: nit: can we simply add this to the step below? Personal preference, feel free to ignore.

ruippeixotog (changes requested):

build.sbt
jcazevedo: I'm not very fond of splitting this in multiple `val`s (which also raises some naming issues I don't really want to tackle right now 😅 ), so I added a format:off guard around `commonSettings` in e07cd93.

.github/workflows/ci.yml
jcazevedo: I separated this because we don't need to run it for every Scala version. I don't have a strong preference either way, so I joined both checks in 130eb56.

ruippeixotog: The same is true for `scalafmtCheckAll` - we expect formatted code to be the same regardless of the Scala version used, and if by some weird reason scalafmt formats it differently we have to abide by one specific version in our codebase. Feel free to re-add the Scala version GitHub Actions condition again :)

ruippeixotog (approved):
Thanks!

.github/workflows/ci.yml
jcazevedo: I'm assuming we run `scalafmtCheckAll` for all versions not because we're expecting differences in the format, but because we have different files compiled for different versions.

ruippeixotog: Makes sense, I forgot about that 👍

PR #841, opened 2020-10-09
Use mdoc variables in website

ruippeixotog:
This makes sure the version on the website is always in sync with the build version (I happened to miss it on our last deploy and had to do [a post-release commit](https://github.com/pureconfig/pureconfig/commit/11ede7e43e62b6158163f3611ae3914f6de955f9) to master 🙁 ).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -131,7 +131,8 @@ lazy val docsSettings = Seq(
     "white-color" -> "#FFFFFF"
   ),
   micrositeGitterChannel := false, // ugly
-  mdocExtraArguments += "--no-link-hygiene"
+  mdocExtraArguments += "--no-link-hygiene",
+  mdocVariables := Map("VERSION" -> version.value)
 )
 
 // add support for Scala version ranges such as "scala-2.12+" in source folders (single version folders such as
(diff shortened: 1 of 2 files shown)


jcazevedo:
It's a good idea to have the website in sync with the build version, but this will cause every PR to have a diff against the current version of the website.

ruippeixotog:
> It's a good idea to have the website in sync with the build version, but this will cause every PR to have a diff against the current version of the website.

The "Diff Website" step compares the PR's website output against `master`, not against https://pureconfig.github.io (otherwise every PR after the first change to the website would fail the build). We'll get Diff Website warnings every time we update the PureConfig version, but that doesn't seem too disruptive to me as we usually do it only to release a new version (which is usually a direct commit to master).

jcazevedo (approved):
> > It's a good idea to have the website in sync with the build version, but this will cause every PR to have a diff against the current version of the website.
> 
> The "Diff Website" step compares the PR's website output against `master`, not against https://pureconfig.github.io (otherwise every PR after the first change to the website would fail the build). We'll get Diff Website warnings every time we update the PureConfig version, but that doesn't seem too disruptive to me as we usually do it only to release a new version (which is usually a direct commit to master).

Ah, right! Nevermind my comment.

LGTM

PR #842, opened 2020-10-09
Add support for mdoc in PureConfig modules

ruippeixotog:
This PR adds support for mdoc on our PureConfig modules while keeping our existing subproject structure (which was a blocker according to my discussion in https://github.com/scalameta/mdoc/issues/155). This is achieved by creating a simple local SBT plugin that generates synthetic SBT subprojects just for documentation.

As a POC I updated `pureconfig-akka` to use mdoc instead of tut. Once this is accepted and merged I'll proceed to update all the remaining modules.

This is part of #574.


project/ModuleMdocPlugin.scala
```diff
@@ -0,0 +1,31 @@
...
+    val docProj =
+      Project(docProjId, docProjRoot)
+        .enablePlugins(MdocPlugin)
+        .dependsOn(LocalProject(proj.id))
+        .dependsOn(LocalProject("generic")) // Allow auto-derivation in documentation
+        .settings(
+          name := docProjId,
+          mdocVariables := Map("VERSION" -> version.value),
+          mdocIn := proj.base / "docs",
+          mdocOut := proj.base
```
jcazevedo: This needs a `skip in publish := true` to avoid having these derived projects from being published.

jcazevedo (changes requested):
This is a nice workaround!

project/ModuleMdocPlugin.scala
ruippeixotog: Good catch, added it!

jcazevedo (approved):

PR #843, opened 2020-10-10
Move docs SBT settings to a separate build.sbt

ruippeixotog:
This will reduce the size of `build.sbt` by moving website-specific settings to the only project that uses them.

I also set a new `releaseNextCommitMessage` for sbt-release as I noticed that on our last release that commit (e2e792edb667b7e07e5f7a8ad2d1d80b64366751) had a message inconsistent with the other commits.


docs/build.sbt
```diff
@@ -1 +1,29 @@
...
+  "gray-light" -> "#E3E2E3", // star back
+  "gray-lighter" -> "#F4F3F4", // code back
+  "white-color" -> "#FFFFFF"
+)
+micrositeGitterChannel := false // ugly
+
+mdocExtraArguments += "--no-link-hygiene"
+mdocVariables := Map("VERSION" -> version.value)
+
+publishArtifact := false
```
jcazevedo: This doesn't seem to prevent ivy files from being published, at least locally (see sbt/sbt#2249). `skip in publish := true` seems more aligned with what we want for this project.

jcazevedo (changes requested):

docs/build.sbt
ruippeixotog: It was defined like this in root `build.sbt`, I just moved it around. It looks like we weren't publishing anything for this project, but I also recall that `skip in publish := true` is the right setting to avoid publishing a project. I changed it in 0a19c3a.

jcazevedo (approved):

PR #845, opened 2020-10-13
Finish migration from tut to mdoc

### LINKED ISSUE #574: Migrate website and documentation to mdoc

ruippeixotog:
In this PR we migrate all SBT projects to use mdoc instead of tut.

Closes #574.


ruippeixotog:
@jcazevedo, can you took a look at this? Should be easy to review, particularly if you go commit by commit.

jcazevedo:
@ruippeixotog: Sorry for the delay. I'll try to review this later today.

modules/enum/docs/README.md
```diff
@@ -27,27 +27,26 @@ import enum.Enum
...
-  case object GoodBye extends Greeting
-  case object ShoutGoodBye extends Greeting
+case object Hello extends Greeting
+case object WhisperHello extends Greeting
+case object GoodBye extends Greeting
+case object ShoutGoodBye extends Greeting
 
-  final implicit val EnumInstance: Enum[Greeting] = Enum.derived[Greeting]
-}
+final implicit val EnumInstance: Enum[Greeting] = Enum.derived[Greeting]
```
jcazevedo, line 35: It seems like this is one of those cases where using the `reset-object` modifier allows us to have these definitions inside the companion object.

jcazevedo (approved):
Looks good! Thanks! I just commented on a location where it seems like using the `reset-object` modifier removes the need to define things differently.

modules/enum/docs/README.md
ruippeixotog: I think I actually prefer not to have a companion object here. In documentation I find it better to have as few boilerplate as possible and stick to the essential bits (sealed family definition and reader instance). Adding a companion object here does not seem very useful - what do you think?

jcazevedo: Having the implicit instance in the companion object allows the example type to be copied to a different file from where the configuration is read without requiring extra modifications, but I'm OK either way.

ruippeixotog: Nearly every example in our documentation has reader instances defined inline outside any companion object. I usually value more the conciseness of REPL-like behavior rather than having example easily copiable to production code (which I wouldn't recommend even if we had companion objects). If anything, let's keep it like this for consistency with the rest of the documentation :)

jcazevedo (approved):

PR #855, opened 2020-10-25
Fix issue with docker build in the "Diff Website" CI check

jcazevedo:
This pull request attempts to fix an issue with building the docker container for the "Comment on pull request" step in the "Diff Website" check.

The issue was fixed in https://github.com/thollander/actions-comment-pull-request/pull/18, but no stable version has been released yet. For the time being, I think we can use the `master` branch of the action.


ruippeixotog (approved):
Perfect!

PR #860, opened 2020-11-01
Use computed latest released version in modules' READMEs

ruippeixotog:
As titled. The strategy of parsing CHANGELOG to get the latest version is kinda hacky, but this is the easiest option I can think of that:

- Computes reliably the latest version that was actually released;
- Does not depend on state external to the repo (i.e. the value of `@VERSION@` should be constant over time in any given revision, even if we check out old commits).

Incidentally, this forces us to always update generated READMEs when we add a new version section to the CHANGELOG. We usually do this right before we release versions and push all changes together, so this doesn't seem like a problem.

### DIFF
--- a/bundle/docs/README.md
+++ b/bundle/docs/README.md
@@ -33,7 +33,7 @@ To use PureConfig in an existing SBT project with Scala 2.11 or a later version,
 `build.sbt`:
 
 ```scala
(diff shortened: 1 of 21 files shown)


jcazevedo (approved):
I suppose this is OK. If we ever start maintaining the CHANGELOG for unreleased versions we can improve the regex to skip them.

PR #881, opened 2020-11-29
Explicitly type implicit definitions

jcazevedo:
This is in preparation for Scala 3 cross-compilation since explicitly typing implicit definitions starts to be required.

### DIFF
--- a/testkit/src/main/scala/pureconfig/arbitrary/package.scala
+++ b/testkit/src/main/scala/pureconfig/arbitrary/package.scala
@@ -3,8 +3,7 @@ package pureconfig
 import java.io.File
 import java.math.BigInteger
 import java.nio.file.Path
-import java.time
-import java.time._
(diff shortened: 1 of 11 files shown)


testkit/src/main/scala/pureconfig/arbitrary/package.scala
```diff
@@ -1,31 +1,41 @@
 package pureconfig
 
+import java.io.File
+import java.math.BigInteger
+import java.nio.file.Path
+import java.time
+import java.time._
```
ruippeixotog: nit: this is probably clearer than `time.Duration` to distinguish it from Scala `Duration`.

```suggestion
import java.time.{Duration => JavaDuration, _}
```

ruippeixotog (approved):
LGTM, it puts us closer to Scala 3 and it's good practice anyway :)

ruippeixotog (approved):

PR #887, opened 2020-12-06
Fix location of test sources for pureconfig-generic

jcazevedo:
The pureconfig-generic test source files weren't under the `scala` folder, so weren't being considered in the build.


melrief (approved):

ruippeixotog (approved):
Nice catch!

PR #890, opened 2020-12-06
Remove dependency of pureconfig-generic from the tests project

jcazevedo:
This PR removes the dependency of pureconfig-generic from the `tests` project, so that we can test pureconfig-core without requiring generic derivation.

Tests specific for pureconfig-generic were moved to the module's tests. Instances of `ConfigReader` and `ConfigWriter` that were being derived using pureconfig-generic are now explicitly created with `ConfigReader.forProductN`.

This also removes the dependency of scalacheck-shapeless from the `tests` project, since we can easily manually generate the `Arbitrary` instances for the types that were previously relying on scalacheck-shapeless, and this eases the process of starting to support Scala 3.


tests/src/test/scala/pureconfig/ConfigReaderFailuresPrettyPrintSuite.scala
```diff
@@ -83,23 +80,6 @@ class ConfigReaderFailuresPrettyPrintSuite extends BaseSuite {
           |  - Expected type OBJECT. Found STRING instead.""".stripMargin
   }
 
-  it should "print a message displaying relevant errors for coproduct derivation" in {
```
ruippeixotog, line 86: Where did this test go?

tests/src/test/scala/pureconfig/ConfigReaderFailuresPrettyPrintSuite.scala
```diff
@@ -175,36 +155,4 @@ class ConfigReaderFailuresPrettyPrintSuite extends BaseSuite {
           |at 'tuple':
           |  - List of wrong size found. Expected 3 elements. Found 6 elements instead.""".stripMargin
   }
-
-  it should "print a message showing the errors of the attempted options when no valid coproduct choice was found" in {
```
ruippeixotog, line 179: Ditto

ruippeixotog (commented):
LGTM overall.

One suggestion: as we're increasing the boilerplate needed to define a test config class, what do you think of moving all test config classes to some `TestConfigs` object or to a separate `data`/`testdata` package so that they can be reused across test suites? We done this to some extent at a file scope (e.g. `ConfigReaderFailureOriginSuite`) and it could potentially take a lot of boilerplate out of tests (at the cost of each test case not being self-contained anymore). Thoughts?

tests/src/test/scala/pureconfig/ConfigReaderFailuresPrettyPrintSuite.scala
jcazevedo: I forgot to reintroduce them. Added them as tests in the pureconfig-generic-base module.

jcazevedo:
> One suggestion: as we're increasing the boilerplate needed to define a test config class, what do you think of moving all test config classes to some `TestConfigs` object or to a separate `data`/`testdata` package so that they can be reused across test suites? We done this to some extent at a file scope (e.g. `ConfigReaderFailureOriginSuite`) and it could potentially take a lot of boilerplate out of tests (at the cost of each test case not being self-contained anymore). Thoughts?

I tend to prefer having test configs self-contained within a suite, both to make the suite easier to read on its own and also to avoid unintentionally messing up with another suite in case of a change. That being said, there are suites (`ApiSuite` and `ConfigSourceSuite`) where the same type could've been used for all test cases, but would've required some changes to the tests, which I purposely tried to avoid in this PR. The change in `ConfigReaderFailureOriginSuite` didn't require any changes because the type was the same.

ruippeixotog (approved):
Fair enough, let's try to do this in a future PR to remove boilerplate then - having them common for the whole test class would already be helpful :)

PR #891, opened 2020-12-08
Start cross-compiling pureconfig-core to Scala 3.0.0-M2

jcazevedo:
Depends on #890.

This PR starts cross-compiling pureconfig-core to Scala 3.0.0-M2. 

This makes the `Derivation` macro a no-op in Scala 3. I'm not sure we should implement it with Scala 3 macros, since errors in implicit chains are much better now, pointing the user to the most specific implicit argument that was missing in the implicit search. I'm actually inclined to drop it entirely, since the `Derivation` type leaks into PureConfig's external API, the macro is hard to maintain and we can recommend users in Scala 2 to use [splain](https://github.com/tek/splain) for better implicit not found errors. @leifwickland and @ruippeixotog: If you agree with this, I'll open a separate PR to remove the `Derivation` type and update the documentation to provide a guide to using [splain](https://github.com/tek/splain) for debugging implicit not found errors.

My plan is to follow-up with the necessary updates to support generic derivation of `ConfigReader` and `ConfigWriter` instances in another PR, hopefully still during this week. I think we should be able to support this directly in pureconfig-core using the type class derivation features of Scala 3 (and taking advantage of the `derives` machinery). We probably still want to keep pureconfig-generic for compatibility purposes and to provide the configuration mechanisms for derivation we currently support with hints.


core/build.sbt
```diff
@@ -1,7 +1,9 @@
-import Dependencies._
+import Dependencies.{Version, _}
```
ruippeixotog: Do we need this change?

core/src/main/scala/pureconfig/DurationUtils.scala
```diff
@@ -115,6 +115,7 @@ private[pureconfig] object DurationUtils {
       case Duration.MinusInf => "MinusInf"
       // We must do an `eq` instead of `==` comparison because `Undefined` is intentionally != itself.
       case d if d eq Duration.Undefined => UndefinedDuration
+      case _ => UndefinedDuration
```
ruippeixotog, line 116: Are we sure we're not missing any case? If so, we can remove the case above this one.

core/build.sbt
```diff
@@ -1,7 +1,9 @@
...
 
 name := "pureconfig-core"
 
+crossScalaVersions += Version.scala30
```
ruippeixotog: It's becoming increasing complex to handle `crossScalaVersions` across projects. Up until recently not all projects supported Scala 2.13; a significant number of them don't support Scala 2.11 anymore and I suppose we'll be introducing Scala 3.0 compilation incrementally over time. Moreover, we don't always use the same syntax, preferring to filter out 2.11 versions using a `~=` rather than `-=`.

At this point, I'm starting to think it may not make sense anymore to have a default for `crossScalaVersions` and we should simply have all projects define it explicitly. What do you think?

build.sbt
```diff
@@ -103,19 +106,25 @@ lazy val commonSettings = Seq(
...
-    List(
-      if (minor.exists(_ <= 12)) unmanagedSrcs.value.map { dir => new File(dir.getPath + "-2.12-") }
-      else Nil,
-      if (minor.exists(_ >= 12)) unmanagedSrcs.value.map { dir => new File(dir.getPath + "-2.12+") }
-      else Nil
-    ).flatten
+    val version = CrossVersion.partialVersion(scalaVersion.value)
+    val expectedVersions = Seq((2, 11), (2, 12), (2, 13), (3, 0))
+    expectedVersions.flatMap { case (major, minor) =>
+      lazy val versionV = major * 100 + minor
```
ruippeixotog: Scala provides an `Ordering[(Int, Int)]` instance. Can we compare pairs directly for conciseness and to avoid introducing assumptions on how versioning is done?

ruippeixotog: Better yet, SBT provides native support for semantic versioning: https://www.scala-sbt.org/1.0/docs/sbt-1.2-Release-Notes.html#Semantic+Version+selector+API

build.sbt
```diff
@@ -124,20 +133,26 @@ lazy val lintFlags = {
     "-encoding",
     "UTF-8", // arg for -encoding
     "-feature",
-    "-unchecked",
+    "-unchecked"
+  )
+
+  lazy val preScala3LintFlags = allVersionLintFlags ++ List(
```
ruippeixotog: Like `crossScalaVersions`, this is another aspect where I think that extracting out common parts is starting to make the build too complex. With four Scala versions, each mixing different subsets of flags, I'd prefer if we took the cost of duplication and simply defined one `List` per major version.

build.sbt
```diff
@@ -103,19 +106,25 @@ lazy val commonSettings = Seq(
...
   unmanagedSrcs ++= {
-    val minor = CrossVersion.partialVersion(scalaVersion.value).map(_._2)
-    List(
-      if (minor.exists(_ <= 12)) unmanagedSrcs.value.map { dir => new File(dir.getPath + "-2.12-") }
-      else Nil,
-      if (minor.exists(_ >= 12)) unmanagedSrcs.value.map { dir => new File(dir.getPath + "-2.12+") }
-      else Nil
-    ).flatten
+    val version = CrossVersion.partialVersion(scalaVersion.value)
+    val expectedVersions = Seq((2, 11), (2, 12), (2, 13), (3, 0))
```
ruippeixotog: Can we extract this from our existing definitions in `Dependencies`?

testkit/src/main/scala/pureconfig/ConfigReaderMatchers.scala
```diff
@@ -26,12 +26,10 @@ trait ConfigReaderMatchers { this: AnyFlatSpec with Matchers =>
   def failWith(failure: ConfigReaderFailure): Matcher[ConfigReader.Result[Any]] =
     be(Left(ConfigReaderFailures(failure)))
 
-  def failWithType[Reason <: FailureReason: ClassTag]: Matcher[ConfigReader.Result[Any]] =
+  def failWithConvertFailureOf[Reason <: FailureReason: ClassTag]: Matcher[ConfigReader.Result[Any]] =
```
ruippeixotog: Do we really have to do this? This is significantly more verbose than before and on this particular instance I don't think we gain much in terms of conciseness.

tests/build.sbt
```diff
@@ -1,7 +1,7 @@
-import Dependencies._
+import Dependencies.{Version, _}
```
ruippeixotog: Ditto

.github/workflows/ci.yml
```diff
@@ -24,12 +24,23 @@ jobs:
         run: sbt "++${{ matrix.scala }} scalafmtCheckAll; scalafmtSbtCheck"
 
       - name: Compile
+        if: startsWith(matrix.scala, '2')
         run: sbt coverage "++${{ matrix.scala }} compile"
 
+      - name: Compile (without coverage)
```
ruippeixotog: This will produce two steps for compile and for tests visible in GitHub Actions, regardless of the Scala version. It's a bit verbose on this config too.

Instead, can we conditionally include `coverage` in the command on the same step? I believe you can do that using Github Actions own Liquid-like syntax and `${{ matrix.scala }}`.

ruippeixotog (changes requested):
Thanks for doing this, I'm excited to start supporting Scala 3.0! Apart from one comment regarding `failWithType` the code changes look good to me, but I think we need to simplify the build.

ruippeixotog:
Regarding `Derivation` I'm up for deprecating it 👍  I'm happy to know better implicit explanation is being introduced in Scala 3 and splain is becoming a mature option, as this shouldn't be left to libraries like PureConfig anyway and the current approach is rather hacky.

testkit/src/main/scala/pureconfig/ConfigReaderMatchers.scala
jcazevedo: The overload resolution rules change in Scala 3 (https://dotty.epfl.ch/docs/reference/changed-features/overload-resolution.html) so we have an ambiguous overload even with the `DummyImplicit` trick. I suppose we can rename this differently for conciseness (e.g. `failWithReason`?).

core/build.sbt
jcazevedo: Not if we refer to `Version` differently, but there's an `autoImport` from the sbt-microsites plugin also putting a `Version` in scope.

core/src/main/scala/pureconfig/DurationUtils.scala
jcazevedo: I think we can remove the case above, yes.

core/build.sbt
jcazevedo: I agree. It makes it easier to understand which versions a given module is being compiled with.

build.sbt
jcazevedo: Nice! Unfortunately, the semantic version support doesn't seem to play very well with versions with a suffix (like the milestone release of Scala 3 "3.0.0-M2"). For the time being, I ended up using the `Ordering` instance for this.

build.sbt
jcazevedo: I agree.

.github/workflows/ci.yml
jcazevedo: I couldn't find a way to do this using GitHub Actions' expression syntax, so ended up using bash in order to have a single step.

modules/magnolia/src/test/scala/pureconfig/module/magnolia/CoproductHintSuite.scala
```diff
@@ -76,7 +76,9 @@ class CoproductHintSuite extends BaseSuite {
       val cc = implicitly[ConfigConvert[Conf]]
 
       val conf = ConfigFactory.parseString("{ typ = ambiguous-conf }")
-      cc.from(conf.root()) should failWithType[KeyNotFound] // "typ" should not be passed to the coproduct option
+      cc.from(conf.root()) should failWithReason[
+        KeyNotFound
+      ] // "typ" should not be passed to the coproduct option
```
ruippeixotog: nit: let's move the comment to a separate line to avoid scalafmt formatting it like this?

testkit/src/main/scala-3.0+/pureconfig/TypeStringCompat.scala
```diff
@@ -0,0 +1,17 @@
+package pureconfig
+
+import scala.reflect.ClassTag
+
+trait TypeStringCompat[T] {
+  def typeName: String
+}
+
+object TypeStringCompat {
+  // FIXME: ClassTag are weaker cases of TypeTags, which are not available in Scala 3. This is sufficient for our use
```
ruippeixotog, line 10: How are we planning to address this in the future? Is there an alternative more strongly typed than `ClassTag` in Scala 3?

build.sbt
```diff
@@ -88,7 +87,10 @@ lazy val commonSettings = Seq(
   scalacOptions ++= lintFlags.value,
 
   scalacOptions in Test ~= { _.filterNot(_.contains("-Ywarn-unused")) },
-  scalacOptions in Test += "-Xmacro-settings:materialize-derivations",
+  scalacOptions in Test ++= forScalaVersions {
+    case (2, _) => List("-Xmacro-settings:materialize-derivations")
```
ruippeixotog: Does Scala 3 complain if you use this flag? If not I'd leave it set on all version for simplicity (`-Xmacro-settings:` are reserved for use by macros anyway).

testkit/build.sbt
```diff
@@ -1,7 +1,16 @@
...
 
 name := "pureconfig-testkit"
 
-libraryDependencies ++= Seq(scalaTest, scalaCheck, scalaTestPlusScalaCheck)
+crossScalaVersions := Seq(scala211, scala212, scala213, scala30)
+
+libraryDependencies ++= Seq(
+  Dependencies.scalaTest,
+  Dependencies.scalaCheck.withDottyCompat(scalaVersion.value),
+  Dependencies.scalaTestPlusScalaCheck.intransitive().withDottyCompat(scalaVersion.value)
```
ruippeixotog, line 13: Why are we doing this? Can you add a comment explaining why this is needed?

build.sbt
```diff
@@ -103,53 +105,77 @@ lazy val commonSettings = Seq(
...
-    "-Ywarn-dead-code",
-    "-Ywarn-numeric-widen"
-  )
-
-  def withCommon(flags: String*) =
-    allVersionLintFlags ++ flags
-
-  forScalaVersions {
+lazy val lintFlags = forScalaVersions { v =>
+  (v: @unchecked) match {
```
ruippeixotog: How about getting rid of this `v` and using a partial function with `forScalaVersions` directly?

build.sbt
```diff
@@ -103,53 +105,77 @@ lazy val commonSettings = Seq(
...
-      if (minor.exists(_ <= 12)) unmanagedSrcs.value.map { dir => new File(dir.getPath + "-2.12-") }
-      else Nil,
-      if (minor.exists(_ >= 12)) unmanagedSrcs.value.map { dir => new File(dir.getPath + "-2.12+") }
-      else Nil
-    ).flatten
+    val versionNumber = CrossVersion.partialVersion(scalaVersion.value)
+    val expectedVersions = Seq(scala211, scala212, scala213, scala30).flatMap(CrossVersion.partialVersion)
+    expectedVersions.flatMap { case v @ (major, minor) =>
+      List(
+        if (versionNumber.exists(Ordering[(Long, Long)].lteq(_, v)))
```
ruippeixotog: nit: if you import `scala.math.Ordering.Implicits._` you can simply do `.exists (_ <= v)` and `.exists(_ >= v)` here.

ruippeixotog (changes requested):

core/build.sbt
ruippeixotog: Much better, thanks!

build.sbt
jcazevedo: We get a warning:

```
[warn] bad option '-Xmacro-settings:materialize-derivations' was ignored
```

This is eventually going away with the deprecation of the `Derivation` type, so I wouldn't worry too much about it.

build.sbt
jcazevedo: I think we lose the selector to put the `@unchecked` annotation that way. The warning is a bit annoying.

testkit/build.sbt
jcazevedo: We need to avoid having scalactic and scalatest-core pulled both for Scala 3.x (via the `scalaTest` dependency) and for Scala 2.x (via the `scalaTestPlusScalaCheck` dependency with the `withDottyCompat` flag).

testkit/src/main/scala-3.0+/pureconfig/TypeStringCompat.scala
jcazevedo: I believe we should be able to use [`quoted.Type`](https://dotty.epfl.ch/api/scala/quoted/Type.html) for this. Unfortunately, to use `quoted.Type`, we need a `QuoteContext` in scope, which involves more than a direct type class translation. Since it requires more deep changes in the places that are using it, I preferred to defer that change to when using `ClassTag`s starts becoming a problem.

modules/magnolia/src/test/scala/pureconfig/module/magnolia/CoproductHintSuite.scala
jcazevedo: I think this was due to a previous rename of that method. The original line should fit without line breaks.

build.sbt
ruippeixotog: I may be missing some detail, but the annotation doesn't seem to be needed here as you're not pattern-matching on any erased type. If you start pattern matching directly you may need to add a new `case _ => Nil` or `case _ => throw new Exception("Unknown version")`, which is good practice anyway, but other than that everything should work fine without warnings.

jcazevedo: Right, sorry! I was assuming you meant using a partial function directly. Adding a fallback works as well, yes.

ruippeixotog (approved):
Thanks @jcazevedo!

PR #912, opened 2021-01-16
Fix link to pureconfig-enumeratum

monktastic:


ruippeixotog (changes requested):
Hi @monktastic! We use [mdoc](https://github.com/scalameta/mdoc) to generate documentation for PureConfig READMEs. You need to update [modules/enum/docs/README.md](https://github.com/pureconfig/pureconfig/blob/master/modules/enum/docs/README.md) instead and run `sbt mdoc` afterwards for the build to pass.

monktastic:
Thanks! Not sure of the process here, but made that change and ran `sbt mdoc` and pushed up a new commit.

ruippeixotog (approved):
Thank you @monktastic!

PR #932, opened 2021-02-07
Disable sbt doc in GitHub Actions on Scala 3.0

ruippeixotog:
`sbc doc` is non-deterministically failing for the last few CI builds. I suspect this is due to Scaladoc's microsites-like static site generation feature on 3.0 (https://dotty.epfl.ch/docs/usage/scaladoc/staticSite.html). This expects Markdown on a `docs`  folder just like we have and outputs a `_site` directory that seems to be causing some trouble. Let's disable this for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,7 @@ jobs:
           COVERALLS_FLAG_NAME: Scala ${{ matrix.scala }}
 
       - name: Build Scaladoc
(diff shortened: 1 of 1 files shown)


ruippeixotog:
Scaladoc is also generated during `publish` and `publishLocal`, so we need another approach... I'll try to figure out how to disable static site generation.

ruippeixotog:
Closed in favor of #933.

Issue #854, opened 2020-10-24
Dotty / Scala 3 Support


bbstilson:
I was chatting with some kind folks in the dotty gitter, and someone pointed out that because the derivation macros are based on Scala 2 macros, pureconfig can't be used with dotty/scala 3 until they're ported.

jcazevedo:
Starting to add support for Scala 3 in PureConfig has been on my backlog for a while, but I haven't started working on it yet. I think I'll be able to have some time for it later this week. I won't get in the way of pull requests if someone wants to start tackling it in the meantime.

jakubjanecek:
Hi, what is the progress on Scala 3 support. If I see it correctly the `master` is already build for Scala 3 but there is not official (SNAPSHOT) release. Could you please make some so that other projects depending on PureConfig can start preparing for Scala 3? Thank you.

ruippeixotog:
Good call @jakubjanecek, I thought we already had a released version for Scala 3. I'll try to do a release this weekend.

ruippeixotog:
`pureconfig-core` v0.14.1 [was released](https://mvnrepository.com/artifact/com.github.pureconfig/pureconfig-core) this weekend for Scala 3.0.0-RC1, you should be able to use it now. Sorry again for the delay.

jakubjanecek:
Thanks but unfortunately `pureconfig-core` is not enough, I need the full `pureconfig` artifact as I am using the `semiauto` derivations in my project. Or is there other way around that?

jcazevedo:
We're still working on adding support for automatic derivation of `ConfigReader` and `ConfigWriter` instances in Scala 3. In the meantime, only pureconfig-core is available, so you won't be able to use semiauto derivations. If you want to immediately start using the new PureConfig version in Scala 3, you can migrate your instance definitions to use the `ConfigReader.forProductN` and `ConfigWriter.forProductN` helpers.

dyercode:
So since you closed this, what specific issue is there to follow for derivations/pureconfig proper? For those of us who were watching this issue expecting it.

jcazevedo:
@dyercode: You can use #970 to keep track of that.

bbstilson:
Thanks for your hard work @jcazevedo, @ruippeixotog

PR #959, opened 2021-02-26
Bump Scala 3 to version 3.0.0-RC1

ruippeixotog:
Bumping Scala 3 ahead of a new PureConfig release (as per #854).

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ object Dependencies {
     val scala211 = "2.11.12"
     val scala212 = "2.12.12"
     val scala213 = "2.13.3"
-    val scala30 = "3.0.0-M3"
+    val scala30 = "3.0.0-RC1"
 
     val shapeless = "2.3.3"
     val typesafeConfig = "1.4.1"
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -6,13 +6,13 @@ object Dependencies {
...
     val scala213 = "2.13.3"
-    val scala30 = "3.0.0-M3"
+    val scala30 = "3.0.0-RC1"
 
     val shapeless = "2.3.3"
     val typesafeConfig = "1.4.1"
 
-    val scalaTest = "3.2.3"
-    val scalaTestPlusScalaCheck = "3.2.2.0"
+    val scalaTest = "3.2.5"
```
ruippeixotog, line 13: Having dependencies structured in this non-standard way prevents scala-steward from detecting updates and bumping versions. I don't think we gain anything in not inlining them, I'll investigate changing it.

jcazevedo: What do you mean by non-standard way? Scala Steward seemed to be able to propose version updates in the past for these dependencies, #816 and #865 being examples of it.

ruippeixotog: Oh, never mind then. Non-standard way as in not in the format `"org" % "artifact" % "version"`, but it looks like scala-steward is smarter than that :)

jcazevedo (approved):

PR #963, opened 2021-02-27
Have more stable status check names in CI

jcazevedo:
This is an attempt to have more stable status check names in CI by not tagging checks all the way to the Scala patch version. This has the benefit of not forcing us to update the branch protection rules in GitHub on every patch update of a Scala version and also allows for more stable version tags in Coveralls.


ruippeixotog (approved):
Cool!

PR #968, opened 2021-03-02
Add support for Scala 3.0 in pureconfig-bundle

ruippeixotog:
We forgot to add it to the bundle project, so the `pureconfig` artifact was not published on our latest release (https://mvnrepository.com/artifact/com.github.pureconfig/pureconfig).

After this is accepted I'm going to to an ad-hoc release of `pureconfig` for v0.14.1.

### DIFF
--- a/bundle/build.sbt
+++ b/bundle/build.sbt
@@ -2,4 +2,4 @@ import Dependencies.Version._
 
 name := "pureconfig"
 
-crossScalaVersions := Seq(scala212, scala213)
+crossScalaVersions := Seq(scala212, scala213, scala30)


leifwickland (approved):

jcazevedo:
I don't think we can do this yet. pureconfig-generic isn't ready for Scala 3 yet.

ruippeixotog:
My bad, I thought shapeless already had a 3.0 version. Let me close this and close #854 too.

Now that I think of it, maybe what we want is to create a generic module using [Scala 3 derivations](https://dotty.epfl.ch/docs/reference/contextual/derivation.html). Now that RC1 is out I think it should be good to create one. Have you investigated them? If not I'll take a look this week.

jcazevedo:
> Now that I think of it, maybe what we want is to create a generic module using [Scala 3 derivations](https://dotty.epfl.ch/docs/reference/contextual/derivation.html). Now that RC1 is out I think it should be good to create one. Have you investigated them? If not I'll take a look this week.

That is my plan, as suggested in #891, but I haven't gotten around to doing it yet. As mentioned in that PR, I think we can provide generic derivation for Scala 3 directly in the core module, to take advantage of the `derives` machinery. pureconfig-generic would be kept for compatibility purposes and to allow the configuration of generic derivations with hints. I'd eventually do that, but feel free to start working on it.

Generic derivation is a big part of PureConfig, so I'd prefer if we keep #854 open until we provide the same functionality for Scala 3, or open another issue to track that.

ruippeixotog:
> That is my plan, as suggested in #891, but I haven't gotten around to doing it yet.

Sorry, I seem to have forgotten much of our previous discussion. For some reason I was really convinced that pureconfig-generic was already on Scala 3.0 🤦 

> As mentioned in that PR, I think we can provide generic derivation for Scala 3 directly in the core module, to take advantage of the derives machinery. pureconfig-generic would be kept for compatibility purposes and to allow the configuration of generic derivations with hints. I'd eventually do that, but feel free to start working on it.

Would it be mandatory to use `derives` in pureconfig-core? Looking at [this](https://dotty.epfl.ch/docs/reference/contextual/derivation.html#deriving-instances-elsewhere) I thought we could have a separate module for native Scala 3 derivation, like `pureconfig-generic-dotty`. I'd prefer to keep as few version-dependent code like `FactoryCompat` as possible, as I find it hard to maintain and clients generally don't expect library code to change when a different Scala version is used.

> Generic derivation is a big part of PureConfig, so I'd prefer if we keep #854 open until we provide the same functionality for Scala 3, or open another issue to track that.

I'd prefer to open a new issue for this. I suppose that providing the same functionality (including hints and generic stuff) will only really be possible when shapeless does a 3.0 release, whereas people might be happy simply with an alternative way to perform derivation.

jcazevedo:
> Would it be mandatory to use `derives` in pureconfig-core? Looking at [this](https://dotty.epfl.ch/docs/reference/contextual/derivation.html#deriving-instances-elsewhere) I thought we could have a separate module for native Scala 3 derivation, like `pureconfig-generic-dotty`. I'd prefer to keep as few version-dependent code like `FactoryCompat` as possible, as I find it hard to maintain and clients generally don't expect library code to change when a different Scala version is used.

It's not mandatory, but it'd be nice to define a type with automatically derived instances for `ConfigReader` and `ConfigWriter` in Scala 3 using `class MyConf derives ConfigReader, ConfigWriter`, and that's only possible if we have a `derived` method in the companion object of the type classes.

jcazevedo:
> I'd prefer to open a new issue for this. I suppose that providing the same functionality (including hints and generic stuff) will only really be possible when shapeless does a 3.0 release, whereas people might be happy simply with an alternative way to perform derivation.

I've opened #970 to track that.

PR #979, opened 2021-03-21
Remove the `Derivation` macro

jcazevedo:
As suggested in #891, this removes the `Derivation` macro and adapts the FAQ so that we refer to [splain](https://github.com/tek/splain) as a tool to debug implicit not found errors in Scala 2.x.


ruippeixotog (approved):
Looks good, thanks for cleaning this up!

PR #981, opened 2021-03-21
Enable Scalafix and the `OrganizeImports` rule

jcazevedo:
This enables [Scalafix](https://scalacenter.github.io/scalafix/) and the [`OrganizeImports` rule](https://github.com/liancheng/scalafix-organize-imports) so that we can guarantee a consistent style for imports. I'm not sure if we had a style we were aiming for previously, but I opted to group imports into 4 groups: java imports, scala imports, non-pureconfig imports and pureconfig imports.


.scalafix.conf
```diff
@@ -0,0 +1,12 @@
...
+
+OrganizeImports {
+  coalesceToWildcardImportThreshold = 2147483647 # Int.MaxValue
+  expandRelative = false
+  groupExplicitlyImportedImplicitsSeparately = false
+  groupedImports = AggressiveMerge
+  groups = ["re:javax?\\.", "scala.", "*", "pureconfig."]
+  importSelectorsOrder = Ascii
+  importsOrder = Ascii
+  removeUnused = false
```
ruippeixotog, line 11: I know the compiler already warn us of unused imports, but what would be the downsides of enabling it here too? Wouldn't we get automated fixing and earlier CI signal (rather than waiting for compilation to finish)?

build.sbt
```diff
@@ -96,6 +99,13 @@ lazy val commonSettings = Seq(
 
   scalafmtOnCompile := true,
 
+  // We can't use Scalafix in Scala 3 yet.
+  libraryDependencies ++= forScalaVersions {
+    case (2, _) => List(compilerPlugin(scalafixSemanticdb))
+    case _ => List.empty
+  }.value,
+  scalafixOnCompile := forScalaVersions { case (2, _) => true; case _ => false }.value,
```
ruippeixotog, line 104: I'm a bit wary of having more and more build stuff conditionally disabled or modified for Scala 3.0... let's keep close track of these and remove them as soon as Scala 3.0 gets support for it.

build.sbt
```diff
@@ -132,6 +142,7 @@ lazy val lintFlags = forScalaVersions {
       "-Xlint:_,-unused",
       "-Xfatal-warnings",
       "-Yno-adapted-args",
+      "-Yrangepos",
```
ruippeixotog: Why did you enable this?

.github/workflows/ci.yml
```diff
@@ -28,7 +28,8 @@ jobs:
           java-version: ${{ matrix.jdk }}
 
       - name: Check formatting
```
ruippeixotog: The name doesn't look accurate anymore. Maybe "Check formatting and run linter"?

ruippeixotog (changes requested):
Could you please fix merge conflicts?

.scalafix.conf
jcazevedo: This won't have any effect with `-Xfatal-warnings` enabled because the build will fail before running Scalafix.

build.sbt
jcazevedo: It's required for the SemanticDB compiler plugin (https://scalacenter.github.io/scalafix/docs/users/installation.html). I added a comment to make it explicit.

.github/workflows/ci.yml
jcazevedo: Yes. That seems like a better name.

ruippeixotog (approved):
Thanks!

PR #996, opened 2021-03-31
Migration to Cats Effect 3.0, FS2 3.0

### LINKED ISSUE #946: Support cats-effect 3.0 series and http4s 1.0 series in modules
`cats-effect` 3 has reached RC2 state. It would be nice to start building support for this new library. This will also involve bumping `http4s` to 1.0 series, which supports `cats-effect` 3.0 series.

Both modules are pretty compact, so, it should be relatively easy to implement. The only significant change will be to undeprecate methods w/o `Blocker` instances (because `Blocker` is gone), so, the API of cats-effect module will change in significantly breaking way.

But the problem is that it will be difficult to maintain versions for both series ce3/http4s1 and ce2/http4s0.21 withing the same tree and keep the current versioning scheme. May be it is worth to extract them to separate project/projects and apply reflect version differences there. But this will require some planning from project maintainers anyway.

arixmkii:
Fixes #946 

At least it compiles and passes tests 🤷 But need to check API changes and decide if something else should be changed as well (like dropping more deprecated methods).

### DIFF
--- a/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
+++ b/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
@@ -158,15 +158,10 @@ package object catseffect {
       )
 
     val check =
-      F.ifElseM(
-        F.blocking(!overrideOutputPath && Files.isRegularFile(outputPath)) -> fileAlreadyExists,
-        F.blocking(Files.isDirectory(outputPath)) -> fileIsDirectory
-      )(F.unit)
+      F.blocking(!overrideOutputPath && Files.isRegularFile(outputPath))
+        .ifM(fileAlreadyExists, F.blocking(Files.isDirectory(outputPath)).ifM(fileIsDirectory, F.unit))
 
-    val outputStream =
-      Resource.make(F.blocking(Files.newOutputStream(outputPath))) { os =>
-        F.blocking(os.close())
-      }
+    val outputStream = Resource.fromAutoCloseable(F.blocking(Files.newOutputStream(outputPath)))
 
     check >> outputStream.use { os =>
       blockingSaveConfigToStreamF(conf, os, options)
--- a/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
+++ b/modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
@@ -8,7 +8,7 @@ import scala.language.higherKinds
 import scala.reflect.ClassTag
 
 import cats.data.{EitherT, NonEmptyList}
-import cats.effect.{Blocker, ContextShift, Resource, Sync}
+import cats.effect.{Resource, Sync}
 import cats.implicits._
 import com.typesafe.config.{ConfigRenderOptions, Config => TypesafeConfig}
 import pureconfig._
@@ -26,29 +26,10 @@ package object catseffect {
     *         `A` from the configuration source, or fail with a ConfigReaderException which in turn contains
     *         details on why it isn't possible
     */
-  @deprecated("Use `loadF[F, A](cs, blocker)` instead", "0.12.3")
   def loadF[F[_], A](
       cs: ConfigSource
-  )(implicit F: Sync[F], reader: ConfigReader[A], ct: ClassTag[A]): F[A] = {
-    val delayedLoad = F.delay {
-      cs.load[A].leftMap[Throwable](ConfigReaderException[A])
-    }
-    delayedLoad.rethrow
-  }
-
-  /** Load a configuration of type `A` from a config source
-    *
-    * @param cs the config source from where the configuration will be loaded
-    * @param blocker the blocking context which will be used to load the configuration.
-    * @return The returned action will complete with `A` if it is possible to create an instance of type
-    *         `A` from the configuration source, or fail with a ConfigReaderException which in turn contains
-    *         details on why it isn't possible
-    */
-  def loadF[F[_], A](
-      cs: ConfigSource,
-      blocker: Blocker
-  )(implicit F: Sync[F], csf: ContextShift[F], reader: ConfigReader[A], ct: ClassTag[A]): F[A] =
-    EitherT(blocker.delay(cs.cursor()))
+  )(implicit F: Sync[F], reader: ConfigReader[A], ct: ClassTag[A]): F[A] =
+    EitherT(F.blocking(cs.cursor()))
       .subflatMap(reader.from)
       .leftMap(ConfigReaderException[A])
       .rethrowT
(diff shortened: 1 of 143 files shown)


rubiconvvv:
@kubukoz @ruippeixotog hey guys. please check this out, we want to convert our project to cats 3 as soon as possible.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -169,42 +137,39 @@ package object catseffect {
...
         new IllegalArgumentException(s"Cannot save configuration in file '$outputPath' because it is a directory")
       )
 
     val check =
-      F.defer {
-        if (!overrideOutputPath && Files.isRegularFile(outputPath)) fileAlreadyExists
-        else if (Files.isDirectory(outputPath)) fileIsDirectory
-        else F.unit
-      }
+      F.ifElseM(
```
kubukoz: I would recommend `F.blocking(...).ifM(ifTrue = ..., ifFalse = ...)` instead, more explicit

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -169,42 +137,39 @@ package object catseffect {
...
+        F.blocking(!overrideOutputPath && Files.isRegularFile(outputPath)) -> fileAlreadyExists,
+        F.blocking(Files.isDirectory(outputPath)) -> fileIsDirectory
+      )(F.unit)
 
     val outputStream =
-      Resource.make(blocker.delay(Files.newOutputStream(outputPath))) { os =>
-        blocker.delay(os.close())
+      Resource.make(F.blocking(Files.newOutputStream(outputPath))) { os =>
```
kubukoz: if the result of `newOutputStream` is a Closeable, you can probably do `Resource.fromAutoCloseable`

kubukoz (commented):
In general it looks good :)

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
arixmkii: Thanks! Didn't think about syntax version of `ifM`. Fits really nice here.

kubukoz:
I'll unsubscribe now but feel free to ping me if you need another round.

project/Dependencies.scala
```diff
@@ -3,7 +3,7 @@ import sbt._
 object Dependencies {
 
   object Version {
-    val scala212 = "2.12.12"
+    val scala212 = "2.12.13"
```
jcazevedo: This update is blocked by an sbt-coverage issue (scoverage/sbt-scoverage#321). Is this required for this PR?

ruippeixotog: We can't migrate to 2.12.13 yey because scoverage doesn't support it yet (see https://github.com/scoverage/sbt-scoverage/issues/321). This is also what's making the build fail.

Is it required for any of the modules?

modules/http4s/build.sbt
```diff
@@ -4,7 +4,7 @@ name := "pureconfig-http4s"
 
 crossScalaVersions := Seq(scala212, scala213)
 
-libraryDependencies ++= Seq("org.http4s" %% "http4s-core" % "0.21.21")
+libraryDependencies ++= Seq("org.http4s" %% "http4s-core" % "1.0.0-M20")
```
jcazevedo: We don't usually publish modules for milestone versions of upstream dependencies. Can we delay this update for when a stable release of http4s in the 1.0 series comes out?

ruippeixotog: How close are we to a RC version here? I'd rather avoid milestone versions if possible.

ruippeixotog (changes requested):
Thanks for the PR @arixmkii! I don't have much experience with cats-effect but I see we already got some eyes on it, so LGTM. I only left a few comments below.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -169,42 +137,34 @@ package object catseffect {
     *
     * @param conf The configuration to save
     * @param outputPath Where to write the configuration
-    * @param blocker the blocking context which will be used to load the configuration.
     * @param overrideOutputPath Override the path if it already exists
     * @param options the config rendering options
     * @return The return action will save out the supplied configuration upon invocation
     */
   def blockingSaveConfigAsPropertyFileF[F[_], A](
```
jcazevedo: We should remove the deprecated `saveConfigAsPropertyFileF` method now, since it has the same signature as this and its implementation doesn't correctly scope blocking operations.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -229,22 +189,20 @@ package object catseffect {
     *
     * @param conf The configuration to write
     * @param outputStream The stream in which the configuration should be written
-    * @param blocker the blocking context which will be used to load the configuration.
     * @param options the config rendering options
     * @return The return action will save out the supplied configuration upon invocation
     */
   def blockingSaveConfigToStreamF[F[_], A](
```
jcazevedo: We should remove the deprecated `saveConfigToStreamF` method now, since it has the same signature as this and its implementation doesn't correctly scope blocking operations.

jcazevedo (changes requested):
@arixmkii: Thanks for this!

Since Cats Effect 3 removes the `Blocker` construct, I don't think it makes much sense to now have both `save...F` and `blockingSave...F` methods. I suggest we move the `blockingSave...F` implementations to the `save...F` ones and deprecate the `blockingSave...F` methods, while removing the deprecation from the `save...F` methods.

project/Dependencies.scala
arixmkii: No, it is not required. I will downgrade.

modules/http4s/build.sbt
arixmkii: It is the first version compatible with cats-effect 3.0. Can't tell how close it is to RC state, because there are more changes than just cats-effect 3.0. Parts related to dependencies looks mostly complete for me, but it will need some time to publish with RC tag.

I checked the source of the module. It is not using any effect related methods. It could be possible to compile module with older version, but it will make it impossible to use http4s and cats-effect module in the same application anyway (because of dependency on incompatible versions of cats-effect). If this is ok (at least temporary), it could be downgraded.

modules/cats-effect/src/main/scala/pureconfig/module/catseffect/package.scala
arixmkii: Removed deprecated methods and reused their names for blocking implementations.

modules/http4s/build.sbt
arixmkii: Downgraded version. Please see my response in different conversation for what it means for compatibility between modules.

project/Dependencies.scala
arixmkii: Downgraded. Bumped it while was looking through code trying to understand the project. Didn't know that there are issues with coverage as used coverage only for Scala 2.13 before.

ruippeixotog (approved):
LGTM now, thanks. Will let @jcazevedo take a second look and merge this.

jcazevedo (approved):
@arixmkii: Looks good. Thanks!

Issue #970, opened 2021-03-02
Add support for generic derivation of `ConfigReader` and `ConfigWriter` instances in Scala 3


jcazevedo:

heksesang:
I have been putting together this during the last week:
https://gitlab.com/heksesang/scala3-derivation/-/tree/master/pureconfig-derivation/src/no/heksesang/derivation/pureconfig

Still need to add the value class and tuple cases, but was thinking once that was done I'd look at embedding this into `pureconfig-core`.

ruippeixotog:
Closed by #1019.

jcazevedo:
I think we should keep this open while we still don't have support for deriving `ConfigWriter` instances in Scala 3.

ruippeixotog:
You're right, I forgot about `ConfigWriter`!

carueda:
Status?

ruippeixotog:
Sorry, nothing so far :/

tjarvstrand:
Is there any hope of this being completed? I love pureconfig but it's the only thing holding me back from going to Scala 3 and at this point I'm having to consider switching to something else.

evis:
Currently, we can derive `ConfigReader` using `derives ConfigReader`. Could we also support deriving it in Scala 3 using `deriveReader[T]`, like in Scala 2? That would simplify migration to Scala 3. Currently we have to replace all `deriveReader` invocations with `derives`, that's quite cumbersome.

Should I create a separate issue for that?

evis:
Created a separate issue: https://github.com/pureconfig/pureconfig/issues/1590

PR #1019, opened 2021-04-20
Added Scala 3 derivation for ConfigReader.

heksesang:
A first step towards solving #970. This only adds derivation for readers, so writers are still a todo.


jcazevedo:
@heksesang: Thanks for you PR! I'm going to need some time to properly review it. 

As a more general note, I would prefer, at least while we're compiling PureConfig for Scala 2.x, to avoid using Scala 3's new syntax, in an effort to make the syntax more consistent throughout the codebase.

heksesang:
Ah, right, I remember you mentioned that. I'll try to find time to fix that during the next couple of days. :)

This implementation doesn't include features that cannot be achieved purely with mirrors, I figured it's better to leave things that require more complex macros to the magnolia- or shapeless-based implementations. So there is no support for default values for parameters.

Also, currently the extension for the derivation syntax lives in the `pureconfig` package – is it preferable to have it under the `pureconfig.generic` package so the syntax is more opt-in, or have it like now, that you get it by default when doing `import pureconfig.*`?

ruippeixotog (changes requested):
Hi @heksesang, this is a great contribution, thanks! The build doesn't seem to be passing for Scala 3.0 and I'm also getting a build error locally (also related to formatting, probably). Can you please fix it?

I am still reviewing the code thoroughly so I don't have concrete comments yet, but I have two general remarks:

- Please be careful with not polluting the `pureconfig.generic` namespace and bear in mind that many users are currently importing it. This PR is introducing a lot of new definitons in `pureconfig.generic` that should be private. As you probably know our shapeless-based derivation code also lives in the `pureconfig.generic` package and many users import it to get important classes like hints, so the simple act of migrating to Scala 3 may introduce clashes for them. Please either use another package name or scope everything inside objects with unique-ish names (i.e. not `Sum` and `Product`).

- At a first glance, I see quite a lot of `.asInstanceOf` casts and type erasure in the form of `ConfigReader[_]`. In both shapeless and magnolia derivation modules we always tried to minimize unsafe casts (shapeless is fully type-safe IIRC) and we should aim to do the same here. For example, in products, `*:` retains the type of each element and both `m.MirroredElemLabels` and `m.MirroredElemTypes` are strongly typed tuples so it should be possible to use a recursive shapeless-like approach to combine them.

heksesang:
It would be nice if you'd say what the build error is – not sure I can do much about fixing an error I cannot reproduce.

* When it comes to the formatting, v3.0.0-RC1 of scalafmt isn't able to parse all Scala 3 code, it simply crashes for some valid Scala 3 code. But now v3.0.0-RC2 is out and should have fixed some of it, so I'll see if I can make it work with that version (though I cannot guarantee anything as there are still bugs in this latest version too).

* When it comes to polluting the namespace, I can probably move the implementation to `pureconfig.generic.derivation` or something like that. Does that sound good?

* I might be able to remove the casts now, the code was originally written for 3.0.0-RC1 where it was not possible to avoid the `instanceOf` casts.

leifwickland:
@heksesang,

Can you view https://github.com/pureconfig/pureconfig/runs/2413899147?

In case you can't, the error reported is:

```
[error] 1 files must be formatted
[error] invariant failed:
[error] when verifying stats.forall(((x$11: scala.meta.Stat) => scala.meta.internal.trees.`package`.XtensionTreesStat(x$11).isTopLevelStat))
[error] found that stats.forall(((x$11: scala.meta.Stat) => scala.meta.internal.trees.`package`.XtensionTreesStat(x$11).isTopLevelStat)) is false
[error] where stats = List(import scala.compiletime.{constValue, constValueTuple, erasedValue, summonInline}, import scala.compiletime.ops.int._, import scala.deriving.Mirror, import pureconfig.error.{
[error]   CannotConvert,
[error]   ConfigReaderFailures,
[error]   ConvertFailure,
[error]   KeyNotFound,
[error]   UnknownKey,
[error]   WrongSizeList
[error] }, export Product.deriveForMirroredProduct, object Product {
[error]   inline def deriveForMirroredProduct[A](using m: Mirror.ProductOf[A]): ConfigReader[A] = {
[error]     val hint = summonInline[ProductHint[A]]
[error] 
[error]     new ConfigReader[A] {
[error]       def from(cur: ConfigCursor): ConfigReader.Result[A] =
[error]         for {
[error]           objCur <- cur.asObjectCursor
[error]           result <- {
[error]             val results =
[error]               for {
[error]                 (field, reader) <- summonFor[A]
[error]               } yield {
[error]                 val action = hint.from(objCur, field)
[error]                 if reader.isInstanceOf[ReadsMissingKeys] || !action.cursor.isUndefined then
[error]                   reader.from(action.cursor)
[error]                 else
[error]                   cur.failed(KeyNotFound.forKeys(action.field, objCur.keys))
[error]               }
[error] 
[error]             ConfigReader.Result.sequence(results)
[error]               .map(
[error]                 _.reverse
[error]                   .foldLeft[Tuple](EmptyTuple)((t, v) => v *: t)
[error]                   .asInstanceOf[m.MirroredElemTypes]
[error]               )
[error]           }
[error]         } yield m.fromProduct(result)
[error]     }
[error]   }
[error] 
[error]   inline def summonFor[A](using m: Mirror.ProductOf[A]): List[(String, ConfigReader[_])] =
[error]     labelsFor[m.MirroredElemLabels].zip(summonConfigReaders[m.MirroredElemTypes])
[error] 
[error]   inline def summonConfigReaders[T <: Tuple]: List[ConfigReader[_]] =
[error]     inline erasedValue[T] match {
[error]       case _: (h *: t) =>
[error]         inline erasedValue[h] match {
[error]           case _: Tuple => deriveForTuple[h & Tuple] :: summonConfigReaders[t]
[error]           case _ => summonInline[ConfigReader[h]] :: summonConfigReaders[t]
[error]         }
[error]           
[error]       case _: EmptyTuple => Nil
[error]     }
[error] 
[error]   inline def deriveForTuple[T <: Tuple]: ConfigReader[T] =
[error]     new ConfigReader[T] {
[error]       def from(cur: ConfigCursor): ConfigReader.Result[T] =
[error]         for {
[error]           listCur <- asList(cur)
[error]           result <- read[T, 0](listCur.list)
[error]         } yield result
[error] 
[error]       def asList(cur: ConfigCursor) =
[error]         cur.asListCursor.flatMap { listCur =>
[error]           if (constValue[Tuple.Size[T]] == listCur.size)
[error]             Right(listCur)
[error]           else
[error]             listCur.failed(
[error]               WrongSizeList(constValue[Tuple.Size[T]], listCur.size)
[error]             )
[error]         }
[error]     }
[error] 
[error]   inline def read[T <: Tuple, N <: Int](cursors: List[ConfigCursor]): Either[ConfigReaderFailures, T] =
[error]     inline erasedValue[T] match {
[error]       case _: (h *: t) =>
[error]         val h = summonInline[ConfigReader[h]].from(cursors(constValue[N]))
[error] 
[error]         h -> read[t, N + 1](cursors) match {
[error]           case (Right(h), Right(t)) => Right((h *: t).asInstanceOf[T])
[error]           case (Left(h), Left(t))   => Left(h ++ t)
[error]           case (_, Left(failures))  => Left(failures)
[error]           case (Left(failures), _)  => Left(failures)
[error]         }
[error] 
[error]       case _: EmptyTuple =>
[error]         Right(EmptyTuple.asInstanceOf[T])
[error]     }
[error] }): /home/runner/work/pureconfig/pureconfig/core/src/main/scala-3/pureconfig/generic/product.scala
[error] (tests / Test / scalafmtCheck) 1 files must be formatted
[error] (core / Compile / scalafmtCheck) invariant failed:
[error] when verifying stats.forall(((x$11: scala.meta.Stat) => scala.meta.internal.trees.`package`.XtensionTreesStat(x$11).isTopLevelStat))
[error] found that stats.forall(((x$11: scala.meta.Stat) => scala.meta.internal.trees.`package`.XtensionTreesStat(x$11).isTopLevelStat)) is false
[error] where stats = List(import scala.compiletime.{constValue, constValueTuple, erasedValue, summonInline}, import scala.compiletime.ops.int._, import scala.deriving.Mirror, import pureconfig.error.{
[error]   CannotConvert,
[error]   ConfigReaderFailures,
[error]   ConvertFailure,
[error]   KeyNotFound,
[error]   UnknownKey,
[error]   WrongSizeList
[error] }, export Product.deriveForMirroredProduct, object Product {
[error]   inline def deriveForMirroredProduct[A](using m: Mirror.ProductOf[A]): ConfigReader[A] = {
[error]     val hint = summonInline[ProductHint[A]]
[error] 
[error]     new ConfigReader[A] {
[error]       def from(cur: ConfigCursor): ConfigReader.Result[A] =
[error]         for {
[error]           objCur <- cur.asObjectCursor
[error]           result <- {
[error]             val results =
[error]               for {
[error]                 (field, reader) <- summonFor[A]
[error]               } yield {
[error]                 val action = hint.from(objCur, field)
[error]                 if reader.isInstanceOf[ReadsMissingKeys] || !action.cursor.isUndefined then
[error]                   reader.from(action.cursor)
[error]                 else
[error]                   cur.failed(KeyNotFound.forKeys(action.field, objCur.keys))
[error]               }
[error] 
[error]             ConfigReader.Result.sequence(results)
[error]               .map(
[error]                 _.reverse
[error]                   .foldLeft[Tuple](EmptyTuple)((t, v) => v *: t)
[error]                   .asInstanceOf[m.MirroredElemTypes]
[error]               )
[error]           }
[error]         } yield m.fromProduct(result)
[error]     }
[error]   }
[error] 
[error]   inline def summonFor[A](using m: Mirror.ProductOf[A]): List[(String, ConfigReader[_])] =
[error]     labelsFor[m.MirroredElemLabels].zip(summonConfigReaders[m.MirroredElemTypes])
[error] 
[error]   inline def summonConfigReaders[T <: Tuple]: List[ConfigReader[_]] =
[error]     inline erasedValue[T] match {
[error]       case _: (h *: t) =>
[error]         inline erasedValue[h] match {
[error]           case _: Tuple => deriveForTuple[h & Tuple] :: summonConfigReaders[t]
[error]           case _ => summonInline[ConfigReader[h]] :: summonConfigReaders[t]
[error]         }
[error]           
[error]       case _: EmptyTuple => Nil
[error]     }
[error] 
[error]   inline def deriveForTuple[T <: Tuple]: ConfigReader[T] =
[error]     new ConfigReader[T] {
[error]       def from(cur: ConfigCursor): ConfigReader.Result[T] =
[error]         for {
[error]           listCur <- asList(cur)
[error]           result <- read[T, 0](listCur.list)
[error]         } yield result
[error] 
[error]       def asList(cur: ConfigCursor) =
[error]         cur.asListCursor.flatMap { listCur =>
[error]           if (constValue[Tuple.Size[T]] == listCur.size)
[error]             Right(listCur)
[error]           else
[error]             listCur.failed(
[error]               WrongSizeList(constValue[Tuple.Size[T]], listCur.size)
[error]             )
[error]         }
[error]     }
[error] 
[error]   inline def read[T <: Tuple, N <: Int](cursors: List[ConfigCursor]): Either[ConfigReaderFailures, T] =
[error]     inline erasedValue[T] match {
[error]       case _: (h *: t) =>
[error]         val h = summonInline[ConfigReader[h]].from(cursors(constValue[N]))
[error] 
[error]         h -> read[t, N + 1](cursors) match {
[error]           case (Right(h), Right(t)) => Right((h *: t).asInstanceOf[T])
[error]           case (Left(h), Left(t))   => Left(h ++ t)
[error]           case (_, Left(failures))  => Left(failures)
[error]           case (Left(failures), _)  => Left(failures)
[error]         }
[error] 
[error]       case _: EmptyTuple =>
[error]         Right(EmptyTuple.asInstanceOf[T])
[error]     }
[error] }): /home/runner/work/pureconfig/pureconfig/core/src/main/scala-3/pureconfig/generic/product.scala
[error] Total time: 9 s, completed Apr 22, 2021 8:04:27 PM
```

heksesang:
Should I change the `ProductConfigReader` and `CoproductConfigReader` to the same kind of pattern that I did for `EnumConfigReader` to provide customization? Depending on a `given CoproductHint[A]` when deriving doesn't seem to provide any benefit compared to "designing" the reader by extending the trait and referring to the type from that object when deriving instead – the latter would be more explicit and avoid errors due to incorrect implicit resolution.

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
```diff
@@ -0,0 +1,47 @@
...
+      productHint
+    )
+
+  def withProductHint(productHint: ProductHint[?]) =
+    new ConfigReaderDerivation(
+      coproductHint,
+      productHint
+    )
+
+  extension (c: ConfigReader.type)
```
ruippeixotog: nit: please use bracket style here for consistency with the rest of the codebase.

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
```diff
@@ -0,0 +1,47 @@
...
+  trait DerivedConfigReader[A] extends ConfigReader[A]
+  object DerivedConfigReader {
+    inline def derived[A](using m: Mirror.Of[A]): DerivedConfigReader[A] =
+      inline m match {
+        case given Mirror.ProductOf[A] => ProductConfigReader.derived
+        case given Mirror.SumOf[A] => CoproductConfigReader.derived
+      }
+  }
+
+  def withCoproductHint(coproductHint: CoproductHint[?]) =
```
ruippeixotog: From what I could see in the tests, the API introduced here is a bit weird to use - we need to create a value, take a dependent type from it and use it in a `derives` clause. More importantly, it's a complete departure from the way other generic libraries use hints, which will be a big source of confusion for users. I think we should make it a goal that generic derivation modules are interchangeable with just an import switch.

Is there any fundamental reason that prevents the existing mechanism (providing implicits of hints) to work?

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
```diff
@@ -0,0 +1,47 @@
...
+
+object ConfigReaderDerivation {
+  object Default
+      extends ConfigReaderDerivation(
+        CoproductHint.default[Any],
+        ProductHint.default[Any]
+      )
+}
+
+val default = ConfigReaderDerivation.Default
```
ruippeixotog, line 24: I'm not fond of defining top-level on packages, particularly in this scenario - users are intended to use `import pureconfig.generic.derivation.default._`, which leads them to the wrong idea that `default` is a package and not an object.

As @jcazevedo mentioned before, please avoid using Scala 3 specific syntax and code organization - at some point in the future once 3.0 is more widespread we'll likely want to change that view, but right now I'd like to keep writing code as we would be for Scala 2 (apart from features needed to provide derivation like `inline`, `derives` and `extension`, obviously). It helps keeping the codebase consistent and reduces the cognitive load.

core/src/main/scala-3/pureconfig/generic/derivation/labels.scala
```diff
@@ -0,0 +1,18 @@
+package pureconfig.generic
+package derivation
+
+import scala.compiletime.{constValue, erasedValue, summonFrom, summonInline}
+import scala.deriving.Mirror
+
+inline def labelsFor[T <: Tuple]: List[String] =
```
ruippeixotog: Same thing here about avoiding Scala 3 specific syntax (top-level definitions).

core/src/main/scala-3/pureconfig/generic/derivation/expandType.scala
```diff
@@ -0,0 +1,6 @@
+package pureconfig.generic.derivation
+
+extension [A](a: A)
```
ruippeixotog: Is it possible to make this an utility function `inline def expandType[A, B](a: A): A & B`? Just like it was generally bad practice to apply extension methods on `Any`, this too seems a bit intrusive.

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,100 @@
...
+                .sequence(results)
+                .map(
+                  _.reverse
+                    .foldLeft[Tuple](EmptyTuple)((t, v) => v *: t)
+                )
+            }
+          } yield m.fromProduct(result)
+      }
+
+    inline def summonFor[A](using m: Mirror.ProductOf[A]): List[(String, ConfigReader[_])] =
```
ruippeixotog: It's unfortunate that we're still erasing types here. IIRC Scala's tuples are just like shapeless `HList`s, so it should be possible to avoid erasure altogether - or is some fundamental API still missing?

Given that this all runs at compile time and we're just adding early support for derivations it should be safe to start with this and improve later, though.

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,100 @@
...
+      labelsFor[m.MirroredElemLabels].zip(summonConfigReaders[m.MirroredElemTypes])
+
+    inline def summonConfigReaders[T <: Tuple]: List[ConfigReader[_]] =
+      inline erasedValue[T] match {
+        case _: (h *: t) =>
```
ruippeixotog: Unless this new type of case requires it, can we use upper case for matched types?

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,100 @@
...
+            if (constValue[Tuple.Size[T]] == listCur.size)
+              Right(listCur)
+            else
+              listCur.failed(
+                WrongSizeList(constValue[Tuple.Size[T]], listCur.size)
+              )
+          }
+      }
+
+    inline def read[T <: Tuple, N <: Int](cursors: List[ConfigCursor]): Either[ConfigReaderFailures, T] =
```
ruippeixotog: Since this is a method specific to tuple reading, can it be an internal method of the anonymous `ConfigReader` in `deriveForTuple`? If that's not possible, can you rename it to `readTuple` or something more concrete?

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,100 @@
...
+      inline erasedValue[T] match {
+        case _: (h *: t) =>
+          val reader =
```
ruippeixotog: This is the same as `summonConfigReader`.

project/Dependencies.scala
```diff
@@ -5,12 +5,12 @@ object Dependencies {
   object Version {
     val scala212 = "2.12.12"
     val scala213 = "2.13.5"
-    val scala30 = "3.0.0-RC2"
+    val scala30 = "3.0.0-RC3"
```
ruippeixotog: Please rebase/merge on top of the current master to start using Scala 3.0.0.

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigReaderDerivation.scala
```diff
@@ -0,0 +1,61 @@
...
+package derivation
+
+import scala.compiletime.{constValue, erasedValue, error, summonInline}
+import scala.deriving.Mirror
+
+import pureconfig.error.{CannotConvert, ConfigReaderFailures}
+import pureconfig.generic.derivation.transformedLabelsFor
+import pureconfig.generic.error.NoValidCoproductOptionFound
+
+type EnumConfigReader[A] = EnumConfigReaderDerivation.Default.EnumConfigReader[A]
```
ruippeixotog, line 11: This looks good to me, but out of curiosity: can we get away with using `ConfigReader` for everything? `enum` are first class now, but do we have a way to know if a type is an enum on `inline` methods, or are they just syntatic sugar for more traditional sealed families?

core/src/main/scala-3/pureconfig/generic/derivation/CoproductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,81 @@
...
+import pureconfig.generic.derivation.ConfigReaderDerivation
+import pureconfig.generic.error.InvalidCoproductOption
+
+trait CoproductConfigReaderDerivation(hint: CoproductHint[?]) { self: ConfigReaderDerivation =>
+  trait CoproductConfigReader[A] extends DerivedConfigReader[A]
+  object CoproductConfigReader {
+    inline def derived[A](using m: Mirror.SumOf[A]): CoproductConfigReader[A] =
+      new CoproductConfigReader[A] {
+        def from(cur: ConfigCursor): ConfigReader.Result[A] =
+          for {
```
ruippeixotog: This `for` seems unnecessary, it looks like you can simply return the contents of `result`.

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import scala.deriving.Mirror
+
+import pureconfig.generic.{CoproductHint, ProductHint}
+
+class ConfigReaderDerivation(coproductHint: CoproductHint[?], productHint: ProductHint[?])
+    extends CoproductConfigReaderDerivation(coproductHint)
+    with ProductConfigReaderDerivation(productHint) {
+
+  trait DerivedConfigReader[A] extends ConfigReader[A]
```
ruippeixotog: Why do we need this hierarchy of `ConfigReader`? Can't we just use `ConfigReader` everywhere?

ruippeixotog (commented):
@heksesang I'm sorry for taking so long to review this. The Scala concepts used here are all new to me so I had to learn how everything worked. I'm still not 100% confident I understand everything added here, so I'm sorry if my questions don't make sense 🙂  This looks good overall and even though it's missing features from other generic libraries it's good enough to provide users as a "beta" implementation. 

One thing I want to make sure of is that introducing this won't negatively affect existing users (using PureConfig with shapeless or magnolia in Scala 2.x or using just pureconfig-core in 3.x). From the test suites, it looks like users still need to add an explicit import to use this new derivation mechanism - is this correct, @heksesang?

@jcazevedo, IIRC one of the reasons why we decided to leave hints out of `core` was that different derivation libraries could support different configuration options. However, after our latest hint refactoring they became fairly generic so I don't mind bringing it to `core`. What do you think?

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
heksesang: At least with 3.0.0-RC2 of scalafmt there is an option that rewrites the syntax from or to braceless – could enable that to ensure that braces are consistently used until we're ready to switch to braceless. As the compiler allows either style, it's easy to miss braces a couple of places when you're used to writing braceless syntax. :p

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
heksesang: I don't think there was anything that prevented the old mechanism from working. I can change it back to the way it was if that is preferable (given there weren't any issues that I have forgotten about). It's also a possibility to keep both patterns (won't confuse exisiting users and it still allows for those that prefer this pattern).

As I mentioned, I think it's a cleaner pattern for configuration this way and would prefer to be able to do it this way at least, even if the exisiting mechanism exists as well.

The way I see it, the fact that the hint is used at the derivation site and not when when you're calling `from` means that there is no benefit of using an implicit hint compared to just declaring a configured reader derivation the way done with this pattern (this goes for the other generic libraries as well). Configuring a reader derivation with this pattern is more explicit and doesn't provide any negative consequences that I am aware of.

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
heksesang: I believe it had something to do with how the derivation mechanism requires the `derived` method to be in the companion object of the type it derives.

So this code will not work:
```scala
object DerivedConfigReader:
  inline def derived[A](using m: Mirror.Of[A]): ConfigReader[A] = ???
```

But possible I can avoid this by doing extension methods on the `ConfigReader` object instead, now that I think about it.

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
heksesang: So it's preferable that I remove the definition and that the user uses `import pureconfig.generic.derivation.ConfigReaderDerivation.Default._` instead?

But I'd just like to point out that if the user is already writing Scala 3 code, they'd know that top level definitions might not be a package. Also, Scala 2 already had package objects that did the same thing (I would have put this definition in a package object in Scala 2). Also not sure why a user would care if it's a package or an object, as long as they get their definitions. So from a user perspective this shouldn't be a problem. Just curious to why you think this might cause any issues, if I am missing something. :)

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigReaderDerivation.scala
heksesang: The `derived` method needs to be defined in the companion object of the type it derives. Which means if I were to change the return type of the `derived` method here to `ConfigReader[A]`, I would also need to extend `object ConfigReader` with this method – but we already have a different method which does the "normal" derivation for sum types. The user would then have to ensure everything is scoped so they can import just the one `derived` method they need and avoid ambiguous definitions. The other alternative is to embed this derivation into the other extension method.

But because `enum` is pretty much syntactic sugar for sealed families (not much reason to use sealed families instead of enums in Scala 3 unless you have some very special case where enum doesn't work, so enum and sealed family should be treated as the same thing), you'd have to first try do this derivation to see if all types of the family are objects (no case classes in the enum) and then try to do the other derivations if that fails (i.e. within the `derived` method the "normal" derivation would act as a fallback if the enum derivation fails). It should be possible to do that, but then it would also not be possible to choose the "normal" derivation for sums if the sum consists of only objects and no case classes (but this might not be a big loss).

Which approach is preferable here? Use `ConfigReader` everywhere and make the choice of derivation for enums "automatic", or keep the choice and also keep the `EnumConfigReader` type?

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
heksesang: You have to use lower case for matched types like this.

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
heksesang: It's not possible to define an inline method inside another inline method, but I can rename it to `readTuple` for example.

core/src/main/scala-3/pureconfig/generic/derivation/expandType.scala
heksesang: I can do that. It's also possible to "hide" it behind an import, keeping it as an extension method – is this still considered bad practice from your point of view?

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
heksesang: I'll see if I can use the `summonConfigReader` method (some things are like they are because the derivation mechanisms have been a bit fickle at times).

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
heksesang: It's maybe possible that the reading can be done recursively and such avoid erasure, but I am not sure there is much benefit. I can try it out and see if it's worth it (I am reluctant to do it if it has too negative effects on code readability, given how small the issue is by doing erasure there).

heksesang:
It doesn't support everything that is supported by the other generic libraries, but I think it might be better that this has fewer features to keep the code as simple as possible (some of the features from the other libraries will need the use of reflection APIs, not just mirrors), you can always use the magnolia- or shapeless-based libraries if it's not a problem to introduce that dependency and you need more features. The default implementation would act more as a basic support for derivation if you don't want to depend on magnolia or shapeless for some reason. I assume it's not planned to phase out the shapeless- and magnolia-based libraries?

And yes, it should be necessary to add an additional import to enable the derivation support now (you have to import the extension method), but that might be for the best if the other implementations are going to support the derived clause too.

heksesang:
Now that the derivation syntax is enabled by a separate import of extension methods, it should also be possible to keep the hints in the generic-base library and move all this derivation code into that module as well. That would maybe be cleaner?

jcazevedo:
> @jcazevedo, IIRC one of the reasons why we decided to leave hints out of `core` was that different derivation libraries could support different configuration options. However, after our latest hint refactoring they became fairly generic so I don't mind bringing it to `core`. What do you think?

I suppose we can do it, but I'd prefer to keep them in the generic family of modules unless absolutely required in the core.

I think we have two things at hand here: one is supporting generic derivation of `ConfigReader` as pureconfig-generic currently does, and another is supporting the `derives` clause from Scala 3. I think the proposed solution in this PR of requiring an import to use the `derives` clause and to customize the derived `ConfigReader` based on a type is excessively verbose and doesn't scale well with additional derivation options.

I think it'd be preferable to only support `derives ConfigReader` in pureconfig-core, without any possible configuration. The produced `ConfigReader` would be the same as if we'd use pureconfig-generic without custom `ProductHint`s or `CoproductHint`s. Users that wanted access to the customization options would use pureconfig-generic with either automatic derivation or the `deriveReader` API. That way we can keep pureconfig-core small, add support for the new `derives` clause (for the default use case) and keep the derivation configurations restricted to the generic family of modules. @heksesang and @ruippeixotog: What do you think?

heksesang:
I like the idea of having a very barebone implementation for derivation, means the code for it can be less complicated. We could provide customization as a separate option in another module.

I still think we should require an import to use the syntax, because otherwise I am not sure we can override the syntax from the other generic libraries (should be possible to do `derives ConfigReader` with another implementation when they become available for Scala 3 too).

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigReaderDerivation.scala
ruippeixotog: If `enum` is just syntactic sugar for sealed families, then I totally agree with not making things automatic - it would be a departure from existing behavior and it could lead to inconsistencies where adding a new option would change completely the expected format.

core/src/main/scala-3/pureconfig/generic/derivation/expandType.scala
ruippeixotog: Hiding it beyond a private/hidden import (for some definition of private/hidden) sounds good. My main concern is users importing `pureconfig.generic.derivation._` (which seems plausible), magically getting in scope new functions defined on every object and starting to rely on them.

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
ruippeixotog: Yeah, I'm not sure of the benefits either. Let's leave it like this for now :+1:

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
ruippeixotog: Fair point. The organization still feels a bit awkward to me, mostly because the intended usage is to import a value while I expect (without a strong reason) package structures to be static. But you're right that it's unrelated to Scala 2 vs Scala 3 and it's a personal choice, so unless other maintainers share this preference please let this as it is.

ruippeixotog:
> The default implementation would act more as a basic support for derivation if you don't want to depend on magnolia or shapeless for some reason. I assume it's not planned to phase out the shapeless- and magnolia-based libraries?

Well, I don't really know. For now shapeless and magnolia implementations are certainly more mature, but PureConfig does not require any of the heavy machinery of shapeless and magnolia seems to be roughly at the same level as Scala 3 in terms of power, so as the Scala 3 derivation implementation gets more mature I wouldn't rule out making it the main one (of course, as long as shapeless and magnolia are around there won't be any problem in maintaining the existing derivation modules we have).

> I think it'd be preferable to only support derives ConfigReader in pureconfig-core, without any possible configuration. The produced ConfigReader would be the same as if we'd use pureconfig-generic without custom ProductHints or CoproductHints. Users that wanted access to the customization options would use pureconfig-generic with either automatic derivation or the deriveReader API.

I assume this will mean having two distinct Scala 3 derivation implementations - a simple one in core and an advanced one in some new generic module - is that what you're suggesting?

If that's what you mean, I like the idea but only as long as the `pureconfig-core` one is _very_ simple and code duplication is minimal. If we see ourselves having too much repeated boilerplate in the two versions I'd rather have the full version in `pureconfig-core` or in another module.

> I still think we should require an import to use the syntax, because otherwise I am not sure we can override the syntax from the other generic libraries (should be possible to do derives ConfigReader with another implementation when they become available for Scala 3 too).

I agree. I think that mid/long-term we may have a Scala 3 implementation mature enough to make it the "canonical" `derives` provider and force other implementations to use some `deriveReader` API. However, I think we should be cautious and consider this implementation to be in a "beta" stage - we place it at the same level as other implementations and let users choose their preferred one (hopefully just by switching import statements in Scala 3). We can always make the import unnecessary in the future as we gather more feedback :)

heksesang:
I have simplified it a bit now. There is now a default implementation that makes a result similar to the defaults for the other generic modules, and the user has to import the `derived` method to use the syntax.

With this change, all the generic stuff is left in `generic-base` like it was. Should the new files be moved from `pureconfig.generic.derivation` to `pureconfig.derivation`, considering it's not actually depending on anything from the generic modules now?

And as it is now, it means that you won't get the same errors from this implementation as you'd get from the other ones because the errors are defined in `generic-base` and not in `core`.

It should also be possible to add an extension method in a generic module that can use given hints for configuring the derivation (by passing arguments from the hints to the constructor for the traits), but it would still be limited to the little configuration that can be done in the core module (which is to set field mapping and the discrimination field for coproducts).

What do you think, @jcazevedo and @ruippeixotog?

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
```diff
@@ -0,0 +1,24 @@
...
+        case given Mirror.ProductOf[A] => derivedProduct
+        case given Mirror.SumOf[A] => derivedSum
+      }
+  }
+}
+
+object ConfigReaderDerivation {
+  object Default
+      extends ConfigReaderDerivation
+      with CoproductConfigReaderDerivation(ConfigFieldMapping(PascalCase, KebabCase), "type")
```
ruippeixotog, line 20: I'm surprised that `ConfigFieldMapping` is defined in core and not in `generic-base`. We don't seem to have any usages in core, so I suppose it was left here accidentally when we moved shapeless derivation code to its separate module. I think it's ok to leave it like this though.

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,83 @@
...
+                      labels.map(objCur.atKeyOrUndefined(_)),
+                      labels.map(KeyNotFound.forKeys(_, objCur.keys))
+                    )
+                  )
+            } yield m.fromProduct(result)
+        }
+    }
+
+  inline def readTuple[T <: Tuple, N <: Int](
+      cursors: List[ConfigCursor],
```
ruippeixotog, line 54: nit: this is a weird-looking implementation, where we have (linked) lists as arguments but we keep indexing them to get each value. Consider destructuring both lists on each recursive call and get rid of `N`.

core/src/main/scala-3/pureconfig/generic/derivation/labels.scala
```diff
@@ -0,0 +1,20 @@
+package pureconfig.generic
+package derivation
+
+import scala.compiletime.{constValue, erasedValue, summonFrom, summonInline}
+import scala.deriving.Mirror
+
+object Labels {
+  inline def of[T <: Tuple]: List[String] =
```
ruippeixotog: Is this being used anywhere?

core/src/main/scala-3/pureconfig/generic/derivation/labels.scala
```diff
@@ -0,0 +1,20 @@
+package pureconfig.generic
+package derivation
+
+import scala.compiletime.{constValue, erasedValue, summonFrom, summonInline}
+import scala.deriving.Mirror
+
+object Labels {
```
ruippeixotog: nit: the name of this file should probably be `Labels.scala`.

tests/src/test/scala-3/pureconfig/ProductReaderDerivationSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+import scala.language.higherKinds
+
+import com.typesafe.config.{ConfigFactory, ConfigRenderOptions, ConfigValueFactory}
+import org.scalacheck.Arbitrary
+import pureconfig._
+import pureconfig.ConfigConvert.catchReadError
+import pureconfig.error.{KeyNotFound, WrongType}
+import pureconfig.generic.derivation.default.derived
+
+class ProductReaderDerivationSuite extends BaseSuite {
```
ruippeixotog, line 15: I think there are two tests missing here:

- A simple success case, where we have a case class with 2-3 primitive or collection parameters and we check that it is read successfully;
- A test verifying that we can read config lists from tuples.

tests/src/test/scala-3/pureconfig/CoproductReaderDerivationSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+import pureconfig.generic._
+import pureconfig.generic.derivation.default.derived
+
+enum AnimalConfig derives ConfigReader {
+  case DogConfig(age: Int)
+  case CatConfig(age: Int)
+  case BirdConfig(canFly: Boolean)
+}
+
+class CoproductReaderDerivationSuite extends BaseSuite {
```
ruippeixotog, line 17: We should also add one test for when the hint field exists but it doesn't correspond to an existing option.

core/src/main/scala-3/pureconfig/generic/derivation/CoproductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,56 @@
...
+          optCur <- objCur.atKey(optionField)
+          option <- optCur.asString
+          result <-
+            readers.get(option) match {
+              case Some(reader) => reader.from(cur)
+              case None =>
+                Left(
+                  ConfigReaderFailures(
+                    optCur.failureFor(
+                      CannotConvert(option, constValue[m.MirroredLabel], "The value is not a valid option.")
```
ruippeixotog, line 28: I understand that we don't want to use `NoValidCoproductOptionFound` as it's defined in `generic-base`, but it's a bit disappointing that we are losing the rich message provided by it. Should we consider copying it to `derivation` or restructure that one in particular? What do you think, @jcazevedo?

ruippeixotog (commented):
Thanks once again for submitting this, @heksesang. The code is much simpler now and I'm happy to go ahead with it. I just left a few nits and a request to add a few more unit tests.

core/src/main/scala-3/pureconfig/generic/derivation/labels.scala
heksesang: Might not be in use any more, I'll remove it and see if that's the case.

core/src/main/scala-3/pureconfig/generic/derivation/CoproductConfigReaderDerivation.scala
heksesang: Some generic derivation is now part of core if we merge this, so it might make sense to have some basic error types for derivation and such also defined in core.

heksesang: An alternative is also to move all the new code into `generic-base`.

core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala
heksesang: It's useful for the derivation code that it exists. So I think it's better if we leave it in `core`, unless we move all the derivation code into `generic-base`.

core/src/main/scala-3/pureconfig/generic/derivation/labels.scala
heksesang: I'll fix that. It's probably been left with that name because changing just case is a hassle in git, it doesn't register the change on my computer unless I commit the file with another name and then commit it with the correct name and case in the next commit.

core/src/main/scala-3/pureconfig/generic/derivation/ProductConfigReaderDerivation.scala
heksesang: Any suggestions on how to do that? Because I am not sure I know how.

heksesang: Alright, so I got it working as long as I don't have any compile time verification of the cursor list size, and return an error of some kind if the list of cursors is too short (is there a correct error there already to use for that?).

ruippeixotog: I realized only now that you're using this function both for both tuple-like and object-like configs, so you're right that we may not be able to get rid of this. I generally find this function heard to read because it assumes too many things about how it's called - reading its implementation I expected `cursors`, `keyNotFound` and the tuple type `T` to always have the same size, but in the tuple-like variant you call this with an empty `keyNotFound` list because you _know_ that cursors won't ever be undefined.

The function works though, and you already spent too much of your time on code style comments and nits from me. Let's keep this first version and think of improvements later 🙂

heksesang: Not able to get it working. I can make the list into an array, if that makes the N parameter anymore bearable for you?

tests/src/test/scala-3/pureconfig/ProductReaderDerivationSuite.scala
heksesang: The last test should verify that `{ values: [ true, 5] }` can be read as `case class Foo(values: (Boolean, Int))`, if I understand you correctly?

heksesang:
I have now added more unit tests.

For products:
* one that checks that it is possible to read primitives and lists of primitives

For coproducts:
* one that verifies that you get the expected error when the hint field is missing

For tuples:
* one that verifies that lists can be interpreted as tuples if the length and types match
* one that verifies that you get the expected error if the list is too short when interpreting it as a tuple
* one that verifies that you get the expected error if the list is too long when interpreting it as a tuple
* one that verifies that you get the expected error if the types in the list do not match the tuple types

heksesang:
What is left now is the question of errors. There are two alternatives that I can see:
* move code to `generic-base` and use the already defined error types there
* keep the code in `core`, and define some new error types for the basic derivation there

tests/src/test/scala-3/pureconfig/ProductReaderDerivationSuite.scala
ruippeixotog: Yes, that's correct!

ruippeixotog (approved):
LGTM. Thank you so much for bearing with my comments and sorry again for the long wait. I'm going to let @jcazevedo take a look at this before merging as he also showed interest in reviewing this.

core/src/main/scala-3/pureconfig/generic/derivation/CoproductConfigReaderDerivation.scala
ruippeixotog: I think `generic-base` is not an adequate place for actual derivation code. That module was created precisely to share common definitions across all derivation libraries. If we were to take this to another module I'd rather have a new `pureconfig-generic-scala3` that depended on `generic-base`, making it on par with shapeless and magnolia. However, given that this uses only Scala 3 native concepts I'm leaning towards keeping this in code in `core` and replicating generic classes to `core` in order to keep Scala 3 derivation self-contained.

jcazevedo: @ruippeixotog: That looks like a sensible way to go about it.

jcazevedo:
@heksesang: Thanks for this PR! This is a great starting point to start supporting generic derivation in Scala 3.

> I still think we should require an import to use the syntax, because otherwise I am not sure we can override the syntax from the other generic libraries (should be possible to do `derives ConfigReader` with another implementation when they become available for Scala 3 too).

Sorry if I'm missing something, but I don't think that offering customizable derivation using the `derives` clause is very convenient. By requiring an import to both configure and enable the use of the `derives` clause, would it be possible for different types within the same file to use different derivation options?

heksesang:
> @heksesang: Thanks for this PR! This is a great starting point to start supporting generic derivation in Scala 3.
> 
> > I still think we should require an import to use the syntax, because otherwise I am not sure we can override the syntax from the other generic libraries (should be possible to do `derives ConfigReader` with another implementation when they become available for Scala 3 too).
> 
> Sorry if I'm missing something, but I don't think that offering customizable derivation using the `derives` clause is very convenient. By requiring an import to both configure and enable the use of the `derives` clause, would it be possible for different types within the same file to use different derivation options?

Currently there is no customization for the core derivation. That's what we agreed on, I thought. So I am not sure what you mean here.

jcazevedo:
Yes, we are currently not offering customization for the core derivation. However, we're still only enabling the use of the `derives` clause behind an explicit import, under the premise that we will use different imports to provide derivation options in the future, no?

heksesang:
The idea was if we want to be able to use shapeless or magnolia with the same syntax. In the future, `case class Foo(i: Int) derives ConfigReader` could use magnolia or shapeless once the modules are ported to Scala 3, by doing `import pureconfig.generic.shapeless.derived` or `import pureconfig.generic.magnolia.derived` for example.

Simply put, I find a bit weird to lock the derives syntax to one particular implementation when we provide multiple implementations.

jcazevedo:
> The idea was if we want to be able to use shapeless or magnolia with the same syntax. In the future, `case class Foo(i: Int) derives ConfigReader` could use magnolia or shapeless once the modules are ported to Scala 3, by doing `import pureconfig.generic.shapeless.derived` or `import pureconfig.generic.magnolia.derived` for example.
> 
> Simply put, I find a bit weird to lock the derives syntax to one particular implementation when we provide multiple implementations.

I understand that. It's just not very clear to me what would be the recommended way to customize the derivation in the scenarios that would support it (what we currently do by providing custom implementations of `ProductHint` and `CoproductHint`). Would we fall back to the `derivedReader` approach, or would we provide different `derived` implementations?

heksesang:
The core one does not support customization, so if you want to customize you will have to switch to one of the other two. And they use the `ProductHint` and `CoproductHint` mechanism.

But I am not really sure what you are asking about here. I am not sure why customization is even a topic when we decided to not add it.

jcazevedo:
> But I am not really sure what you are asking about here. I am not sure why customization is even a topic when we decided to not add it.

Sorry. I brought up customization because of the requirement of an explicit import for the `derives` clause to be available when using pureconfig-core. The reasoning for having that explicit import was so that we could have different derivation implementations for the `derives` clause, particularly those that will support customization, hence my question. Since this PR hints toward allowing different `derived` implementations, I was wondering if that would be the approach we'd recommend for customizing derivation, which didn't seem very convenient to me.

heksesang:
Well, I assume for now the recommended approach would be to make the hints and the derived extension for shapeless/magnolia would respect those hints. Not a big fant of implicit-based configuration, but if we don't want to change it, there is no need to.

But that's a discussion that could be taken separately from this PR in any case – the outcome wouldn't change this PR.

jcazevedo (approved):
Sure. @heksesang: Thanks again for this! I just synced this with master and will merge after the CI build.

ruippeixotog:
@jcazevedo the way I thought about multiple `derives` was that once shapeless and magnolia were available on Scala 3 users should be able to choose which implementation they want and switch between them as easily as they do now - by switching imports. For example, folks may prefer shapeless if it's less buggy, magnolia if it's faster and the core one because it's dependency-free.

As I mentioned above, I can see in the future the Scala 3 derivation becoming a _de facto_ standard if it's stable and fast enough. If that happens, we can just remove the import requirement and force shapeless and magnolia libraries to define readers explicitly. This would be fairly disruptive, so it should happen only if Scala 3 derivation is really a clear winner.

This is all independent of which customization mechanism is used - for all I know we can keep implicit-based configurations on shapeless and magnolia for backwards compatibility, add `deriveReader` implementations for Scala 3 derivation and think of nicer ways to do it for users on Scala 3 using `derives`. But like @heksesang I also think of this as a separate concern.

YourPsychiatrist:
Checking the dates, I am thinking this change should already be included in 0.16.0 (and potentially earlier ones). Isn't it? Because I am having some troubles finding the derivation package (`pureconfig.generic.derivation`) and can't get it to run with Scala 3. Any ideas why that could be?

PR #1041, opened 2021-05-13
Backport module for cats effect 2 series

Ssstlis:
Due to the fact that the migration of the ecosystem to CE3 will take a long time, it is necessary to have integration with CE2 as a separate module, so that the upgrade of pureconfig does not force the entire application to migrate to CE3.

For example, I migrated distage-testkit to version `1.0.6` and this lib depends on pureconfig `0.15.0`, so when we bump pureconfig we have an unnecessary cats-effect 3 module, which leads to the stub on specific versions.


catostrophe:
@ruippeixotog can this be released? We'd also like to have this module for `derevo` and other projects dependent on `pureconfig`.

Ssstlis:
@ruippeixotog bump this PR

modules/cats-effect2/docs/README.md
```diff
@@ -0,0 +1,77 @@
+# Cats-effect2 module for PureConfig
+
+Adds support for loading configuration using [cats-effect 2.* series](https://github.com/typelevel/cats-effect) to control side effects.
+This module written for backport pureconfig to old `cats-effect 2.* series`.
```
ruippeixotog: nit

```suggestion
This is a backport of `pureconfig-cats-effect` to the old 2.* series.
```

modules/cats-effect2/src/main/scala/pureconfig/module/catseffect/package.scala
```diff
@@ -0,0 +1,115 @@
...
+
+import cats.data.EitherT
+import cats.effect.{Blocker, ContextShift, Resource, Sync}
+import cats.implicits._
+import com.typesafe.config.ConfigRenderOptions
+
+import pureconfig._
+import pureconfig.error.ConfigReaderException
+
+package object catseffect {
```
ruippeixotog: In order to be extra clear that this is a backport, I'd rather use a separate `catseffect2` package for this. Having artifacts with different (org, name) pairs containing the same package is confusing and may lead to runtime problems since one does not necessarily evict the other. 

(using both `catseffect` and `catseffect2` modules in the same project is most likely an error anyway, but at least it will be clear if that happens accidentally)

ruippeixotog (changes requested):
Hi @Ssstlis, thanks for the PR and sorry for the wait. We don't usually accept backports and multi-version releases for PureConfig modules, but after some discussion we decided to accept this one given cats-effect's importance in the Scala ecossystem and the fact that the two release series are still actively being maintained.

I left a few comments below, other than that LGTM.

Ssstlis:
@ruippeixotog I probably fixed all of your requests, can you check it out? Thanks.

modules/cats-effect2/build.sbt
```diff
@@ -0,0 +1,21 @@
...
+
+libraryDependencies ++= Seq(
+  "org.typelevel" %% "cats-effect" % "2.5.1"
+)
+
+developers := List(Developer("keirlawson", "Keir Lawson", "keirlawson@gmail.com", url("https://github.com/keirlawson")))
+
+osgiSettings
+
+OsgiKeys.exportPackage := Seq("pureconfig.module.catseffect.*")
```
ruippeixotog: This also needs to be updated.

```suggestion
OsgiKeys.exportPackage := Seq("pureconfig.module.catseffect2.*")
```

modules/cats-effect2/src/main/scala/pureconfig/module/catseffect/syntax/package.scala
```diff
@@ -0,0 +1,19 @@
+package pureconfig.module.catseffect
```
ruippeixotog: Changing the package object is not enough, you also need to update all packages here and rename the folder to `catseffect2`.

modules/cats-effect2/src/test/scala/pureconfig/module/catseffect/CatsEffectSuite.scala
```diff
@@ -0,0 +1,106 @@
+package pureconfig.module.catseffect
```
ruippeixotog: Don't forget the test files, too.

modules/cats-effect2/src/main/scala/pureconfig/module/catseffect/syntax/package.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import scala.language.higherKinds
+import scala.reflect.ClassTag
+
+import cats.effect.{Blocker, ContextShift, Sync}
+
+import pureconfig.module.catseffect
```
ruippeixotog: This import seems unnecessary.

ruippeixotog (changes requested):
Thanks for the update, but there are still some package references missing.

Ssstlis:
@ruippeixotog sorry for the problems, this is all due the IDEA bugs =( pipeline in my forked repository now running well

ruippeixotog (approved):
Everything good now, thanks!

ruippeixotog (approved):

PR #1046, opened 2021-05-14
Update PureConfig to Scala 3.0

ruippeixotog:
Scala 3.0 final is out!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
           - scala: '2.13'
             scala-version: 2.13.5
           - scala: '3.0'
-            scala-version: 3.0.0-RC2
+            scala-version: 3.0.0
 
     steps:
       - name: Checkout repository
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = 3.0.0-RC1
+version = 3.0.0
 
 align.preset = none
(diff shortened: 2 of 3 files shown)


timothyklim:
`org.scalatestplus %% scalacheck-1-15` is available now

jcazevedo (approved):
Thanks for this! 🎉

sergiuszkierat:
@jcazevedo, [scala3-migrate](https://github.com/scalacenter/scala3-migrate) raises an error, when can we expect artifacts for `3.0.0`?
```
sbt> migrate-libs root
[info] 
[info] 
[info] Starting to migrate libDependencies for root
[info] 
[info] X             : Cannot be updated to scala 3
[info] Valid         : Already a valid version for Scala 3
[info] To be updated : Need to be updated to the following version
[info] 
[info] com.github.pureconfig:pureconfig:0.15.0                  -> X : Contains Macros and is not yet published for 3.0.0
```

jcazevedo:
@sergiuszkierat: We should be making a release for Scala 3.0.0 in the near future. Note that it might still take a while before we support generic derivation of `ConfigReader`s and `ConfigWriter`s for Scala 3. For now, only pureconfig-core should be available.

sergiuszkierat:
Thanks for updating me. As far as I know, we should follow https://github.com/pureconfig/pureconfig/issues/970 :eye:

PR #1062, opened 2021-05-27
upgrade scalacheck-shapeless

SethTisue:


SethTisue:
this repo is already on ScalaCheck 1.15, so I think it was just an oversight that `_1.14` was still being used for scalacheck-shapeless

jcazevedo (approved):
Yeah. Thanks for this!

PR #1072, opened 2021-06-09
Remove code deprecated on or before v0.11.x

ruippeixotog:
The cutoff version was somewhat arbitrary. There is still code deprecated in v0.12 that we can remove next version.


jcazevedo (approved):
Thanks!

PR #1075, opened 2021-06-14
Update sbt-scoverage to official version

ruippeixotog:
https://github.com/scoverage/sbt-coveralls/pull/128 was finally merged and a new version was released yesterday, so we can go back to using the artifact published to `org.scoverage`.


jcazevedo (approved):
Thanks!

PR #1077, opened 2021-06-15
Use while loop for "CapitalizedWordsNamingConvention.toTokens" instead of regex to improve performance

pwliwanow:
Thanks for your work on PureConfig, it's a great lib!

We use PureConfig to read and write quite big configs (~7MB), so I checked the performance the other day. 

For testing purposes, I used an app that in a loop performs:
```scala
val rendered = implicitly[ConfigWriter[AppConfig]].to(appConfig).render(ConfigRenderOptions.concise())
implicitly[ConfigReader[AppConfig]].from(ConfigFactory.parseString(rendered).root()) match {
  case Left(error) => throw new IllegalStateException(s"Invalid config: $error")
  case Right(v) => v
}
```

It seems that the majority of time is spent in regex that is used by `CapitalizedWordsNamingConvention.toTokens`:

pureconfig-master

and here's a fragment of a Flame Graph:

pureconfig-master-flamegraph

This PR rewrites `CapitalizedWordsNamingConvention.toTokens` to use while loop instead of regex. After the change, "own time" of `CapitalizedWordsNamingConvention.toTokens` is over 8x lower:

image

and similarly, here's fragment of a Flame Graph:

pureconfig-fix-flamegraph

When it comes to reading our real-life config from JSON and writing that config to JSON, here are the results before and after the change:
| Type | Avg Time |
|------|------|
| Writing config before | 458ms |
| Reading config before | 905ms |
| Writing config after | 177ms |
| Reading config after | 522ms |

For our use case, the change means that reading configs from JSON to a case class representation is ~40% faster and writing configs from case class to JSON is ~60% faster.

### DIFF
--- a/core/src/main/scala/pureconfig/NamingConvention.scala
+++ b/core/src/main/scala/pureconfig/NamingConvention.scala
@@ -7,13 +7,62 @@ trait NamingConvention {
 
 trait CapitalizedWordsNamingConvention extends NamingConvention {
   def toTokens(s: String): Seq[String] = {
-    CapitalizedWordsNamingConvention.wordBreakPattern.split(s).map(_.toLowerCase)
+    if (s.isEmpty) List.empty
+    else {
+      val buffer = List.newBuilder[String]
+      val currentTokenBuilder = new StringBuilder
+      val noneCtx = 0
+      val lowerCaseCtx = 1
+      val upperCaseCtx = 2
+      val notLetterCtx = 3
+      var ctx = noneCtx
+      val _ = {
+        val c = s.charAt(0)
+        if ('a' <= c && c <= 'z') {
+          currentTokenBuilder.append(c)
+          ctx = lowerCaseCtx
+        } else if ('A' <= c && c <= 'Z') {
+          currentTokenBuilder.append(c.toLower)
+          ctx = upperCaseCtx
+        } else {
+          currentTokenBuilder.append(c)
+          ctx = notLetterCtx
+        }
+      }
+      var i = 1
+      while (i < s.length) {
+        val c = s.charAt(i)
+        if ('a' <= c && c <= 'z') {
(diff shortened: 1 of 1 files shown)


leifwickland:
In principle, I like this change. Thanks for the contribution, @pwliwanow.
Before approving it, I'd like to see:
- Attempt to convert the looping to a tailrec function to eliminate the explicit `var`s.
- A new (possibly transient) unit test that generates thousands of random strings and runs them through both implementations, verifying both produce the same chunks. Perhaps it'd be sensible to commit the input and expected output to a file and have that become an ongoing test of this new algorithm.

jcazevedo:
I'm a bit wary of moving this to a noticeable more complicated implementation. Performance isn't something we're super focused on in PureConfig, and I don't think these particular performance gains justify the change. The most typical use case of PureConfig is loading configurations at the start of an application, and at that point, the extra 400ms don't seem very troublesome to me.

For situations where performance at this point is relevant, we provide ways to customize the `ConfigFieldMapping` used in derived readers and writers (https://pureconfig.github.io/docs/overriding-behavior-for-case-classes.html#field-mappings) so you can have a more performant custom implementation for your application.

ruippeixotog:
+1 for what @jcazevedo said. In many cases (and even more in functional programming) performance is usually traded off for code conciseness, depending on how low-level the library is supposed to sit. In PureConfig performance is definitely not a core requirement and switching a 2 LOC implementation with a 50+ LOC implementation of a FSM is quite an extreme case of losing conciseness and readability.

As @jcazevedo mentioned, this is a perfect use case for you to have your own implementation of `ConfigFieldMapping` and use it by exporting product and coproduct hints. However, I believe that between a regex and a full FSM there are a lot of intermediate solutions that produce faster results without sacrificing on readability (e.g. by folding the string?).

pwliwanow:
Understandable, we'll customize `ConfigFieldMapping` then

Issue #1082, opened 2021-06-20
SBT project structure and Scala 3


ruippeixotog:
In addition to the issues identified in #595, Our build needs a significant refactoring before we can further adopt Scala 3:

- We currently have a hard dependency on project `generic` on all modules, both for testing and for documentation. Shapeless 3 is already available, but it's a completely different implementation and having both in the same `generic` project under different source folders would be considerably difficult to maintain (besides being confusing for users). On some projects we additionally use scalacheck-shapeless for testing, which obviously isn't available on Scala 3.0 either.

- The mechanism for cross-compiling against different Scala versions has its limitations. We need to have a "main" `scalaVersion` to be used when we do e.g. `compile` and and additional `crossScalaVersions` when we do `+compile`. Right now, SBT doesn't allow us to have a Scala 3 only module (with `scalaVersion := "3.0.0"`) because the main version in `core` is currently set to Scala 2.12. `compile` fails in that case, even though `+compile` or `++3.0.0 compile` still works. Moreover, we cannot set projects with conditional subproject dependencies - we can't say "bundle depends on generic on Scala 2.x but not on 3.0" or "bundle depends on generic on Scala 2.x and on shapeless3 on 3.0" because `dependsOn` works at the project definition level, not as an SBT setting.

I think we need to work on 1) removing the special status of `generic` and shapeless as much as we possibly can and 2) try alternative approaches to handling different Scala versions. I've been exploring https://github.com/sbt/sbt-projectmatrix and opened a few issues there and I'll start by removing 
scalacheck-shapeless and trying to create a shapeless3 version of `generic`.

TimoMeijer:
@ruippeixotog I would like to help out and contribute, is there anything I could do that would be helpful to work towards further Scala 3 compatibility?

PR #1083, opened 2021-06-20
Remove dependency on scalacheck-shapeless

ruippeixotog:
It's fairly easy to create `Arbitrary` instances for case classes by generating tuples and mapping. Removing this dependency helps with #1082.

### DIFF
--- a/testkit/src/main/scala/pureconfig/BaseSuite.scala
+++ b/testkit/src/main/scala/pureconfig/BaseSuite.scala
@@ -7,8 +7,8 @@ import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
 
 class BaseSuite
     extends AnyFlatSpec
-    with ConfigConvertChecks
     with Matchers
-    with ConfigReaderMatchers
     with EitherValues
     with ScalaCheckDrivenPropertyChecks
+    with ConfigConvertChecks
+    with ConfigReaderMatchers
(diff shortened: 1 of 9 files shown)


testkit/src/main/scala/pureconfig/BaseSuite.scala
```diff
@@ -7,8 +7,8 @@ import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
 
 class BaseSuite
     extends AnyFlatSpec
-    with ConfigConvertChecks
     with Matchers
-    with ConfigReaderMatchers
     with EitherValues
     with ScalaCheckDrivenPropertyChecks
+    with ConfigConvertChecks
```
ruippeixotog, line 13: IntelliJ didn't like these two traits appearing before the first four (which the former require), so I just reordered them.

jcazevedo:
I'm OK with this change, but for #1082 this won't help much if we're still relying on `HList` from shapeless 2 (at least until milessabin/shapeless#1043 is ready). Wouldn't it be preferable to provide explicit instances (even if their implementation is more verbose) for `Foo`, `FlatConfig`, `MyType` and `ConfigWithUnknownType`?

ruippeixotog:
What do you mean with "relying on HList"? Our current implementation is totally dependent on shapeless 2 and it's a non-goal to make it Scala 3 ready, and on tests the only reference I can see is on `HListConvertersSuite` which I imagine it will always be shapeless 2 dependent. For everything else, we rely only on Scala tuples to generate arbitraries for case classes. Am I missing something?

jcazevedo:
Oh. Sorry. I read this wrong.

jcazevedo (approved):

PR #1095, opened 2021-07-11
add Spark DataType and StructType converters

yangzai:
This PR adds a new `spark` module with Spark `DataType` and `StructType` converters. Please refer to [README.md](https://github.com/pureconfig/pureconfig/blob/43e2485bb67a57e5d0f4344fe170dabd36d4bab8/modules/spark/README.md) for more details.

### DIFF
--- a/modules/spark/README.md
+++ b/modules/spark/README.md
@@ -7,7 +7,7 @@ Adds support for selected [Spark](http://spark.apache.org/) classes to PureConfi
 In addition to [core PureConfig](https://github.com/pureconfig/pureconfig), you'll need:
 
 ```scala
-libraryDependencies += "com.github.pureconfig" %% "pureconfig-spark" % "0.16.0"
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-spark" % "0.17.0"
 ```
 
 Also, `pureconfig-spark` depends on `spark-sql` with `provided` scope.
--- a/modules/spark/README.md
+++ b/modules/spark/README.md
@@ -36,6 +36,7 @@ import org.apache.spark.sql.types._
 import pureconfig._
 import pureconfig.generic.auto._
 import pureconfig.module.spark.sql._
+import com.typesafe.config.ConfigRenderOptions
 
 case class MySchema(name: String, fields: List[StructField], someOtherSetting: Option[String])
 
@@ -44,6 +45,8 @@ def mySchemaToSparkSchema(schema: MySchema): StructType =
 
 def sparkSchemaToMySchema(name: String, schema: StructType): MySchema =
   MySchema(name, schema.fields.toList, None)
+
+val renderOpt = ConfigRenderOptions.defaults.setOriginComments(false)
 ```
 
 Convert custom schema to Spark and back to custom schema. Resultant string schema should match original source.
@@ -52,7 +55,7 @@ val mySchemaRes = ConfigSource.string(
   """name: Employee,
     |fields: [
     |  { name: name, data-type: string }, #types are case-insensitive and some types have variations/truncations
-    |  { name: age, data-type: integer }, #also note that `nullable` and `metadata` are optional fields with Spark defaults
+    |  { name: age, data-type: integer, nullable = false, metadata = "{\"k\": \"v\"}" }, #also note that `nullable` and `metadata` are optional fields with Spark defaults
     |  { name: salary, data-type: "decimal(6,2)" },
     |  { name: address, data-type: "line1 string, line2 string" } #outer `struct` is optional
     |]
@@ -62,7 +65,7 @@ val mySchemaRes = ConfigSource.string(
 //     "Employee",
 //     List(
 //       StructField("name", StringType, true, {}),
-//       StructField("age", IntegerType, true, {}),
+//       StructField("age", IntegerType, false, {"k":"v"}),
 //       StructField("salary", DecimalType(6, 2), true, {}),
 //       StructField(
 //         "address",
@@ -82,7 +85,7 @@ val sparkSchemaRes = mySchemaRes.map(mySchemaToSparkSchema)
 // sparkSchemaRes: Either[error.ConfigReaderFailures, StructType] = Right(
 //   StructType(
 //     StructField("name", StringType, true, {}),
-//     StructField("age", IntegerType, true, {}),
+//     StructField("age", IntegerType, false, {"k":"v"}),
 //     StructField("salary", DecimalType(6, 2), true, {}),
 //     StructField(
 //       "address",
@@ -106,7 +109,7 @@ val mySchemaRes2 =
 //     "Employee",
 //     List(
 //       StructField("name", StringType, true, {}),
-//       StructField("age", IntegerType, true, {}),
+//       StructField("age", IntegerType, false, {"k":"v"}),
 //       StructField("salary", DecimalType(6, 2), true, {}),
 //       StructField(
 //         "address",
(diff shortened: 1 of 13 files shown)


modules/spark/build.sbt
```diff
@@ -0,0 +1,17 @@
+import Dependencies.Version._
+
+name := "pureconfig-spark"
+
+crossScalaVersions := Seq(scala212) //Spark does not support Scala 2.13 yet
```
yangzai, line 5: Getting CI errors from the exclusion of `scala213` here.

yangzai (commented):
https://github.com/pureconfig/pureconfig/pull/1095/checks?check_run_id=3041795063

modules/spark/build.sbt
yangzai: https://github.com/pureconfig/pureconfig/pull/1095/checks?check_run_id=3041795063

modules/spark/docs/README.md
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.spark.sql.types._
+import pureconfig._
+import pureconfig.generic.auto._
+import pureconfig.module.spark._
+
+case class MySchema(name: String, fields: List[Field])
+case class Field(name: String, dataType: DataType)
+
+def mySchemaToSparkSchema(schema: MySchema): StructType =
+  StructType(schema.fields.map { case Field(n, dt) => StructField(n, dt) })
```
ruippeixotog: I'm curious about this example; why don't we also add a converter for `StructField` for completion? We can read from config objects with `name` and `dataType` fields, with the other two `nullable` and `metadata` as optional (choosing defaults for each).

modules/spark/docs/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+def sparkSchemaToMySchema(name: String, schema: StructType): MySchema =
+  MySchema(name, schema.fields.toList.map { case StructField(n, dt, _, _) => Field(n, dt) })
+```
+
+Convert custom schema to spark and back to custom schema. Resultant string schema should match original source.
```
ruippeixotog:
```suggestion
Convert custom schema to Spark and back to custom schema. Resultant string schema should match original source.
```

modules/spark/src/main/scala/pureconfig/module/spark/spark.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import scala.util.Try
+
+import org.apache.spark.sql.types.{DataType, StructType}
+
+import pureconfig.{ConfigConvert, ConfigReader, ConfigWriter}
+
+/** `ConfigConvert` instances for Spark data structures.
+  */
+package object spark {
```
ruippeixotog: These are all `spark-sql` types. Do we expect `pureconfig-spark` to have non-sql converters too in the future?

If that's the case, I'd rather have define these in a separate `pureconfig/module/spark/sql.scala` object. While you correctly set `spark-sql` as provided, as it is now any user trying to use this module will need to provide this dependency, even if they're using things that don't depend on `spark-sql`.

ruippeixotog (changes requested):
Thanks for this PR @yangzai, this looks useful!

One thing that's missing here are tests. I know the implementations are very simple, but I'd rather see some round trip tests for your converters to ensure that readers can read what writers write. Adding tests for some more complex strings could be useful too.

modules/spark/docs/README.md
yangzai: @ruippeixotog `StructField` can be derivable but it requires `Metadata`. Maybe I could try providing a converter for `Metadata` instead of fixing it to default?
The custom `Field` in the example is kind of trivial but it could also contain other settings meant for other data stores that may not be relevant for Spark.

modules/spark/src/main/scala/pureconfig/module/spark/spark.scala
yangzai: Sure I can move it down a level.

yangzai:
> Thanks for this PR @yangzai, this looks useful!
> 
> One thing that's missing here are tests. I know the implementations are very simple, but I'd rather see some round trip tests for your converters to ensure that readers can read what writers write. Adding tests for some more complex strings could be useful too.

@ruippeixotog yup, will be adding some test cases as well.

modules/spark/build.sbt
yangzai: @ruippeixotog could you also have a look a this issue? It seems like if `scala213` is removed from any module it would fail.

modules/spark/docs/README.md
yangzai: `Metadata` can be serialised as Json strings. I will also simplify my example to use `StructField` directly instead.

yangzai: @ruippeixotog resolving with f125110d9a0911fe774ef12425086e13263f4caa

yangzai: @ruippeixotog moved 5957f793b7f9fa116277c9c29a256bf12e4350ba

modules/spark/src/main/scala/pureconfig/module/spark/spark.scala
yangzai: @ruippeixotog moved 5957f793b7f9fa116277c9c29a256bf12e4350ba

yangzai:
@ruippeixotog added some test cases 29928641ec67b4865609fd53b5f16b11a2bfd2f4

modules/spark/src/main/scala/pureconfig/module/spark/sql/package.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import pureconfig.ConfigConvert
+
+/** `ConfigConvert` instances for `spark-sql` data structures.
+  */
+package object sql {
+  implicit val dataTypeConvert: ConfigConvert[DataType] =
+    ConfigConvert.viaNonEmptyStringTry[DataType](s => Try(DataType.fromDDL(s)), _.sql)
+
+  implicit val structTypeReader: ConfigConvert[StructType] =
```
ruippeixotog:
```suggestion
  implicit val structTypeConvert: ConfigConvert[StructType] =
```

modules/spark/src/test/scala/pureconfig/module/spark/sql/SparkSqlTest.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import com.typesafe.config.ConfigValueFactory
+import org.apache.spark.sql.types._
+
+import pureconfig.BaseSuite
+import pureconfig.error.CannotConvert
+import pureconfig.generic.auto._
+import pureconfig.module.spark.sql.arbitrary._
+
+class SparkSqlTest extends BaseSuite {
```
ruippeixotog: Just for consistency with all other tests in PureConfig.

```suggestion
class SparkSqlSuite extends BaseSuite {
```

modules/spark/src/test/scala/pureconfig/module/spark/sql/arbitrary/package.scala
```diff
@@ -0,0 +1,93 @@
+package pureconfig.module.spark.sql
+
+import org.apache.spark.sql.types._
+import org.json4s._
+import org.json4s.jackson.JsonMethods.{compact, render}
+import org.scalacheck.{Arbitrary, Gen}
+
+package object arbitrary {
+  //Valid `Metadata` field types: Boolean, Long, Double, String, Metadata, Array[Boolean], Array[Long], Array[Double], Array[String], and Array[Metadata]
+  val genJBool: Gen[JBool] = Arbitrary.arbitrary[Boolean].map(JBool.apply)
```
ruippeixotog: nit: in other places in PureConfig (e.g. in `testkit`) we usually split `Gen` and `Arbitrary` instances into separate `gen` and `arbitrary` packages. Could you do the same here?

modules/spark/README.md
```diff
@@ -0,0 +1,233 @@
...
+```scala
+import pureconfig.module.spark._
+```
+
+## Supported classes
+
+* `org.apache.spark.sql.types.DataType`
+* `org.apache.spark.sql.types.StructType`
+* `org.apache.spark.sql.types.Metadata`
+* `org.apache.spark.sql.types.StructField` (derivable)
```
ruippeixotog: Hmm, it's probably confusing to list it like this and then require a generic derivation library to use it. Some users opt to not use generic derivation and this may be misleading for them.

Would you consider using `ConfigReader.forProduct4` and explicitly define the reader for `StructField` to make this module self-sufficient? Otherwise I'd rather not mention `StructField` here at all.

ruippeixotog (changes requested):

ruippeixotog:
@yangzai great work on the tests, thanks! :) I left one final batch of nits and comments.

modules/spark/src/test/scala/pureconfig/module/spark/sql/arbitrary/package.scala
yangzai: Sure, will do.

modules/spark/README.md
yangzai: Will look into `ConfigReader.forProduct4`.

modules/spark/src/test/scala/pureconfig/module/spark/sql/arbitrary/package.scala
yangzai: resolving with ab197aaae287f2a3da62db9422966315894f2cb8

modules/spark/README.md
yangzai: resolving with d942c9cfa8188fb9d04f01e84032e9640dd61a2b

yangzai:
@ruippeixotog sorry, I had to be away for a bit. Rebased since the branch has been stale for awhile. All change request should have been resolved, let me know if any other changes are required.

modules/spark/src/main/scala/pureconfig/module/spark/sql/package.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  implicit val structTypeConvert: ConfigConvert[StructType] =
+    ConfigConvert.viaNonEmptyStringTry[StructType](s => Try(StructType.fromDDL(s)), _.toDDL)
+
+  implicit val metadataConvert: ConfigConvert[Metadata] =
+    ConfigConvert.viaNonEmptyStringTry[Metadata](s => Try(Metadata.fromJson(s)), _.json)
+
+  private val (key0, key1, key2, key3) = ("name", "data-type", "nullable", "metadata")
+
+  implicit val structFieldReader: ConfigReader[StructField] = ConfigReader.forProduct4(key0, key1, key2, key3) {
```
ruippeixotog: nit: I'd rather inline all field names, even if it leads to them being repeated in two places. Putting variables with generic names like `key0, key1, key2, key3` in the scope of the whole module (even if private) seems a bit untidy to me. But if you prefer like this I'm ok with leaving it like that.

ruippeixotog (commented):
Hi @yangzai, thank you for putting the effort to update the PR and going the extra mile by investigating the issue with scalafix/sbt by yourself!

Your code looks good to me as it is now. However, I'd rather hold on a bit before merging this. I'm not really fond of the `isScala212` hack, mostly because our project structure and reliance on `crossScalaVersions` is not causing problems only on scalafix (see #1082). I'd like to try and do a larger refactoring of the SBT build to see if I can avoid it altogether, but it may take some time for me to be able to pick the task.

If I can't find a good solution in a timely manner, I'll happily merge this before the next PureConfig release. Does that sound good to you?

yangzai:
> Hi @yangzai, thank you for putting the effort to update the PR and going the extra mile by investigating the issue with scalafix/sbt by yourself!
> 
> Your code looks good to me as it is now. However, I'd rather hold on a bit before merging this. I'm not really fond of the `isScala212` hack, mostly because our project structure and reliance on `crossScalaVersions` is not causing problems only on scalafix (see #1082). I'd like to try and do a larger refactoring of the SBT build to see if I can avoid it altogether, but it may take some time for me to be able to pick the task.
> 
> If I can't find a good solution in a timely manner, I'll happily merge this before the next PureConfig release. Does that sound good to you?

Sure you can merge this anytime, or we could split this out as a separate project if needed.

modules/spark/src/main/scala/pureconfig/module/spark/sql/package.scala
yangzai: resolving with 12a3c096f1a82487c8aee836e8e29356a495d755

yangzai:
@ruippeixotog BTW Spark 3.2 with Scala 2.13 support is out https://mvnrepository.com/artifact/org.apache.spark/spark-core

ruippeixotog:
> @ruippeixotog BTW Spark 3.2 with Scala 2.13 support is out https://mvnrepository.com/artifact/org.apache.spark/spark-core

Good to know! The main problem with the build in this PR was fixed in #1188, there are just some final tweaks needed now. Adding support for Scala 2.13 would avoid the initial problem, but we would be just hiding it instead of fixing it - it would show up later in some other module. @scala-steward will bump Spark's version to 3.2 right after we merge this :)

ruippeixotog (approved):

PR #1097, opened 2021-07-12
Fix linter step in GitHub Actions

ruippeixotog:
Fixes the "Run linter" command, which is setting up the target Scala version in the wrong way.

The difference is subtle:
- `sbt ++2.13.6 <cmd>` makes SBT run two separate commands: `++2.13.6` changes _all_ projects to use Scala version 2.13.6, and `<cmd>` runs the given command after the change;
- `sbt "++2.13.6 <cmd>"` makes SBT run a single command, `++2.13.6 <cmd>`, meaning "run cmd on all projects that have 2.13.6 in their cross-version list".

This is the reason why #1095 is failing CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
         # Consider joining this action with the "Check formatting" one once Scalafix is supported for Scala 3.
       - name: Run linter
         if: startsWith(matrix.scala, '2')
-        run: sbt "++${{ matrix.scala-version }} scalafixAll --check"
+        run: sbt ++${{ matrix.scala-version }} "scalafixAll --check"
 
       - name: Compile
         run: sbt "++${{ matrix.scala-version }} compile"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
         # Consider joining this action with the "Check formatting" one once Scalafix is supported for Scala 3.
       - name: Run linter
         if: startsWith(matrix.scala, '2')
-        run: sbt ++${{ matrix.scala-version }} "scalafixAll --check"
+        run: sbt "++${{ matrix.scala-version }} scalafixAll --check"
 
       - name: Compile
         run: sbt "++${{ matrix.scala-version }} compile"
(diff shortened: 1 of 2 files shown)


ruippeixotog:
While it should have worked, it seems that Scalafix is doing some weird stuff with input arguments that prevents us from using `scalafixAll --check` in the way we intend... I'm going to investigate more tomorrow.

jcazevedo:
I vaguely remember having some issues with this in the past. I'll also try to investigate it today.

yangzai:
Related issues: https://github.com/sbt/sbt/issues/4181 https://github.com/sbt/sbt/issues/3465

ruippeixotog:
Thanks for the tip @yangzai, it looks like the workaround mentioned there worked!

jcazevedo (approved):
Thanks @ruippeixotog and @yangzai!

yangzai:
@ruippeixotog @jcazevedo https://github.com/pureconfig/pureconfig/pull/1095/checks?check_run_id=3090360533 oh no, I think its . We could try https://github.com/sbt/sbt/issues/3465#issuecomment-471714505 or https://github.com/sbt/sbt/issues/3465#issuecomment-504407867 (which is probably talking about https://github.com/twitter/algebird/blob/17a2a8de8b396b49a658604fa9e04c3bdbf55f19/build.sbt#L172)

yangzai:
@ruippeixotog @jcazevedo I was trying out the approach in https://github.com/sbt/sbt/issues/3465#issuecomment-471714505 but in the end all that was needed to be disabled was scaafix. So it was scalafix all along and not SBT...

Fixed in c1589f9d9d269ed1ae19aab67e597814cacdb410

ruippeixotog:
Hi @yangzai, thanks once more for helping with this. I was aware that the problem was with Scalafix, but I don't know if it's something they can fix or a limitation of SBT `InputTask` (tasks with arguments). Scalafix seems to fail when we try `sbt "++2.13.6 scalafixAll --check"`, but not when we do `sbt "++2.13.6 scalafixAll"` (which fixes files in addition to checking).

This is yet another problem with relying on the flawed SBT cross-version mechanism (see #1082 for other problems).

PR #1117, opened 2021-08-10
Add ChronoUnit ConfigReader

mdedetrich:
This PR adds a `ConfigReader` for `ChronoUnit`. Not sure if a test is necessary since its reusing an existing `javaEnumReader`.

### DIFF
--- a/core/src/main/scala/pureconfig/BasicReaders.scala
+++ b/core/src/main/scala/pureconfig/BasicReaders.scala
@@ -5,6 +5,7 @@ import java.math.{BigDecimal => JavaBigDecimal, BigInteger}
 import java.net.{URI, URL}
 import java.nio.file.{Path, Paths}
 import java.time._
+import java.time.temporal.ChronoUnit
 import java.time.{Duration => JavaDuration}
 import java.util.UUID
 import java.util.regex.Pattern
@@ -67,10 +68,7 @@ trait PrimitiveReaders {
 trait JavaEnumReader {
 
   implicit def javaEnumReader[A <: java.lang.Enum[A]](implicit tag: ClassTag[A]): ConfigReader[A] =
-    ConfigReader.fromString(catchReadError(s => {
-      val enumClass = tag.runtimeClass.asInstanceOf[Class[A]]
-      Enum.valueOf(enumClass, s)
-    }))
+    configurable.genericJavaEnumReader[A](identity)
 }
 
 /** Trait containing `ConfigReader` instances for classes related to file system paths and URIs.
@@ -108,6 +106,9 @@ trait JavaTimeReaders {
 
   implicit val periodConfigReader: ConfigReader[Period] =
     ConfigReader.fromNonEmptyString[Period](PeriodUtils.fromString)
+
+  implicit val chronoUnitReader: ConfigReader[ChronoUnit] =
+    configurable.genericJavaEnumReader[ChronoUnit](ConfigFieldMapping(KebabCase, ScreamingSnakeCase))
 
   implicit val javaDurationConfigReader: ConfigReader[JavaDuration] =
     ConfigReader.fromNonEmptyString[JavaDuration](catchReadError(JavaDuration.parse))
--- a/core/src/main/scala/pureconfig/configurable/package.scala
+++ b/core/src/main/scala/pureconfig/configurable/package.scala
@@ -4,10 +4,12 @@ import java.time._
 import java.time.format.DateTimeFormatter
 
 import scala.collection.JavaConverters._
+import scala.reflect.ClassTag
 
 import com.typesafe.config.ConfigValueFactory
 
-import pureconfig.ConfigConvert.{catchReadError, viaNonEmptyString}
+import pureconfig.ConfigConvert.viaNonEmptyString
+import pureconfig.ConvertHelpers.catchReadError
 import pureconfig.error.FailureReason
 
 /** Provides methods that create [[ConfigConvert]] instances from a set of parameters used to configure the instances.
@@ -74,4 +76,12 @@ package object configurable {
         keyFormatter(key) -> writerV.to(value)
       }.asJava)
     )
+
+  def genericJavaEnumReader[A <: java.lang.Enum[A]](
+      transformValue: String => String
+  )(implicit tag: ClassTag[A]): ConfigReader[A] =
+    ConfigReader.fromString(catchReadError(s => {
+      val enumClass = tag.runtimeClass.asInstanceOf[Class[A]]
+      Enum.valueOf(enumClass, transformValue(s))
+    }))
 }
--- a/core/src/main/scala/pureconfig/BasicWriters.scala
+++ b/core/src/main/scala/pureconfig/BasicWriters.scala
@@ -5,6 +5,7 @@ import java.math.{BigDecimal => JavaBigDecimal, BigInteger}
 import java.net.{URI, URL}
 import java.nio.file.Path
 import java.time._
+import java.time.temporal.ChronoUnit
 import java.time.{Duration => JavaDuration}
 import java.util.UUID
 import java.util.regex.Pattern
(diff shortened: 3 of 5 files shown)


core/src/main/scala/pureconfig/BasicReaders.scala
```diff
@@ -95,7 +96,7 @@ trait RegexReaders {
 
 /** Trait containing `ConfigReader` instances for `java.time` classes.
   */
-trait JavaTimeReaders {
+trait JavaTimeReaders extends JavaEnumReader {
```
mdedetrich: An alternative which avoids having `JavaTimeReaders` extend `JavaEnumReader` is to make a companion object for `JavaEnumReader` , i.e.

```scala
object JavaEnumReader extends JavaEnumReader
```

at which point you can just `import JavaEnumReader._`. In general it can be argued that these traits should have companion objects since they don't require any parameters and allow fine grained control of imports but I think thats a separate discussion/point.

core/src/main/scala/pureconfig/BasicReaders.scala
```diff
@@ -66,11 +67,19 @@ trait PrimitiveReaders {
...
     ConfigReader.fromString(catchReadError(s => {
       val enumClass = tag.runtimeClass.asInstanceOf[Class[A]]
-      Enum.valueOf(enumClass, s)
+      Enum.valueOf(enumClass, transformValue(s))
     }))
+
+  implicit def javaEnumReader[A <: java.lang.Enum[A]](implicit tag: ClassTag[A]): ConfigReader[A] =
+    baseJavaEnumReader[A](identity)
+
+  private[pureconfig] def mixedCaseJavaEnumReader[A <: java.lang.Enum[A]](implicit tag: ClassTag[A]): ConfigReader[A] =
```
mdedetrich: So I realized that there is already an `implicit def javaEnumReader` however the `Enum.valueOf` function that is used for Java readers typically only work with upper case values since that is the standard for Java enums.

When dealing with typesafe config howerver, its not typical to use uppercase values, i.e. in the context of this PR you would write a config something like

```
my-chrono-unit=millis
```
and not
```
my-chrono-unit=MILLIS
```

Hence why I created a private `mixedCaseJavaEnumReader` which also handles cases such as `Millis` and `millis`. There is an argument that the default `javaEnumReader` should have been a mixed case one rather than uppercase however this would break existing code out there so I left it as is. There is also merit in having `mixedCaseJavaEnumReader` as public however I am not sure if you want to expand the trait's namespace like this.

ruippeixotog: I like the idea of adding a public reader for Java enums with a configurable field mapping, but I have a few comments:

- `mixedCaseJavaEnumReader` is probably too specific for your case. For example, using kebab case (`MY_ENUM` -> `my–enum`) is probably more standard than simply having everything lower case. I'd rather have `baseJavaEnumReader` public and use it directly in `chronoUnitReader`;
- We have a special implementation for `String => String` functions with the purpose of `transformValue`, please see `ConfigFieldMapping`. Just like in product/coproduct hints, we should probably use `ConfigFieldMapping` directly instead of a general function (in your specific case, it would be `ConfigFieldMapping(ScreamingSnakeCase, KebabCase)`);
- `BasicReaders` should include only ready-to-use reader instances, not instances that need to be "configured". We have a package `pureconfig.configurable` to hold readers that need to be constructed explicitly. Can you please move your `baseJavaEnumReader` there, renaming it to something like `javaEnumReaderWithMapping`?

core/src/main/scala/pureconfig/BasicReaders.scala
ruippeixotog: If you follow my suggestion above of moving `baseJavaEnumReader` to `pureconfig.configurable` you won't have this problem :)

ruippeixotog (changes requested):
Thanks for the contribution, @mdedetrich!

mdedetrich:
Thanks for the reply, I have a couple of questions

> , we should probably use ConfigFieldMapping directly instead of a general function (in your specific case, it would be ConfigFieldMapping(ScreamingSnakeCase, KebabCase));

I just had a look at `ConfigFieldMapping` and from what I can tell reading the documentation it seems its only  meant to apply to `case class` fields and not actual values (i.e. the rhs of the `=`), i.e.

```
myField=something
```

vs

```
my-field=something
```

Wouldn't it be confusing to use a `ConfigFieldMapping` for both keys and values? In my case I want the transformation to *only* apply to values and not keys/fields.

> BasicReaders should include only ready-to-use reader instances, not instances that need to be "configured". We have a package pureconfig.configurable to hold readers that need to be constructed explicitly. Can you please move your baseJavaEnumReader there, renaming it to something like javaEnumReaderWithMapping

Agreed however should we still have a main `implicit val chronoUnitReader: ConfigReader[ChronoUnit]` in `JavaTimeReaders`? I believe there is a difference between configurable readers for things like date time formatters (where there are legitimately different unique formats for date time) and something like `ChronoUnit` where you have different cases for the same thing (i.e. `half-days` vs `HALF_DAYS` vs `half_days`). Wouldn't having a single kebab case (as you suggested) reader for `ChronoUnit` be more preferable since any other kind of configuration would be simply be non standard for typesafe config?

ruippeixotog:
> Wouldn't it be confusing to use a ConfigFieldMapping for both keys and values? In my case I want the transformation to only apply to values and not keys/fields.

I'm afraid the documentation (and the class name itself) are a bit misleading. `ConfigFieldMapping` is used generally as a way to map between naming conventions, regardless of whether they're applied in keys or values. For example, besides case class field mapping, it's also used to choose the values for the disambiguation `type` field in sealed families (e.g. `type = blue-car` to select a `BlueCar` subclass). It's ok to use it for this use case 👍 

> Agreed however should we still have a main implicit val chronoUnitReader: ConfigReader[ChronoUnit] in JavaTimeReaders?

Yes, `chronoUnitReader` should remain in `JavaTimeReaders` just like you did, and its implementation should call our new configurable reader - I see no problem with a "non-configurable" reader being a particular instance of a configurable one (particularly one so general like a reader for all Java enums).

mdedetrich:
Thanks for the responses, I will work on the PR tomorrow!

mdedetrich:
@ruippeixotog I believe I have done the changes (along with re basing on master), let me know if anything else needs to be done.

core/src/main/scala/pureconfig/configurable/package.scala
```diff
@@ -74,4 +76,12 @@ package object configurable {
         keyFormatter(key) -> writerV.to(value)
       }.asJava)
     )
+
+  def baseJavaEnumReader[A <: java.lang.Enum[A]](
```
ruippeixotog: Can you please name this `genericJavaEnumReader` for consistency with other readers here?

core/src/main/scala/pureconfig/BasicReaders.scala
```diff
@@ -109,6 +107,9 @@ trait JavaTimeReaders {
   implicit val periodConfigReader: ConfigReader[Period] =
     ConfigReader.fromNonEmptyString[Period](PeriodUtils.fromString)
 
+  implicit val chronoUnitReader: ConfigReader[ChronoUnit] =
+    configurable.baseJavaEnumReader[ChronoUnit](ConfigFieldMapping(ScreamingSnakeCase, KebabCase))
```
ruippeixotog: Can you add some tests for this new reader?

You may also consider replicating this for the write side so you can do roundtrip tests and keep parity (not required to merge this, though).

ruippeixotog (changes requested):
Sorry for taking long to reply. Almost there, here are some final comments :)

mdedetrich:
@ruippeixotog I have resolved all of your remarks, I also ended up adding a round trip test for `ChronoUnit`

ruippeixotog (approved):
Thanks again, @mdedetrich!

mdedetrich:
@ruippeixotog Sorry for disturbing but would it be possible to make a release of pureconfig with this change in there?

ruippeixotog:
Hi @mdedetrich, I'll do it next week 👍

mdedetrich:
Awesome, thank you!

PR #1135, opened 2021-08-29
Dont redundantly scalafmt on compile

mdedetrich:
This opts for checking scalafmt when a PR is opened rather than doing it on compile. This can be considered a subjective change however here is my personal reasoning

* Doing format on compile significantly impacts IDE/editor performance when working on the codebase. Since editors/IDE's index/track source files, its more ideal to do this once then every time you compile (especially for a language like Scala where you do frequently compile just to check your code is correct). Furthemore some editors such as Intellij can sometimes report file conflicts when a formatted file is quickly edited.
* Doing a double check of both formatting and checking on compile can be considered redundant. In terms of practicality you don't really care that much if a PR in progress is not formatted, you only want to make sure that code is formatted in the final step before merging.

Note that the PR removes using `scalafmt` in `ci.yaml` and instead creates it as a separate pipeline that uses scalafmt-native. This has the following advantages
* scalafmt-native is significantly faster than standard scalafmt in sbt (I haven't seen it ever take longer than 10 seconds on massive codebases)
* it runs in parallel to the main compile/test rather than sequentially. For the same reason it also doesn't block the build/test if your code is formatted (it only blocks the merging of the PR). This gives you quick feedback that the code isn't formatted.

Note that I have submitted this kind of scalafmt checking on numerous Scala projects and there haven't been any issues.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,26 @@
+name: Scalafmt
+
+on:
+  pull_request:
+    branches: ['**']
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  build:
+    name: Code is formatted
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout current branch (full)
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Check project is formatted
+        uses: jrouly/scalafmt-native-action@v1
+        with:
+          version: '3.0.1'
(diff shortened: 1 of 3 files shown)


.github/workflows/format.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Checkout current branch (full)
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Check project is formatted
+        uses: jrouly/scalafmt-native-action@v1
+        with:
+          version: '3.0.1'
```
mdedetrich, line 26: This specific github action will automatically work when a new scalafmt version comes out, i.e. you are not blocked when a new scalafmt is released due to the maintainer having to manually release a new version of the plugin.

PR #1143, opened 2021-09-05
add module zio-config

yangzai:
New ZIO Config module that derives `ConfigConvert` from `ConfigDescriptor` from ZIO Config.

### DIFF
--- a/modules/zio-config/docs/README.md
+++ b/modules/zio-config/docs/README.md
@@ -20,10 +20,13 @@ import com.typesafe.config.ConfigRenderOptions
 import pureconfig._
 import pureconfig.module.zioconfig._
 import zio.config.ConfigDescriptor
-import zio.config.magnolia.DeriveConfigDescriptor.descriptor
+import zio.config.ConfigDescriptor._
 
 case class Person(name: String, age: Int, children: List[Person])
-object Person { implicit val confDesc: ConfigDescriptor[Person] = descriptor }
+object Person {
+  implicit val confDesc: ConfigDescriptor[Person] =
+    (string("name") |@| int("age") |@| list("children")(confDesc))(Person.apply, Person.unapply)
+}
 ```
 
 You can now read and write `Person` without re-implementing or re-deriving `ConfigConvert`.
@@ -35,4 +38,3 @@ val res = ConfigWriter[Person].to(alice)
 
 val maybeAlice = ConfigSource.string(res).load[Person]
 ```
-
--- a/modules/zio-config/docs/README.md
+++ b/modules/zio-config/docs/README.md
@@ -14,13 +14,14 @@ libraryDependencies += "com.github.pureconfig" %% "pureconfig-zio-config" % "@VE
 
 `Person` already has a `ConfigDescriptor` instance.
 By importing `pureconfig.module.zioconfig._`, it would also have a `ConfigConvert` instance.
-
 ```scala mdoc:silent
 import com.typesafe.config.ConfigRenderOptions
 import pureconfig._
 import pureconfig.module.zioconfig._
 import zio.config.ConfigDescriptor
 import zio.config.ConfigDescriptor._
+
+val configOpt = ConfigRenderOptions.defaults.setOriginComments(false)
 
 case class Person(name: String, age: Int, children: List[Person])
 object Person {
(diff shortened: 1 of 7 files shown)


modules/zio-config/docs/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+In addition to the [PureConfig core](https://github.com/pureconfig/pureconfig), you'll need:
+
+```scala
+libraryDependencies += "com.github.pureconfig" %% "pureconfig-zio-config" % "@VERSION@"
+```
+
+## Example
+
+`Person` already has a `ConfigDescriptor` instance.
```
ruippeixotog, line 15: There seems to be a bit of context lacking here. Users may not understand right away why `Person` already has a `ConfigDescriptor` instance, particularly if they don't use automatic derivation or `zio-config-magnolia` in particular.

I'd rather not introduce any additional dependencies, neither here nor in tests. From [ZIO Config docs](https://zio.github.io/zio-config/docs/configdescriptor/configdescriptor_index#a-simple-example), creating a `ConfigDescriptor` manually seems rather straightforward and it would make the requirements for this module clearer. Would you mind defining it manually here?

If you really feel that using automatic derivation is important here and in tests, please improve this paragraph to explain that `DeriveConfigDescriptor` triggers automatic derivation, possibly linking to appropriate ZIO Config documentation.

modules/zio-config/docs/README.md
```diff
@@ -0,0 +1,38 @@
+# ZIO Config module for PureConfig
+
+Support for providing instances of `ConfigCovert` given instances of ZIO Config `ConfigDescriptor`
```
ruippeixotog: nit, a link to the library would help.

```suggestion
Support for providing instances of `ConfigCovert` given instances of [ZIO Config](https://zio.github.io/zio-config/) `ConfigDescriptor`.
```

ruippeixotog (changes requested):
The implementation LGTM, @yangzai! Please take a look at the comments below. Also, please add a reference to this module in docs/docs/docs/library-integrations.md.

modules/zio-config/docs/README.md
yangzai: @ruippeixotog I didn't want users to bother about how existing `ConfigDescriptor`s was implemented but it is also true that users should not need to know how automatic derivation works to understand the example. I'm removing `zio-config-magnolia` from the test dependencies as well since we are removing it from docs.

modules/zio-config/src/main/scala/pureconfig/module/zioconfig/package.scala
```diff
@@ -0,0 +1,22 @@
...
+    ConfigConvert[Config].xemap(
+      c =>
+        TypesafeConfigSource
+          .fromConfig(c)
+          .flatMap(cs => read(cd.from(cs)))
+          .left
+          .map(ZioConfigReadError.apply),
+      //`zio-config` write should not fail but is allowed to fail
+      //defaulting write failures to empty `Config` so it fits `pureconfig` writer
+      _.toHocon(cd).map(_.toConfig).getOrElse(ConfigFactory.empty)
```
ruippeixotog: One thought: should we throw an exception in case of an error instead of returning an empty object? Write failures may be unexpected and even technically impossible, but if some error ends up occurring I'd rather fail fast and loud rather than using a default that may make things very confusing (it may even be a valid success result in some cases). What are your thoughts?

ruippeixotog (commented):
Looks good, thank you again @yangzai. I left one last comment below - let me know your thoughts. I'm happy to merge this either way.

modules/zio-config/src/main/scala/pureconfig/module/zioconfig/package.scala
yangzai: @ruippeixotog I understand the issue of silent failures but on the other hand users might be depending on `pureconfig` writes to be safe and pure. I will switch to throwing to conform to what has been done in other modules for write errors, e.g. the `shapeless` writers in `pureconfig-generic`.

`zio-config` writes can still be made to fail for arbitrary reasons, e.g:
```scala
val confDescChild: ConfigDescriptor[Person] =
(string("name") |@| int("age") |@| list("children")(confDescChild))(Person.apply, Person.unapply)
  .transformOrFailRight(identity, {
    case Person(_, age, _) if age > 18 => Left("Too old to be a child.")
    case Person(_, _, children) if children.nonEmpty => Left("Child cannot have children.")
    case p => Right(p)
  })
}
```
Although personally, in cases like that, I would prefer refining the types. Maybe I should also add that as another example for write failures.

ruippeixotog (approved):

PR #1146, opened 2021-09-09
Update library-integrations for zio-config module

yangzai:
@ruippeixotog Sorry I forgot about [this](https://github.com/pureconfig/pureconfig/pull/1143#pullrequestreview-746686022).


ruippeixotog (approved):
Right, I forgot too. Thanks!

PR #1168, opened 2021-10-01
Documentation fixes

mattkohl:
Thank you for pureconfig!


leifwickland (approved):
@mattkohl Thanks for using PureConfig. Thanks for taking the time to make these improvements. They look good to me.

PR #1179, opened 2021-10-08
Remove CONTRIBUTORS file

ruippeixotog:
It's been a long time since we updated this file and even when we did we haven't been very consistent (sometimes asking contributors to add their name, other times not).

Our git history is the source of truth for who contributed with commits to this repository, so I think we should delete this file as redundant.


jcazevedo (approved):
I agree. Thanks for this.

PR #1180, opened 2021-10-08
Update Build Website CI dependencies

ruippeixotog:
As titled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,12 +87,12 @@ jobs:
       - name: Set up Ruby
         uses: actions/setup-ruby@v1
         with:
-          ruby-version: 2.6
+          ruby-version: 3.0
 
       - name: Install dependencies
         run: >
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -87,12 +87,12 @@ jobs:
       - name: Set up Ruby
         uses: actions/setup-ruby@v1
```
jcazevedo: We're getting a warning that this action is deprecated and no longer maintained and that we should use [ruby/setup-ruby](https://github.com/ruby/setup-ruby) instead. Do you want to change it in this PR since you're already updating the workflow?

jcazevedo (approved):

.github/workflows/ci.yml
ruippeixotog: I hadn't noticed the warning, thanks. Updated this in 4e8c817.

PR #1181, opened 2021-10-09
Adds combinator `ensure` to `ConfigReader`

queimadus:
The `ensure` combinator allows a user to quickly fail a reader if a condition does not hold without having to deal with `Either`s nor the `emap` boilerplate.

### DIFF
--- a/core/src/main/scala/pureconfig/error/FailureReason.scala
+++ b/core/src/main/scala/pureconfig/error/FailureReason.scala
@@ -144,3 +144,10 @@ final case class WrongSizeList(expected: Int, found: Int) extends FailureReason 
 final case class WrongSizeString(expected: Int, found: Int) extends FailureReason {
   def description = s"String of wrong size found. Expected $expected characters. Found $found characters instead."
 }
+
+/** A failure reason given when a validation failed.
+  *
+  * @param description
+  *   the validation failed description
+  */
+final case class ValidationFailed(description: String) extends FailureReason
--- a/core/src/main/scala/pureconfig/error/FailureReason.scala
+++ b/core/src/main/scala/pureconfig/error/FailureReason.scala
@@ -145,9 +145,9 @@ final case class WrongSizeString(expected: Int, found: Int) extends FailureReaso
   def description = s"String of wrong size found. Expected $expected characters. Found $found characters instead."
 }
 
-/** A failure reason given when a validation failed.
+/** A failure reason given when a user validation failed.
   *
   * @param description
   *   the validation failed description
   */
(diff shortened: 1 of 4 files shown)


core/src/main/scala/pureconfig/error/FailureReason.scala
```diff
@@ -144,3 +144,10 @@ final case class WrongSizeList(expected: Int, found: Int) extends FailureReason
 final case class WrongSizeString(expected: Int, found: Int) extends FailureReason {
   def description = s"String of wrong size found. Expected $expected characters. Found $found characters instead."
 }
+
+/** A failure reason given when a validation failed.
+  *
+  * @param description
+  *   the validation failed description
+  */
+final case class ValidationFailed(description: String) extends FailureReason
```
ruippeixotog: nit: maybe `UserValidationFailed`, to emphasize that this is not emitted by internal code?

docs/docs/docs/combinators.md
```diff
@@ -49,6 +49,23 @@ ConfigSource.string("{ port = 8080 }").load[PortConf]
...
 ```
 
+`ensure` allows users to quickly fail a reader if a condition does not hold:
+
+```scala mdoc:silent
+case class Age(years: Int)
+case class User(age: Age)
+
+// reads a valid age
+implicit val userReader = ConfigReader[Int]
```
ruippeixotog: This should be called `ageReader`.

I have some minor concerns with this example. In this scenario, I think the recommendation would be to add a `require` call in the body of `Age` to ensure there's no way to create an invalid `Age` - from configs or any other method. 

A combinator like this really shines when you use more than one field of the same object (e.g. to check that a registration date is after the user's date of birth), as there's no way to create a "wrapper" like `Age` in that scenario. To keep this example simple but useful, an alternative would be to define `case class User(age: Int)` and create a user reader with `ensure` (in this case `ensure` saves the user the trouble of creating a wrapper). What do you think?

ruippeixotog (changes requested):
This combinator is a great idea, thanks for the contribution @queimadus! :)

core/src/main/scala/pureconfig/error/FailureReason.scala
queimadus: I agree. Done in e8e1404.

docs/docs/docs/combinators.md
queimadus: I like the idea of having a validation between two fields of the same class! Please take a look at ab60f63.

ruippeixotog (approved):
Thanks again!

PR #1183, opened 2021-10-11
Remove dependency on generic for modules' tests

jcazevedo:
This PR aims to reduce the special status pureconfig-generic currently has in the codebase, by removing its dependency from the modules' test scope. This is helpful for #1082.


modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
```diff
@@ -6,64 +6,79 @@ import cats.data._
...
+import pureconfig.{BaseSuite, ConfigReader, ConfigSource, ConfigWriter}
 
 class CatsSuite extends BaseSuite {
 
-  case class Numbers(numbers: NonEmptyList[Int])
-  case class NumVec(numbers: NonEmptyVector[Int])
-  case class NumSet(numbers: NonEmptySet[Int])
-  case class NumMap(numbers: NonEmptyMap[String, Int])
-  case class NumChain(numbers: NonEmptyChain[Int])
+  case class Numbers[T](numbers: T)
```
ruippeixotog, line 15: Can we avoid this unnecessary wrapper too? You can use `ConfigSource` here to navigate through a config and extract a `ConfigValue` for `checkReadWrite`, I think. You can also keep two configs `numberListConf` and `numberMapConf` to avoid repetition.

modules/magnolia/build.sbt
```diff
@@ -6,7 +6,9 @@ crossScalaVersions := Seq(scala212, scala213)
 
 libraryDependencies ++= Seq(
   "com.propensive" %% "magnolia" % "0.17.0",
-  "org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided
+  "org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided,
+  // We're using shapeless for illTyped in tests.
+  Dependencies.shapeless % Test
```
ruippeixotog: Slightly related with this PR: as we are trying to remove the special status of generic, can we remove shapeless from `Dependencies`? Ideally we should be using shapeless only inside generic (here it's ok as we're using it for completely different reasons, just to use utils).

modules/magnolia/src/test/scala-2.12+/pureconfig/CoproductConvertersSuite2_12.scala
```diff
@@ -1,8 +1,11 @@
 package pureconfig
 
+import scala.language.higherKinds
+
 import com.typesafe.config.ConfigFactory
 
-import pureconfig.generic.auto._
```
ruippeixotog, line 5: How was this available here? 😮 I thought `magnolia` didn't depend on `generic`?

testkit/src/main/scala-2.13-/pureconfig/TypeStringCompat.scala
```diff
@@ -1,13 +1,22 @@
...
   implicit def fromTypeTag[T](implicit t: TypeTag[T]): TypeStringCompat[T] = new TypeStringCompat[T] {
     def typeName = t.tpe.toString
   }
 }
+
+trait TypeStringCompatLowerPriority {
+  // A fallback for types that don't have a TypeTag. ClassTags are weaker cases of TypeTags, not knowing about the
+  // argument types of a given type. They therefore produce type strings with less information.
+  implicit def fromClassTag[T](implicit t: ClassTag[T]): TypeStringCompat[T] = new TypeStringCompat[T] {
+    def typeName = t.runtimeClass.getName.toString
```
ruippeixotog, line 20: Nit: is it worth changing the type description here to clarify that the type may be erased? (e.g. by appending "erased" or being smarter in case `ClassTag` knows if it's type has type parameters). I'm afraid this might be misleading if it's unclear if errors are using a type tag or class tag.

ruippeixotog (changes requested):
LGTM overall.

One general suggestion: after seeing 15+ files using this pattern, it seems to be that we should create some test util `def configValueSource(conf: String) = ConfigSource.string(s"""{ key: $conf }""").at("key")` on `BaseSuite` or other central place. In fact, we already have a very similar util for `ConfigValue` and cursors (https://github.com/pureconfig/pureconfig/blob/master/tests/src/test/scala/pureconfig/FluentConfigCursorSuite.scala#L12-L14).

testkit/src/main/scala-2.13-/pureconfig/TypeStringCompat.scala
jcazevedo: Is it worth it? Note that this is only used for test descriptions in the `checkRead` and `checkWrite` (and variants) methods.

ruippeixotog: Oh, never mind then. I thought it was being used to generate exception messages in library code, I recall that we have similar code working with type/class tags there.

modules/magnolia/build.sbt
jcazevedo: Sure. That makes sense.

modules/magnolia/src/test/scala-2.12+/pureconfig/CoproductConvertersSuite2_12.scala
jcazevedo: It used the `module` helper, which ended up including generic as a test dependency.

modules/cats/src/test/scala/pureconfig/module/cats/CatsSuite.scala
jcazevedo: > Can we avoid this unnecessary wrapper too?

The problem here is that some of these cats types (`NonEmptySet`, `NonEmptyMap` and `NonEmptyChain`) don't have `TypeTag` nor `ClassTag` instances (see https://github.com/typelevel/cats/issues/2928), so they're not directly usable in `checkReadWrite`. I didn't want to disrupt the rest of the spec much, so I kept the usage of a wrapper here for us to have access to, at least, a `ClassTag` instance.

> You can also keep two configs `numberListConf` and `numberMapConf` to avoid repetition.

I don't mind the repetition much here, and I think it helps with readability given the proximity to the produced values.

jcazevedo:
> One general suggestion: after seeing 15+ files using this pattern, it seems to be that we should create some test util `def configValueSource(conf: String) = ConfigSource.string(s"""{ key: $conf }""").at("key")` on `BaseSuite` or other central place. In fact, we already have a very similar util for `ConfigValue` and cursors (https://github.com/pureconfig/pureconfig/blob/master/tests/src/test/scala/pureconfig/FluentConfigCursorSuite.scala#L12-L14).

Thanks for this suggestion! I ended up pushing the `conf` method to `BaseSuite`, renaming it to `configValue`, and using it throughout these specs instead of the `ConfigSource`-based loading.

ruippeixotog (approved):
Thanks @jcazevedo, this will be very useful later!

One nit: now that we define `configValue` like that, it makes me think that we should have a `configValueStr` variant to avoid those awkward `""""` on many definitions. What do you think? Leaving up to you, I accepted this in case you want to do it later.

jcazevedo:
> One nit: now that we define `configValue` like that, it makes me think that we should have a `configValueStr` variant to avoid those awkward `""""` on many definitions. What do you think? Leaving up to you, I accepted this in case you want to do it later.

Makes sense. I've added a `configString` helper for that.

ruippeixotog (approved):
Thanks for all the changes!

PR #1184, opened 2021-10-12
Specify dialects for Scalafmt

jcazevedo:
We were getting some warnings when running scalafmt like:

```
Default dialect is deprecated; use explicit: [Scala211,scala212,Scala212Source3,scala213,Scala213Source3,Sbt0137,Sbt1,scala3]
```

This PR modifies the Scalafmt configuration to specify a dialect for .scala and .sbt files.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,9 +1,13 @@
 version = 3.0.6
 
+runner.dialect = scala213
 align.preset = none
 maxColumn = 120
 
 fileOverride {
+  "glob:**/*.sbt" {
+    runner.dialect = sbt1
+  }
   "glob:**/scala-3*/**" {
     runner.dialect = scala3
   }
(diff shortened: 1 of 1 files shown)


.scalafmt.conf
```diff
@@ -1,9 +1,13 @@
 version = 3.0.6
 
+runner.dialect = scala213
 align.preset = none
 maxColumn = 120
 
 fileOverride {
+  "glob:**/*.sbt" {
```
ruippeixotog: Shoudn't scalafmt choose sbt out of the box for .sbt files? It seems a bit verbose as this setting probably applies to every Scala project out there.

ruippeixotog (approved):
LGTM

.scalafmt.conf
jcazevedo: I'm not sure. I was assuming that this would use the globally defined dialect. Removing this override yields exactly the same format, so I think we're safe with removing it nonetheless.

ruippeixotog (approved):

PR #1172, opened 2021-10-05
Update olafurpg/setup-scala GitHub action to v13

jcazevedo:

PR #1185, opened 2021-10-12
Add a .jvmopts file

jcazevedo:
This PR adds a .jvmopts file to specify some sensible defaults for sbt. Should help with the build failures in #1172.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,0 +1,4 @@
+-Dfile.encoding=UTF8
+-Xms1g
+-Xmx3g
+-Xss2m


ruippeixotog:
Have you checked that these settings work with #1172? It probably makes sense to ship both of these changes in a single PR, as these settings are only needed because of the CI change (assuming individual developers take care of their own configuration or use sbt-extras).

jcazevedo:
> Have you checked that these settings work with #1172? 

A successful build here should indicate that, since both sbt-extras (used in the olafurpg/setup-scala version of this PR) and the default sbt launcher (used in the olafurpg/setup-scala version on #1172) append the settings in .jvmopts to `JAVA_OPTS`.

> It probably makes sense to ship both of these changes in a single PR, as these settings are only needed because of the CI change (assuming individual developers take care of their own configuration or use sbt-extras).

Sure. I've merged this into #1172.

PR #1188, opened 2021-10-15
Make CI commands adequately scoped on Scala version

ruippeixotog:
Fixes all CI commands to be correctly scoped on the Scala version being tested, introducing a new SBT task to work around https://github.com/scalacenter/scalafix/issues/1488.

As I mentioned in a previous PR, the difference is subtle but important:

- `sbt ++2.13.6 <cmd>` makes SBT run two separate commands, one after the other: `++2.13.6` changes the `scalaVersion` projects that have 2.13.6 in their `crossScalaVersions` to that one, ignoring all the other projects. It then runs `<cmd>`, which runs *for all subprojects* (since they are all children of the root project). The consequence of this is that if a project doesn't support Scala 2.13, `<cmd>` will still run using the original `scalaVersion` (currently 2.12).
- `sbt "++2.13.6 <cmd>"` makes SBT run a single command, `++2.13.6 <cmd>`, which means "run `<cmd>` on all projects that have 2.13.6 in their cross-version list" and it explicitly leaves some subprojects out (running `+2.13.6 -v` shows more details on which projects are left out).

Our current CI pipeline only works now because all projects support at least Scala 2.12 and 2.13, but it's the main reason why #1095 is failing CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,12 +28,12 @@ jobs:
           java-version: ${{ matrix.jdk }}
 
       - name: Check formatting
-        run: sbt ++${{ matrix.scala-version }} scalafmtCheckAll scalafmtSbtCheck
+        run: sbt "++${{ matrix.scala-version }} scalafmtCheckAll" scalafmtSbtCheck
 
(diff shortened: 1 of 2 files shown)


jcazevedo (approved):
Thanks for this!

PR #1195, opened 2021-10-22
Add Hostname and Port implicits in Http4s module.

geirolz:
Http4s use often `Hostname` and `Port`, having those two types directly in this module would be very useful.


modules/http4s/src/main/scala/pureconfig/module/http4s/package.scala
```diff
@@ -1,16 +1,40 @@
...
+  implicit val hostnameReader: ConfigReader[Hostname] =
+    ConfigReader.fromString(str =>
+      Hostname.fromString(str) match {
+        case Some(hostname) => Right(hostname)
+        case None => Left(CannotConvert(str, "Hostname", "Invalid hostname"))
+      }
+    )
+
+  implicit val portWriter: ConfigWriter[Port] =
+    ConfigWriter[String].contramap(_.toString)
```
ruippeixotog: Reading and writing ports from strings doesn't seem like a natural choice. Why don't we read and write from integers?

ruippeixotog (commented):
Thanks for the contribution @geirolz, these converters look useful.

modules/http4s/src/main/scala/pureconfig/module/http4s/package.scala
geirolz: Totally agree! 👍

geirolz: Done

modules/http4s/src/main/scala/pureconfig/module/http4s/package.scala
```diff
@@ -1,16 +1,40 @@
...
+        case Some(hostname) => Right(hostname)
+        case None => Left(CannotConvert(str, "Hostname", "Invalid hostname"))
+      }
+    )
+
+  implicit val portWriter: ConfigWriter[Port] =
+    ConfigWriter[Int].contramap(_.value)
+
+  implicit val portReader: ConfigReader[Port] =
+    ConfigReader.intConfigReader.emap(str =>
```
ruippeixotog: You shouldn't rely on `ConfigReader.intConfigReader` here, as `intConfigReader` is meant to be consumed as an implicit. You should use `ConfigReader[Int]` instead, similarly to how you summon a writer above.

```suggestion
    ConfigReader[Int].emap(str =>
```

ruippeixotog (changes requested):

modules/http4s/src/main/scala/pureconfig/module/http4s/package.scala
geirolz: Sorry, good to know, just updated the PR

geirolz: Done

jcazevedo (changes requested):
@geirolz: Thanks for your contribution! These converters are helpful! 

Since ip4s is a dependency of http4s, I think it's preferable to scope them under their own module, so that users of ip4s alone can benefit from them without forcing a dependency on http4s.

geirolz:
@jcazevedo Actually I thought the same, if you agree I can do it 👍

geirolz:
@jcazevedo done!

geirolz:
Suggestion for the broken test ? I can understand the error

jcazevedo:
> Suggestion for the broken test ? I can understand the error

The Markdown files under modules/ip4s are expected to be compiled versions of the ones under modules/ip4s/docs. CI is complaining because they're copies of one another. You can run `mdoc` on the ip4s module to generate the final Markdown files (under modules/ip4s) and commit those.

ruippeixotog:
> Since ip4s is a dependency of http4s, I think it's preferable to scope them under their own module, so that users of ip4s alone can benefit from them without forcing a dependency on http4s.

I don't think this is desirable as we have it right now, @jcazevedo. I don't think it should be a requirement that we create one module for every single Maven dependency - it will make dependency management very difficult, as either us or PureConfig users will have to mirror the dependency tree. As we have it right now, users would have to figure out that `Hostname` and `Port` comes from a different library depended upon by http4s and manually add this new module to their dependencies.

As a proxy metric: [ip4s](https://github.com/Comcast/ip4s) has 168 stars while [http4s](https://github.com/http4s/http4s) has 2.1K. I'd say that it's highly unlikely we'll find users using ip4s alone, but if we end up having use cases for it I'd rather move this to a new module later rather than doing it now.

jcazevedo:
> I don't think this is desirable as we have it right now, @jcazevedo. I don't think it should be a requirement that we create one module for every single Maven dependency - it will make dependency management very difficult, as either us or PureConfig users will have to mirror the dependency tree. 

I'm not advocating for one module per Maven dependency, but ip4s doesn't seem that tied to http4s to justify having its converters in that module. The converters could just as well be put under the fs2 module, which also depends on ip4s.

The feeling that I get is that ip4s provides data types that are useful for describing the configuration of an application, regardless of the libraries that make use of them.

> As we have it right now, users would have to figure out that `Hostname` and `Port` comes from a different library depended upon by http4s and manually add this new module to their dependencies.

I actually find it less surprising that the converters for entities under the `com.comcast.ip4s.*` package are in a different library than the ones for the entities under the `org.http4s.*` package.

leifwickland:
I'm with @jcazevedo on this one. I can imagine needing `ip4s` in my apps that don't use `http4s`.

ruippeixotog:
@jcazevedo, @leifwickland my argument wasn't that having a separate ip4s module didn't make sense or that it was too tied to http4s, it was about not splitting modules prematurely until there's a real demand for it.

From your comments I may have misjudged the value of having a separate ip4s module, though. If you really see value in it, what do you think of having `pureconfig-http4s` depending on `pureconfig-ip4s`? What bothers me the most in the current setup is the scenario where:
1. Someone decides to use http4s and wants to use a PureConfig module for it
2. They find out that they are lacking important converters for `Hostname` and `Port`
3. They have to manually find out what dependency those classes came from and manually add new PureConfig modules to their build (or if they are lazy they may not even bother to find it)

This kind of dependency management doesn't scale (what if `ip4s` started splitting things further?). Users add `http4s` and SBT makes sure all dependencies are there; I'd argue that adding `pureconfig-http4s` should do the same, even if we don't have explicit compile-time dependencies (though we could have the http4s module re-exporting converters from the ip4s module in that scenario).

jcazevedo:
> If you really see value in it, what do you think of having `pureconfig-http4s` depending on `pureconfig-ip4s`?

I'm OK with that.

leifwickland:
> If you really see value in it, what do you think of having `pureconfig-http4s` depending on `pureconfig-ip4s`?

I'm also OK with that.

geirolz:
> > Suggestion for the broken test ? I can understand the error
> 
> The Markdown files under modules/ip4s are expected to be compiled versions of the ones under modules/ip4s/docs. CI is complaining because they're copies of one another. You can run `mdoc` on the ip4s module to generate the final Markdown files (under modules/ip4s) and commit those.

Running `docs/mdoc` I can't find the compilated version of `README.md` honestly

jcazevedo:
> Running `docs/mdoc` I can't find the compilated version of `README.md` honestly

`docs/mdoc` will only build the Markdown files we use to generate PureConfig's website. You can either run `mdoc` on the root (which will build the docs of all modules) or `ip4s-docs/mdoc` (for just the ip4s docs). `ip4s-docs` is a synthetic project for the module's docs (the files you place under ip4s/docs).

geirolz:
>  If you really see value in it, what do you think of having pureconfig-http4s depending on pureconfig-ip4s? 

@ruippeixotog how should I explicitly this dependency? Do you mean module dependency or library dependency?

geirolz:
@jcazevedo @ruippeixotog if you can help me with the dependency we can merge this, can be very useful for us

leifwickland:
@geirolz A diff like this would do the job:
```
diff --git a/build.sbt b/build.sbt
index a51cefa7..67a33178 100644
--- a/build.sbt
+++ b/build.sbt
@@ -55,8 +55,8 @@ lazy val fs2 = module(project) in file("modules/fs2")
 lazy val generic = genericModule(project) in file("modules/generic") dependsOn `generic-base`
 lazy val `generic-base` = genericModule(project) in file("modules/generic-base")
 lazy val hadoop = module(project) in file("modules/hadoop")
-lazy val http4s = module(project) in file("modules/http4s")
 lazy val ip4s = module(project) in file("modules/ip4s")
+lazy val http4s = module(project) in file("modules/http4s") dependsOn ip4s
 lazy val javax = module(project) in file("modules/javax")
 lazy val joda = module(project) in file("modules/joda")
 lazy val magnolia = module(project) in file("modules/magnolia") dependsOn `generic-base`
 ```

You can verify that adds the dependency by doing a `publishLocal` and then look for `ip4s` in  `~/.ivy2/local/com.github.pureconfig/pureconfig-http4s_2.12/0.17.1-SNAPSHOT/poms/pureconfig-http4s_2.12.pom`.

**Comment by geirolz** [comment]
@leifwickland  thanks! 

@jcazevedo @ruippeixotog done 👍

File: build.sbt [hunk 3]
```diff
@@ -54,7 +54,8 @@ lazy val fs2 = module(project) in file("modules/fs2")
 lazy val generic = genericModule(project) in file("modules/generic") dependsOn `generic-base`
 lazy val `generic-base` = genericModule(project) in file("modules/generic-base")
 lazy val hadoop = module(project) in file("modules/hadoop")
-lazy val http4s = module(project) in file("modules/http4s")
+lazy val ip4s = module(project) in file("modules/ip4s")
+lazy val http4s = module(project) in file("modules/http4s") dependsOn ip4s
```
ruippeixotog: nit: please keep the list of modules sorted alphabetically.

```suggestion
lazy val http4s = module(project) in file("modules/http4s") dependsOn ip4s
lazy val ip4s = module(project) in file("modules/ip4s")
```

modules/ip4s/build.sbt
```diff
@@ -0,0 +1,20 @@
...
+
+crossScalaVersions := Seq(scala212, scala213)
+
+libraryDependencies ++= Seq("com.comcast" %% "ip4s-core" % "3.0.4")
+
+developers := List(
+  Developer("geirolz", "David Geirola", "david.geirolz@gmail.com", url("https://github.com/geirolz"))
+)
+
+osgiSettings
```
ruippeixotog: Can you please remove these keys? This won't merge cleanly with master as we dropped support for OSGi in #1202.

modules/ip4s/docs/README.md
```diff
@@ -0,0 +1,39 @@
...
+We can read a `MyConfig` with the following code:
+
+```scala mdoc:to-string
+val conf = parseString(
+  """{ 
+    |hostname: "0.0.0.0" 
+    |port: 8080 
+    |}""".stripMargin)
+    
+ConfigSource.fromConfig(conf).load[MyConfig]
```
ruippeixotog: With PureConfig we don't recommend using `ConfigFactory`; please prefer creating a `ConfigSource` directly from a string as suggested. You can also remove the `com.typesafe.config.ConfigFactory.parseString` import.

```suggestion
val source = ConfigSource.string(
  """{ 
    |hostname: "0.0.0.0" 
    |port: 8080 
    |}""".stripMargin)
    
source.load[MyConfig]
```

ruippeixotog (changes requested):
LGTM overall, there are just some last things to take care of. Thanks for addressing the changes so far, @geirolz :)

geirolz:
@ruippeixotog thanks! I've just rebased with master and applied your requests 👍

modules/ip4s/build.sbt
```diff
@@ -0,0 +1,11 @@
...
+
+name := "pureconfig-ip4s"
+
+crossScalaVersions := Seq(scala212, scala213)
+
+libraryDependencies ++= Seq("com.comcast" %% "ip4s-core" % "3.0.4")
+
+developers := List(
+  Developer("geirolz", "David Geirola", "david.geirolz@gmail.com", url("https://github.com/geirolz"))
+)
```
ruippeixotog: nit: please end file with a newline.

ruippeixotog (approved):
LGTM! I'll let @jcazevedo give a final review before merging.

modules/ip4s/build.sbt
geirolz: Done 👍

modules/ip4s/docs/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+```scala mdoc:to-string
+val source = ConfigSource.string(
+  """{ 
+    |hostname: "0.0.0.0" 
+    |port: 8080 
+    |}""".stripMargin)
+    
+source.load[MyConfig]
+```
```
jcazevedo: Unfortunately, `source` doesn't have a stable `toString` representation, so the evaluated output will be different every time we build the docs. Can we replace the first part by a silent mdoc block?

jcazevedo (changes requested):
@geirolz: Thanks for this! There's only one small issue with the docs. The rest looks good to me.

modules/ip4s/docs/README.md
geirolz: Done 👍

jcazevedo: Thanks! We could've just silenced the `ConfigSource.string` call and kept the output of `source.load[MyConfig]`, but I suppose this is also OK.

jcazevedo (approved):

geirolz:
@jcazevedo when you plan to release this?

jcazevedo:
@geirolz: I can cut a release this weekend that includes the ip4s module.

jcazevedo:
@geirolz: PureConfig 0.17.1, which includes pureconfig-ip4s, should now be available in Sonatype.

PR #1203, opened 2021-10-31
Review Scala 3 publishing settings for better compatibility

jcazevedo:
Following the [Scala 3.1.0 release announcement](https://scala-lang.org/blog/2021/10/21/scala-3.1.0-released.html), the official recommendation is to publish libraries with Scala 3.0, while still testing them with Scala 3.1:

> If you are a library maintainer, updating to 3.1.0 will force all of your users to update to 3.1.0 as well. We understand that the current state of binary compatibility may be unsatisfactory. We are actively working on technical solutions to support forward-compatibility in 3.2.0. In the meantime, we recommend testing your library with Scala 3.1.0 and 3.0.2, but publishing it with 3.0.2. This will allow downstream users to use your library even if they can’t update to Scala 3.1.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,12 +9,14 @@ jobs:
       fail-fast: false
       matrix:
         jdk: [adopt@1.8]
-        scala: ['2.12', '2.13', '3.1']
+        scala: ['2.12', '2.13', '3.0', '3.1']
         include:
           - scala: '2.12'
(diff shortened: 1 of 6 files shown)


ruippeixotog (approved):
Looks good @jcazevedo, I'm glad there was an official recommendation from the Scala team for library maintainers.

leifwickland:
Thanks for doing this, @jcazevedo!

PR #1215, opened 2021-11-14
Support Scala 3 in cats-effect and cats-effect2

limansky:
Hi,

  I've added Scala 3 support for cats-effect and cats-effect2 modules. However I wasn't able to add it in cats module, because of `for3Use2_13` dependencies in the testkit module. Also I faced with a weird behavior of crossbuild, when sbt tries to build for scala3 *-docs modules, even though they doesn't have Scala 3 enabled in their settings. I'll try to improve this PR, to add Scala 3 in cats as well, or I can create another one, since it looks like it will required significant changes in build.

### DIFF
--- a/modules/cats-effect/build.sbt
+++ b/modules/cats-effect/build.sbt
@@ -3,7 +3,7 @@ import Utilities._
 
 name := "pureconfig-cats-effect"
 
-crossScalaVersions := Seq(scala212, scala213)
+crossScalaVersions := Seq(scala212, scala213, scala30)
 
 libraryDependencies ++= Seq(
   "org.typelevel" %% "cats-effect" % "3.2.9"
--- a/modules/cats-effect/build.sbt
+++ b/modules/cats-effect/build.sbt
@@ -3,7 +3,7 @@ import Utilities._
 
 name := "pureconfig-cats-effect"
 
-crossScalaVersions := Seq(scala212, scala213, scala30)
+crossScalaVersions := Seq(scala212, scala213, scala30, scala31)
 
 libraryDependencies ++= Seq(
   "org.typelevel" %% "cats-effect" % "3.2.9"
--- a/modules/cats-effect/src/test/scala/pureconfig/module/catseffect/CatsEffectSuite.scala
+++ b/modules/cats-effect/src/test/scala/pureconfig/module/catseffect/CatsEffectSuite.scala
@@ -14,6 +14,11 @@ import pureconfig.{BaseSuite, ConfigReader, ConfigSource, ConfigWriter}
 final class CatsEffectSuite extends BaseSuite {
 
   case class SomeCaseClass(somefield: Int, anotherfield: String)
+
+  object SomeCaseClass {
+    def unapply(c: SomeCaseClass): Option[(Int, String)] = Some((c.somefield, c.anotherfield))
+  }
+
   implicit val reader: ConfigReader[SomeCaseClass] =
     ConfigReader.forProduct2("somefield", "anotherfield")(SomeCaseClass.apply)
   implicit val writer: ConfigWriter[SomeCaseClass] =
@@ -31,7 +36,7 @@ final class CatsEffectSuite extends BaseSuite {
   }
 
   "ConfigSource#loadF" should "fail when the expected file is not present" in {
-    val load = ConfigSource.default.at("non-existent-namespace").loadF[IO, SomeCaseClass]
+    val load = ConfigSource.default.at("non-existent-namespace").loadF[IO, SomeCaseClass]()
 
     a[ConfigReaderException[SomeCaseClass]] should be thrownBy load.unsafeRunSync()
   }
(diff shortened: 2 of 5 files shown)


limansky:
BTW, I don't understand if build for 3.1 is really required. As I understand the Scala suffix is the same (_3), so it will produce the same artifact name as Scala 3.0. However it will not be compatible with Scala 3.0.x. Hence, it looks like only Scala 3.0.x build is needed.

modules/cats-effect/build.sbt
```diff
@@ -3,7 +3,7 @@ import Utilities._
 
 name := "pureconfig-cats-effect"
 
-crossScalaVersions := Seq(scala212, scala213)
+crossScalaVersions := Seq(scala212, scala213, scala30)
```
jcazevedo: We're currently following Scala team's recommendation (see https://scala-lang.org/blog/2021/10/21/scala-3.1.0-released.html#compatibility-notice) of publishing libraries with Scala 3.0 while still attempting to build them using the latest Scala 3.x version. See https://github.com/pureconfig/pureconfig/pull/1203 for more context. 

As such, we should also be compiling this for Scala 3.1.

```suggestion
crossScalaVersions := Seq(scala212, scala213, scala30, scala31)
```

modules/cats-effect/src/test/scala/pureconfig/module/catseffect/CatsEffectSuite.scala
```diff
@@ -14,6 +14,11 @@ import pureconfig.{BaseSuite, ConfigReader, ConfigSource, ConfigWriter}
 final class CatsEffectSuite extends BaseSuite {
 
   case class SomeCaseClass(somefield: Int, anotherfield: String)
+
+  object SomeCaseClass {
+    def unapply(c: SomeCaseClass): Option[(Int, String)] = Some((c.somefield, c.anotherfield))
+  }
```
jcazevedo: Instead of creating an `unapply` implementation that attempts to mimic Scala 2's behavior in Scala 3, I think it's preferable to define the `ConfigWriter` for `SomeCaseClass` in a way that is compatible with both Scala 2 and 3. Something like the following should do it:

```scala
implicit val writer: ConfigWriter[SomeCaseClass] =
  ConfigWriter.forProduct2("somefield", "anotherfield") { case SomeCaseClass(somefield, anotherfield) =>
    (somefield, anotherfield)
  }
```

modules/cats-effect2/build.sbt
```diff
@@ -3,7 +3,7 @@ import Utilities._
 
 name := "pureconfig-cats-effect2"
 
-crossScalaVersions := Seq(scala212, scala213)
+crossScalaVersions := Seq(scala212, scala213, scala30)
```
jcazevedo: Same as above.

```suggestion
crossScalaVersions := Seq(scala212, scala213, scala30, scala31)
```

modules/cats-effect2/src/test/scala/pureconfig/module/catseffect2/CatsEffectSuite.scala
```diff
@@ -16,14 +16,18 @@ import pureconfig.{BaseSuite, ConfigReader, ConfigSource, ConfigWriter}
 final class CatsEffectSuite extends BaseSuite {
 
   case class SomeCaseClass(somefield: Int, anotherfield: String)
+  object SomeCaseClass {
+    def unapply(c: SomeCaseClass): Option[(Int, String)] = Some((c.somefield, c.anotherfield))
+  }
```
jcazevedo: As with the `CatsEffectSuite` of pureconfig-cats-effect, I think we can avoid this by defining the `ConfigWriter` differently.

jcazevedo (changes requested):
Thanks for your PR!

> Also I faced with a weird behavior of crossbuild, when sbt tries to build for scala3 *-docs modules, even though they doesn't have Scala 3 enabled in their settings.

Did you have those issues by running `+mdoc`?

modules/cats-effect/src/test/scala/pureconfig/module/catseffect/CatsEffectSuite.scala
limansky: Fixed

modules/cats-effect2/src/test/scala/pureconfig/module/catseffect2/CatsEffectSuite.scala
limansky: Done

limansky:
> Did you have those issues by running +mdoc?

No, the issues were on resolve stage already. I think it was because I removed cross-versioning in `testkit/build.sbt`, to fix incompatibilities in cats module.

jcazevedo (approved):
Thanks! Let's handle the support for Scala 3 in pureconfig-cats in a different PR.

PR #1220, opened 2021-11-27
Support Scala 3 in cats module

limansky:
I've solved issues with building `cats` module for Scala 3. The only thing I'm not sure if `publishMavenStyle` should be kept set false.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -70,6 +70,10 @@ lazy val `zio-config` = module(project) in file("modules/zio-config")
 // Workaround for https://github.com/scalacenter/scalafix/issues/1488
 val scalafixCheckAll = taskKey[Unit]("No-arg alias for 'scalafixAll --check'")
 
+val minScala3Version = settingKey[Option[(Long, Long)]]("Minimal Scala 3 version supported")
+
+Global / excludeLintKeys += minScala3Version
+
 lazy val commonSettings = Seq(
   // format: off
   homepage := Some(url("https://github.com/pureconfig/pureconfig")),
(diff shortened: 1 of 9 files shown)


limansky:
I've added Scala 3 to fs2, http4s, ip4s and sttp as well. All of these modules are published using scala 3.1, so I added Scala 3.1 as a compiler. To make them publishing, I created a setting minScala3Version and set `publish / skip` according to its value.

limansky:
JFYI https://github.com/Comcast/ip4s/issues/341

limansky:
I have no idea why sbt tries to build modules supporting Scala 3.1 only with Scala 3.0. It works fine when runnig `+compile` or `+test`, but fails if the version is specified via `++`. I'm trying to figure out what's going on, but it looks like I've stuck with it.

pslaski:
What is the status of this PR? Are you planning publishing pureconfig modules for scala 3?

ruippeixotog:
@limansky I believe this PR is failing because of https://github.com/sbt/sbt/issues/6915 - SRT isn't dealing correctly with Scala 3 in cross-compilation and projects that only support Scala 3.1 are being complied with Scala 3.0.x too.

limansky:
@ruippeixotog yep, it seems like the root cause is the issue in sbt. If it acceptable for pureconfig team, we can just use Scala 3.1 for all Scala 3 modules.

ruippeixotog:
No, that's not ok since it will force every library that depends on PureConfig to upgrade to Scala 3.1 as well (even if they are not using any of the affected modules). I'd say we can wait for SBT to be fixed and released before we merge this.

ruippeixotog:
Superseded by #1352.

PR #1227, opened 2021-12-06
Enable scalafix on scala 3

mlachkar:
This pull request enable scalafix on Scala3. 
I have also organised a little the build, I can remove this from my change. 
Thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,12 +29,10 @@ jobs:
         with:
           java-version: ${{ matrix.jdk }}
 
-      - name: Check formatting
+      - name: Check formatting and scalafmt
         run: sbt "++${{ matrix.scala-version }} scalafmtCheckAll" scalafmtSbtCheck
 
-        # Consider joining this action with the "Check formatting" one once Scalafix is supported for Scala 3.
       - name: Run linter
-        if: startsWith(matrix.scala, '2')
         run: sbt "++${{ matrix.scala-version }} scalafixCheckAll"
 
       - name: Compile
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
         with:
           java-version: ${{ matrix.jdk }}
 
-      - name: Check formatting and scalafmt
+      - name: Check formatting
         run: sbt "++${{ matrix.scala-version }} scalafmtCheckAll" scalafmtSbtCheck
 
       - name: Run linter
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -29,13 +29,8 @@ jobs:
...
 
-      - name: Check formatting
-        run: sbt "++${{ matrix.scala-version }} scalafmtCheckAll" scalafmtSbtCheck
-
-        # Consider joining this action with the "Check formatting" one once Scalafix is supported for Scala 3.
-      - name: Run linter
-        if: startsWith(matrix.scala, '2')
-        run: sbt "++${{ matrix.scala-version }} scalafixCheckAll"
+      - name: Check formatting and Scalafix
+        run: sbt "++${{ matrix.scala-version }}" scalafmtCheckAll scalafmtSbtCheck scalafixCheckAll
```
ruippeixotog: The difference in quotes is subtle, but significant. Here we actually want both `scalafmtCheckAll` and `scalafixCheckAll` in the same command as the version change - see my explanation about the differences between them in #1188. Can you please change this to either:

- `sbt "++${{ matrix.scala-version }} scalafmtCheckAll scalafixCheckAll" scalafmtSbtCheck`
- `sbt "++${{ matrix.scala-version }} scalafmtCheckAll" "++${{ matrix.scala-version }} scalafixCheckAll" scalafmtSbtCheck`

(I don't know if the first version works as is, so please test before submitting)

ruippeixotog (changes requested):
I'm sorry for the long turnaround time, @mlachkar! Thanks for creating this pull request 🙂 

This LGTM overall, but the build reorganization makes it a bit difficult to understand and I personally prefer using the top-level scope to define `ThisBuild` settings rather than wrapping them in `inThisBuild`. Looks like the project changes also caused CI errors.


Do you mind reverting that part and leave only the Scala 3 enabling? Thanks again!

.github/workflows/ci.yml
mlachkar: The second one works!

mlachkar:
I kept the linting test and formatting test separated even if the second command runs correctly!
I think it's interesting to be able to figure out quickly if it's a formatting issue or scalafix one, but I can change this PR to do that `sbt "++${{ matrix.scala-version }} scalafmtCheckAll" "++${{ matrix.scala-version }} scalafixCheckAll" scalafmtSbtCheck`.

mlachkar:
By the way, pureconfig is one of my favourite libraries! Thanks for your work!

ruippeixotog (approved):
Thanks again!

PR #1260, opened 2022-01-22
[Proposal] Add secret

geirolz:
I've opened this PR to share my solution about secret configs. In every microservice, I've copied and pasted this class `Secret` which is "transparent" to Pureconfig but since you have a wrapper type you can change some behaviours like `toString` or JSON encoding. 

In every project after the config load I print it during the startup, In this way, I can avoid the printing of secrets.
So, I wonder if this solution can be part of PureConfig. I've tried to add it with this PR.
I guess I'm not the only one with this "problem", in this way you can provide an out of the box solution.

Let me know what do you think about it 👍

### DIFF
--- a/core/src/main/scala/pureconfig/Secret.scala
+++ b/core/src/main/scala/pureconfig/Secret.scala
@@ -1,0 +1,33 @@
+package pureconfig
+
+import pureconfig.Secret.SecretBytes
+
+import java.nio.charset.{Charset, StandardCharsets}
+
+case class Secret[+T](value: T) extends AnyVal {
+
+  def map[U](f: T => U): Secret[U] =
+    Secret(f(value))
+
+  override final def toString: String = Secret.placeHolder
+}
+object Secret extends SecretSyntax {
+
+  type SecretBytes = Secret[Array[Byte]]
+
+  val placeHolder: String = "** MASKED **"
+
+  implicit def secretConfigReader[T: ConfigReader]: ConfigReader[Secret[T]] =
+    ConfigReader[T].map(n => Secret(n))
+
+  implicit def secretConfigWriter[T: ConfigWriter]: ConfigWriter[Secret[T]] =
+    ConfigWriter[T].contramap(_.value)
+}
+
+sealed trait SecretSyntax {
+
+  implicit class SecretBytesOps(sb: SecretBytes) {
+    def stringValue(cs: Charset = StandardCharsets.UTF_8): String =
+      new String(sb.value, cs)
+  }
+}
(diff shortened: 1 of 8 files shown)


leifwickland (commented):
@ruippeixotog @jcazevedo How do you feel about this?

I like this proposal. I generally have something similar in my projects. However, I have a vague recollection that something like this has previously been suggested and rejected. I don't recall why.

geirolz:
@leifwickland If not in the core module a dedicated module would be ok as well. For sure in core would be better IMHO

ruippeixotog:
Hey @leifwickland, @geirolz! I'm sorry for the delay in reviewing this. I remember a concept like this `Secret` being brought before and I remember that you added this to the FAQ, @leifwickland (https://pureconfig.github.io/docs/faq.html#how-can-i-avoid-displaying-sensitive-parts-of-my-configuration-such-as-passwords-when-i-print-the-result-of-pureconfig).

I'm finding it very hard to justify this being in PureConfig core:
- `pureconfig-core` contains all type classes, error models and utilities essential to provide config reading and writing. This would be the first data model there, which usually sits one layer above core.
- This case class is not specific to PureConfig, and even to configs, at all. A wrapper that changes the behavior of `toString` is a pretty generic thing and it could be used on many other use cases, such as user profile data models, internal crypto secrets, classes for a transport layer like GRPC... Such a generic utility could easily be in their own library without a direct PureConfig dependency (any user using a PureConfig generic library would get config readers/writers for free without this PR).
- I find the specific use case that we're tackling - printing a config to a sensitive medium - not very common. On my projects I never personally found the need to dump a full config as they are either encoded in env vars of a Docker (following [12 factor app](https://12factor.net/config) recommendations) or kept as a frozen version with the binary somewhere static. I don't know how frequent this happens on other code bases, but it's fair to say it's a thing many `pureconfig-core` users don't need.

I'm ok with having it on a separate module, as on modules we prioritize having many use cases covered rather than keeping their number small - though in this case I'd recommend keeping a completely separate library that can be extended with integration modules itself 🙂

geirolz:
@ruippeixotog thanks for your explanation!  

- I would personally put it in core ( other libraries like [Ciris](https://github.com/vlovgr/ciris/blob/master/modules/core/src/main/scala/ciris/Secret.scala) are offering `Secret` as a data type in the core ). But I understand your motivation.
-  It's generic for sure but, the purpose is not just to change `toString` behaviour but to provide a box to data in order to prevent unwanted prints in a type-safe way.
- I don't know actually if this is only our problem or it's common, @leifwickland I think did something similar. I guess it's not so infrequent in microservice architectures ( based on my experience for sure ). In any case, having an out-of-the-box system to prevent secret printing is not so bad IMHO.

It would be nice if this case is covered by Pureconfig ( core or dedicated module ) even to guarantee support and popularity to this feature. A dedicated library for this IMHO is overkill since it's not providing so many features but I'll think about it 😄

leifwickland:
I'm fine with having a separate lib for Secret. I don't think it's ideal,
but since this topic keeps coming up, it may be the least bad compromise.

PR #1294, opened 2022-03-14
Add `ConfigReader` and `ConfigWriter` for `Uri.Path` in akka-http

duartepinto:


ruippeixotog (approved):
Thanks for adding this, @duartepinto!

PR #1328, opened 2022-06-03
Support Scala 3 in http4s module

mhodovaniuk:
Added Scala 3 cross-compilation to http4s module.

### DIFF
--- a/modules/http4s/build.sbt
+++ b/modules/http4s/build.sbt
@@ -2,7 +2,7 @@ import Dependencies.Version._
 
 name := "pureconfig-http4s"
 
-crossScalaVersions := Seq(scala212, scala213)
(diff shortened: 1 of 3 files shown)


ruippeixotog:
@mhodovaniuk I believe this PR is failing because of https://github.com/sbt/sbt/issues/6915 - SRT isn't dealing correctly with Scala 3 in cross-compilation and projects that only support Scala 3.1 are being complied with Scala 3.0.x too.

mhodovaniuk:
Now it makes sense, why it tries to build with 3.0.x. @ruippeixotog thanks a lot!

ruippeixotog (approved):
Looks good, thanks for updating the PR!

PR #1352, opened 2022-07-24
Start compiling and publishing PureConfig against Scala 3.1

ruippeixotog:
The latest scalatest version (3.2.12) is compiled against Scala 3.1, forcing us and many other projects that depend on Scala 3 to Scala 3.1 as well. Given how widely used scalatest is, I think it's time for us to drop Scala 3.0 as well and simplify our build assumptions (for now). Compared to scalatest, this PureConfig change has a much smaller blast radius (I suspect most PureConfig users are end projects, not libraries), so I think it won't have a lot of consequences.

On this PR I'm dropping Scala 3.0 and combining a lot of existing open PRs that depend on this, such as bumping scalatest, using SBT 1.7.x and a few other build nits.

Supersedes #1348, #1317 and #1220.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,16 +9,14 @@ jobs:
       fail-fast: false
       matrix:
         jdk: [adopt@1.8]
-        scala: ['2.12', '2.13', '3.0', '3.1']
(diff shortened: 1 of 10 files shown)


jcazevedo (approved):
Thanks!

PR #1365, opened 2022-08-18
Fix spark-sql tests

ruippeixotog:
For some time now, PureConfig spark-sql tests have been flaky with due to an incorrect scalacheck `Gen` implementation. I'm fixing it in this PR by ensuring generated struct fields always have valid identifiers.


jcazevedo (approved):
Thanks!

PR #1389, opened 2022-10-06
Support for Either[A, B]

eadgbear:
Adding support for `Either[A,B]`. For example:

```scala
case class Foobert(
  maybeList: Either[String, List[String]]
)

val x = ConfigSource.string("maybe-list = a").loadOrThrow[Foobert]
assert(x.maybeList.isLeft)
val y = ConfigSource.string("maybe-list = [a, b, c]").loadOrThrow[Foobert]
assert(y.maybeList.isRight)
```

### DIFF
--- a/core/src/main/scala/pureconfig/CollectionReaders.scala
+++ b/core/src/main/scala/pureconfig/CollectionReaders.scala
@@ -26,13 +26,13 @@ trait CollectionReaders {
   implicit def eitherReader[A, B](implicit convA: ConfigReader[A], convB: ConfigReader[B]): ConfigReader[Either[A, B]] =
     new ConfigReader[Either[A, B]] {
       override def from(cur: ConfigCursor): ConfigReader.Result[Either[A, B]] = {
-        convA.from(cur) match {
-          case Left(aErr) ⇒
-            convB.from(cur) match {
-              case Left(bErr) ⇒ Left(aErr ++ bErr)
-              case Right(bType) ⇒ Right(Right[A, B](bType))
+        convB.from(cur) match {
+          case Left(bErr) ⇒
+            convA.from(cur) match {
+              case Left(aErr) ⇒ Left(bErr ++ aErr)
+              case Right(aType) ⇒ Right(Left[A, B](aType))
             }
-          case Right(aType) ⇒ Right(Left[A, B](aType))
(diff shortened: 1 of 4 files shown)


core/src/main/scala/pureconfig/CollectionReaders.scala
```diff
@@ -23,6 +23,20 @@ trait CollectionReaders {
       }
     }
 
+  implicit def eitherReader[A, B](implicit convA: ConfigReader[A], convB: ConfigReader[B]): ConfigReader[Either[A, B]] =
+    new ConfigReader[Either[A, B]] {
+      override def from(cur: ConfigCursor): ConfigReader.Result[Either[A, B]] = {
+        convA.from(cur) match {
```
arixmkii: Should it be right biased? That we first try if we can read as right value and only after fall back to left one..

eadgbear: Fair point! Addressed this in a change.

ruippeixotog:
Hey there! Sorry it took so long to review this.

I believe we considered adding readers for `Either` before, but ultimately decided against it for the same reasons as [Circe did](https://github.com/circe/circe/issues/751): depending on the types you choose, this reader may behave in unexpected ways and we can't expect a write-read roundtrip to yield the same results.

`ConfigReader[Either[T, T]]` is the simplest example, where as you discussed it's unclear if users want a left-biased reader or a right-biased one (right bias is often used for success-failure representations, but there are plenty of balanced use cases as well). Weirder cases can happen if the types are different but still accept the same representations, though. For example, if we decided to go with a right-biased reader, our implementation of `ConfigReader[Either[Int, String]]` would read a config like `field = 42` as `Right("42")`, not `Left(42)` as most people would expect! A `ConfigReader[String]` is able to parse any primitive value, so it will always work for numbers or booleans.

It's actually very simple to create a `ConfigReader[Either[A, B]]` for everyone that wants this behavior; it can be defined in a one liner as:

```scala
implicit val eitherReader: ConfigReader[Either[String, Int]] = 
  ConfigReader[Int].map(Right.apply) orElse ConfigReader[String].map(Left.apply)
```

Given how simple it is to provide this behavior, I'd rather have users express their intent explicitly rather than having a general `Either` reader that may be misused.

ruippeixotog:

PR #1403, opened 2022-10-24
Cross-compile Cats, FS2 for Scala 3

armanbilge:
Thanks!

### DIFF
--- a/modules/fs2/src/test/scala/pureconfig/module/fs2Suite.scala
+++ b/modules/fs2/src/test/scala/pureconfig/module/fs2Suite.scala
@@ -26,7 +26,9 @@ class fs2Suite extends AnyFlatSpec with Matchers {
   implicit val reader: ConfigReader[SomeCaseClass] =
     ConfigReader.forProduct2("somefield", "anotherfield")(SomeCaseClass.apply)
   implicit val writer: ConfigWriter[SomeCaseClass] =
-    ConfigWriter.forProduct2("somefield", "anotherfield")(SomeCaseClass.unapply(_).get)
+    ConfigWriter.forProduct2("somefield", "anotherfield") { case x @ SomeCaseClass(_, _) =>
+      x
+    }
 
   "streamConfig" should "load a case class from a byte stream" in {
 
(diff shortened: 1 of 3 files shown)


modules/fs2/src/test/scala/pureconfig/module/fs2Suite.scala
```diff
@@ -26,7 +26,7 @@ class fs2Suite extends AnyFlatSpec with Matchers {
   implicit val reader: ConfigReader[SomeCaseClass] =
     ConfigReader.forProduct2("somefield", "anotherfield")(SomeCaseClass.apply)
   implicit val writer: ConfigWriter[SomeCaseClass] =
-    ConfigWriter.forProduct2("somefield", "anotherfield")(SomeCaseClass.unapply(_).get)
+    ConfigWriter.forProduct2("somefield", "anotherfield")(x => (x.somefield, x.anotherfield))
```
ruippeixotog, line 29: Out of curiosity, why did you have to change this to be an explicit application? Is there a way to keep using `unapply`?

ruippeixotog (approved):
Thanks for the contribution, @armanbilge!

modules/fs2/src/test/scala/pureconfig/module/fs2Suite.scala
armanbilge: You can keep using `unapply` in both Scala 2 and Scala 3, but the code will not cross-compile across Scalas. This is because in Scala 2 it returns an `Option` but in Scala 3 it is "option-less" .

https://docs.scala-lang.org/scala3/reference/changed-features/pattern-matching.html

jcazevedo (approved):

armanbilge:
@ruippeixotog @jcazevedo would it be possible to get this into a release sometime soon? Or do you happen to publish snapshots (I couldn't see any in CI). Thanks!

ruippeixotog:
@armanbilge I can try to cut a release this weekend.

armanbilge:
@ruippeixotog thanks! Would it be helpful at all if I set up this repository for CI release? You would have to install your sonatype credentials as secrets but then releasing can be easy.

See: https://typelevel.org/sbt-typelevel/secrets.html

armanbilge:
Thank you for releasing!!

Issue #1086, opened 2021-06-30
Upgrade http4s module from 0.21 to 0.23


Ali-Mahmood:
Not upgrading will prevent projects who use pure config from upgrading to http4s 0.23, example below:

`[error]         * org.http4s:http4s-core_2.13:0.23.0-RC1 (early-semver) is selected over 0.21.24
[error]             +- org.http4s:http4s-dsl_2.13:0.23.0-RC1 ()           (depends on 0.23.0-RC1)
[error]             +- org.http4s:http4s-server_2.13:0.23.0-RC1 ()        (depends on 0.23.0-RC1)
[error]             +- org.http4s:http4s-blaze-core_2.13:0.23.0-RC1 ()    (depends on 0.23.0-RC1)
[error]             +- org.http4s:http4s-client_2.13:0.23.0-RC1 ()        (depends on 0.23.0-RC1)
[error]             +- com.github.pureconfig:pureconfig-http4s_2.13:0.16.0 (depends on 0.21.24)
`

ruippeixotog:
Hi! We'll likely wait for a stable version to be released before the module is upgraded. In the meantime, you can avoid that error by temporarily removing pureconfig-http4s from your dependencies. That module contains only an `Uri` converter, so it should be easy to replace.

Ali-Mahmood:
Hi! I see, that sounds good. Ill do that for now, thank you!

augi:
It would be nice to release also a version of pureconfig depending on http4s `0.22` in the meantime 🙏

satorg:
@ruippeixotog just fyi: looks like this issue can be closed since the v0.23 is supported since pureconfig v0.17.0, isn't it?

satorg:
> It would be nice to release also a version of pureconfig depending on http4s 0.22 in the meantime 🙏

Totally agree :) Would a PR with a new pureconfig-http4s-022 module be accepted if I go ahead and create it? Does not seem to be something tremendously difficult, but will make lives of some folks easier.

ruippeixotog:
Hey @satorg, you're right, `pureconfig-http4s` is currently using v0.23, we can close this 👍 

The goal of releasing PureConfig depending on 0.22 was just a temporary workaround while we didn't have 0.21 IIUC, so there shouldn't be any need to create additional forks, right?

satorg:
@ruippeixotog thank you for looking into it.

[...]

PR #1421, opened 2022-11-19
Add http4s022 compatibility module

satorg:
Based on this discussion: https://github.com/pureconfig/pureconfig/issues/1086#issuecomment-885428880

Adds support for the legacy **http4s v0.22** by introducing the **pureconfig-http4s022** module.
This might be useful for projects that are still stuck to Cats Effect 2 and therefore cannot migrate to the newer **http4s v0.23+** right away.

The introduced module is simply a copy of **pureconfig-http4s** but with a downgraded dependency to the older **http4s v0.22.x** library. Most likely it is going to be used along with the other compatibility module **pureconfig-cats-effect2**.


build.sbt
```diff
@@ -62,6 +62,7 @@ lazy val generic = genericModule(project) in file("modules/generic") dependsOn `
 lazy val `generic-base` = genericModule(project) in file("modules/generic-base")
 lazy val hadoop = module(project) in file("modules/hadoop")
 lazy val http4s = module(project) in file("modules/http4s") dependsOn ip4s
+lazy val http4s022 = module(project) in file("modules/http4s022")
```
satorg, line 65: removed dependency on `ip4s` since it is not necessary but brings a newer [ip4s](https://github.com/Comcast/ip4s) which is not compatible with **http4s v0.22.x**.

ruippeixotog: If it's not necessary, can you please remove it from `http4s` as well to keep both modules in sync? This can also make it easier for users to switch between `http4s` and `http4s022` without hitting any unexpected errors (e.g. someone relying on `ip4s` through `http4s` that wants to move to `http4s022`).

ruippeixotog (changes requested):
LGTM overall @satorg, thanks for taking the time to push this through!

I left just a small nit; let me know if it's fine by you to make that change.

build.sbt
satorg: @ruippeixotog done, check it out please. Also just to doublecheck with you: I think this can be a potentially breaking change because there can be projects around that may have `ip4s` as a transient dependency via `http4s` module. But if you think it is not a big deal, then I'm fine with it too. Personally, I think it is always better to declare required dependencies explicitly after all.

ruippeixotog (approved):

satorg:
The CI failed with 
```
At docs/target/site/index.html:7:

  External link https://maven-badges.herokuapp.com/maven-central/com.github.pureconfig/pureconfig_2.12/badge.svg failed: got a time out (response code 0) (status code 0)
```
perhaps, some GHW flakes?

ruippeixotog:
Just a transient error, it seems. Merging now, thanks again!

PR #1444, opened 2023-01-08
Update jekyll and html-proofer CI checks

ruippeixotog:
CI checks are currently failing on all PRs because of errors with jekyll and htmlproofer. I haven't dug deep into root causes, but:

- htmlproofer is failing installation because it's failing to build some native gem extensions;
- Jekyll seems to have some incompatibility with deprecated Ruby APIs (https://github.com/jekyll/jekyll/issues/9231).

In both cases, updating versions fixed all issues. The latest html-proofer version added a few more checks to its default configuration and they seemed worth paying attention to, so this PR also includes some minor website link changes.


jcazevedo (approved):
Thanks for this!

PR #1459, opened 2023-02-01
Build modules/enumeratum for scala3

froth:
Had to rename some variables as `enum` is a reserved keyword in scala3.

### DIFF
--- a/modules/enumeratum/build.sbt
+++ b/modules/enumeratum/build.sbt
@@ -1,9 +1,20 @@
 import Dependencies.Version._
+import Utilities.forScalaVersions
 
 name := "pureconfig-enumeratum"
 
-crossScalaVersions := Seq(scala212, scala213)
+crossScalaVersions := Seq(scala212, scala213, scala3)
 
 libraryDependencies ++= Seq("com.beachape" %% "enumeratum" % "1.7.2")
 
 developers := List(Developer("aeons", "Bjørn Madsen", "bm@aeons.dk", url("https://github.com/aeons")))
+
+lazy val extraScalacOptions = forScalaVersions {
+  case (3, _) =>
+    List(
+      "-Yretain-trees" // needed for using ValueEnums on scala3 https://github.com/lloydmeta/enumeratum/blob/master/README.md?plain=1#L28
+    )
+  case _ => List.empty
+}
+
+scalacOptions ++= extraScalacOptions.value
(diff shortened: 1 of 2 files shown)


modules/enumeratum/build.sbt
```diff
@@ -1,9 +1,17 @@
...
 
-crossScalaVersions := Seq(scala212, scala213)
+crossScalaVersions := Seq(scala212, scala213, scala3)
 
 libraryDependencies ++= Seq("com.beachape" %% "enumeratum" % "1.7.2")
 
 developers := List(Developer("aeons", "Bjørn Madsen", "bm@aeons.dk", url("https://github.com/aeons")))
+
+lazy val extraScalacOptions = forScalaVersions {
+  case (3, _) => List("-Yretain-trees")
```
ruippeixotog: Why do we need this here? If it's is really needed, can you please add a comment with an explanation for it?

modules/enumeratum/src/main/scala/pureconfig/module/enumeratum/package.scala
```diff
@@ -11,47 +11,47 @@ import pureconfig.error.CannotConvert
 
 package object enumeratum {
 
-  implicit def enumeratumConfigConvert[A <: EnumEntry](implicit enum: Enum[A], ct: ClassTag[A]): ConfigConvert[A] =
-    viaNonEmptyStringOpt[A](enum.withNameOption, _.entryName)
+  implicit def enumeratumConfigConvert[A <: EnumEntry](implicit `enum`: Enum[A], ct: ClassTag[A]): ConfigConvert[A] =
```
ruippeixotog: Rather than quoting all the local variables named "enum" here, why don't you rename it to something else, like `e` or `en`? It leaves the code cleaner and closer to the way one would write it for Scala 3.

ruippeixotog (changes requested):
Thanks for the PR, @froth! LGTM overall, just a few comments.

modules/enumeratum/src/main/scala/pureconfig/module/enumeratum/package.scala
froth: I agree. Done :)

modules/enumeratum/build.sbt
froth: Without it the compiler complains that one should enable "-Yretain-trees". I added a comment linking to the enumeratum documentation.

froth:
Thanks for the review @ruippeixotog, I changed the parts you commented on.

ruippeixotog (approved):
Thanks again!

PR #1461, opened 2023-02-05
Adds more config readers and writers for ip4s models

satorg:
This PR fills in the whole hierarchy of `Host` with all its descendants:

- `Host` (_added_)
    - `Hostname` (existing)
    - `IpAddress` (_added_)
        - `Ipv4Address` (_added_)
        - `Ipv6Address` (_added_)
    - `IDN` (_added_)

### DIFF
--- a/modules/ip4s/docs/README.md
+++ b/modules/ip4s/docs/README.md
@@ -23,7 +23,7 @@ libraryDependencies += "com.github.pureconfig" %% "pureconfig-ip4s" % "@VERSION@
 
 ## Example
 
-To load an `Host` or a `Port` into a configuration, create a class to hold it:
+To load a `Host` or a `Port` into a configuration, create a class to hold it:
 
 ```scala mdoc:silent
 import com.comcast.ip4s._
--- a/modules/ip4s/docs/README.md
+++ b/modules/ip4s/docs/README.md
@@ -1,8 +1,17 @@
 
 # Ip4s module for PureConfig
 
-Adds support for [Ip4s](https://github.com/Comcast/ip4s)'s Hostname and Port class to PureConfig. PRs adding support
-for other classes are welcome :)
+Adds support for [Ip4s](https://github.com/Comcast/ip4s)'s models to PureConfig.
+Currently the following models are supported:
+  - `Host`:
+    - `Hostname`
+    - `IpAddress`:
+      - `Ipv4Address`
+      - `Ipv6Address`
+    - `IDN`
+  - `Port`
+
+PRs adding support for other classes are welcome :)
 
 ## Add pureconfig-ip4s to your project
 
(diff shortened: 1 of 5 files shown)


modules/ip4s/docs/README.md
```diff
@@ -14,25 +23,28 @@ libraryDependencies += "com.github.pureconfig" %% "pureconfig-ip4s" % "@VERSION@
 
 ## Example
 
-To load an `Hostname` or a `Port` into a configuration, create a class to hold it:
+To load an `Host` or a `Port` into a configuration, create a class to hold it:
```
ruippeixotog: nit

```suggestion
To load a `Host` or a `Port` into a configuration, create a class to hold it:
```

modules/ip4s/src/main/scala/pureconfig/module/ip4s/package.scala
```diff
@@ -1,31 +1,58 @@
...
 import pureconfig.error.CannotConvert
 import pureconfig.{ConfigReader, ConfigWriter}
 
 package object ip4s {
 
+  private def mkConfigReader[From, To](f: From => Option[To])(
+      cname: String,
+      uname: String
+  )(implicit From: ConfigReader[From]): ConfigReader[To] = From.emap { from =>
+    f(from).toRight(CannotConvert(from.toString, cname, s"Invalid $uname"))
```
ruippeixotog: Instead of manually writing the class name of `To` every time, can you please make the compiler do that for you by requiring an implicit `ClassTag[To]`?

With a way to get the class name implicitly, I also question the value of asking for a human-readable name just to put in error messages. I'd much rather have an API that simply takes the reader function and handles the rest automatically.

modules/ip4s/src/test/scala/pureconfig/module/ip4s/Ip4sTest.scala
```diff
@@ -1,12 +1,134 @@
...
+    Gen.oneOf(
+      ipGenerator,
+      hostnameGenerator,
+      // `idnGenerator` sometimes produces values that can be parsed as a valid `Hostname`.
+      // However, `Hostname` has a higher priority over `IDN` in `Host.fromString`.
+      // Therefore filter such values out.
+      idnGenerator.filter(idn => Hostname.fromString(idn.toString).isEmpty)
+    )
+
+  private def testsFor[A: ConfigWriter: ConfigReader, B: Arbitrary](
```
ruippeixotog: This is quite a lot of custom test code - have you checked the utilities PureConfig provides in [ConfigConvertChecks](https://github.com/pureconfig/pureconfig/blob/master/testkit/src/main/scala/pureconfig/ConfigConvertChecks.scala) (available though `BaseSuite`)? Can we try to reuse them instead of having to maintain this alternative?

ruippeixotog (changes requested):
Apologies for taking so long to review this, @satorg. This looks good overall, I just left a few comments for your consideration.

satorg:
@ruippeixotog , thank you for your feedback! I'll be back to this PR soon (hopefully).

modules/ip4s/docs/README.md
satorg: fixed

modules/ip4s/src/main/scala/pureconfig/module/ip4s/package.scala
satorg: agree. fixed.

modules/ip4s/src/test/scala/pureconfig/module/ip4s/Ip4sTest.scala
satorg: Thank you for the hint – I've checked it out.

I found that I can make use of the `checkArbitrary` method in the tests and so did. However, it looks like there are no methods in `ConfigConvertChecks` that can help to verify failure scenarios, as well as verify that a particular `ConfigWriter` writes to an expected `ConfigValueType`.

So I re-used the `checkArbitrary` method in the tests but preserved checks for `ConfigValueType` and failure cases.

Let me know if you have concerns or objections about that please .

ruippeixotog (approved):
LGTM, thanks!

Issue #1489, opened 2023-04-17
Dependency issue with http4s (0.22 vs 0.23)


danebert:
After updating to 0.17.3 (from 0.17.2) I am seeing this dependcy conflict (we are using http4s version 0.23.11):

com.github.pureconfig:pureconfig-http4s_2.12:0.17.3 (depends on 0.22.15)

I think the 0.17.2 version of pureconfig-http4s depended on a 0.23.X version of http4s.

Snippet from sbt dependencyTree output:

[info]   +-com.github.pureconfig:pureconfig-http4s_2.12:0.17.2 [S]
[info]   | |   
[info]   | +-org.http4s:http4s-core_2.12:0.23.16

cosminci:
[Maven Central POM for pureconfig-http4s_2.12](https://repo1.maven.org/maven2/com/github/pureconfig/pureconfig-http4s_2.12/0.17.3/pureconfig-http4s_2.12-0.17.3.pom): `http4s-core` has version `0.22.15`.

This issue is only present for Scala 2.12. Scala 2.13 pureconfig modules have the correct http4s dependency version (0.23.18).

ruippeixotog:
Hey guys, sorry about that. The issue that led to that was fixed in #1492 and I have just released v0.17.4 ([pureconfig-http4s](https://repo1.maven.org/maven2/com/github/pureconfig/pureconfig-http4s_2.12/0.17.4/pureconfig-http4s_2.12-0.17.4.pom), [pureconfig-http4s022](https://repo1.maven.org/maven2/com/github/pureconfig/pureconfig-http4s022_2.12/0.17.4/pureconfig-http4s022_2.12-0.17.4.pom)).

satorg:
oops, that was all my fault, sorry about that. I completely missed the project name out.

PR #1492, opened 2023-04-25
Fix project name of http4s022 module

ruippeixotog:
When this module was created as a copy of http4s, the project name should have been updated as well. Having two subprojects with the same name caused published artifacts to override one another, leading to issues like #1489.

Running `sbt name` on master and on this branch and computing the diff shows that the only affected module was http4s022, as expected:

```diff
26c26
< [info] 	pureconfig-http4s
---
> [info] 	pureconfig-http4s022
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -44,6 +44,7 @@ def genericModule(proj: Project) = proj
   .dependsOn(core)
   .dependsOn(testkit % "test")
   .settings(commonSettings)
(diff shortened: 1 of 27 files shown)


jcazevedo (approved):
Thanks for this!

satorg:
oops, that was all my fault, sorry for that. I missed the project name out 😊

PR #1515, opened 2023-06-10
Remove sass gem from CI setup steps

ruippeixotog:
CI jobs are all failing because "sass from sass-embedded conflicts with installed executable from sass", e.g. [this job](https://github.com/pureconfig/pureconfig/actions/runs/5217381536/jobs/9446955287?pr=1514). It seems [Jekyll now comes with sass bundled](https://jekyllrb.com/docs/configuration/sass/), so we don't need it anymore.


jcazevedo (approved):
Thanks!

PR #1516, opened 2023-06-10
Clean up dependency definitions

ruippeixotog:
https://github.com/scoverage/sbt-scoverage/issues/439 is fixed now, so it's safe now to use the latest scalatest version with every Scala major version.

I also inlined Typesafe config SBT entry in `core` as it should only really be included there.

Supersedes #1500. Failures on website CI jobs are due to #1515.


jcazevedo (approved):
Thanks!

PR #1575, opened 2023-10-01
Simplify and document ProductConfigReaderDerivation

ruippeixotog:
Simplifies some parts of `ProductConfigReaderDerivation` by avoiding index tracking when reading tuples and by reusing existing utility functions to deal with `ConfigReader.Result`. Also documents relevant parts of the code for better understandability.

Depends on #1574.


jcazevedo (approved):
Better late than never!

Issue #1590, opened 2023-11-09
deriveReader for Scala 3


evis:
Currently, we can derive `ConfigReader` using `derives ConfigReader`. Could we also support deriving it in Scala 3 using `deriveReader[T]`, like in Scala 2? That would simplify migration to Scala 3. Currently we have to replace all `deriveReader` invocations with `derives`, that's quite cumbersome.

grzegorz-bielski:
Done in here: https://github.com/pureconfig/pureconfig/pull/1599, but the PR is stuck for the time being.

Edit:
Done in https://github.com/pureconfig/pureconfig/pull/1636 and merged

Issue #1488, opened 2023-04-17
Scala 3 derivation of ConfigReader doesn't work with default values


sideeffffect:
```scala
import pureconfig._
import pureconfig.generic.derivation.default._

case class Person(name: String, age: Int = -1) derives ConfigReader

println(ConfigSource.string("{ name = foo }").load[Person])
```
prints
```scala
Left(ConfigReaderFailures(ConvertFailure(KeyNotFound(age,Set()),None,age)))
```

https://scastie.scala-lang.org/HdRYEUzCRme8LH4zffzlPQ

(Expected result is `Right(Person(foo,-1))`)

ruippeixotog:
Fixed by #1636.

bcarter97:
Is it possible to use this with `derives`?

As far as I can see you have to create a semiauto `deriveReader` for every case class, whereas before you could add `derives ConfigReader`. Is that the case or am i missing something?

sideeffffect:
Nice, this works :tada: 
```scala
import pureconfig._
import pureconfig.generic.semiauto.*

case class Person(name: String, age: Int = -1)
object Person:
  given ConfigReader[Person] = deriveReader

println(ConfigSource.string("{ name = foo }").load[Person])

// Right(Person(foo,-1))
```
https://scastie.scala-lang.org/gqFJABbnQD6UnSl9I38NfQ

sideeffffect:
But this still doesn't
```scala
import pureconfig.*
import pureconfig.generic.derivation.default.*

case class Person(name: String, age: Int = -1) derives ConfigReader

println(ConfigSource.string("{ name = foo }").load[Person])

// Left(ConfigReaderFailures(ConvertFailure(KeyNotFound(age,Set()),None,age)))
```
Not even with `"com.github.pureconfig" %% "pureconfig-generic-scala3" % "0.17.7"`
https://scastie.scala-lang.org/T3SR4at8Sy6IQfPLvfVr1g

I'll create a new ticket.

Issue #1330, opened 2022-06-19
How to use PureConfig in Scala 3?


beneyal:
Hi 🙂 

I couldn't find an answer, so I hope it's okay I'm asking here.

I'm migrating a project to Scala 3.1.2, and have this line in my `Dependencies.scala`:

```scala
lazy val pureConfig: Seq[ModuleID] = Seq(
  "com.github.pureconfig" %% "pureconfig" % "0.17.1"
)
```

When I try to compile using SBT, I get an error `Error downloading com.github.pureconfig:pureconfig_3:0.17.1` because it's not found.

My `resolvers` look like this:
```scala
resolvers ++= Seq(
  // a bunch of local resolvers
  Opts.resolver.mavenLocalFile,
  Resolver.defaultLocal,
  Resolver.mavenLocal,
  Resolver.sonatypeRepo("releases"),
  Resolver.sonatypeRepo("snapshots"),
  DefaultMavenRepository
)
```

What am I missing?

Thank you very much!

bwbecker:
I'm attempting the same thing.  I note that on Maven, `pureconfig-core` has an artifact for Scala3 (https://mvnrepository.com/artifact/com.github.pureconfig/pureconfig-core) but that just `pureconfig` does not (https://mvnrepository.com/artifact/com.github.pureconfig/pureconfig).  `pureconfig-generic` also does not have a Scala3 artifact.

So my conclusion is that maintainers are in the process of converting the library but aren't done yet and we'll either need to get involved or be patient.

leifwickland:
@ruippeixotog Does https://github.com/pureconfig/pureconfig/issues/1082 accurately describe what's holding up a Scala 3 version of `pureconfig-generic`?

ruippeixotog:
Hey everyone! #1082 describes the shortcomings of our project structure, which isn't very user-friendly when it comes to Scala 3.

However, while #1082 makes things somewhat harder, I don't think it's the main reason why we don't have `pureconfig-generic` for Scala 3. Shapeless itself isn't published for Scala 3 and Scala 3's philosophy is a bit different when it comes to derived instances. There's https://github.com/typelevel/shapeless-3, but it's not a drop-in replacement for Scala 2 in any sense.

[...]

PR #1599, opened 2023-11-22
Further Scala 3 support 

### LINKED ISSUE #970: Add support for generic derivation of `ConfigReader` and `ConfigWriter` instances in Scala 3

grzegorz-bielski:
Aims to resolve: https://github.com/pureconfig/pureconfig/issues/970, https://github.com/pureconfig/pureconfig/issues/1330, https://github.com/pureconfig/pureconfig/issues/1488, https://github.com/pureconfig/pureconfig/issues/1590 and possibly other issues, closing the gap between existing Scala 2.13 shapeless based `pureconfig-generic` and Scala 3 derivation in `core`.

## Main changes introduced in this PR:
- Refactored `ConfigReader` derivation with default params support for Scala 3 
- `ConfigWriter`, `ConfigConvert` derivation for Scala 3
- `EnumConfigWriter` and `EnumConfigConvert` derivation for Scala 3
- `ProductHint`, `CoproductHint` and new `EnumHint` to make the API consistent

## Other changes:
- code of `generic-base` was put into `core`, so classes like `ProductHint` could be used in Scala 3 derivation
- cross compiled `pureconfig` (bundle) and `pureconfig-generic` for Scala 3 with compatibility layer in `semiauto` for easier migration which is a proxy to the `core`.
- deprecated `import pureconfig.generic.derivation.default.*` import in favour of `import pureconfig.generic.derivation.syntax.*`. Old import was bringing in only the `ConfigReader` extension with applied default config. New import brings extensions for deriving `ConfigReader`, `ConfigWriter`, `ConfigConvert` and default values are taken from implicit conversions (`implicit def`) in companion objects of `ProductHint`, `CoproductHint` and `EnumHint` which is similar to behaviour from Scala 2. This makes the `default` part of the import meaningless. 
Since the implicit conversions in Scala 2 and 3 seem to be working a little bit differently, then once `implicit def` is removed from Scala 3, it might be necessary to use another auxiliary import in form on `import pureconfig.generic.derivation.defaults.given`. 
- in addition cross compiled `hadoop`, `javax`, `joda`, `scala-xml`, `scalaz` and `squants` modules for Scala 3. I think that leaves only `enum` and `cron4s` for which there is no Scala 3 version and `zio-config` which I'd leave for separate MR since we are on version `1.0.10`, and there soon will be `4.0.0`, so it needs more consideration.

### TODO:
- [x] Testing against existing project(s)
- [ ] ... ?
- [ ] Docs update


grzegorz-bielski:
Hi @ruippeixotog, @jcazevedo, @leifwickland 

I know this is a bigger code piece, but please take a look if you have some time. 🙇 

I think this will give us feature parity with Scala 3, but I also want to test it with some projects on the side and start the review process before writing any docs so this still can be considered a WIP.

tjarvstrand:
Pureconfig is the last thing blocking my project from Scala 3. It would be awesome to see this land!

Andrapyre:
I see that releasing the pureconfig-yaml module for Scala 3 is not part of this and that there's not much (or anything) about yaml in the documentation. Is this module still planned for long-term support? How much effort would it take to get this module to where it can be released for Scala 3? I'm open to contributing on that.

grzegorz-bielski:
@Andrapyre This PR is focused mostly at improving derivation support in Scala 3.
`pureconfig-yaml` is available for Scala 3 for 6 months now, and can be used as follows:
(ported from existing module readme)
```Scala
//> using scala 3.3.1
//> using lib com.github.pureconfig::pureconfig-core:0.17.5-SNAPSHOT
//> using lib com.github.pureconfig::pureconfig-yaml:0.17.5-SNAPSHOT

import pureconfig.generic.derivation.syntax.*
// or `import pureconfig.generic.derivation.default.*` from before 0.17.5-SNAPSHOT (this PR)
import pureconfig.module.yaml.*
import pureconfig.*
import java.nio.file.Files

case class Person(name: String, age: Int, children: List[Person]) derives ConfigReader

@main
def run =
  val yamlFile = Files.createTempFile("conf-", ".yaml")

  Files.write(
    yamlFile,
    """
    | name: John
    | age: 42
    | children:
    |   - name: Sarah
    |     age: 7
    |     children: []
    |   - name: Andy
    |     age: 10
    |     children: []
    | """.stripMargin.getBytes
  )

  println(
     YamlConfigSource.file(yamlFile).load[Person]
  ) // Right(Person(John,42,List(Person(Sarah,7,List()), Person(Andy,10,List()))))

  println(
    YamlConfigSource.file(yamlFile).at("age").load[Int]
  ) // Right(42)
```

For some reason I don't see it on maven though 🤔 

I guess documentation here is not the best. I was hoping to improve this as well, but first I'd like some feedback on these derivation changes. 😄

Andrapyre:
@grzegorz-bielski , ah it's not in Maven because there hasn't been a release since June 4th, when support was added. https://github.com/pureconfig/pureconfig/commit/608634b080511b70cc2f7bbf0105aeaa2e3d8cee

ruippeixotog:
Hi @grzegorz-bielski, sorry it took so long to get to this, and thanks for the interest in ramping up Scala 3 support! I'm afraid this will _definitely_ need to be split in smaller pieces, both from a reviewability and from an VCS tidiness standpoint.

At 70 changed files and 1300+ lines added, it's going to be virtually impossible for me to review this. It has a lot of forward non-breaking changes (e.g. adding getting more modules to use Scala 3 and adding support for `ConfigWriter` derivation) which are straightforward improvements at easy to review, but which are mixed with breaking changes that are much more arguable and warrant a lot more discussion (e.g. moving files to `core`, usage of `implicit` hints in derivation).

From a VCS standpoint: in this repo we squash and merge PRs, so everything here would be going into a single commit. In order to make sure we can trace changes back later on, we need to make sure that each PR is a complete but atomic feature or improvement.

grzegorz-bielski:
@ruippeixotog Thanks for taking a look!

Please note that most of the changed lines in this PR come from tests copied almost verbatim from `pureconfig-generic`, and most of the changed files are about shuffling them from one place to another without changing any code. 
If we want to keep the VCS clean we can always _not_ squash commits for this particular PR and instead rebase them before merging.

I believe splitting it into multiple PRs will be more work, which will delay the complete functionality even further, but if that's necessary... then it can be done. 😅 

I think we would need following MRs:
1. **putting hints in `core`** - This is necessary to be able to preserve compatibility between Scala 2 and Scala 3 derivation config, since Scala 3 derivation is in `core`. Alternatively we can move the new Scala 3 derivation to `pureconfig-generic` and leave the old experimental Scala 3 derivation in core. 
2. **new `ConfigReader`** - We can perhaps keep old experimental Scala 3 `ConfigReader` without implicit hints for backwards compatibility under `pureconfig.generic.derivation.default.*`. On the other hand I think solution from this MR is cleaner and it doesn't change default. It's a breaking change only if you were using `ConfigReaderDerivation.Default` directly, and since Scala 3 derivation was considered experimental and it's easy to fix then maybe we should not care.
3. **new `ConfigWriter`**
4. **new `ConfigConvert`**
5. **new `EnumConfigReader` and `EnumConfigWriter`** - I believe the addition of `EnumHint` introduced in this PR could be skipped, but I think that's a nice improvement which makes the API more consistent
6. **Scala 3 compatibility layer in `pureconfig-generic`** - so we can use `pureconfig.generic.semiauto.deriveReader` and just import `pureconfig`
7. **any other extra changes** like cross compiling auxiliary modules

Before embarking on that journey though, I need to know if we share the same vision. 😄 

My goal is to have the user download a single dependency, that is `"com.github.pureconfig" %% "pureconfig" % version`, and be able to cross compile the following code against Scala 2.12, 2.13 and 3.

```scala
import pureconfig.generic.semiauto.deriveReader

case class Cat(name: Option[String] = None) 

implicit val catReader: ConfigReader[Cat] = deriveReader
```

This PR already accomplishes that.

Do you have any concerns regarding that? 👀

Andrapyre:
This looks like a lot of great work by @grzegorz-bielski on a top-prio initiative that would unblock users for Scala 3 (or at least for this lib in Scala 3 as in my case). Not to be obnoxious, but could we get some more momentum on this?

evis:
@ruippeixotog could you approve @grzegorz-bielski's plan, please? https://github.com/pureconfig/pureconfig/pull/1599#issuecomment-1858805071

tjarvstrand:
@grzegorz-bielski If it's not too much work, maybe it's more feasible to get this to move forward by starting to follow the plan you outlined above? It'd be a shame to see all your effort go wasted!

grzegorz-bielski:
@tjarvstrand I was able to get in touch with @ruippeixotog some time ago. He won't be available for the time being, but should return soon(ish) and hopefully some things will move forward. 

I don't want to start working on the outlined plan just yet. Looks like from the maintainers perspective, the first step - putting derivation hints in `core` is rather controversial, but this is the basis of my changes. Building on top of that, without any approval from maintainers could lead to even more wasted effort...

Please give it a little bit more time. I have my finger on the pulse!

tjarvstrand:
Great! Just wanted to make sure this is not left to rot😊

core/src/main/scala-3/pureconfig/generic/derivation/ConfigConvertDerivaion.scala
maciej-arciuch-allegro, line 1: I see a typo in the file name

ramytanios:
Passionately waiting for this PR ... :rocket:

ruippeixotog:
Hi @grzegorz-bielski, apologies for taking so long to reply.

I discussed the changes you proposed here with @jcazevedo and we're aligned that they'd be too disruptive for existing users. Moreover, we aren't really fond of with some of the concrete design proposals here. In particular:

* We're still keen on keeping `core` APIs simple and we don't think full-blown, configurable generic derivation (or any supporting classes, like hints) fits that role.
* We're not fond of mixing `deriving` clauses and implicit summoning together, as they result in a configuration model that's hard to reason about. The scoping rules are at times confusing and in some cases it may actually limit what users can do (e.g. if you have a series of classes defined inside the same object, it's hard or impossible to define different hints for each one).

All existing public APIs should be kept as they are (i.e. the instructions in [the documentation](https://pureconfig.github.io/docs/scala-3-derivation.html) should continue to work). One way we could incorporate the changes in this PR would be for you to create some `pureconfig-generic-scala3` module which would be a drop-in (source-compatible) replacement for `pureconfig-generic` implemented using Scala 3 macros. That module could be included in the `pureconfig` bundle, making it easy for users to cross-compile projects as you wanted. We would end with two different derivation APIs in Scala 3:

* The default one, provided by core, would use `deriving` and be simple and non-configurable by design. The idea is for users to have a quick start into generic class derivation following the conventions chosen by the library.
* The advanced one, provided by `pureconfig-generic-scala3`, would still be fully implicit-based and would work exactly like `pureconfig-generic`. It's ok to reuse code from core here, or even to push derivation config capabilities to core in order to reuse as much code as possible, as long as the public APIs remain as described here.

How does that sound to you?

ruippeixotog (changes requested):
Sending back to author to address the comment above.

grzegorz-bielski:
Thank you @ruippeixotog for providing this feedback, it's really invaluable!

Having two derivation APIs - simple and more configurable one for advanced usage and cross-compilation sounds good to me. 
But two distinct implementations would be a maintenance burden and I agree that it's better to push as much as possible to core. Ideally, we should try to keep the full Scala 3 derivation implementation in one place only. 

In that case, I think the internals of the Scala 3 derivation in `core` should be fully configurable and cover all the use cases of the proposed `pureconfig-generic-scala3`. 
Instead of moving existing, public hints to core we can have a separate, `private[pureconfig]` hierarchy of hints which mirrors current `ProductHint` and `CoproductHint`. 
Then, the public API in `core` would have some hardcoded configuration values and work like in the [existing Scala 3 derivation API](https://pureconfig.github.io/docs/scala-3-derivation.html) and the one in `pureconfig-generic-scala3` would automatically map the old `ProductHint` and `CoproductHint` from `generic-base` to the new internal hints providing the implicits based configuration like in `pureconfig-generic`. 
With this approach we should end up with a flexible and (internally) extendibly derivation model, quite similar to the one implemented in this PR, which should also cover the use cases of both `core` and the `pureconfig-generic-scala3`.

However, in order to implement this we would need to modify public traits like `pureconfig.generic.derivation.CoproductConfigReaderDerivation` , which would be a breaking change, but I think not many people are using them directly and with some migration notes and a [high level API](https://pureconfig.github.io/docs/scala-3-derivation.html) that still works, a breaking change in an API that is described in the docs as being in alpha/beta stage, it should be fine.

What's your opinion on this?

ruippeixotog:
I'm aligned with your overview, with one caveat - `core` shouldn't need to use any form of implicit-based hints, and I'm keen on avoiding mixing arbitrary implicit summoning with macro derivation - as I mentioned above, scoping rules may get strange and mid-macro implicit summoning makes derivation harder to understand (it breaks some of the advantages of functional programming as function calls start depending on the environment rather than just function parameters).

Ideally, `pureconfig-genreric-scala3` would be the only module doing implicit summoning and passing relevant bits (e.g. references to the relevant hint methods, or an internal mirror of hint as you suggested) down to "functional" macro code. In our suggested approach the macro code would be called directly (as opposed to `deriving`, which is routed directly into a no-arg function), and as such we have control over how we call it.

Note that the current Scala 3 derivation is already configurable by following [the cake pattern](https://github.com/pureconfig/pureconfig/blob/7d15967d12b3bee0fc7bbd13325ebb787a8a3a31/core/src/main/scala-3/pureconfig/generic/derivation/ConfigReaderDerivation.scala#L28-L31), and I'm confident that with some effort it could be further simplified to be configurable by composition or simple parameter passing.

grzegorz-bielski:
> core shouldn't need to use any form of implicit-based hints

That was the idea! 

I just think we need to keep the notion of hints in the `core`'s derivation code, otherwise we might end up with two different implementations. Whether they are passed explicitly or summoned inline is less important to me.

Just so there is no misunderstanding, please consider the following example.

Given the function signature from this PR like this:
```scala
inline def deriveReader[A](using m: Mirror.Of[A], ph: ProductHint[A], cph: CoproductHint[A]): ConfigReader[A] = ???
```
we can change it to:
```scala
inline def deriveReader[A](ph: InternalProductHint[A], cph: InternalCoproductHint[A])(using m: Mirror.Of[A]): ConfigReader[A] = ???
```
The `InternalCoproductHint[A]` and `InternalProductHint[A]` would be `private[pureconfig]` mirrors of the old `ProductHint` and be just hardcoded in `core` for any `A` using existing cake pattern. 

In `pureconfig-genreric-scala3` however, they would come from:
```scala
deriveReader[A](summonInline[ProductHint[A]].toInternal, summonInline[CoproductHint[A]].toInternal)(using m)
```

Do you think we can proceed with such implementation?

grzegorz-bielski:
I tried to introduce an internal mirror of hits in core, but it wasn't enough. I would also need error types and other helpers...

Instead, it might be easier to just take the code in this PR as-is, move it to `pureconfig-generic-scala3` and try to push as much as possible to core (except implicit hints) piece by piece.

This is what I did in: https://github.com/pureconfig/pureconfig/pull/1636 (still a WIP) for `ConfigReader`.

I think we can close this PR. It's clear now that this approach won't be considered.

Issue #1218, opened 2021-11-23
case class default values are non-lazy


dontgitit:
```scala
import pureconfig._
import pureconfig.generic.auto._

case class SampleConf(foo: String = ???)

ConfigSource.string("{ foo: bar }").load[SampleConf]
```

the above produces:
```
scala.NotImplementedError: an implementation is missing
  at scala.Predef$.$qmark$qmark$qmark(Predef.scala:347)
  at SampleConf$.apply$default$1(<console>:1)
  at anon$exportedReader$macro$16$1.inst$macro$1$lzycompute(<console>:1)
  at anon$exportedReader$macro$16$1.inst$macro$1(<console>:1)
  ... 35 elided
```

I would expect it to produce `Right(SampleConf(bar))` since `foo` is defined. It seems the default values are being invoked even if they're unneeded. This example is a little egregious (using `???`) but the default value might involve a complex operation that we don't want to be computed if it's unnecessary.

dontgitit:
P.S. It breaks even when it's nested inside a nullable field;

```scala
import pureconfig._
import pureconfig.generic.auto._

case class Foo(bar: String, baz: Option[String] = ???)

case class Blah(hello: String, world: Option[Foo])

ConfigSource.string("{ hello: hey }").load[Blah]
```

which again makes:
```
scala.NotImplementedError: an implementation is missing
  at scala.Predef$.$qmark$qmark$qmark(Predef.scala:347)
  at Foo$.apply$default$2(<console>:1)
  at anon$exportedReader$macro$99$1.inst$macro$49$lzycompute(<console>:1)
  at anon$exportedReader$macro$99$1.inst$macro$49(<console>:1)
  at anon$exportedReader$macro$99$1.$anonfun$inst$macro$21$1(<console>:1)
  at shapeless.Lazy$$anon$1.value$lzycompute(lazy.scala:123)
  at shapeless.Lazy$$anon$1.value(lazy.scala:123)
  at shapeless.lazily$.apply(lazy.scala:142)
  at anon$exportedReader$macro$99$1.inst$macro$21$lzycompute(<console>:1)
  at anon$exportedReader$macro$99$1.inst$macro$21(<console>:1)
  at anon$exportedReader$macro$99$1.$anonfun$inst$macro$20$1(<console>:1)
  at shapeless.Lazy$$anon$1.value$lzycompute(lazy.scala:123)
  at shapeless.Lazy$$anon$1.value(lazy.scala:123)
[...]

PR #1636, opened 2024-03-19
Introduce hints aware ConfigReader derivation in a new pureconfig-generic-scala3 module

grzegorz-bielski:
A different approach to Scala 3 generic derivation originally started in: https://github.com/pureconfig/pureconfig/pull/1599 

This PR introduces a new `pureconfig-generic-scala3` which should be a drop-in replacement for `pureconfig-generic` from Scala 2.13 and include derivation with a support for implicit hints.

The scope of this particular PR is only the `ConfigReader` derivation, and tests are taken from existing `pureconfig-generic` module.

To reduce the amount of changes in a single PR I didn't modify the original Scala 3 derivation in `core`. There is a however an opportunity for code reuse - for example the deriving `ConfigReader` for tuples could be pretty much the same in both versions.
I believe this, and other things can be refactored in separate, subsequent PRs.


grzegorz-bielski:
Hello again  @ruippeixotog , @jcazevedo !

Can you please take a look at this, reduced in scope PR?

SethTisue:
FYI @alexandru... would this be enough for your needs?

jcazevedo:
Thank you for your work on this, @grzegorz-bielski! I'll take a look at the PR during this week.

modules/generic-scala3/build.sbt
```diff
@@ -0,0 +1,3 @@
+import Dependencies.Version._
+
```
ruippeixotog, line 2: Even though you specified only Scala 3 in `crossScalaVersions` the main `scalaVersion` is still set to the default value (2.12.x). The module only really supports Scala 3, so I think it's better to override it here.

```suggestion

scalaVersion := scala3

```

modules/generic-scala3/src/main/scala-3/pureconfig/generic/derivation/HintsAwareConfigReaderDerivation.scala
```diff
@@ -0,0 +1,18 @@
+package pureconfig
+package generic
+package derivation
+
+import scala.deriving.Mirror
+
+trait HintsAwareConfigReaderDerivation
+    extends HintsAwareCoproductConfigReaderDerivation,
+      HintsAwareProductConfigReaderDerivation:
```
ruippeixotog: General comment: as with the Scala 3 derivation in core, I strongly prefer if all code in this repository uses the same style for scoping. Can you please use braces instead of just indentation to mark all scopes (class, function, conditionals, `match`...)?

modules/generic-scala3/src/main/scala-3/pureconfig/generic/derivation/HintsAwareCoproductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,52 @@
+package pureconfig
+package generic
+package derivation
+
+import scala.compiletime.*
```
ruippeixotog: Same thing as above - please maintain the same import style as in the rest of the repository.

```suggestion
import scala.compiletime._
```

modules/generic-scala3/src/main/scala-3/pureconfig/generic/derivation/HintsAwareConfigReaderDerivation.scala
```diff
@@ -0,0 +1,18 @@
+package pureconfig
+package generic
+package derivation
```
ruippeixotog: nit: I'd probably prefer to keep code under `pureconfig.generic.scala3` to align with the Maven artifact name (with the exception of `semiauto` as we want this to be a drop-in replacement for `generic`).

Some of the packaging decisions within `generic` aren't consistent at the moment as back then we didn't expect there to be other generic derivation modules other than `generic` (which should have been called `generic-shapeless`), but I think we can make this one follow a more future-proof structure.

modules/generic-scala3/src/main/scala-3/pureconfig/generic/semiauto.scala
```diff
@@ -0,0 +1,6 @@
+package pureconfig.generic
+
+import pureconfig.generic.derivation.*
+
+object semiauto:
+  export HintsAwareConfigReaderDerivation.deriveReader
```
ruippeixotog: I like that this module is kept as simple as possible, delegating all logic to an internal implementation 👍

modules/generic-scala3/src/test/scala-3/pureconfig/generic/ProductHintSuite.scala
```diff
@@ -0,0 +1,217 @@
...
+      camel-case-int = 1
+      camel-case-string = "bar"
+      camel-case-conf {
+        this-is-an-int = 3
+        this-is-another-int = 10
+      }
+    }""")
+
+    conf.to[ConfWithCamelCase] shouldBe Right(ConfWithCamelCase(1, "bar", ConfWithCamelCaseInner(3, 10)))
+  }
```
ruippeixotog: Is there a reason why you chose not to add name conversion tests like "write kebab case config keys..." and "write camel case config keys..." in `modules/generic/src/test/scala/pureconfig/ProductHintSuite.scala`?

ruippeixotog (commented):
Thanks for submitting this, @grzegorz-bielski! This is a very good start and by isolating changes to a new module we can be much more confident we won't negatively affect existing users.

I read most of the code and LGTM in general, I just left a few comments with minor questions/requests. I still need more time to review `HintsAwareProductConfigReaderDerivation` - looking at the tests it seems to work as expected but I'd like to take a closer look at the implementation before I merge this. I'll also let @jcazevedo take a look before merging.

modules/generic-scala3/src/main/scala-3/pureconfig/generic/derivation/HintsAwareCoproductConfigReaderDerivation.scala
grzegorz-bielski: Sure, I can change it!

But let's keep in mind that [Scala 3 reference](https://docs.scala-lang.org/scala3/reference/changed-features/imports.html) mentions the `_` imports as something to be `dropped in a future versions`, so I think we should treat it more as a `givens` / `using` rather than optional braces and at some point commit to `*`.

modules/generic-scala3/build.sbt
grzegorz-bielski: I tried that, but it broke the build. Let me check again...

modules/generic-scala3/src/test/scala-3/pureconfig/generic/ProductHintSuite.scala
grzegorz-bielski: Those tests require `ConfigWriter` which derivation is not implemented in this PR. I'll add it in some subsequent one.

modules/generic-scala3/build.sbt
grzegorz-bielski: Yes. It breaks the `2.12` build in CI, see: https://github.com/grzegorz-bielski/pureconfig/actions/runs/8473601089/job/23218298773#step:5:264

jcazevedo: Apparently this is an issue with strict aggregation in PureConfig's build. Trying to run `++ 2.12.19 compile` ends up attempting to compile `generic-scala3` since 2.12.19 is included in `crossScalaVersions` of the root project (and the root project aggregates `generic-scala3`). I think #1643 should fix that.

alexandru:
@SethTisue love seeing this. For our Scala 3 migration, we'll need `ConfigWriter`, too.

modules/generic-scala3/build.sbt
grzegorz-bielski: @jcazevedo  yeah, looks like this fixes it. Thx!

grzegorz-bielski:
Hello, @ruippeixotog, @jcazevedo. Did you have a chance to take a deeper look? 👀 
I'd be nice to move this forward.

jcazevedo:
@grzegorz-bielski: Not yet, sorry. I'll try to make a push this week.

modules/generic-scala3/src/main/scala-3/pureconfig/generic/derivation/HintsAwareCoproductConfigReaderDerivation.scala
ruippeixotog: Yeah, I saw that the latest Scala 3 version already treats it as deprecated... I'm curious to see how it affects projects targeting Scala 2 too, though. We'll continue cross-compiling `core` with Scala 2 and 3 for the foreseeable future, which prevents us from adopting Scala 3 specific syntax there.

grzegorz-bielski: There are `-Xsource:3` and `-Xsource:3-cross` [flags](https://docs.scala-lang.org/scala3/guides/migration/tooling-scala2-xsource3.html) which can help with that, but only for Scala 2.13, so I guess it can be leveraged only if pureconfig drops Scala 2.12.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,137 @@
...
+      case _: EmptyTuple =>
+        Right(widen[EmptyTuple, T](EmptyTuple))
+    }
+
+}
+
+private[scala3] object ProductDerivationMacros {
+  inline def getDefaults[T](inline size: Int): Vector[Option[Any]] = ${ getDefaultsImpl[T]('size) }
+
+  def getDefaultsImpl[T](size: Expr[Int])(using Quotes, Type[T]): Expr[Vector[Option[Any]]] = {
```
ruippeixotog: I think there's an opportunity here to fix #1218. If we're defining our own macros to access default values, we can also make them lazy by returning here an `Expr[Vector[Option[() => Any]]]` instead of evaluating all defaults eagerly. This should help a lot cases where computing a default may be slow or throw exceptions.

Feel free to do it here or in a follow-up PR. If you do it here, please add tests covering this.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,137 @@
...
+      case _: EmptyTuple =>
+        Right(widen[EmptyTuple, T](EmptyTuple))
+    }
+
+  private inline def readTuple[T <: Tuple, N <: Int](cursors: List[ConfigCursor]): Either[ConfigReaderFailures, T] =
+    inline erasedValue[T] match {
+      case _: (h *: t) =>
+        val n = constValue[N]
+        val reader = summonConfigReader[h]
+        val cursor = cursors(n)
```
ruippeixotog, line 97: Same thing here - probably better to convert `cursors` to a `Vector[ConfigCursor]`.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,137 @@
...
+      labels: List[String],
+      actions: Map[String, ProductHint.Action],
+      defaults: Vector[Option[Any]]
+  ): Either[ConfigReaderFailures, T] =
+    inline erasedValue[T] match {
+      case _: (h *: t) =>
+        val n = constValue[N]
+        lazy val reader = summonConfigReader[h]
+        val default = defaults(n)
+        val label = labels(n)
```
ruippeixotog, line 66: This indexing will happen at runtime, right? If that's the case, maybe it's better to use a `Vector[ConfigCursor]` for `labels`?

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,137 @@
...
+
+import pureconfig.error.{ConfigReaderFailures, KeyNotFound, WrongSizeList}
+import pureconfig.generic.ProductHint.UseOrDefault
+import pureconfig.generic.derivation.Utils
+import pureconfig.generic.derivation.Utils.widen
+
+trait HintsAwareProductConfigReaderDerivation { self: HintsAwareConfigReaderDerivation =>
+  inline def deriveProductReader[A](using
+      m: Mirror.ProductOf[A],
+      ch: CoproductHint[A],
```
ruippeixotog: Same thing as `CoproductConfigReader`, we shouldn't require a `CoproductHint` here.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareCoproductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import scala.compiletime._
+import scala.deriving.Mirror
+
+import pureconfig.error.{CannotConvert, ConfigReaderFailures}
+import pureconfig.generic.derivation.Utils
+import pureconfig.generic.error.InvalidCoproductOption
+
+trait HintsAwareCoproductConfigReaderDerivation { self: HintsAwareConfigReaderDerivation =>
+  inline def deriveSumReader[A](using m: Mirror.SumOf[A], ch: CoproductHint[A], ph: ProductHint[A]): ConfigReader[A] =
```
ruippeixotog: We shouldn't require a `ProductHint` here.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,137 @@
...
+              objCursor <- cur.asObjectCursor
+              labels = Utils.transformedLabels(identity)
+              actions = labels.map { label => label -> ph.from(objCursor, label) }.toMap
+              result <- readCaseClass[m.MirroredElemTypes, 0, A](objCursor, labels, actions, defaults)
+            } yield m.fromProduct(result)
+          }
+        }
+    }
+
+  private inline def readCaseClass[T <: Tuple, N <: Int, A: ProductHint: CoproductHint](
```
ruippeixotog: From this point on we shouldn't need any implicit hints, as we already extracted actions for each key.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
```diff
@@ -0,0 +1,137 @@
...
+            case _ =>
+              objCursor.failed(KeyNotFound.forKeys(fieldHint.field, objCursor.keys))
+          }
+
+        val tail = readCaseClass[t, N + 1, A](objCursor, labels, actions, defaults)
+
+        val resultTuple = ConfigReader.Result.zipWith(head, tail)((h, t) => widen[h *: t, T](h *: t))
+
+        val usedFields = actions.map(_._2.field).toSet
+        val hintFailures = summon[ProductHint[A]].bottom(objCursor, usedFields).toLeft(())
```
ruippeixotog, line 84: This looks correct, but I noticed that no tests cover the behavior of `bottom`. Could you please add a test for derived readers dealing with unknown keys?

build.sbt
```diff
@@ -180,6 +181,9 @@ lazy val lintFlags = forScalaVersions {
 // Use the same Scala 2.12 version in the root project as in subprojects
 scalaVersion := scala212
 
+// temp, will be done in: https://github.com/pureconfig/pureconfig/pull/1643
+crossScalaVersions := Nil
```
ruippeixotog: Let's ensure we don't forget to remove this 🙂

ruippeixotog (changes requested):
Thanks again for this work @grzegorz-bielski, it looks really good 🎉 I'm excited to see next steps to further improve derivation.

I left some comments, mostly focused on product derivation. Once they're addressed I'm happy with merging this, assuming @jcazevedo is happy with the PR too.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
grzegorz-bielski: There are tests for this, [here](https://github.com/pureconfig/pureconfig/pull/1636/files#diff-f40c59d3ad8b4e6265edbc0f6fadda20d376eeab8ab2df82be379030bd9ec767R96) and [here](https://github.com/pureconfig/pureconfig/pull/1636/files#diff-f40c59d3ad8b4e6265edbc0f6fadda20d376eeab8ab2df82be379030bd9ec767R112). 
If you'll change this line to `val hintFailures: ConfigReader.Result[Unit] = Right(())`, they will fail.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
grzegorz-bielski: We don't need `CoproductHint`, but `ProductHint` is needed to access the `.bottom`, even with extracted actions. If you prefer, I can pass it explicitly.

build.sbt
grzegorz-bielski: I'll remove it after https://github.com/pureconfig/pureconfig/pull/1643 is merged

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
grzegorz-bielski: Yeah, let's do it now!

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigReaderDerivation.scala
```diff
@@ -0,0 +1,31 @@
...
+import scala.compiletime._
+import scala.deriving.Mirror
+
+trait HintsAwareConfigReaderDerivation
+    extends HintsAwareCoproductConfigReaderDerivation,
+      HintsAwareProductConfigReaderDerivation {
+  inline def deriveReader[A](using m: Mirror.Of[A], ph: ProductHint[A], cph: CoproductHint[A]): ConfigReader[A] =
+    inline m match {
+      case pm: Mirror.ProductOf[A] => deriveProductReader[A](using pm, ph)
+      case sm: Mirror.SumOf[A] => deriveSumReader[A](using sm, cph)
```
grzegorz-bielski: After the change from two implicit hints per method to only one there was an `wrong number of arguments at posttyper` compiler error here. Passing arguments explicitly helped to alleviate that.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
ruippeixotog: Right, makes sense :+1:

ruippeixotog (approved):
LGTM 👍 @jcazevedo, this is now dependent on us merging #1643.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigReaderDerivation.scala
```diff
@@ -0,0 +1,31 @@
...
+package generic
+package scala3
+
+import scala.compiletime._
+import scala.deriving.Mirror
+
+trait HintsAwareConfigReaderDerivation
+    extends HintsAwareCoproductConfigReaderDerivation,
+      HintsAwareProductConfigReaderDerivation {
+  inline def deriveReader[A](using m: Mirror.Of[A], ph: ProductHint[A], cph: CoproductHint[A]): ConfigReader[A] =
```
jcazevedo: This is relying on the fact that we always have default `ProductHint` and `CoproductHint` instances for all types, even for those that aren't products nor coproducts. Can we only require a `Mirror` and `summonInline` either `ProductHint` or `CoproductHint`?

```scala
inline def deriveReader[A](using m: Mirror.Of[A]): ConfigReader[A] =
  inline m match {
    case pm: Mirror.ProductOf[A] => deriveProductReader[A](using pm, summonInline[ProductHint[A]])
    case sm: Mirror.SumOf[A] => deriveSumReader[A](using sm, summonInline[CoproductHint[A]])
  }
```

ruippeixotog:
@grzegorz-bielski, do you mind rebasing or merging with master and removing the `crossScalaVersions` setting? Once that's done let's finally close this 🙂

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigReaderDerivation.scala
```diff
@@ -0,0 +1,31 @@
...
+  protected inline def summonConfigReader[A]: ConfigReader[A] =
+    summonFrom {
+      case reader: ConfigReader[A] => reader
+      case m: Mirror.Of[A] =>
+        deriveReader[A](using m, summonInline[ProductHint[A]], summonInline[CoproductHint[A]])
+    }
+}
+
+object HintsAwareConfigReaderDerivation extends HintsAwareConfigReaderDerivation {
+  object syntax {
```
jcazevedo: I'm not convinced we should be providing this. Following up on @ruippeixotog's comments in https://github.com/pureconfig/pureconfig/pull/1599#issuecomment-1998631196, I think that providing access to the `derives` clause via explicit imports when the actual derivation is controlled by placing hints in scope can make the configuration model very hard to reason about. I would prefer not providing access to the `derives` clause in pureconfig-generic.

I don't think this is a blocker for this PR. We can iterate on this as a follow-up.

jcazevedo (dismissed):
Thanks @grzegorz-bielski! This is excellent!

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigReaderDerivation.scala
ruippeixotog: We do have a default instance of `ProductHint` and `CoproductHint` defined in their companion objects, so I don't think it's possible for a type not to have an instance... and given these macros are expanded in compile time I don't think there's a runtime cost to always requiring them.

I'm fine either way, but I have a slight preference for avoiding `summonInline` as it makes the macro function more "pure" (in the sense of being fully defined by its arguments).

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigReaderDerivation.scala
ruippeixotog: Yeah, agreed, I missed this one.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigReaderDerivation.scala
grzegorz-bielski: We cannot escape `summonInline`. It has to be in either `summonConfigReader` or `deriveReader`.
I think moving it to `deriveReader`, as suggested by @jcazevedo makes more sense.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigReaderDerivation.scala
grzegorz-bielski: I removed it. It's easier to bring something back, rather than trying to work around it in already published artifact. 😄 

In general however, I disagree with the notion that `derives` + implicit hints is harder to reason about than `deriveReader` + implicit hints. I could be missing something, but to me the first version is just a nicer syntax for the same thing. 
I suspect people will be adding `extension (c: ConfigReader.type) inline def derived[A](using Mirror.Of[A]): ConfigReader[A] = deriveReader[A]` to their codebases anyway.

One thing that is definitely hard to reason about is that we rely on _default_ implicit config. This is how it works in `pureconfig-generic` so we shouldn't change that for this module, but personally I much prefer the model from new Scala 3 configured derivation in circe-core where you have to explicitly state the config.
But the `implicit def` that pureconfig relies on will at some point be removed in Scala 3, so it might force the users to state the config with `import defaults.given` anyway. 😅

grzegorz-bielski:
@jcazevedo , @ruippeixotog Can we merge it now? 👀

ruippeixotog (approved):

ruippeixotog:
@grzegorz-bielski done! I'll release a new version later today or during the weekend.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigReaderDerivation.scala
jcazevedo: > One thing that is definitely hard to reason about is that we rely on _default_ implicit config. This is how it works in `pureconfig-generic` so we shouldn't change that for this module, but personally I much prefer the model from new Scala 3 configured derivation in circe-core where you have to explicitly state the config.

Circe makes it more obvious that some configuration is taking place by using the `ConfiguredEncoder`, `ConfiguredDecoder` and `ConfiguredCodec` types, but I would prefer a simpler model for PureConfig. For example, let's say that we have the following coproduct definitions:

```scala
sealed trait Foo
case class Bar(a: Int, b: String) extends Foo
case class Baz(c: Double, e: Boolean) extends Foo
```

If we want to add default `Codec` instances for each of these types we can do the following:

```scala
sealed trait Foo derives Codec
case class Bar(a: Int, b: String) extends Foo derives Codec
case class Baz(c: Double, e: Boolean) extends Foo derives Codec
```

Or, if we want to configure the way serialization/deserialization works, use a `ConfiguredCodec` instead and have a `Configuration` in scope:

```scala
given conf: Configuration = Configuration.default

sealed trait Foo derives ConfiguredCodec
case class Bar(a: Int, b: String) extends Foo derives ConfiguredCodec
case class Baz(c: Double, e: Boolean) extends Foo derives ConfiguredCodec
```

However, if we want to have a different configuration in just one of the types we would have to do something like the following:

```scala
given conf: Configuration = Configuration.default

sealed trait Foo derives ConfiguredCodec

case class Bar(a: Int, b: String) extends Foo

object Bar {
  given codec: ConfiguredCodec[Bar] =
    ConfiguredCodec.derived(using Configuration.default.withPascalCaseMemberNames)
}

case class Baz(c: Double, e: Boolean) extends Foo derives ConfiguredCodec
```

The two main things that I don't like with this approach is that (1) we need to be aware of more types and (2) we might not always be able to use the `derives` clause when configuration is to take place.

In PureConfig, I would prefer reducing the number of types we would need to know to `ConfigReader` and `ConfigWriter`, and have this process simplified to the point where `derives` would only produce default instances and explicit given instances were to be defined whenever some configuration was to take place.

evis:
@ruippeixotog did you manage to make a release? Can't find it here https://mvnrepository.com/artifact/com.github.pureconfig

ruippeixotog:
I haven't, I ended up forgetting about it... I have just released v1.17.7, which also includes @grzegorz-bielski 's latest PR 👍

tjarvstrand:
@grzegorz-bielski @ruippeixotog Awesome work guys! I - and many others, I believe - have been waiting for this for a long time!

I'm at a point now where I I want to start migrating some projects to Scala 3. Is there any chance the documentation could be updated?

grzegorz-bielski:
Hi @tjarvstrand !

Yes, we definitely need more docs on it... soon(ish)!

While the Scala 3 derivation is almost feature-complete now, I'm still developing some features for it such as: https://github.com/pureconfig/pureconfig/pull/1671 
Some things in core and best practices could change too: https://github.com/pureconfig/pureconfig/pull/1674

I don't know about you @ruippeixotog, but in my opinion It might be yet a little bit too soon to write full featured docs for Scala 3 derivation. It's still a little bit scattered.

For now I'd recommend looking at the tests for usage scenarios of `pureconfig-generic-scala3`: https://github.com/pureconfig/pureconfig/tree/master/modules/generic-scala3/src/test/scala/pureconfig/generic

tjarvstrand:
Thanks! I actually managed to get things working pretty easily, albeit with semi-automatic derivation. There is no automatic derivation at this point, right?

ruippeixotog:
> I don't know about you @ruippeixotog, but in my opinion It might be yet a little bit too soon to write full featured docs for Scala 3 derivation. It's still a little bit scattered.

Ideally the documentation won't change much - we're going in the direction of having plain, no frills derivation using `deriving` for which we [already have documentation](https://pureconfig.github.io/docs/scala-3-derivation.html) (though it may need to be updated in some parts) and more complete derivation logic that should be a drop-in replacement for Scala 2 (meaning all the existing documentation should apply).

@tjarvstrand if you want to try it out I'd go for adding the extra import, doing things like you'd do in Scala 2 and reporting back stuff that doesn't work. Even if they're limitations already known by us, having user reports flags that they're important gaps :)

chenjianjx:
Seems this feature is not in the stable version `0.17.7` yet, but we sort of need it now. 

Could you please let us know when `0.17.8` can be released?   Or is there a workaround if I just want to have `camel case -> camel case` mapping in scala3 ?

PR #1638, opened 2024-03-24
Update index.md to add information for Scala 3

ChidiRnweke:
The current import linked in the documentation does not work for Scala 3. It results in a quite opaque error while trying to compile. Adding this line makes it clearer to Scala 3 users that their import is slightly different.


ruippeixotog (approved):
Thanks @ChidiRnweke! I'll merge this for now, though I think it's likely that it'll probably be reverted later on. Now that we have a `pureconfig-generic-scala3` module I'm keen on publishing a `pureconfig` bundle with it before the next release, making the transition from Scala 2 to Scala 3 more seamless. I'll accept this just in case we can't get it done before the next release 🙂

PR #1653, opened 2024-04-26
Introduce hints aware ConfigWriter and ConfigConvert derivation for Scala 3

grzegorz-bielski:
Another batch of changes extracted from https://github.com/pureconfig/pureconfig/pull/1599

Adding derivation for `ConfigWriter` and `ConfigConvert`, analogous to the https://github.com/pureconfig/pureconfig/pull/1636 and porting some more tests from `pureconfig-generic`.

`EnumConfigWriter` and `EnumConfigConvert` will be delivered in separate PR.


grzegorz-bielski:
Hello, @ruippeixotog , @jcazevedo !
With readers merged, it's time for writers.
Can you take a look? 😃

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigConvertDerivation.scala
```diff
@@ -0,0 +1,11 @@
+package pureconfig
+package generic
+package scala3
+
+import scala.deriving.Mirror
+
+trait HintsAwareConfigConvertDerivation extends HintsAwareConfigWriterDerivation, HintsAwareConfigReaderDerivation:
+  inline def deriveConvert[A: Mirror.Of]: ConfigConvert[A] =
```
ruippeixotog: Is there a need for a separate trait and object for deriving `ConfigConvert`? Through the rest of the codebase we always kept the concept of `ConfigConvert` very lightweight - we always try to define it as a function of a `ConfigReader` + `ConfigWriter` rather than giving it its own logic. What do you think of putting this definition of `deriveConvert` directly inside `semiauto` to simplify things?

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigWriterDerivation.scala
```diff
@@ -0,0 +1,73 @@
...
+      product: Product,
+      labels: Vector[String]
+  ): List[(String, ConfigValue)] =
+    inline erasedValue[T] match {
+      case _: (h *: t) =>
+        val n = constValue[N]
+        def value[T] = product.productElement(n).asInstanceOf[T]
+
+        val valueOpt = summonConfigWriter[h] match {
+          case writer: WritesMissingKeys[h] => writer.toOpt(value)
```
ruippeixotog: As written here the code is somewhat dangerous; `value[T]` can return any type whatsoever so it can be used anywhere, and if I'm reading this correctly it runs at runtime, not at compile time. Using `h` consistently across this function is probably safer.

```suggestion
        val value = product.productElement(n).asInstanceOf[h]

        val valueOpt = summonConfigWriter[h] match {
          case writer: WritesMissingKeys[`h` @unchecked] => writer.toOpt(value)
```

ruippeixotog (changes requested):
LGTM overall, thank you again!

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigConvertDerivation.scala
grzegorz-bielski: I intentionally left the `semiauto` as simple as possible to treat it as place for exports, but sure, I can do that.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigWriterDerivation.scala
grzegorz-bielski: Good point, fixed!

grzegorz-bielski:
Hi, @ruippeixotog! I incorporated your suggestions, please recheck.

jcazevedo (approved):
Looks good. Thanks!

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigConvertDerivation.scala
ruippeixotog: Yeah, I understand, it's mostly a question of taste 🙂 I consider those cases different as `deriveConvert` (and creation of `ConfigConvert` instances in general) is supposed to be just a convenience method, equivalent to providing `given` instances for `ConfigReader` and `ConfigWriter`. Hiding that away into a full fledged trait equivalent to the other two could open a path to it being more than that.

ruippeixotog (approved):

PR #1654, opened 2024-05-01
Add EnumConfigWriter and EnumConfigConvert derivation for Scala 3

grzegorz-bielski:
Extending the derivation in core for `ConfigWriter` and `ConfigConvert` for Enums.

I also added some convenience methods for a better compatibility with the old `pureconfig-generic`.

Together with, https://github.com/pureconfig/pureconfig/pull/1653 and https://github.com/pureconfig/pureconfig/pull/1636, it should bring the Scala 3 derivation to _almost_ the same place as the Scala 2 one. 

We are still missing:
-  derivation for value classes, which I'm not sure how to approach at the moment 🤔 
-  tuples read as product with keys like `_1`, `_2` (the behaviour of `pureconfig.generic.DerivedConfigReader.tupleReader`).  This seems like a niche use case and I'm not sure if it's worth porting it at all.


core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigConvertDerivation.scala
```diff
@@ -0,0 +1,48 @@
...
+    object Instance
+        extends EnumConfigConvertDerivation,
+          EnumConfigReaderDerivation(transformName),
+          EnumConfigWriterDerivation(transformName)
+
+    Instance.EnumConfigConvert.derived[A]
+  }
+
+  inline def deriveEnumerationConvert[A: Mirror.SumOf]: ConfigConvert[A] =
+    Default.EnumConfigConvert.derived[A]
```
jcazevedo: I would prefer if we would keep these `derive*` methods exposed only in pureconfig-generic-scala3. We're going to include both pureconfig-core and pureconfig-generic-scala3 in the pureconfig bundle, so it can be confusing for downstream users of the bundle to have the multiple locations.

jcazevedo (dismissed):
Thanks!

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigConvertDerivation.scala
grzegorz-bielski: Makes sense. I pushed all of them to `EnumDerivationUtils` object in `pureconfig-generic-scala3`.

grzegorz-bielski:
Hi @jcazevedo, @ruippeixotog. It's ready for a (re)review now.

@ruippeixotog, I don't think I could get rid of `EnumConfigConvertDerivation` trait in this case as we need it for the `derives` clause.

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigWriterDerivation.scala
```diff
@@ -0,0 +1,43 @@
...
+
+trait EnumConfigWriterDerivation(transformName: String => String) {
+
+  trait EnumConfigWriter[A] extends ConfigWriter[A]
+
+  object EnumConfigWriter {
+    inline def derived[A](using m: Mirror.SumOf[A]): EnumConfigWriter[A] =
+      new EnumConfigWriter[A] {
+        assertIsEnum[m.MirroredElemTypes]
+        val labels = transformedLabels[A](transformName)
```
ruippeixotog: Can you please make this a `Vector`, to avoid a O(n) label search on every write?

grzegorz-bielski: Yeah, certainly! I forgot to port it 😄

modules/generic-scala3/src/test/scala/pureconfig/generic/EnumDerivationUtilsSuite.scala
```diff
@@ -0,0 +1,103 @@
...
+import scala.deriving.Mirror
+import scala.language.higherKinds
+
+import com.typesafe.config.{ConfigFactory, ConfigValueFactory, ConfigValueType}
+
+import pureconfig._
+import pureconfig.error.{CannotConvert, WrongType}
+import pureconfig.generic.semiauto._
+
+class EnumDerivationUtilsSuite extends BaseSuite {
```
ruippeixotog: I'm not super familiar with the test tooling offered by Scala 3, but is there any way for us to test that enum derivation fails if enum options have parameters? Something akin to `shapeless.test.illTyped` for Scala 3 inline functions?

tests/src/test/scala-3/pureconfig/EnumerationConvertDerivationSuite.scala
```diff
@@ -0,0 +1,101 @@
...
+import scala.deriving.Mirror
+import scala.language.higherKinds
+
+import com.typesafe.config.{ConfigFactory, ConfigValueFactory, ConfigValueType}
+
+import pureconfig._
+import pureconfig.error.{CannotConvert, WrongType}
+import pureconfig.generic.derivation._
+
+class EnumerationConvertDerivationSuite extends BaseSuite {
```
ruippeixotog: nit, just to use the same terminology as library code.

```suggestion
class EnumConvertDerivationSuite extends BaseSuite {
```

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigWriterDerivation.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.typesafe.config.{ConfigValue, ConfigValueFactory}
+
+import pureconfig.error.{CannotConvert, ConfigReaderFailures}
+import pureconfig.generic.derivation.Utils.*
+
+type EnumConfigWriter[A] = EnumConfigWriterDerivation.Default.EnumConfigWriter[A]
+
+trait EnumConfigWriterDerivation(transformName: String => String) {
+
+  trait EnumConfigWriter[A] extends ConfigWriter[A]
```
ruippeixotog, line 17: I know that you copied the general structure from `EnumConfigReader`, but we seem to have a significant amount of unnecessary complexity when defining these traits. Why exactly do we need to have `EnumConfigWriter` defined inside the derivation trait? As far as I can see it doesn't need anything from the context of the outer trait, so it can be moved outside.

I don't really like cake patterns like this as they make it easy for developers to create a lot of "parallel" type worlds that aren't compatible. For example, if you add a `derives pureconfig.generic.derivation.EnumConfigWriter[T]` clause to a class and import a non-default `EnumConfigWriterDerivation` (e.g. to use a custom transform name), this won't work because the two `EnumConfigWriter` types aren't actually the same.

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigConvertDerivation.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import com.typesafe.config.ConfigValue
+
+import pureconfig.error.{CannotConvert, ConfigReaderFailures}
+import pureconfig.generic.derivation.Utils.*
+
+type EnumConfigConvert[A] = EnumConfigConvertDerivation.Default.EnumConfigConvert[A]
+
+trait EnumConfigConvertDerivation extends EnumConfigReaderDerivation, EnumConfigWriterDerivation {
+  trait EnumConfigConvert[A] extends ConfigConvert[A]
```
ruippeixotog: As I mentioned in the other PR, I'm not very fond of maintaining lots of logic for `ConfigConvert`, even if it's just boilerplate. `ConfigConvert` is meant to be a very thin class on top of `ConfigReader` and `ConfigWriter`...

I guess my first question is: do we really need to provide users the convenience of writing `derives ConfigConvert` instead of `derives ConfigReader, ConfigReader`? Is is worth having additional non-trivial (i.e. not just a one liner calling `ConfigConvert.fromReaderAndWriter`) derivation code here? I'd like to hear your thoughts too, @jcazevedo.

If we really want to be able to write `derives ConfigConvert`, does it need to have a class structure as complex as the other derivation cake patterns? Can `derived` implementations summon instances for other classes using the same `derived` mechanism? (searching imports for the appropriate `derived` implementation rather than statically choosing the inline function to call)

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/EnumDerivationUtils.scala
```diff
@@ -0,0 +1,33 @@
+package pureconfig
+package generic
+package scala3
+
+import scala.deriving.Mirror
+
+import pureconfig.generic.derivation._
+
+private[generic] object EnumDerivationUtils {
```
ruippeixotog: nit: this doesn't look like utilities at all, this is more like a facade for all enum derivation. What do you think of simply naming it `EnumDerivation`?

```suggestion
private[generic] object EnumDerivation {
```

ruippeixotog (changes requested):

modules/generic-scala3/src/test/scala/pureconfig/generic/EnumDerivationUtilsSuite.scala
grzegorz-bielski: I know nothing of sorts. However, I believe we can use `compile` assertions from ScalaTest. It's less precise, but It'll give us some coverage.

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigConvertDerivation.scala
grzegorz-bielski: > do we really need to provide users the convenience of writing derives ConfigConvert instead of derives ConfigReader, ConfigReader

I think that's useful to have.

> If we really want to be able to write derives ConfigConvert, does it need to have a class structure as complex as the other derivation cake patterns?

Maybe not 🤔 
I'll try experimenting with it a little bit more.

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigWriterDerivation.scala
grzegorz-bielski: I don't like it either, I'd much rather use the approach [like in here](https://github.com/pureconfig/pureconfig/pull/1599/files#diff-c20fb1e4550742d25cf2925bba77e13cf4fca645eb4543bb30dad39c5a55a2fc) (maybe minus the EnumHint 😄).

Having said that - this is already an established pattern and I'd like to continue with it in this PR. 
If we want to change the approach I'd recommend doing it in separate PR and refactor every Scala 3 derivations in core to use it.

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigConvertDerivation.scala
grzegorz-bielski: I made it a tad simpler. We still need a separate type to distinguish it from regular ConfigConvert derivation, which complicates things a bit.

grzegorz-bielski:
@jcazevedo , @ruippeixotog Please have another look! 🙇

core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigConvertDerivation.scala
ruippeixotog: Looks nice now! We can't avoid having a separate type but at least it's a straightforward `derived` implementation rather than a web of traits.

ruippeixotog (approved):
LGTM! Thanks for bearing with us and the long review wait times 🙂

PR #1662, opened 2024-05-12
Fix CI badge URL

jcazevedo:
The URL for the badge we display in the README and in the website was incorrectly built as per https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge#using-the-workflow-file-name.


ruippeixotog (approved):
LGTM

PR #1671, opened 2024-06-08
Scala 3 value class derivation

grzegorz-bielski:
This PR introduces value class derivation in `generic-scala3` module. It's handled transparently using the same `deriveReader` / `deriveWriter` / `deriveConvert` methods.

I think this is the last missing big feature relevant for cross-compilation. 

On the Scala 3 - only side we still lack the ability to derive opaque types, but there might be some [changes in this area](https://contributors.scala-lang.org/t/pre-sip-derives-clauses-for-opaque-types/5924).

### DIFF
--- a/core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
+++ b/core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
@@ -3,6 +3,7 @@ package derivation
 
 import scala.compiletime.{constValue, erasedValue}
 import scala.deriving.Mirror
+import scala.quoted._
 
 object Utils {
 
@@ -32,4 +33,11 @@ object Utils {
       case _: (h *: t) => transform(constValue[h & String]) :: transformedLabelsTuple[t](transform)
       case _: EmptyTuple => Nil
     }
+
+  object TypeName {
+    inline def apply[A]: String = ${ typeNameImpl[A] }
+
+    def typeNameImpl[A](using Type[A], Quotes): Expr[String] = Expr(Type.show[A])
+  }
+
 }
--- a/core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
+++ b/core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
@@ -3,7 +3,7 @@ package derivation
 
 import scala.compiletime.{constValue, erasedValue}
 import scala.deriving.Mirror
-import scala.quoted._
+import scala.quoted.*
 
 object Utils {
 
@@ -34,10 +34,13 @@ object Utils {
       case _: EmptyTuple => Nil
     }
 
-  object TypeName {
-    inline def apply[A]: String = ${ typeNameImpl[A] }
+  inline def typeName[A]: String = ${ typeNameImpl[A] }
 
-    def typeNameImpl[A](using Type[A], Quotes): Expr[String] = Expr(Type.show[A])
-  }
+  // object TypeName {
+  //   inline def apply[A]: String = ${ typeNameImpl[A] }
+
+  //   private def typeNameImpl[A](using Type[A], Quotes): Expr[String] = Expr(Type.show[A])
+  // }
+  private def typeNameImpl[A](using Type[A], Quotes): Expr[String] = Expr(Type.show[A])
 
 }
--- a/core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
+++ b/core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
@@ -36,11 +36,6 @@ object Utils {
 
   inline def typeName[A]: String = ${ typeNameImpl[A] }
 
-  // object TypeName {
-  //   inline def apply[A]: String = ${ typeNameImpl[A] }
-
-  //   private def typeNameImpl[A](using Type[A], Quotes): Expr[String] = Expr(Type.show[A])
-  // }
   private def typeNameImpl[A](using Type[A], Quotes): Expr[String] = Expr(Type.show[A])
 
 }
--- a/core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
+++ b/core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
@@ -3,7 +3,7 @@ package derivation
 
 import scala.compiletime.{constValue, erasedValue}
 import scala.deriving.Mirror
-import scala.quoted.*
+import scala.quoted._
 
 object Utils {
 
--- a/modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/AnyValDerivationMacros.scala
+++ b/modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/AnyValDerivationMacros.scala
@@ -11,31 +11,67 @@ private[pureconfig] object AnyValDerivationMacros {
   def deriveAnyValReaderImpl[A <: AnyVal: Type](using Quotes): Expr[ConfigReader[A]] = {
     import quotes.reflect._
 
-    val typeRepr = TypeRepr.of[A]
-    val wrapperSymbol = typeRepr.typeSymbol
+    val wrapperTypeRepr = TypeRepr.of[A]
+    val wrapperSymbol = wrapperTypeRepr.typeSymbol
     val underlyingField = wrapperSymbol.declaredFields.head
-    val underlyingTpe = typeRepr.memberType(underlyingField)
+    val underlyingTypeRepr = wrapperTypeRepr.memberType(underlyingField)
 
     // derive underlying reader and wrap it in a value class constructor
-    underlyingTpe.asType match {
+    underlyingTypeRepr.asType match {
       case '[t] =>
         val underlyingReader = Expr
           .summon[ConfigReader[t]]
           .getOrElse {
+            // TODO: derive reader for underlying type if it's not found ?
             report.errorAndAbort(
-              s"Cannot summon ConfigReader for value class ${typeRepr.show} wrapping ${underlyingTpe.show}."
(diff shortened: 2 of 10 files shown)


grzegorz-bielski:
Hello @ruippeixotog , @jcazevedo, this is yet another improvement for Scala 3 derivation. 
Please take a look in your spare time! 🙇

core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
```diff
@@ -32,4 +33,11 @@ object Utils {
       case _: (h *: t) => transform(constValue[h & String]) :: transformedLabelsTuple[t](transform)
       case _: EmptyTuple => Nil
     }
+
+  object TypeName {
+    inline def apply[A]: String = ${ typeNameImpl[A] }
+
+    def typeNameImpl[A](using Type[A], Quotes): Expr[String] = Expr(Type.show[A])
```
ruippeixotog:
```suggestion
    private def typeNameImpl[A](using Type[A], Quotes): Expr[String] = Expr(Type.show[A])
```

core/src/main/scala-3/pureconfig/generic/derivation/Utils.scala
```diff
@@ -32,4 +33,11 @@ object Utils {
       case _: (h *: t) => transform(constValue[h & String]) :: transformedLabelsTuple[t](transform)
       case _: EmptyTuple => Nil
     }
+
+  object TypeName {
+    inline def apply[A]: String = ${ typeNameImpl[A] }
```
ruippeixotog: nit: it feels a bit weird to have an object with an `apply` method for a regular function that doesn't create any custom type or data format. Are there any blockers with simply creating a `def typeName[A]` function in `Utils`?

modules/generic-scala3/src/main/scala/pureconfig/generic/semiauto.scala
```diff
@@ -12,6 +12,9 @@ object semiauto {
   export HintsAwareConfigWriterDerivation.deriveWriter
   export EnumDerivation._
 
+  inline def deriveConvert[A <: AnyVal]: ConfigConvert[A] =
```
ruippeixotog: Why are we overloading the method if the default implementation already has compile-time checks for whether the type is an `AnyVal`? If this is redundant, can we delete it and simply the user-facing interface?

Same thing for `deriveReader` and `deriveWriter`.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareConfigWriterDerivation.scala
```diff
@@ -5,19 +5,27 @@ package scala3
...
     }
 
-  protected inline def summonConfigWriter[A]: ConfigWriter[A] =
+  private[pureconfig] inline def summonConfigWriter[A]: ConfigWriter[A] =
     summonFrom {
       case writer: ConfigWriter[A] => writer
       case given Mirror.Of[A] => deriveWriter[A]
+      case _ =>
+        inline if AnyValDerivationMacros.isAnyVal[A]
+        then AnyValDerivationMacros.unsafeDeriveAnyValWriter[A]
```
ruippeixotog: Can you please use Scala 2 compatible syntax here?

We're starting to enforce it in https://github.com/pureconfig/pureconfig/pull/1677, so you may want to use the configs in that PR as a guide.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/AnyValDerivationMacros.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.deriving.Mirror
+import scala.quoted._
+
+private[pureconfig] object AnyValDerivationMacros {
+
+  /** Derive a `ConfigReader` for a value class. Can only be called after checking that `A` is a value class.
+    */
+  inline def unsafeDeriveAnyValReader[A]: ConfigReader[A] = ${ deriveAnyValReaderImpl[A] }
+
+  def deriveAnyValReaderImpl[A: Type](using Quotes): Expr[ConfigReader[A]] = {
```
ruippeixotog: Can we make all these private?

```suggestion
  private def deriveAnyValReaderImpl[A: Type](using Quotes): Expr[ConfigReader[A]] = {
```

ruippeixotog (changes requested):

modules/generic-scala3/src/main/scala/pureconfig/generic/semiauto.scala
grzegorz-bielski: It's not redundant. We have a compile-time check, but only for the recursive case of `summonConfigReader` where we don't know anything about the `A`. 
Mirrors are not synthesized for value classes, so it's not possible to call the `inline def deriveReader[A](using m: Mirror.Of[A]): ConfigReader[A]` version on them directly. We need either use the overload with `[A <: AnyVal]` or change the signature to the `inline def deriveReader[A]: ConfigReader[A]` and do the check inside, similar to the `summonConfigReader`, but without the first case in `summonFrom`. 
I think the solution with overload is cleaner.

Maybe only this `deriveConvert` is a little bit weird, because the implementation on the first glance looks the same 😄.

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/AnyValDerivationMacros.scala
```diff
@@ -0,0 +1,78 @@
...
+
+import scala.compiletime._
+import scala.deriving.Mirror
+import scala.quoted._
+
+private[scala3] object AnyValDerivationMacros {
+
+  /** Derive a `ConfigReader` for a value class. Can only be called after checking that `A` is a value class.
+    */
+  inline def unsafeDeriveAnyValReader[A](derivationImpl: HintsAwareConfigReaderDerivation): ConfigReader[A] =
```
grzegorz-bielski: Passing `derivationImpl` here enables the usage of the implementation provided by the `HintsAwareConfigReaderDerivation` trait, which hypothetically could be be further extended with some custom logic, for example - default static hints instead of the ones summoned inline.

grzegorz-bielski:
Hello @ruippeixotog , please have another look! 🙇

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/AnyValDerivationMacros.scala
ruippeixotog: I think I don't really want this unless there are real use cases and real users demanding it... And if such requests appeared, incorporating them in hints would be my primary choice.

None of our derivation code (Scala 2 or Scala 3) is intended to be extended by users other than by the ways we allow (i.e. hints), and I don't want to give users the impression it's any other way. Sometimes we don't declare classes as private out of convenience. Derivation code in general is already quite complex as it is - let's not add make it more complex than it needs to :)

modules/generic-scala3/src/main/scala/pureconfig/generic/semiauto.scala
grzegorz-bielski: After giving it some more thought, I think we might actually need to hide everything behind `inline def deriveReader[A]: ConfigReader[A]` anyway at some point, as it might become more cluttered with potential future opaque types derivation and auto derivation 🤔 .

I'll try experimenting with it a little bit more.

grzegorz-bielski: In the end I exposed in under simplified interface.

grzegorz-bielski:
@ruippeixotog please re-review 🙇

ruippeixotog:
Hey @grzegorz-bielski, the PR looks good to me. Can you just take a look at my comment above on allowing `derivationImpl` to be passed as an argument?

grzegorz-bielski:
> Hey @grzegorz-bielski, the PR looks good to me. Can you just take a look at my comment above on allowing `derivationImpl` to be passed as an argument?

@ruippeixotog I don't see any new comments 🤔

ruippeixotog (changes requested):
Apologies... I wrote this on mobile and forgot to publish the review 😞

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/AnyValDerivationMacros.scala
grzegorz-bielski: I removed it, it's indeed not yet used internally. 

The main use case for that was going to be implementing an internal variant of derivation without summoning inline hints that could unblock the `derives` syntax for `pureconfig-generic-scala3`, but that was going to be in some subsequent PR.
When experimenting with it I noticed that the mechanism is quite generic and flexible which might be useful to users, but you are right - it's better to use hints for this.
However, I strongly believe that if something is not intended to be used by users directly then we ought to make it private, even if it's less convenient!

ruippeixotog: Thanks for changing it!

> However, I strongly believe that if something is not intended to be used by users directly then we ought to make it private, even if it's less convenient!

I don't think functions and classes were deliberately left public, they're public only because no one bothered much with visibility modifiers. I'm aligned with you, feel free to make all derivation internals private - I'd personally not treat it as a breaking change.

ruippeixotog (approved):
LGTM!

grzegorz-bielski:
Hi @ruippeixotog !
I noticed that latest (and only) release `0.17.7` of `pureconfig-generic-scala3` doesn't include this feature.
When you have an opportunity, can you please prepare a new release? 🙏

ruippeixotog:
Apologies @grzegorz-bielski, I'll cut a release this weekend :)

ruippeixotog:
@grzegorz-bielski, I have just released v0.17.8, apologies once again for the delay.

PR #1672, opened 2024-06-10
fix CHANGELOG

grzegorz-bielski:
@ruippeixotog I'm afraid we don't support general derivation of `ConfigWriter` and `ConfigConvert` using `derives` clause just yet! We do have it for enums though.

If we want to support the `derives` clause we need to either reconsider [this](https://github.com/pureconfig/pureconfig/pull/1636#discussion_r1580104314) approach or reimplement the `ConfigWriter` and `ConfigConvert`  derivation in core without hints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ### 0.17.7 (Jun 9, 2024)
 
 - New features
-  - Added support for Scala 3 derivation of `ConfigWriter` and `ConfigConvert` using a `derives` clause.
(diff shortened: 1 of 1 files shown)


ruippeixotog:
Well spotted, I meant to talk about enum readers/writers but I ended up missing the `Enum` part :)

I think our approach hasn't changed - `derives` should be used for behavior locked in to our current defaults, while users may opt to use implicit-based derivation if they need configurable behavior.

ruippeixotog (approved):

PR #1674, opened 2024-06-11
Allow Scala 3 `derives` without an import and deprecate direct usage of derivation classes

ruippeixotog:
When the original Scala 3 derivation was designed, it was built such that users could customize it if they went through the trouble of creating their own derivation object and mixing in `ConfigReaderDerivation` traits to their liking. This resulted in a complex setup of traits using a cake pattern and the need for an explicit import.

`pureconfig-generic-scala3` is a more capable alternative for deriving readers in Scala 3 with support for configuration. Now that we have it, we can simplify the `pureconfig-core` implementation by locking in `derives` behavior to the defaults used by PureConfig. This means that we can make `derives` work without an import and, after a deprecation period, simplify derivation internals.

### DIFF
--- a/core/src/main/scala-3/pureconfig/ReaderDerives.scala
+++ b/core/src/main/scala-3/pureconfig/ReaderDerives.scala
@@ -2,7 +2,7 @@ package pureconfig
 
 import scala.deriving.Mirror
 
-import pureconfig.generic.derivation.*
+import pureconfig.generic.derivation._
 
 trait ReaderDerives {
   inline def derived[A](using m: Mirror.Of[A]): ConfigReader[A] =
--- a/core/src/main/scala-3/pureconfig/ReaderDerives.scala
+++ b/core/src/main/scala-3/pureconfig/ReaderDerives.scala
@@ -1,0 +1,10 @@
+package pureconfig
+
+import scala.deriving.Mirror
+
+import pureconfig.generic.derivation.*
+
+trait ReaderDerives {
+  inline def derived[A](using m: Mirror.Of[A]): ConfigReader[A] =
+    ConfigReaderDerivation.Default.deriveConfigReader[A]
+}
--- a/core/src/main/scala-2/pureconfig/ReaderDerives.scala
+++ b/core/src/main/scala-2/pureconfig/ReaderDerives.scala
@@ -1,0 +1,5 @@
+package pureconfig
+
+trait ReaderDerives {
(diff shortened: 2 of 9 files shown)


ruippeixotog:
@grzegorz-bielski, I'd appreciate your input on this as well.

grzegorz-bielski:
Hi @ruippeixotog, I think this is a good direction!

I would even take it a step further and at some point deprecate entire derivation in core (maybe besides enums).

Recently I've been [experimenting](https://github.com/grzegorz-bielski/pureconfig/pull/2/files) with `derives` clause based derivation in `generic-scala3` without the implicit hints summoning. 
Surprisingly, not a whole lot of changes is needed. A think this might be a good alternative to porting changes from `generic-scala3` into core, as it would result with less code to maintain.

core/src/main/scala-3/pureconfig/ReaderDerives.scala
```diff
@@ -0,0 +1,10 @@
+package pureconfig
+
+import scala.deriving.Mirror
+
+import pureconfig.generic.derivation._
+
+trait ReaderDerives {
+  inline def derived[A](using m: Mirror.Of[A]): ConfigReader[A] =
```
grzegorz-bielski, line 8: nit: Could be now `inline def derived[A: Mirror.Of]: ConfigReader[A] =`

ruippeixotog:
> I would even take it a step further and at some point deprecate entire derivation in core (maybe besides enums).

I'd consider proposals where we don't provide derivation capabilities in core, but I don't think you can enable derivation without an import unless you plug `derives` directly into a class's companion object (which in our case is in `core`), can you? If that's the case, I'd be less leaning towards a solution like that.

grzegorz-bielski:
>I don't think you can enable derivation without an import unless you plug `derives` directly into a class's companion object (which in our case is in `core`), can you?

No, but what we can do is to expose the extension method with `derived` in the `pureconfig` package, so it's always imported whenever you do `import pureconfig.*` / `import.pureconfig._` like in [here](https://github.com/grzegorz-bielski/pureconfig/pull/2/commits/9197e6f6a2d158c11ca4cfab01bc52e043e9bcb6).
It's not the same thing, but it's still pretty ergonomic. I believe it's a good trade off.

jcazevedo (approved):
Very cool. Thanks for this!

PR #1675, opened 2024-06-11
fix #1673 pureconfig-generic-scala3 derivation of ConfigReader doesn't work with default values

### LINKED ISSUE #1673: pureconfig-generic-scala3 derivation of ConfigReader doesn't work with default values
Continuation from https://github.com/pureconfig/pureconfig/issues/1488#issuecomment-2158917879

This still doesn't work as expected (should return Right)
```scala
import pureconfig.*
import pureconfig.generic.derivation.default.*

case class Person(name: String, age: Int = -1) derives ConfigReader

println(ConfigSource.string("{ name = foo }").load[Person])

// Left(ConfigReaderFailures(ConvertFailure(KeyNotFound(age,Set()),None,age)))
```
with `"com.github.pureconfig" %% "pureconfig-generic-scala3" % "0.17.7"`
https://scastie.scala-lang.org/T3SR4at8Sy6IQfPLvfVr1g

/cc @grzegorz-bielski

sideeffffect:
fixes https://github.com/pureconfig/pureconfig/issues/1673

### DIFF
--- a/modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/derived.scala
+++ b/modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/derived.scala
@@ -1,0 +1,16 @@
+package pureconfig
+package generic
+package scala3
+
+import HintsAwareConfigReaderDerivation.deriveReader
+import HintsAwareConfigWriterDerivation.deriveWriter
+import scala.deriving.Mirror
+
+extension (c: ConfigReader.type)
+  inline def derived[A: Mirror.Of: ProductHint: CoproductHint]: ConfigReader[A] = deriveReader
+
+extension (c: ConfigWriter.type)
+  inline def derived[A: Mirror.Of: ProductHint: CoproductHint]: ConfigWriter[A] = deriveWriter
+
+extension (c: ConfigConvert.type)
+  inline def derived[A: Mirror.Of: ProductHint: CoproductHint]: ConfigConvert[A] = semiauto.deriveConvert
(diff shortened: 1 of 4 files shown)


grzegorz-bielski:
Ouh, you won't get an approval for that from @ruippeixotog and @jcazevedo 😄 
See: https://github.com/pureconfig/pureconfig/pull/1672#issuecomment-2159752017 and https://github.com/pureconfig/pureconfig/pull/1636#discussion_r1580104314

sideeffffect:
:scream_cat: That would be very unfortunate...

The current situation isn't great. Being able to work with default parameters is very important. And doing that requires jumping through unnecessarily large number of hoops. Let's at least decrease them :pray:

sideeffffect:
But the ultimate goal, of course, should be to make `derives ConfigReader` work with default parameter values out of the box, without any magic imports.

ruippeixotog (changes requested):
@grzegorz-bielski already pointed out to the relevant discussions. The default `derives` implementation should be able to handle default values out of the box as you said, and now that Grzegorz already did the hard work of adding support for it, it's just a question of moving relevant bits to `pureconfig-core` and reuse in both sides. Let's do that instead?

> And doing that requires jumping through unnecessarily large number of hoops.

Come on, the number of hoops is not _that_ large 🙂 And by adding that simple 16-line file to `pureconfig-generic` you're only getting rid of one anyway, you still need the import (which tends to be a common practice for PureConfig users that need to change PureConfig's default conventions globally).

sideeffffect:
> The default `derives` implementation should be able to handle default values out of the box

:100: 

> just a question of moving relevant bits to `pureconfig-core`

Of course, that would be the best :+1: :+1: :+1:

But at this point, I'm a bit confused... If I may ask, why wasn't _this_ done in the first place?

PR #1677, opened 2024-06-13
Enforce Scala 2 compatible syntax

ruippeixotog:
In past PRs we've been asking contributors to  use Scala 2 compatible syntax. On this PR I'm starting to enforce these choices using scalac and scalafix.

Most PureConfig code is cross-compiled against Scala 2.12, 2.13 and 3 and in those we have no choice but to use syntax compatible with all versions. Some Scala3-only code _could_ technically use the new syntax, but this would create a discrepancy in code style that I'd rather avoid as long as we can (i.e. until Scala 3 drops support for some Scala 2 syntax).

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,7 +1,6 @@
 rules = [OrganizeImports]
 
 OrganizeImports {
-  targetDialect = Scala2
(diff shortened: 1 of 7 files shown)


jcazevedo (approved):
Thank you for this!

PR #1703, opened 2024-09-04
Update reference of scala-3-derivation doc in index.md

elabeca:
Currently the scala-3-derivation link points to an .html file which is non-existent. This pull request updates this reference to point to the correct docs/docs/docs/scala-3-derivation.md markdown documentation file.

### DIFF
--- a/docs/docs/docs/index.md
+++ b/docs/docs/docs/index.md
@@ -20,7 +20,7 @@ Users of Scala 3 need to add the following dependency to their `build.sbt`:
 libraryDependencies += "com.github.pureconfig" %% "pureconfig-core" % "@VERSION@"
 ```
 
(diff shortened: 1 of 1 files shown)


ruippeixotog (changes requested):
Hi @elabeca, thanks for spotting this! However, the docs in the repo seem to be in some inconsistent state. The file you changed is the source for https://pureconfig.github.io/docs/index.html (where referencing the html version is correct). The toplevel `README.md` is built from `bundle/docs/README.md`, but for some reason they're not in sync.

I believe you want to run the missing paragraphs to `bundle/docs/README.md` ("For Scala 3, add (...) Scala 3's derivation that you can find here.") with the full URL (https://pureconfig.github.io/docs/scala-3-derivation.html) and run `sbt mdoc` to generate `README.md`.

I'll check the CI config later, as it should have detected and blocked this from happening.

elabeca:
Thanks @ruippeixotog

PR #1718, opened 2024-09-25
Generate `Automatic-Module-Name` attribute

hubertp:
It is currently impossible to define `pureconfig` on a module path. This is because Java fails to automatically infer Module name from jar's containing invalid characters (e.g. `_` or `-` in `_2.13` cross version). It's a common problem with Scala artifacts.

This change adds `Automatic-Module-Name` attribute to MANIFEST that contains a valid Java identifier that can be used to define appropriate module dependencies.

Also fixed a warning that was stopping the compilation for those who are on JDK 20 already.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,9 +15,18 @@ ThisBuild / libraryDependencySchemes ++= Seq(
   "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always
 )
 
+lazy val javaModuleName = settingKey[String]("Java Module name")
+lazy val publishSettingAutomaticModuleName = Seq(
+  packageOptions += Package.ManifestAttributes("Automatic-Module-Name" -> javaModuleName.value)
+)
+
 lazy val core = (project in file("core"))
   .enablePlugins(BoilerplatePlugin)
-  .settings(commonSettings)
+  .settings(
+    commonSettings,
+    javaModuleName := "pureconfig.core",
+    publishSettingAutomaticModuleName
+  )
 
 lazy val testkit = (project in file("testkit"))
   .settings(commonSettings)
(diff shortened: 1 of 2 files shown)


hubertp:
Any problem with this PR? This should unblock anyone who wants to use `pureconfig` as a Java module, with a minimal effort.

grzegorz-bielski:
This is a nice addition.

cc @ruippeixotog, @jcazevedo

build.sbt
```diff
@@ -15,9 +15,18 @@ ThisBuild / libraryDependencySchemes ++= Seq(
   "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always
 )
 
+lazy val javaModuleName = settingKey[String]("Java Module name")
```
ruippeixotog, line 18: nit: can you please move this to after the projects are defined? This seems to be too niche and minor to be defined at the top of build.sbt.

build.sbt
```diff
@@ -44,6 +53,7 @@ def genericModule(proj: Project) = proj
   .dependsOn(testkit % "test")
   .settings(commonSettings)
   .settings(name := s"pureconfig-${baseDirectory.value.getName}")
+  .settings(Seq(javaModuleName := s"${name.value.replace("-", ".")}") ++ publishSettingAutomaticModuleName)
```
ruippeixotog, line 56: I don't understand why we're defining this only for `core` and generic modules. Don't other modules, like `pekko-http` and `scala-xml`, suffer from the same problem? If we can, we should skip the SBT setting boilerplate and just inline this into `commonSettings`.

ruippeixotog (changes requested):
Hey @hubertp, thanks for the contribution. Is there anywhere where I can read more about Java module usages and Scala's recommended approach to this? As you said this seems to be a problem common to all Scala artifacts, so is there general guidance from the Scala team or the community? It seems strange for every project to require that small amount of boilerplate.

This LGTM in general, sending back you just just for minor adjustments.

ruippeixotog:

PR #1791, opened 2025-03-23
Drop support for Scala 2.12

ruippeixotog:
The latest versions of Scala 3 have been slowly deprecating/removing support for [some Scala 2 syntax](https://www.scala-lang.org/blog/2024/02/29/scala-3.4.0-and-3.3.3-released.html#legacy-syntax-deprecation). The latest version is 3.6.x, but we're still in 3.3.x as dependency update PRs have been consistently failing. This is something I've seen in other projects as well.

The Scala team has been retrofitting Scala 3 syntax into Scala 2.13 behind a compiler flag. Scala 2.12 didn't get such support, meaning that in order to continue to support the latest Scala 3 we need to either 1) fork a lot of source code, or 2) drop support for Scala 2.12.

The amount of downloads from Sonatype shows that only 5% of last month's pureconfig_2.12 downloads were for v0.17.8, our latest version released in Nov 2024. This seems to indicate that many 2.12 users are stuck in an older version and wouldn't be affected with us stopping new releases.

image

By comparison, Scala 2.13 downloads are predominantly for the latest versions, as expected:

image

Dropping support for Scala 2.12 seems to be the best course of action here.


jcazevedo (changes requested):
I also think it's appropriate to drop support for Scala 2.12. The changes look good. There's only one suite in the Magnolia submodule (`CoproductConvertersSuite2_12`) which we can now incorporate into the main spec.

> The latest version is 3.6.x, but we're still in 3.3.x as dependency update PRs have been consistently failing.

Has PureConfig been getting updates updates to Scala 3 for versions above 3.3? The [recommendation](https://www.scala-lang.org/development/) is to use the LTS line (currently 3.3.x) for libraries.

ruippeixotog:
> I also think it's appropriate to drop support for Scala 2.12. The changes look good. There's only one suite in the Magnolia submodule (`CoproductConvertersSuite2_12`) which we can now incorporate into the main spec.

Nice catch, it looks like it was a Scala 2.11 carveout, so we could have merged it before. I'm planning to submit a follow up PR  clearing the codebase of Scala 2.13 warnings after this so I may catch a few more of these :)

> Has PureConfig been getting updates updates to Scala 3 for versions above 3.3? The [recommendation](https://www.scala-lang.org/development/) is to use the LTS line (currently 3.3.x) for libraries.

Oh, that's nice to know. I've been getting updates in other libraries (e.g. https://github.com/ruippeixotog/scala-scraper/pull/561), so I thought it was expected - but it's true that we haven't been receiving pureconfig PRs past Scala 3.3. Do you know if this is a scala-steward setting?

mzuehlke:
> Do you know if this is a scala-steward setting?

Scala Steward does not creates PR from Scala LTS to Scala Next.
But once a Scala version greater than 3.3.x has been picked manually Scala Steward will creates PRs to follow the latest versions.

jcazevedo (approved):
Thank you!

gabrieljones:
Unfortunately https://github.com/pureconfig/pureconfig/commit/4834622ed19334f3d9c5a1370efb013178a03a21 broke (https://github.com/pureconfig/pureconfig/issues/1818) runtime compatibility for 2.13 libs compiled against 0.17.8 running on 0.17.9.  Per Semver this should have been released with a semver major increment. When you are still in 0.x land is a semver increment going from `0.17.x` to `0.18.x`?

gabrieljones:
How hard would it be to:
- create a release of this on v0.18.x
- create a branch for 0.17.x
- revert this change on the o.17.x branch
- release 0.17.9
?

ruippeixotog:
Hi @gabrieljones, in semver you should have no expectations whatsoever about breaking changes in v0.x.y versions ([spec](https://semver.org/#spec-item-4)).

The only thing we usually guarantee in PureConfig is source compatibility between patch versions (v0.x.y -> v0.x.z). We don't guarantee binary compatibility between any version - you should always compile your code with the PureConfig version you'll be using at runtime.

PR #1792, opened 2025-03-23
Remove installation of libxslt-dev from CI build

jcazevedo:
It looks like at some point in time the PureConfig website required libxslt-dev to build. However, that no longer seems to be the case. Seeing as it is failing the installation in recent builds (https://github.com/pureconfig/pureconfig/actions/runs/14014181748/job/39247022750), we might as well remove it.

Incidentally, this also doesn't pin on a specific Jekyll 4.x version, as that seems to be what's [recommended by sbt-microsites](https://47degrees.github.io/sbt-microsites/docs/getting-started/#continuous-integration---travis).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,9 +82,7 @@ jobs:
           ruby-version: 3.2
 
       - name: Install dependencies
-        run: >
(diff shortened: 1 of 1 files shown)


ruippeixotog (approved):
Thanks!

PR #1794, opened 2025-03-24
Avoid multiple calls to `ProductHint#bottom` in derived `ConfigReader`s for Scala 3

jcazevedo:
`ProductHint#bottom` is meant to be called only once when producing an instance for a product of type `A` within `ConfigReader#from`. The current code was calling it once per field of the product, resulting in possibly multiple repeated failures. This is problematic for strict derivations, as we get as many `UnknownKey` failures as valid fields in the product being read.


ruippeixotog (approved):
LGTM

PR #1802, opened 2025-04-15
Clear all Scala 2.13 warnings

ruippeixotog:
We couldn't fix a lot of these warnings before due to the need to keep source code compatible with Scala 2.12. Now that we dropped Scala 2.12, it's time to clean up code.

Besides fixing existing warnings, on this PR I'm also modifying some scalac options to issue more of them (e.g. adding `-deprecation`, removing `-Ywarn-unused:-implicits` in favor of explicit `@unused` annotations, `-Xsource:3` to warn about behavioral differences with Scala 3).

I fixed all `-Xsource:3` issues in mdoc snippets as well (as they're being flagged as errors), but I haven't fixed all warnings there as code simplicity and conciseness is generally more important for documentation.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -106,7 +106,7 @@ lazy val commonSettings = Seq(
 
   Test / scalacOptions ~= { _.filterNot(_.contains("-Ywarn-unused")) },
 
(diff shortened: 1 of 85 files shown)


jcazevedo (approved):
Thank you!

PR #1817, opened 2025-05-30
Ignore StackOverflow URLs when checking PureConfig website

ruippeixotog:
The latest dependency PRs have all been failing CI because htmlproofer can't fetch a StackOverflow link in the website. This has been the case for a while, so it doesn't seem transient - SO likely implemented anti-scraping measures affecting GitHub Actions.

The SO question still seems relevant, so I think we should just exclude stackoverflow.com from liveness checks in CI.


jcazevedo (approved):
Thank you!

PR #1820, opened 2025-06-04
feat(yaml): changing inheritance of the CustomConstructor to EnvScalarConstructor to enable env var interpolation

bruno-de-queiroz:
Reasons:
Although HOCON provides the replacement by env var, by default, the loader removes duplicated keys. Of course, we could simply `allowDuplicates` in the `LoaderOptions`, but having the `EnvScalarConstructor` already there, I feel it's overall a more standardized approach to env var interpolations in yaml files.

Not adding all the test cases for the possible cases covered by the EnvScalarConstructor, because I feel I would be re-testing the snakeyaml framework, but let me know.


bruno-de-queiroz:
Bump @ruippeixotog :)

ruippeixotog (changes requested):
Hi @bruno-de-queiroz, the PR sounds good to me overall.

My only ask is that this behavior should be optional and configurable by users. Contrary to Typesafe Config, where env var support is encoded in the spec, loading an YAML file doesn't usually come with an expectation that env vars are going to be replaced, and even if it was chaging it now would be a breaking change.

Could you please change your code such that users can enable env var resolution by writing something like this?

```scala
YamlConfigSource.file(resourcePath("multi3.yaml"))
  .withEnvironmentOverrides()
  .load[Conf]
```

This would more closely emulate `ConfigFactory.systemEnvironmentOverrides()`.

modules/yaml/src/main/scala/pureconfig/module/yaml/YamlConfigSource.scala
```diff
@@ -80,6 +97,10 @@ final class YamlConfigSource private (
     yamlConstructors.put(Tag.TIMESTAMP, new ConstructYamlStr())
   }
 
+  private[this] class CustomEnvironmentOverrideConstructor extends EnvScalarConstructor {
+    yamlConstructors.put(Tag.TIMESTAMP, new ConstructYamlStr())
```
bruno-de-queiroz, line 100: Couldn't find a better way to reduce repetition.  yamlConstructors is protected, and the ConstructYamlStr is not statically exposed. My initial idea was to leverage a trait: 
```scala
trait StringTimestampConstruct { 
  this: SafeConstructor => 
  this.yamlConstructors.put(Tag.TIMESTAMP, new this.ConstrctYamlStr())
}
```

ct-brunoqueiroz:
bump @ruippeixotog :)

modules/yaml/src/main/scala/pureconfig/module/yaml/YamlConfigSource.scala
ruippeixotog: Yeah, it's a shame, but I think this is fine.

ruippeixotog (approved):
Thanks for reminding me of this, LGTM!

modules/yaml/src/main/scala/pureconfig/module/yaml/YamlConfigSource.scala
```diff
@@ -10,16 +10,19 @@ import scala.util.Try
...
 import org.yaml.snakeyaml.error.{Mark, MarkedYAMLException, YAMLException}
 import org.yaml.snakeyaml.nodes.Tag
-import org.yaml.snakeyaml.{LoaderOptions, Yaml}
 
 import pureconfig.ConfigReader.Result
 import pureconfig.error._
 import pureconfig.module.yaml.error.{NonStringKeyFound, UnsupportedYamlType}
 import pureconfig.{ConfigObjectSource, ConfigSource}
 
+import scala.util.chaining.scalaUtilChainingOps
```
ruippeixotog: Actually, I believe you need to run scalafmt and scalafix before we merge this, as this import is out of order. That is failing the Scala 2.13 build.

ruippeixotog (changes requested):

modules/yaml/src/main/scala/pureconfig/module/yaml/YamlConfigSource.scala
ct-brunoqueiroz: Done :)

bruno-de-queiroz: Done :)

ruippeixotog (approved):

PR #1824, opened 2025-06-27
Prepare publishing to Sonatype Central Portal

jcazevedo:
This prepares the publishing settings so that we can start pushing artifacts to [Sonatype Central Portal](https://central.sonatype.com/). I have already migrated the com.github.pureconfig namespace to the Central Portal and published a snapshot of PureConfig there. sbt-sonatype has been removed since [sbt 1.11.0](https://github.com/sbt/sbt/releases/tag/v1.11.0) already supports central repository publishing.

I have used similar settings to publish [a test project of mine](https://github.com/jcazevedo/sonatype-central-scala/blob/main/build.sbt) to the Central Portal. Note that publishers need to generate [a user token](https://central.sonatype.com/account) and use that instead of the existing username and password.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -117,7 +117,11 @@ lazy val commonSettings = Seq(
 
   publishMavenStyle := true,
   Test / publishArtifact := false,
-  publishTo := sonatypePublishToBundle.value,
+  publishTo := {
+    val centralSnapshots = "https://central.sonatype.com/repository/maven-snapshots/"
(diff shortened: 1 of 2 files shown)


jcazevedo:
Once this is merged, I'll probably publish a new release, to make sure everything is working as expected.

ruippeixotog (approved):
Thank you!

PR #1838, opened 2025-08-09
Rid of deprecated URL constructor

satorg:
`java.net.URL` constructors are [officially deprecated](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/net/URL.html#constructor-summary) since JDK 20.
The recommended way to create `URL` is via `URI`:
```
new URI("https://github.com").toURL()
```


mzuehlke:
I think the pull request description should be: new URI(...

satorg:
Yes, sure, good catch, thank you! Fixed.

mzuehlke (approved):
Thanks

ruippeixotog (approved):
Thank you!

Issue #1795, opened 2025-03-26
Can't parse option of complex type - key not found 'type'


mwisnicki:
Can't read this simple config:

```scala
case class RootWithNested(nested: Option[Nested])
case class Nested(enabled: Boolean)
```

It works when I replace `Nested` with `Map[String, String]`.

Fails with:
```
pureconfig.error.ConfigReaderException: Cannot convert configuration to a RootWithNested. Failures are:
  at 'nested':
    - (String: 2) Key not found: 'type'.


	at pureconfig.ConfigSource.loadOrThrow(ConfigSource.scala:81)
	at pureconfig.ConfigSource.loadOrThrow$(ConfigSource.scala:24)
	at pureconfig.ConfigObjectSource.loadOrThrow(ConfigSource.scala:92)
	at ConfigSpec.$init$$$anonfun$2(ConfigSpec.scala:34)
```

<details>
<summary>
Full repro
</summary>

```scala
//> using scala "3.3.5"
//> using dep "org.scalameta::munit::1.1.0"
//> using dep "com.github.pureconfig::pureconfig-generic-scala3::0.17.8"

import munit.FunSuite

import pureconfig.*
import pureconfig.generic.semiauto.*

case class RootWithNested(nested: Option[Nested])
case class Nested(enabled: Boolean)

case class RootWithMap(nested: Option[Map[String, String]])


class ConfigSpec extends FunSuite  {

  given ConfigReader[RootWithMap] = deriveReader
  given ConfigReader[RootWithNested] = deriveReader

  val conf =
    """
      |nested {
      |  enabled = false
      |}
      |""".stripMargin

  test("RootWithMap") {
    val parsed = ConfigSource.string(conf).loadOrThrow[RootWithMap]()
    println(s"Parsed: $parsed")
  }

  test("RootWithNested") {
    val parsed = ConfigSource.string(conf).loadOrThrow[RootWithNested]()
    println(s"Parsed: $parsed")
  }
}
```
</details>

mwisnicki:
What I found is that adding

```scala
given ConfigReader[Nested] = deriveReader
```

fixes the problem so looks like recursive derivation does not work correctly?

thereisnospoon:
We are currently facing exactly the same issue while migrating our applications to Scala 3.
Thanks for creating this issue @mwisnicki !

ruippeixotog:
[...]

PR #1839, opened 2025-08-12
Fix semiauto derivation

Odomontois:
Targets basically the same issue as #1795
Provide specific methods for even more semiauto derivation in scala 3. 

The proposed method adds two more derivation methods (for Reader and Writer). Those disallows auto-derivation of any enum or sealed trait (i.e. having `Mirror.SumOf`) after auto-derivation flow came through some case class or tuple (i.e. having `Mirror.ProductOf`), effectively stopping derivation at any fields of type `Option[Foo]`, `List[Foo]`, etc. 

We add additional contextual parameter, containing two flags: 
* `auto` that distincts between current behaviour and new behaviour
* `allowAutoSum` that turns false only when we meet first product
And check the second one whenever we would like to derive instance for the sum type.

### DIFF
--- a/modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/AnyValDerivationMacros.scala
+++ b/modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/AnyValDerivationMacros.scala
@@ -10,9 +10,8 @@ private[scala3] object AnyValDerivationMacros {
 
   /** Derive a `ConfigReader` for a value class. Can only be called after checking that `A` is a value class.
     */
-  inline def unsafeDeriveAnyValReader[A](using inline df: DerivationFlow): ConfigReader[A] = ${
-    deriveAnyValReaderImpl[A]('df)
-  }
+  inline def unsafeDeriveAnyValReader[A](using inline df: DerivationFlow): ConfigReader[A] =
+    ${ deriveAnyValReaderImpl[A]('df) }
 
   private def deriveAnyValReaderImpl[A: Type](df: Expr[DerivationFlow])(using Quotes): Expr[ConfigReader[A]] = {
     import quotes.reflect._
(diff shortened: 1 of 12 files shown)


ruippeixotog (changes requested):
Hey @Odomontois, thanks for the PR.

While I'm personally in favor of fixing the unexpected behavior described in #1795, I'm not a fan of the complexity this PR introduces for library users.

Inside our `semiauto` module we'd now have two methods, some of them confusingly having a `*Semiauto` suffix. Are the non-semiauto methods in the semiauto module in fact auto? No, because their behavior doesn't match the auto derivation we have in Scala 2, and they don't match the semiauto methods either. Granting users the level of control provided by `DerivationFlow` can lead to even weirder behaviors that I'm not keen on introducing.

I suppose your intention with adding `DerivationFlow` was to maintain backwards compatibility, which is understandable. But I consider the current behavior of Scala 3 semiauto derivation to be a bug, not something we want to keep. Our goal with these native derivations should be to get them as close to our original Scala 2 one as much as possible and we never claimed our Scala 3 implementation is stable, so I'd be in favor of breaking backwards compatibility and fixing existing methods to align with Scala 2 semiauto. @jcazevedo, do you agree?

Odomontois:
@ruippeixotog 
Am I getting it correctly, you're advising changing the current scala 3 derivation towards real semiauto behaviour, which also means changing most of the scala 3 derivation unit tests?

ruippeixotog:
@Odomontois, if by "real semiauto" you mean changing the Scala 3 behavior to match the one in Scala 2, then yes, that's what I'm advising.

I think the changes in Scala 3 derivation tests will be limited to adding givens for case classes or sealed families that don't have them.

Odomontois:
@ruippeixotog I've tried removing everything and doing fair semiauto first, but it felt like having auto-style descend to embedded case-classes and tuples could be very useful. And we need to descend to low-level sealed traits in case of complex sealed hierarchies.
So I kept the "DerivationFlow" stuff as a bare contextual compile-time wrapper around a boolean constant.

PR #1841, opened 2025-08-21
Use `shields.io` Maven Central badge in README

bdmendes:
Fixes the broken badge. Also, updates the URL to Maven Central, without the need to rely on the (possibly fading) redirect from the old repository.


ruippeixotog (approved):
Thanks!

PR #1894, opened 2026-01-23
Fix Scala 3 unknown keys not being enforced for empty case classes and case objects

afonsoafonsoafonso:
When using `ProductHint` with `allowUnknownKeys = false`, the validation was not being performed for case objects and zero-arity / empty case classes. Unknown keys in the configuration were silently ignored instead of producing an error.

Reproducible `scala-cli` snippet:
```
//> using scala "3.3.7"
//> using dep "com.github.pureconfig::pureconfig-core:0.17.9"
//> using dep "com.github.pureconfig::pureconfig-generic-scala3:0.17.9"

import pureconfig._
import pureconfig.syntax._
import pureconfig.generic.semiauto._
import pureconfig.generic.ProductHint

@main def bug() =
  case class EmptyConf()

  given ProductHint[EmptyConf] = ProductHint(allowUnknownKeys = false)
  given ConfigReader[EmptyConf] = deriveReader

  val conf = ConfigFactory.parseString("""a = 1""")

  println(conf.to[EmptyConf])
  // Prints: Right(EmptyConf())
  // Expected: Left(ConfigReaderFailures(ConvertFailure(UnknownKey(a),Some(ConfigOrigin(String)),a)))
```

Although not included in regression tests, this bug is not present in the Scala 2 Shapeless-based implementation, as seen in the newly-implemented specs and snippet below:
```
//> using scala "2.13.18"
//> using dep "com.github.pureconfig::pureconfig-core:0.17.9"
//> using dep "com.github.pureconfig::pureconfig-generic:0.17.9"

import pureconfig._
import pureconfig.syntax._
import pureconfig.generic.semiauto._
import pureconfig.generic.ProductHint

object Bug extends App {
  case class EmptyConf()

  implicit val hint: ProductHint[EmptyConf] = ProductHint(allowUnknownKeys = false)
  implicit val reader: ConfigReader[EmptyConf] = deriveReader

  val conf = ConfigFactory.parseString("""a = 1""")

  println(conf.to[EmptyConf])
  // Correctly prints: Left(ConfigReaderFailures(ConvertFailure(UnknownKey(a),Some(ConfigOrigin(String)),a)))
}
```

The bug was due to empty case classes or case objects immediately entering the empty tuple pattern match case, skipping the bottom validations. The fix was to essentially iterate through the tuple as we already do in the `readTuple` method.

Also extracted the combination of the head and tail results into a new method to avoid the duplicate logic.

### DIFF
--- a/modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
+++ b/modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
@@ -15,13 +15,6 @@ import pureconfig.generic.derivation.Utils.widen
 import ProductDerivationMacros._
 
 trait HintsAwareProductConfigReaderDerivation { self: HintsAwareConfigReaderDerivation =>
-
-  private inline def combineResults[H, T <: Tuple, R <: Tuple](
-      head: Either[ConfigReaderFailures, H],
-      tail: Either[ConfigReaderFailures, T]
-  ): Either[ConfigReaderFailures, R] =
-    ConfigReader.Result.zipWith(head, tail)((h, t) => widen[H *: T, R](h *: t))
-
   inline def deriveProductReader[A](using pm: Mirror.ProductOf[A], ph: ProductHint[A]): ConfigReader[A] =
     inline erasedValue[A] match {
       case _: Tuple =>
@@ -85,7 +78,7 @@ trait HintsAwareProductConfigReaderDerivation { self: HintsAwareConfigReaderDeri
 
         val tail = readCaseClass[t, N + 1, A](objCursor, labels, actions, defaults)
 
-        combineResults[h, t, T](head, tail)
+        ConfigReader.Result.zipWith(head, tail)((h, t) => widen[h *: t, T](h *: t))
 
       case _: EmptyTuple =>
         val usedFields = actions.map(_._2.field).toSet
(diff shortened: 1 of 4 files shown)


modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
```diff
@@ -15,6 +15,13 @@ import pureconfig.generic.derivation.Utils.widen
 import ProductDerivationMacros._
 
 trait HintsAwareProductConfigReaderDerivation { self: HintsAwareConfigReaderDerivation =>
+
+  private inline def combineResults[H, T <: Tuple, R <: Tuple](
+      head: Either[ConfigReaderFailures, H],
+      tail: Either[ConfigReaderFailures, T]
+  ): Either[ConfigReaderFailures, R] =
+    ConfigReader.Result.zipWith(head, tail)((h, t) => widen[H *: T, R](h *: t))
```
jcazevedo: I don't think this expression warrants its own dedicated method. Would it be OK to inline it?

jcazevedo (changes requested):
Thank you for the fix!

modules/generic-scala3/src/main/scala/pureconfig/generic/scala3/HintsAwareProductConfigReaderDerivation.scala
afonsoafonsoafonso: Sure! See 415ae421

jcazevedo (approved):
Thank you!

Issue #1742, opened 2024-11-21
Is it possible to have `allowUnknownKeys = false` with Scala 3 derivation


gabrieljones:
I saw that Scala 3 derivation does not yet consider `ProductHint`. Is some other way to achieve `allowUnknownKeys = false` across my object graph with Scala 3 derivation?

gabrieljones:
My current workaround is to separate my pureconfig code into a subproject that uses Scala 2.13 while the main project uses Scala 3. It would be nice to bring them all under the same roof.

ruippeixotog:
Hi @gabrieljones, @grzegorz-bielski recently developed a `pureconfig-generic-scala3` module that should work (mostly) as a drop-in replacement for `pureconfig-generic` in Scala 2.13 ([code](https://github.com/pureconfig/pureconfig/tree/master/modules/generic-scala3)). While there isn't a lot of documentation for it, as long as you add it as a dependency alongside `pureconfig-core`, hints should be respected. Note that the module only supports semiauto derivation, so you'll need to explicitly create readers for your top-level configs at least.

ruippeixotog:
We still intend the derives implementation in `pureconfig-core` non-configurable, with behavior matching the no-hint implementation in Scala 2. `generic-scala3` is here for advanced users that need customization. Closing as won't fix.

PR #1943, opened 2026-06-05
Use case class defaults in Scala 3 `derives` derivation

### LINKED ISSUE #1673: pureconfig-generic-scala3 derivation of ConfigReader doesn't work with default values

lrytz:
Fixes #1673: `ConfigReader`s derived in Scala 3 via `derives` clauses fail with `KeyNotFound` when the key for a parameter with a default value is missing from the config.

This moves the `getDefaults` macro from `pureconfig-generic-scala3` to `pureconfig-core` and uses it in the case-class branch of the core derivation, with `pureconfig-generic-scala3` reusing it from its new location — the approach suggested in the review of #1675, and one of the code-reuse follow-ups anticipated in #1636.

The semantics follow the hints-aware derivation under the default `ProductHint`: a default is used exactly when its key is missing, taking precedence over `ReadsMissingKeys` readers, and defaults are evaluated lazily (preserving the #1218 fix). With this change, `derives` parses the same configs to the same values as `pureconfig-generic-scala3` under default hints. Hint support in `derives` remains intentionally out of scope (#1742).

### DIFF
--- a/core/src/main/scala-3/pureconfig/generic/derivation/ProductDerivationMacros.scala
+++ b/core/src/main/scala-3/pureconfig/generic/derivation/ProductDerivationMacros.scala
@@ -1,0 +1,40 @@
+package pureconfig
+package generic
+package derivation
+
+import scala.quoted._
+
+private[pureconfig] object ProductDerivationMacros {
+  type DefaultValue = Option[() => Any]
+
+  /** Materializes the default values of the constructor parameters of `T` as a vector of thunks, with `None` for
+    * parameters without a default.
+    *
+    * @param size
+    *   the number of constructor parameters of `T`
+    */
+  inline def getDefaults[T](inline size: Int): Vector[DefaultValue] = ${ getDefaultsImpl[T]('size) }
+
+  private def getDefaultsImpl[T](size: Expr[Int])(using Quotes, Type[T]): Expr[Vector[DefaultValue]] = {
+    import quotes.reflect._
+
+    val n = size.valueOrAbort
+    val typeRepr = TypeRepr.of[T]
+
+    def defaultMethodAt(i: Int) =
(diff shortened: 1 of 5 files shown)


lrytz:
LLM-assisted, I iterated on it, tested it etc, so not a blind vibe.

Should #1742 be closed as won't-fix?

core/src/main/scala-3/pureconfig/generic/derivation/ProductDerivationMacros.scala
```diff
@@ -0,0 +1,40 @@
+package pureconfig
+package generic
+package derivation
+
+import scala.quoted._
+
+private[pureconfig] object ProductDerivationMacros {
```
lrytz, line 7: Code in this file moved from `modules/generic-scala3` unchanged

ruippeixotog:
Hey @lrytz, sorry for the long wait - I'll review it by the end of this week 🙂

ruippeixotog (approved):
LGTM. Nice and clean, thank you!

> Should #1742 be closed as won't-fix?

Yeah, I think we still want the `derives` implementation in `pureconfig-core` non-configurable, with behavior matching the no-hint implementation in Scala 2. `generic-scala3` is here for advanced users that need customization. I'll close it.

